PR #98, opened 2013-03-09
Moved private and public archives to /res/raw for android library usage.

leonrd:
* Moved archives to res/raw since android library projects cannot contain files in assets/ .

(Just in case someone wants to use the distribution as a library for another project.)


tito:
This will be reworked soon, there is no need of this pr now :)

PR #130, opened 2013-06-08
Coldstart speedup

tito:
**DO NOT MERGE**

This is an experimentation for redirecting all the C file handler to the right place. If the user refer to kivy/lang.py, we will open libs/py_kivy_lang.so instead, even if the .so is a .py. We use Android libs folder to put our own data.

The approach used is to prepend a .c file for every compiled file that define redirection:
```c
#define __open __android_opem
#define open __android_open
#define fopen __android_fopen
#define freopen __android_freopen
#define stat __android_stat
#define lstat __android_lstat
#define dlopen __android_dlopen
#define access __android_access
#define opendir __android_opendir
#define readdir __android_readdir
#define closedir __android_closedir
```

And we do a external shared library that implement thoses methods.
It's ugly, but the decompression is done during the installation, not during the first start.

So the real deal here is: 
- did i miss any low-level function to wrap?
- what's the real impact?


tito:
Test has been done, but because we're still unsure about the whole implication, it has been abandonned.

FeralBytes:
May I ask what did the tests yield? Would you be willing to allow further testing by other Apps? Or do you have a better plan to try? Ultimately it is not a huge deal to me, because it is only a factor for the first time start up.

PR #200, opened 2014-01-22
Add recipe for requests library

b3ni:


requestsuser:
Requests don't works without this recipe. ImportError: No module named 'requests.packages.urllib3'

PR #222, opened 2014-03-25
Support agave froyo api 8 android2.2

joranbeasley:
resolves issues that prevent building against API 8 for android 2.2 Agave-Froyo.  easy changes to do manually but it was suggested on the freenode IRC channel that a do a pull request with my changes and you could potentially make a branch for (limited) 2.2 support?

### DIFF
--- a/src/src/org/renpy/android/SDLSurfaceView.java
+++ b/src/src/org/renpy/android/SDLSurfaceView.java
@@ -1223,7 +1223,7 @@ public class SDLSurfaceView extends SurfaceView implements SurfaceHolder.Callbac
                 return super.sendKeyEvent(event);
             }
 
-            @Override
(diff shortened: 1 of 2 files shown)


tito:
@akshayaurora What is the impact of @Override for later devices?

If there is no impact, i guess this PR would be good if the change `""` could be conditional.

Otherwise, please status on the PR (either work, ask the author to update, branch it, or nothing). Unsure if we still want to support 8 API as a minimum.

davelab6:
Should this be closed?

inclement:
This API 8 issue may still exist, but since this PR is inactive and was never merged I'm going to close it. An appropriate fix in the new toolchain would be welcome, though I'm not sure what's the best way to do it.

PR #249, opened 2014-06-08
Multi-window support and Two new recipes: WebOb and CJK

hinesmr:
1. cjklib ('cjklib' on pypi) is a pure-python swiss-army knife translation library for character-oriented languages like Chinese.
2. WebOb ('webob' on pypi) is a popular python WSGI middleware library for building reusable python web applications.
3. One extra line in AndroidManifest.xml.tmpl to support Samsung Multi-window support.
    
I've tested both of these in my buildozer-created APK and they work great. No patches required as they do not have C-extensions.
    
One warning: cjklib uses sqlite and creates a pretty "hefty" 45 MB database upon it's 'setup.py' installation for the purposes of translating different languages. I've submitted a request on their mailing list for options for pruning the size of this database. Hope to get a response from them soon.

Comments welcome!

### DIFF
--- a/recipes/cjklib/recipe.sh
+++ b/recipes/cjklib/recipe.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+VERSION_cjklib=${VERSION_cjklib:-0.3.2}
+DEPS_cjklib=(setuptools android)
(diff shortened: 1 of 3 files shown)


tito:
Need another approach for #255, or redo this PR without the multiwindow changes.

tito:
Now it's the #305

davelab6:
I guess this can be closed?

Issue #234, opened 2014-05-10
*.so is too small to be an ELF executable


trivedigaurav:
Running a packaged kivy app on Android causes: "ImportError: dlopen failed: "/data/data/org.test.sms/files/lib/python2.7/site-packages/kivy/properties.so" is too small to be an ELF executable.

Run Log (using logcat): https://gist.github.com/trivedigaurav/74197a06807c4abdb485
Build Log (using P4A): https://gist.github.com/trivedigaurav/8203502e17bd1cf52e31

*Related:*
* Groups Discussion: https://groups.google.com/d/topic/kivy-users/jvnpObU-Ukc/discussion
* (Possibly) Related Bug: https://github.com/kivy/python-for-android/issues/77

thegrymek:
Hi, 

I have the same error when I set P4A_pyjinius_DIR.

cbenhagen:
I am running into the same error with NO P4A_pyjinius_DIR set. https://github.com/kivy/python-for-android/pull/251 did not fix it.

cbenhagen:
I finally found what triggered this error in my case: pyyaml. Actually the pyyaml recipe. After removing it from p4a (`rm -f .buildozer/android/platform/python-for-android/recipes/pyyaml`) my app worked again. While this is only a workaround it might help narrow down the actual bug. Or at least it would help someone who got bitten by it.

I tested on Ubuntu 14.04 and OSX 10.10.2 with a minimal hello world example: https://gist.github.com/cbenhagen/e16da236c3aa821d0a81

dl1ksv:
I get the same problem when building ( and running ) the compass example of kivy from a fresh git checkout
distribute.sh -m "kivy"
The only adaptation I  made was setting the toolchain version to 4.6 instead of 4.4.3 in distribute.sh
The file jnius.so is empty ( 0 byte).
Looking into the build messages I got

creating build/lib.linux-x86_64-2.7/jnius
/home/schroer/python-for-android/src/tools/liblink -lm -L/home/schroer/python-for-android/build/libs -DANDROID -mandroid -fomit-frame-pointer --sysroot /opt/android-ndk/platforms/android-14/arch-arm build/temp.linux-x86_64-2.7/jnius/jnius.o -Llibs/armeabi -L/home/schroer/python-for-android/build/python-install/lib -lsdl -llog -lpython2.7 -o build/lib.linux-x86_64-2.7/jnius/jnius.so
[...]

PR #251, opened 2014-06-22
Fixed kivy recipe to work on repeated runs with P4A_kivy_DIR

inclement:
Not for merging without discussion!

This would fix https://github.com/kivy/python-for-android/issues/234

I'm not sure if this is the right fix, but I'm doing a pr in the hope that someone who knows what's going on will see it. As far as I can tell, it *does* resolve the linked issue.

Without this pr, something doesn't work - the build_ext lines don't place object files where the biglink script can see them. I think the issue is that they recognise the pyx source files haven't changed and therefore don't rebuild the object files, whereas with this fix cython is always called and the files are always rebuilt.

As discussed with tito, maybe there is a better solution, but I'm not even completely sure what's going on and I couldn't track it down better than this.

### DIFF
--- a/recipes/kivy/recipe.sh
+++ b/recipes/kivy/recipe.sh
@@ -27,7 +27,7 @@ function build_kivy() {
 
 	# fake try to be able to cythonize generated files
 	$HOSTPYTHON setup.py build_ext
-	try find . -iname '*.pyx' -exec $CYTHON {} \;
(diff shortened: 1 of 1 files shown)


tito:
Ok, then, let's go. At least, it will resolve current issue that people don't clean their own kivy between 2 different arch build.

Issue #255, opened 2014-07-21
Full control of the AndroidManifest.xml [$20]


vidplace7:
Allow for full control over the contents of the AndroidManifest.xml.

This is crucial to allow developers to limit devices that can install their APKs, for instance, [apps that require BLE](https://developer.android.com/guide/topics/connectivity/bluetooth-le.html#permissions) using the <uses-feature> tag.

Perhaps allow for custom XML to be passed to the build script, or add new fields for each XML tag.
##

kamathln:
+1 for re-opening

We can definitely pass along the manifest as a element tree or something. Recipes and such can play around with the element tree adding and modifying nodes.

daytipper1989:
> You can overwrite templates/AndroidManifest.tmpl.xml on your own. There is also a manifest extras option inside that template near the bottom. If you want a buildozer option to make it easier to overwrite the template in its entirety, you would need to file the bug on buildozer. I'm pretty sure there should already be options for including the manifest extras.

When I overwrite the AndroidManifest.tmpl.xml, it is regenerated everytime I deploy and my changes are lost. Why is this happening with me? any idea? please help.

Issue #172, opened 2013-11-14
Own Activities in AndroidManifest.xml


RudolfVonKrugstein:
Hey,

I want to mix python and java activities. I have written my java activity, and I want to start it from python. For android to accept this, the activity must be registered in the AndroidManifest.xml. Is there any way to get this declaration into the manifest when using python-for-android?

PR #305, opened 2014-12-04
Add --manifest-hook for transforming AndroidManifest

bionoid:
Warning: Do not merge without review/test!

Possible solution for #172, #255. 

I don't have time to play with standalone p4a for several more weeks, if anyone can test and provide feedback in the mean time that would be great. **This has not been tested at all** and could have syntax or other errors

### DIFF
--- a/docs/source/example_manifest.rst
+++ b/docs/source/example_manifest.rst
@@ -1,0 +1,51 @@
+Manipulating the Android Manifest
+---------------------------------
+
+AndroidManifest.xml is built from a template file in python-for-android
+distribution. The template is rendered using data from various command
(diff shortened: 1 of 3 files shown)


tito:
The idea is good, but not really integratable within other tools (like buildozer). I'm not completly in favor of this, a generic xpath/xml tool would be needed to select a part of the xml via xpath and:

- remove a tag + children
- append a tag / text chunk
- remove a property
- append a property

inclement:
I'm going to close this as inactive, and not mergeable under its current state. If anyone would like the feature to be added, it would still be welcome under a new PR.

PR #316, opened 2015-01-02
Don't search for the pip command

mmariani:
When pip is needed, it's provided by the mandatory virtual environment.
Therefore, there is no need to require an external one.


tito:
good catch, thanks

PR #320, opened 2015-01-27
save and restore environment when running recipe functions

kived:
Prevents env vars set in one recipe build function from affecting other recipe build functions.

### DIFF
--- a/distribute.sh
+++ b/distribute.sh
@@ -666,13 +666,23 @@ function run_get_packages() {
 	done
 }
 
+function envfn() {
+	envsave=$(mktemp)
+	envrestore=$(mktemp)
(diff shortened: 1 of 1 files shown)


distribute.sh
```diff
@@ -666,13 +666,23 @@ function run_get_packages() {
 	done
 }
 
+function envfn() {
+	envsave=$(mktemp)
+	envrestore=$(mktemp)
+	set > $envsave
+	$1
+	set > $envrestore
+	eval $(grep -v -F -f$envrestore $envsave)
```
FeralBytes: Shouldn't there be a space between:
```shell
-f $envrestore
```

Issue #322, opened 2015-02-02
Python3 Branch: jinja2 traceback on buildozer --verbose android debug


FeralBytes:
This issue is specific to inclement's Python3 branch of Kivy Py4A.
```
Traceback (most recent call last):
  File "build.py", line 18, in <module>
    import jinja2
  File "buildlib/jinja2-py3.egg/jinja2/__init__.py", line 33, in <module>
    from jinja2.environment import Environment, Template
  File "buildlib/jinja2-py3.egg/jinja2/environment.py", line 13, in <module>
    from jinja2 import nodes
  File "buildlib/jinja2-py3.egg/jinja2/nodes.py", line 18, in <module>
    from jinja2.utils import Markup
  File "buildlib/jinja2-py3.egg/jinja2/utils.py", line 520, in <module>
    from markupsafe import Markup, escape, soft_unicode
ImportError: No module named 'markupsafe'
```

FeralBytes:
Fixed by running:
```shell
sudo pip3 install markupsafe
```
Recommend add dependency for install of markupsafe.

FeralBytes:
Fixed by running:
```shell
sudo pip3 install markupsafe
```
Recommend add dependency for install of markupsafe.

FeralBytes:
Next Traceback:
```shell
Traceback (most recent call last):
  File "build.py", line 506, in <module>
    make_package(args)
  File "build.py", line 286, in make_package
    manifest_extra=manifest_extra,
  File "build.py", line 72, in render
    text = template.render(**kwargs)
  File "buildlib/jinja2-py3.egg/jinja2/environment.py", line 969, in render
    return self.environment.handle_exception(exc_info, True)
  File "buildlib/jinja2-py3.egg/jinja2/environment.py", line 742, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "buildlib/jinja2-py3.egg/jinja2/_compat.py", line 36, in reraise
    raise value.with_traceback(tb)
  File "templates/AndroidManifest.tmpl.xml", line 14, in <module>
    {% if args.min_sdk_version >= 9 %}
TypeError: unorderable types: str() >= int()
```
Patch to Fix is:
Line 14 Becomes:
```xml
{% if args.min_sdk_version|int >= 9 %}
```

FeralBytes:
Next Traceback:
```
Traceback (most recent call last):
  File "build.py", line 506, in <module>
    make_package(args)
  File "build.py", line 286, in make_package
[...]

PR #323, opened 2015-02-03
Rebase Python3 Branch

FeralBytes:
Other tweaks are noted [here](https://github.com/kivy/python-for-android/issues/322).
One main thing that is missing but mentioned in that issue is that I had to manually add markupsafe to my system and I am not sure how Kivy manages dependencies like that.


tito:
Good work!! At this point, maybe you want access to our repo for maintenance as well?

FeralBytes:
@tito Thanks. I would definitely like to help!

PR #334, opened 2015-03-07
Add warning and description for P4A in a venv...

FeralBytes:
Adds a full description of the error caused by using P4A from within a
virtual enviroment to the FAQ. Also adds a Warning in usuage that links
to the FAQ in reference to the command that could result in the error.
Tells users how to avoid OSError: [Errno 40] Too many levels of 
symbolic links.

The only thing I am uncertian about is whether the relative link will 
map correctly but I wrote it with readthedocs.org in mind.


tito:
I didn't knew about it, thanks :)

PR #336, opened 2015-03-13
Apply ignore_path filter to private data

emanuele-f:
Currently the --ignore-path option only applies for public data.
Since installing "release" apk version on /data requires using private data, the ignore filter is also needed here.


tito:
Thanks!

PR #344, opened 2015-03-21
fix mktemp usage for OS X

r1chardj0n3s:
This fixes the invocation of mktemp so it works on OS X also.

### DIFF
--- a/distribute.sh
+++ b/distribute.sh
@@ -654,8 +654,8 @@ function run_get_packages() {
 }
 
 function envfn() {
-	envsave=$(mktemp)
-	envrestore=$(mktemp)
+	envsave=$(mktemp -t envsave)
+	envrestore=$(mktemp -t envsave)
 	set > $envsave
 	$1
 	set > $envrestore


kived:
I've reverted this as it breaks mktemp on Linux. Please find and test a solution which works on both OSX and Linux, or make the script check for which platform is in use. I don't know what problem this was supposed to fix.

r1chardj0n3s:
It was my understanding that the mktemp -t option is supported under Linux. I guess it's only supported under some flavours of Linux :/

How does it fail for you - what is the error? I see that on some mktemps the -t option must take a specifically-formatted string with a number of X in it (ie. ensaveXXXXX). Is that the case for you?

The problem this is solving is that the mktemp command does not work on OS X.

kived:
Yes, the `-t` option works on Linux and it requires XXXXXX in the template. But plain `mktemp` works just fine, and it seems odd that it wouldn't on OSX. So I didn't just make the change to fix, because I can't test on OSX and don't know why plain `mktemp` wouldn't work. :)

If you fix this to work on both Linux and OSX, I will gladly merge it. :)

r1chardj0n3s:
Unfortunately the mktemp version on OS X requires a template. Hilariously
the -t option does different things on the different versions.

denys-duchier:
on linux you need `XXXXXX` at the end of envsave, i.e. `envsaveXXXXXX`.

kived:
I don't know a ton about OSX, but from what I understand it's not horribly uncommon for OSX to have standard commands which work different from their Linux counterparts. Apparently they take that "Think Differently" very seriously. :P

Feel free to add an OS check (maybe via `uname`, that works on Linux anyway :P) if necessary. You could also do something like the following:

    envsave=$(mktemp)
    if [ $? -ne 0 ]; then
        envsave=$(mktemp -t envsave)
    fi

But only if you *can't* specify the XXXXXX part on OSX. Though I'd think that the worst case scenario would be that the filename would end up being `/tmp/envsaveXXXXXXdHGlsh` or something.

PR #345, opened 2015-03-21
check the calculated NDK platform path for validity

r1chardj0n3s:
This simple addition checks that the calculated NDK path is correct. It wasn't for me, and resulted in bizarre errors down the track when it's simple enough to check at this point.


tshirtman:
:+1:

PR #347, opened 2015-03-24
make get_keyboard_height more dynamic

denys-duchier:
when using WIndow.softinput_mode='pan' with kivy. I noticed that changing the orientation of the screen (on android) (e.g. from portrait to landscape), caused the display to misbehave: eventhough the system keyboard had not be requested, the whole screen seem to be offset as if it had been.

The problem was that get_keyboard_height from the android module had a static notion of height.  My patch causes it to reevaluate the height (this is especially important in case of dispaly orientation change).

With my app, I now get the expected behaviour.


akshayaurora:
:+1:

PR #351, opened 2015-03-28
reciepes for freetype and harfbuzz

olymk2:
reciepe for freetype and harfbuzz libraries allows freetype python file to be imported.

### DIFF
--- a/recipes/freetype/recipe.sh
+++ b/recipes/freetype/recipe.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+VERSION_freetype=${VERSION_freetype:-2.5.5}
+DEPS_freetype=(harfbuzz)
+URL_freetype=http://download.savannah.gnu.org/releases/freetype/freetype-2.5.5.tar.gz
+MD5_freetype=7448edfbd40c7aa5088684b0a3edb2b8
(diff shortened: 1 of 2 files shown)


tito:
Freetype is one of the already checkout in src/jni. Why do you need to have it again?

olymk2:
not sure what you mean from what i could figure out i could not import freetype into my python program i tried different required libraries etc and asked around but got the impression it was not available so ported this.
I am using it specifically for creating font outlines if there is something i can add to buildozer.spec and import into python then great :)

olymk2:
on second thought are you saying i can just change the reciepe to look for freetype in another location ?

olymk2:
@tito okay i tried this looks like the freetype code in src/jni is different some how.

they will not build outside of py4a 

copying the files to python-for-android/build/freetype and then entering that folder before running

autogen.sh gives me 

You must be in the same directory as `autogen.sh'.
Bootstrapping doesn't work if srcdir != builddir.

running configure gives me 

Makefile:32: builds/toplevel.mk: No such file or directory
make: *** No rule to make target 'builds/toplevel.mk'. Stop.

so using the existing code may not be an option unless you might know why ?

tito:
Thanks!

PR #355, opened 2015-04-04
bump opencv version to 2.4.11

cbenhagen:
STOP please do not merge this yet!

i somehow can't get it to compile anymore.


cbenhagen:
```
*** Bad machine code: Using an undefined physical register ***
- function:    _ZN2cv6kmeansERKNS_11_InputArrayEiRKNS_12_OutputArrayENS_12TermCriteriaEiiS5_
- basic block: BB#129 if.end.i.invoke.cont243_crit_edge (0x745d300)
- instruction: %R0<def,tied2>, %CPSR<def> = tADDi8 %R0<tied0>, 156, pred:14, pred:%noreg
- operand 2:   %R0<tied0>
```
Never seen this kind of errors. Maybe someone can help.. 
you can find the full output here: https://gist.github.com/cbenhagen/50609d60bb447738e23c

cbenhagen:
this is what changed between 2.4.10.1 and 2.4.11:
https://github.com/Itseez/opencv/compare/2.4.11...2.4.10.1

AndreMiras:
Thanks @cbenhagen for that pull request. I'm sorry it never made it to the tree. Now it's a bit too old and not compatible with the new toolchain. I hope you don't mind me closing it.
Doing/testing the same change on the new toolchain should be rather easy. Basically we just need to update that line https://github.com/kivy/python-for-android/blob/b3d3d45/pythonforandroid/recipes/opencv/__init__.py#L12

PR #358, opened 2015-04-06
some python 3 compatibility

pohmelie:


src/build.py
```diff
@@ -68,7 +68,11 @@ def render(template, dest, **kwargs):
     template = environment.get_template(template)
     text = template.render(**kwargs)
 
-    f = file(dest, 'wb')
+    if sys.version_info[0] < 3:
+        f = file(dest, "wb")
+    else:
+        f = open(dest, 'wb')
+
```
olymk2: just wondering open works in 2 and 3 so why do we even need file ?
in fact we could probably do just these two lines ? but i am curious if there is a reason for file 

with open(dest, 'wb') as f:
    f.write(text.encode('utf-8'))

pohmelie: I don't really know about 2, but I think you are 99% right. My way was "not to brake", so I just turned off my mind in version 2 branch of this if statement :innocent:

tito:
This has been fixed in another pr, thanks anyway :)

PR #361, opened 2015-04-10
for some python3 compatibility 

a358003542:
for some python3 compatibility , and change the jinja2 to the newest version

the modification list here:

1. build.py some print function 
2. build.py line 277-278
        if sys.version_info[0] == 2:
        args.name = args.name.decode('utf-8')
3. templates/AndroidManifest.tmpl.xml  
    14         {% if args.min_sdk_version|int >= 9 %}
    32			  android:configChanges="mcc|mnc|locale|touchscreen|keyboard|keyboardHidden|navigation|orientation|screenLayout|fontScale|uiMode{% if args.min_sdk_version|int >= 13 %}|screenSize{% endif %}" 

4. buildlib/jinja2.egg/jinja2/

just copy from the jinja2 version 2.7.2 for better python3 compatibility .


tito:
Thanks!

PR #371, opened 2015-05-08
recipes: update django to 1.8.1

franciscod:
hope you like it! :)


franciscod:
I've been reading a bit this issuetracker and somebody said that this recipe is kind of useless since it's a pure python module... should it be removed?

akshayaurora:
Pure python modules can simply be specified in bulldozer and the module will be installed without the need for a recipe. Thanks for all your efforts though. We appreciate every bit of help.

PR #372, opened 2015-05-09
Make presplash image fit the screen

Mystic-Mirage:
I made my presplash image to look good on all my devices with different screen resolutions.
Tested image sizes: 416x530, 256x256, 512x512, 1024x1024.
Tested screen resolutions (both -- portrait and landscape): 320x480, 480x800.

I am not math expert so more testing needed with different image sizes and screen resolutions. May be I just lucky. :)

### DIFF
--- a/src/src/org/renpy/android/SDLSurfaceView.java
+++ b/src/src/org/renpy/android/SDLSurfaceView.java
@@ -808,8 +808,11 @@ public class SDLSurfaceView extends SurfaceView implements SurfaceHolder.Callbac
         GLES20.glViewport(0, 0, mWidth, mHeight);
 
         if (bitmap != null) {
-            float mx = ((float)mWidth / bitmap.getWidth()) / 2.0f;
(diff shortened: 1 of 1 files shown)


tito:
This is an interesting patch, however, it need to be configurable. Some splashscreen are not meant to be fitted. I'll let this PR open until somebody add an environmnent configuration that can be used to select the behavior.

I suggest:

ANDROID_SPLASHSCREEN=fit
ANDROID_SPLASHSCREEN=default

j-santander:
@tito I can try my hand at adding those configuration options, but .... could you explain how the environment configuration works? is there somewhere I could read the details? or other configuration item that is currently handled in the same way?

I understand this would be an option in the build.py + some entry in the buildozer configuration file.
... and then that somehow gets added to the package and read from the Java/JNI code?

j-santander:
Ok, I did it based on metadata, see pull request #408

inclement:
Closing in favour of #408

PR #375, opened 2015-05-15
fix ctypes build errors

kived:


olymk2:
Just tested this against my project and it build fine thanks for improving it :)

tito:
Thanks kived! That's explain a lot why some people didn't make it works :)

PR #390, opened 2015-06-12
add leveldb and plyvel recipes

mathewvermeer:
LevelDB is Google's key-value storage library. Plyvel is a Python interface to LevelDB.

### DIFF
--- a/recipes/leveldb/recipe.sh
+++ b/recipes/leveldb/recipe.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+VERSION_leveldb=${VERSION_leveldb:-1.18}
+URL_leveldb=https://github.com/google/leveldb/archive/v${VERSION_leveldb}.tar.gz
+DEPS_leveldb=()
+BUILD_leveldb=$BUILD_PATH/leveldb/$(get_directory $URL_leveldb)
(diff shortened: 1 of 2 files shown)


recipes/leveldb/recipe.sh
```diff
@@ -0,0 +1,48 @@
...
+	export CFLAGS="$CFLAGS -fpic -shared"
+	export CXXFLAGS=$CFLAGS
+	export LDFLAGS="$LDFLAGS -L$LIBS_PATH"
+	export LDFLAGS="$LDFLAGS -lgnustl_shared"
+	export LDSHARED=$LIBLINK
+
+	# Make sure leveldb is compiled for Android and does not include versioned numbers
+	export TARGET_OS=OS_ANDROID_CROSSCOMPILE
+	#FIXME: Not idempotent
+	sed -i "127i\ \ \ \ \ \ \ \ PLATFORM_SHARED_VERSIONED=" build_detect_platform
```
tito: Could you add a "try" before the sed? That if it fail for any reason, it stop :)

mathewvermeer:
Done!

akshayaurora:
@mathewvermeer thanks :+1:

PR #396, opened 2015-06-14
update openssl version to 1.0.2c

cbenhagen:


akshayaurora:
@cbenhagen thanks :+1:

PR #398, opened 2015-06-17
Updated Twisted and Zope recipes

torfinnberset:
Tested building new libraries with official Kivy Python-for-Android VM


tito:
Thanks!

PR #404, opened 2015-06-28
Update paramiko recipe

extra84:
- Update paramiko recipe to a newer version 
- Add ecdsa recipe (dependency)


tito:
Thanks a lot!

PR #405, opened 2015-06-29
Add libtribler recipe

ontwikkelaar:
Add a recipe for libtribler. Tribler can be found at https://github.com/Tribler/tribler.

### DIFF
--- a/recipes/libtribler/recipe.sh
+++ b/recipes/libtribler/recipe.sh
@@ -1,18 +1,37 @@
 #!/bin/bash
+VERSION_libtribler=${VERSION_libtribler:-devel}
 DEPS_libtribler=(kivy python openssl m2crypto twisted sqlite3 pyasn1 apsw cherrypy netifaces libtorrent libnacl libsodium pil plyvel requests)
-BUILD_libtribler=$BUILD_PATH/libtribler/tribler-git
+BUILD_libtribler=$BUILD_PATH/libtribler
 RECIPE_libtribler=$RECIPES_PATH/libtribler
 
 function prebuild_libtribler() {
 
-	# Clone repo so all submodules are included:
-	cd $BUILD_PATH/libtribler
-	git clone --recursive https://github.com/tribler/tribler.git tribler-git
+	cd $PACKAGES_PATH/libtribler
+	LIBTRIBLER_CHANGED=1
+	if [ ! -d ./tribler-git ]; then
+		# Clone repo so all submodules are included:
+		try git clone --recursive https://github.com/tribler/tribler.git tribler-git
+	else
+		# Pull latest changes and check whether anything changed at all:
+		pushd ./tribler-git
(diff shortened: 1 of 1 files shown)


recipes/libtribler/recipe.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+DEPS_libtribler=(kivy python openssl m2crypto twisted sqlite3 pyasn1 apsw cherrypy netifaces libtorrent libnacl libsodium pil plyvel requests)
+BUILD_libtribler=$BUILD_PATH/libtribler/tribler-git
+RECIPE_libtribler=$RECIPES_PATH/libtribler
+
+function prebuild_libtribler() {
+
+	# Clone repo so all submodules are included:
+	cd $BUILD_PATH/libtribler
+	git clone --recursive https://github.com/tribler/tribler.git tribler-git
```
tito: Can't the recipe use the zipball from github? Aka https://github.com/Tribler/tribler/archive/devel.zip
Otherwise, please enhance the recipe to prevent downloading /cloning the repo everytime :)

Darsestheus: The problem with the archive is that it doesn't download the git submodules. I'll have a look at this and perhaps package the submodules separately, probably the preferred option, or see if I can do something about the cloning.

ontwikkelaar: Tribler uses git submodules for Dispersy and pymdht. It then makes imports such as Tribler.dispersy.etc. and something similar for pydmht. So it expects these two modules to be present inside libtribler (and not externally). Using the direct archive from GitHub means that these submodules will be missing, like Darsestheus stated.
The approach in my latest commit is that libtribler is recursively cloned (as to include submodules) to $PACKAGES_PATH/libtribler. This is only done if it hasn't been cloned before. If it has been cloned before than a git pull command is executed instead to fetch the latest changes. In case any changes were fetched (it checks this by comparing the latest commit hash before and after pulling), or in the case that the build directory is empty, the libtribler directory (in $PACKAGES_PATH/libtribler) is copied to the build directory (in $BUILD_libtribler).

brussee:
Picking up this work again for my master thesis project.

I need the following missing recipes for Tribler that need to be recreated for the new p4a-toolchain:

libsodium, 
libtorrent, {libboost}
pil, {jpeg,png}
cherrypy, 
m2crypto, {openssl}
apsw, {sqlite3}
libnacl, <<
plyvel, {||python-leveldb}{leveldb}
netifaces, 
sqlite3, 
requests <<

brussee:
All required recipes mentioned above are now done.
This PR is superseded by #726 so can be closed.

inclement:
@brussee Thanks for all these, and I'm sorry for the delay with merging them, I've had less time than I hoped for p4a recently but expect to be able to look at everything properly soon.

Issue #321, opened 2015-02-01
Python 2.7.2 don't build cleanly with GCC ≥ 4.8


rdoursenaud:
Due to http://bugs.python.org/issue17547, python (and hostpython) targets don't build cleanly with GCC ≥ 4.8.

This makes buildozer useless here on my Arch Linux.

Is there any plan to update Python or should I backport the patch ?
##

FeralBytes:
Is this a p4a issue or is it really a Python issue, I see the issue you linked and it is mark resolved. So what really is the issue? Sounds to me like an Arch issue as you are waiting for Arch to push the patched versions of Python, is that right?
Also the issue that you reference against buildozer #153 does not seem related to me, because it is a Java Exception that kevinkjt2000 is experiencing, but the python issue from above is a GCC compiler flag error.
I am really just trying to understand better so, please expand as able, thank you.

rdoursenaud:
This is a python-gcc compatibility issue that's bleeding into p4a because of the outdated python version used. It's fixed upstream but not in the version used by p4a.
The buildozer bug may not be related but its p4a build log clearly show the same symptoms I have.

inclement:
No longer an issue with the build tools moving to clang.

PR #407, opened 2015-07-04
Python 2.7.2 doesn't build cleanly with GCC ≥ 4.8

kronenpj:
I propose this patch as a solution.to issue #321.


inclement:
Looks like I'm using gcc 5.1, and I haven't had a problem - how do you find it doesn't build cleanly?

The patch seems fairly innocuous, but I'm not sure if I can test it.

kronenpj:
I too am using GCC 5.1 (5.1.1 actually).  It's related to a python issue that's been fixed in more recent versions, but not in Python 2.7.2: http://bugs.python.org/issue17547.

Can you try on a different machine / VM or in some other way make sure that you're not re-using a working build or environment? 

I'd started to update to Python 2.7.4 or later but the codebase was too different for the time I had available.

inclement:
I'm fairly sure I'm not using an old working build as I've rebuilt
python many times recently. I'll check out the issue though, and may
come across it anyway since I was going to test some builds in a VM.

kronenpj:
Fair enough.  I had another thought, it may be that Fedora (my distro) and others mentioned in the Python issue adopted removal of -Wformat but your distribution re-added it to avoid such breakage.

inclement:
For what it's worth, I use Arch Linux - I wouldn't usually expect such a patch but it's not impossible.

I read the python bug and it does seem like I should see it, and probably so should other people, but I haven't seen anyone report the problem. Weird.

kronenpj:
Agreed, I'm usually not the one to find these things the first time :smile:

AndreMiras:
Hello @kronenpj thanks for the patch and sorry we couldn't make it to the tree.
This is rather old and built on the old toolchain.
Also that patch was integrated in 34e269e52 since then.

PR #410, opened 2015-07-07
Enabling android-21 builds

dl1ksv:
This patch enables python to be build for android-21 and fixes the problems for using gcc >=4.8, too

### DIFF
--- a/distribute.sh
+++ b/distribute.sh
@@ -168,7 +168,7 @@ function push_arm() {
 	export CXXFLAGS="$CFLAGS"
 
 	# that could be done only for darwin platform, but it doesn't hurt.
-	export LDFLAGS="-lm"
+	# export LDFLAGS="-lm"
 
 	# this must be something depending of the API level of Android
 	PYPLATFORM=$($PYTHON -c 'from __future__ import print_function; import sys; print(sys.platform)')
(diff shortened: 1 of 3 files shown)


inclement:
Thanks for this patch, it looks good but I guess someone should test it before merging - I'll try to do it soon if nobody else does as I'll need to address the archs thing anyway for the p4a revamp.

About the arch, is the move to armeabi-v7a only an official change in android 5.0? I didn't know that, is there a doc page about it?

dl1ksv:
Without the arch change I get
/opt/android-ndk/platforms/android-21/arch-arm/usr/include/machine/cpu-features.h:52:6: error: #error Unknown or unsupported ARM architecture
 #    error Unknown or unsupported ARM architecture
      ^
/home/schroer/python-for-android/src/jni/../jni/jpeg/jddctmgr.c:31:8: warning: #warning "ANDROID_ARMV6_IDCT is disabled" [-Wcpp]
       #warning "ANDROID_ARMV6_IDCT is disabled"
        ^

I tested with the compass example on an nexus 7.

inclement:
That's interesting, I'll check it at the time. I guess it should be okay though, my understanding was that most devices support it now and maybe if you target that api they *must* support it.

There might be an issue with ctypes, I remember a mention that it didn't work with v7a compilation, though that may be out of date or refer to something slightly different. Just noting it here for reerence.

thopiekar:
+1 for this pull :dancer:

inclement:
I'll merge this, sorry for the delay, I was hoping someone who knows more about it could try it. Since it's currently broken with that api, I guess it's fine. I'm still concerned about the ctypes thing, has anyone happened to try it?

PR #412, opened 2015-07-08
Fix crash on presplash screen if back button pressed

Mystic-Mirage:


inclement:
Looks good, thanks, and sorry for the delay in merging it.

Issue #409, opened 2015-07-07
numpy fails to build 


dl1ksv:
Numpy fails to build with a lot of error messages like

/opt/android-ndk/toolchains/arm-linux-androideabi-4.8/prebuilt/linux-x86_64/bin/../lib/gcc/arm-linux-androideabi/4.8/../../../../arm-linux-androideabi/bin/ld: error: npymath: no archive symb
ol table (run ranlib)
/opt/android-ndk/toolchains/arm-linux-androideabi-4.8/prebuilt/linux-x86_64/bin/../lib/gcc/arm-linux-androideabi/4.8/../../../../arm-linux-androideabi/bin/ld: error: npysort: no archive symb
ol table (run ranlib)
/opt/android-ndk/toolchains/arm-linux-androideabi-4.8/prebuilt/linux-x86_64/bin/../lib/gcc/arm-linux-androideabi/4.8/../../../../arm-linux-androideabi/bin/ld: error: hidden symbol 'introsele
ct_bool' is not defined locally
/opt/android-ndk/toolchains/arm-linux-androideabi-4.8/prebuilt/linux-x86_64/bin/../lib/gcc/arm-linux-androideabi/4.8/../../../../arm-linux-androideabi/bin/ld: error: hidden symbol 'aintrosel
ect_bool' is not defined locally
/opt/android-ndk/toolchains/arm-linux-androideabi-4.8/prebuilt/linux-x86_64/bin/../lib/gcc/arm-linux-androideabi/4.8/../../../../arm-linux-androideabi/bin/ld: error: hidden symbol 'introsele
ct_byte' is not defined locally
/opt/android-ndk/toolchains/arm-linux-androideabi-4.8/prebuilt/linux-x86_64/bin/../lib/gcc/arm-linux-androideabi/4.8/../../../../arm-linux-androideabi/bin/ld: error: hidden symbol 'aintrosel
ect_byte' is not defined locally
/opt/android-ndk/toolchains/arm-linux-androideabi-4.8/prebuilt/linux-x86_64/bin/../lib/gcc/arm-linux-androideabi/4.8/../../../../arm-linux-androideabi/bin/ld: error: hidden symbol 'introsele
ct_ubyte' is not defined locally
[...]

PR #414, opened 2015-07-10
Fixing wrong use of ar when building numpy

dl1ksv:
numpy does not build ( see issue 409 ).

https://github.com/kivy/python-for-android/issues/409

Though the correct version of ar is found

Found executable /usr/bin/gfortran
Found executable /opt/android-ndk/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/arm-linux-androideabi-ld
Found executable /opt/android-ndk/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/arm-linux-androideabi-ar
Found executable /opt/android-ndk/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/arm-linux-androideabi-ranlib

the build process uses ar instead of arm-linux-androideabi-ar.

I fixed this by using $AR which is set by distribute.sh

In addition I found that lapack_lite.so and some other libs require linking against libm.so when running on android-21.

I fixed this, too.

### DIFF
--- a/recipes/numpy/patches/ar.patch
+++ b/recipes/numpy/patches/ar.patch
@@ -1,0 +1,11 @@
+--- numpy/distutils/unixccompiler.orig.py	2015-02-01 17:38:21.000000000 +0100
++++ numpy/distutils/unixccompiler.py	2015-07-08 17:21:05.742468485 +0200
+@@ -82,6 +82,8 @@
+             pass
+         self.mkpath(os.path.dirname(output_filename))
+         tmp_objects = objects + self.objects
++        self.archiver[0]="$AR"
++        print("#####Archiver: " + ', '.join(self.archiver) ) 
+         while tmp_objects:
+             objects = tmp_objects[:50]
+             tmp_objects = tmp_objects[50:]
--- a/recipes/numpy/patches/ar.patch
+++ b/recipes/numpy/patches/ar.patch
@@ -1,11 +1,11 @@
---- numpy/distutils/unixccompiler.orig.py	2015-02-01 17:38:21.000000000 +0100
-+++ numpy/distutils/unixccompiler.py	2015-07-08 17:21:05.742468485 +0200
+--- a/numpy/distutils/unixccompiler.py	2015-02-01 17:38:21.000000000 +0100
++++ b/numpy/distutils/unixccompiler.py	2015-07-08 17:21:05.742468485 +0200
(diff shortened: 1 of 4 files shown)


inclement:
This patch seems good in principle, but I had some issues with it. It's possible that they're related to how I tried it though (with the new p4a toolchain):

- I don't think the patch files were quite the right format, though I'm not sure how this works - I had to add the a/ b/ thing in the diff names and change the filenames they referred to.
- Setting the ar executable to $AR didn't work for me, I had to set it directly, even accessing the env var with python didn't work
- The numpy build still doesn't work for me (on my main system - it works in a VM with nothing else installed) - maybe it's another similar issue.

I'll keep testing, or this seems good to merge if someone else can confirm it works.

Can you or anyone else confirm that this patch definitely works as it stands?

dl1ksv:
If you are in the numpy build directory , you can apply the patches witth
patch -p0 
like this is done in the modified numpy/recipes.sh

I'm working on recipes for gnuradio which requires numpy.
In this context I tested the patched version of numpy and it works.

inclement:
I got it working (including further fixes) using the patch set at https://github.com/kivy/python-for-android/tree/revamp/pythonforandroid/recipes/numpy/patches, which includes the one from this issue (but a modified patch file) plus another to stop numpy trying to pull in system libraries.

I still haven't tested this specific PR, but it looks good to me if someone else (or me when I get time) can run the build to check.

dl1ksv:
It cleanly compiles with this patch set.

inclement:
I didn't merge this only because I'm not sure about hardcoding 'arm-linux-androideabi-ar' - can anyone on OS X comment on whether this is the correct binary name there?

kunguz:
This one worked like a charm! Thanks @dl1ksv !

inclement:
I've merged this into the old_toolchain branch. Thanks for the fixes, and sorry for the delay with it - I'm still worried about this not working on osx, but since nobody ever commented about it let's just fix it for everyone on linux at least (and maybe it works on osx anyway).

Vivaldo-Roque:
Sorry I'm newbie how do I apply this patch?

AndreMiras:
You ask on support channel https://github.com/kivy/python-for-android#support

Vivaldo-Roque:
No, but I'm going to ask now. thank you.

PR #415, opened 2015-07-11
Update greenlet/recipe.sh

jitendranair:
Remove stray URL from bash script (greenlet/recipe.sh).


inclement:
I've merged this manually to the old_toolchain branch. Thanks for the fix!

PR #416, opened 2015-07-12
Fixed version number logic for non-int versions

Cheaterman:
build.py would fail here on versions such as 0.1.1-r4. This fixed it for me - I also checked the resulting string in AndroidManifest, it seems proper.


inclement:
Seems good, thanks.

PR #418, opened 2015-07-15
Add support for KivyMD

mixedCase:
Adds the recipe to compile KivyMD, a set of Kivy widgets inspired by Google's Material Design.

Repo is here: https://github.com/kivymd/kivymd


akshayaurora:
You don't need a recipe for pure python modules. just put them in your project or set them up at pip and mention them in bulldozer.spec .

PR #420, opened 2015-07-25
Replace hardcoded "-jX" for make

thopiekar:
I've noticed that "make -j5" was set globally when push_arm was done.
There might be other places were -j5 is missing now but I think they can be fixed very quickly.


thopiekar:
Just added now -j$MAKE_JOBS to all make commands (untested but beleave it works) and replaced all hardcoded -jX options with -j$MAKE_JOBS

distribute.sh
```diff
@@ -93,6 +93,11 @@ case $OSTYPE in
 		;;
 esac
 
+if [ -f /proc/cpuinfo ]; then
```
kived: Maybe this should check if `MAKE_JOBS` is already set, so you can tell it how many jobs to use? Like `if [ -z "$MAKE_JOBS" -a -f /proc/cpuinfo ]; then` ...

thopiekar: Didn't think about this situation. I'm not so good in bash just found a variation of this on the web and rewritten it.
Would it be a problem to merge the PR and you could make this tweak?

inclement:
I guess the only reason this wasn't merged was kived's observation, but it still seems like an improvement so I'll merge it anyway. Sorry for the delay.

PR #422, opened 2015-07-26
[FIX] AttributeError: 'AnsiCodes' object has no attribute 'LIGHTBLUE_EX'

alanjds:
"colorama<0.3" have no LIGHTBLUE_EX. Fixes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ setup(name='python-for-android',
       author_email='kivy-dev@googlegroups.com',
       url='https://github.com/inclement/python-for-android-revamp', 
       license='MIT', 
-      install_requires=['appdirs', 'colorama', 'sh', 'jinja2', 'argparse'],
+      install_requires=['appdirs', 'colorama>0.3', 'sh', 'jinja2', 'argparse'],
       entry_points={
           'console_scripts': [
               'python-for-android = pythonforandroid.toolchain:ToolchainCL',


inclement:
Nice fix, thanks.

Giszmo:
if this patch fixes the error in the issue name, then what is it causing here?

    python-for-android$ grep col ./setup.py 
          install_requires=['appdirs', 'colorama>0.3', 'sh', 'jinja2', 'argparse'],
    python-for-android$ sudo python setup.py install

    ...

    Finished processing dependencies for python-for-android==0.3
    $ p4a recipes
    [INFO]:    This python-for-android revamp is an experimental alpha release!
    [INFO]:    It should work (mostly), but you may experience missing features or bugs.
      File "/usr/local/bin/p4a", line 9, in <module>
        load_entry_point('python-for-android==0.3', 'console_scripts', 'p4a')()
      File "/usr/local/lib/python2.7/dist-packages/python_for_android-0.3-py2.7.egg/pythonforandroid/toolchain.py", line 2909, in main
        ToolchainCL()
      File "/usr/local/lib/python2.7/dist-packages/python_for_android-0.3-py2.7.egg/pythonforandroid/toolchain.py", line 2534, in __init__
        getattr(self, args.command)(unknown)
      File "/usr/local/lib/python2.7/dist-packages/python_for_android-0.3-py2.7.egg/pythonforandroid/toolchain.py", line 2579, in recipes
        version=version))
    AttributeError: 'AnsiCodes' object has no attribute 'LIGHTBLUE_EX'

Note that I did first try to install p4a via `sudo pip install git+https://github.com/kivy/python-for-android.git`.

inclement:
Do you definitely have colorama>0.3 actually installed?

Giszmo:
Well ... probably not. Hadn't checked. I ran the scripts according to [how to install p4a](http://python-for-android.readthedocs.org/en/latest/quickstart/) and this is where it errored out.

... ok, turns out I do have it installed:

    $ sudo aptitude show python-colorama
    Package: python-colorama                 
    State: installed
    Automatically installed: yes
    Version: 0.3.2-1

I did now also install `python3-colorama (0.3.2-1)` just in case p4a does use python 3 but I get the same error.

inclement:
What colours does it actually have? I'm not sure how best to check, but it should be clear in e.g. the `__dict__` of the Style, Fore and Back objects.

kak2zz:
You can install colorama from https://github.com/tartley/colorama. In my case this solved the problem

normoes:
This also worked for me.

kived:
This PR is incorrect. Those colors were renamed (ex. `LIGHTBLUE` -> `LIGHTBLUE_EX`) in version 0.3.3, not 0.3. Changing the dep to `colorama>=0.3.3` should work.

cvergari:
Hi, I am getting the same error with python-for-android 0.5.3, Pyton 2.7.13 and colorama 0.3.9.

This is my command line:
`p4a apk --private /mnt/d/Dropbox/Python/myproject --arch=x86 --ndk-dir "/mnt/e/Documenti/Android_Projects/crystax-ndk-10.3.2" --package=org.test.myproject --name="myproject WebView" --version 0.1 --bootstrap=webview --requirements=python3crystax,flask,beautifulsoup4,python3 --permission INTERNET --port=5000`

Any ideas?

inclement:
@cvergari Are you sure you only have 0.3.9 installed, and that p4a is using it? I could not reproduce this error with 0.3.9. I'm not sure it should even be possible, as colorama's internal structure has changed such that Fore, Style etc. are no longer AnsiCodes objects but would appear as AnsiFore etc.

cvergari:
@inclement Yes I have 0.3.9 installed, or at least it correctly appears in pip list. However, I think I simply broke my whole system by upgrading Python to 2.7.13, then upgrading colorama in python3, then in python2, then doing only god knows what.
Starting from scratch!

alanjds:
@Giszmo @cvergari: You could get a more sane ground using a virtualenv. Mixing the system Python with SO-Installed packages and pip-installed packages is not always good. I would advise to `aptitude purge python-colorama` then `pip uninstall colorama` TWICE, just to see a "Not installed" error message.

Then install via `pip install colorama -U` and see if it solves for you.

With that working, you can bisect what versions works and what does not and update the `setup.py`

cvergari:
@alanjds I started using virtualenv after that. I am new to linux and I was waiting for a crisis just like this one to find the motivation to learn virtualenv.
Anyway, after having prepared a virtual environment for p4a I found out that p4a cannot be run from within it because p4a uses virtualenv itself, and environments cannot be nested! Or did I misunderstand something?

alanjds:
Strange. I doubt that because there is no "nesting" of venvs. You can activate one from inside another with no problem. The new "activation" effectively ignores the old one.

But maybe you need to install the "virtualenv" package on the "outer" virtualenv too.

cvergari:
I receive this message when running from virtualenv:

`[ERROR]:   virtualenv is needed to install pure-Python modules, but
[ERROR]:   virtualenv does not support nesting, and you are running
[ERROR]:   python-for-android in one. Please run p4a outside of a
[ERROR]:   virtualenv instead.`

I tried installing virtualenv within the outer one, but get the same error.

alanjds:
This is weird. I will not doubt the author of this line, but is very strange.

Anyway, if you need to install a module, usually prefer via `pip` instead of `aptitude`. 

Please try to uninstall from all sources and install again via `pip install -U`.

cvergari:
@alanjds I always use pip when packages are available on pypi!

alanjds:
:( So I am out of ideas for now

cvergari:
No worries! After starting from scratch, colorama is not a problem anymore. P4a is still not working...but I'll bang my head against it some more on my own before asking for help here. Thanks guys!

alanjds:
Came to the IRC if needing: 

Server: irc.freenode.net
Port: 6667, 6697 (SSL only)
Channel: #kivy

or Web: https://webchat.freenode.net/

inclement:
The virtualenv nesting used to cause an error. Possibly it doesn't any more. Updating this would be welcome if anyone can confirm it.

PR #428, opened 2015-08-06
Converted --minsdk option input to int

inclement:
This should fix a python3 error where this value is compared with an int in the AndroidManifest.xml template - currently it's a str vs int comparison which only works in python2.


akshayaurora:
:+1:

Issue #431, opened 2015-08-16
SQLite gets compiled without Full Text Search (FTS3) support


matphysrin:
Hello! I'm trying to make my classical music database app run on Android, but I'm struggling with the full text search (FTS) support in the `sqlite3` library. On my Mac and Ubuntu boxes, as well as on Android using Python with SL4A, SQLite is by default compiled with the FTS support turned on. Not so within py4a. As it is a very handy feature and I'm not aware of any drawbacks of turning the support on, I believe it should be allowed in py4a as well.

According to [the official SQLite documentation](https://www.sqlite.org/fts3.html#section_2) the FTS support gets turned on by setting the preprocessor flags `-DSQLITE_ENABLE_FTS3 -DSQLITE_ENABLE_FTS3_PARENTHESIS` at compilation.

According to [this](http://stackoverflow.com/questions/13323655/controlling-compiler-flags-in-the-android-ndk) and [this](http://stackoverflow.com/questions/16473845/android-ndk-how-to-specify-different-cflags-to-different-platform-using-androi) sites, I've tried myself adding these flags into the `python-for-android/src/jni/sqlite3/Android.mk` file, by adding the lines

```
APP_CFLAGS += -DSQLITE_ENABLE_FTS3
APP_CFLAGS += -DSQLITE_ENABLE_FTS3_PARENTHESIS
```

or the same with `LOCAL_CFLAGS` instead of `APP_CFLAGS`. With no success so far, though. Apparently I don't really see deeper under the hood to be able to make it work.

In case you agree that the FTS support turned on is desirable, could you please look into this? Thanks in advance!
##

inclement:
Haven't looked into this but just a comment: have you checked the sl4a source (I think it's OSS?)? That might be a useful check to see how they set the CFLAGS. If it's the same as you (which does look reasonable), perhaps there is another issue in what you've done.

matphysrin:
Thanks for the prompt reply! No, that didn't occur to me, I shall have a look into the SL4A code.

AndreMiras:
I think this was enabled long ago with https://github.com/kivy/python-for-android/pull/672 and https://github.com/kivy/python-for-android/pull/936

PR #432, opened 2015-08-16
Added flags to support SQLite full text search (FTS) at compilation

matphysrin:
In the SQLite (sqlite3) library, the full text search (FTS3) is not turned on at compilation. This commit turns it on by default. Please see [Issue #431](https://github.com/kivy/python-for-android/issues/431). I've tested these flags using OS X Yosemite 10.10.4 and Android 4.4.2.


inclement:
Sorry for the delay in replying about this. I'm not familiar with database technicalities, do you know if there's any significant downside likely to come with enabling this, like a larger binary size? Assuming not, I'll be happy to merge it.

matphysrin:
Sorry for the delay, this time on my side. I'm no expert either, but I tried myself compiling the library without and with FTS3 enabled and the resulting size of the binary is 774K and 912K, respectively. No idea whether you may consider this a significant size increase. FTS3 is enabled by default when using autoconf of sqlite. Consequently, most of the sqlite3 distributions around have it enabled. I can't imagine it could cause any significant drag on performance for non-full-text queries, as FTS3 is AFAIK a completely separate subsystem (virtual tables).

AndreMiras:
Thank you @matphysrin for improving p4a. I'm sorry your pull request didn't make it to the tree.
The good news is full text search FTS3 has been enabled since then in https://github.com/kivy/python-for-android/pull/672 and even updated to FTS4 in https://github.com/kivy/python-for-android/pull/936.

PR #433, opened 2015-08-18
Updated pylibpd recipe.

Stocarson:
The old version of the patch refers to wrong line numbers and fails to patch the setup.py file.
I've updated the patch file creating a diff with the github repo version of setup.py.
Now the recipe uses git clone --recursive for downloading libpd so we can include all the submodules in one step.
The libpd makefile has been patched to be coherent with the pylibpd patch.


inclement:
Thanks for the contribution. Just to be clear, the [revamp] tag refers to the new revamp branch which is a somewhat different rewrite of python-for-android, but your changes are for the master branch, so I've removed the tag. We'll include your changes in the revamp as and when the recipes are ported, so don't worry about them being missed.

Stocarson:
Ok, thanks for the informations, have a nice day!

inclement:
I've merged this manually to the old toolchain. Thanks for the contribution!

PR #437, opened 2015-09-01
[revamp] Correcting check for libtool

thopiekar:
It is correct that we need libtool, but the executable is called libtoolize. Therefore sh.which(executable) won't work in this case. I renamed the variable "tool" to "executable", so other developers won't get confused and also add package names here.


inclement:
Great, thanks!

PR #441, opened 2015-09-02
[revamp] (WIP) Adding possibility to collect multiple recipes in one {recipe}/__init__.py

thopiekar:
* In detail this means that if you have multiple recipe in one file you'll need to create a tuple with all recipe objects and create a variable called 'recipes' owning that tuple. If there is only one recipe just leave it as recipe = *Recipe*()

! Will probably need further testing !

### DIFF
--- a/pythonforandroid/recipes/pyjnius/__init__.py
+++ b/pythonforandroid/recipes/pyjnius/__init__.py
@@ -11,6 +11,11 @@ class PyjniusRecipe(CythonRecipe):
     name = 'pyjnius'
     depends = ['python2', ('sdl2', 'sdl')]
     site_packages_name = 'jnius'
+
+    def should_build(self):
+        # The last step in postbuild_arch() is to copy the java_classes - beleave the recipe has finished there
+        return not exists(join(self.ctx.javaclass_dir, 'jnius', 'src', 'org'))
+
     def prebuild_arch(self, arch):
         super(PyjniusRecipe, self).prebuild_arch(arch)
         if 'sdl2' in self.ctx.recipe_build_order:
--- a/pythonforandroid/recipes/pyjnius/__init__.py
+++ b/pythonforandroid/recipes/pyjnius/__init__.py
@@ -14,7 +14,7 @@ class PyjniusRecipe(CythonRecipe):
 
     def should_build(self):
         # The last step in postbuild_arch() is to copy the java_classes - beleave the recipe has finished there
-        return not exists(join(self.ctx.javaclass_dir, 'jnius', 'src', 'org'))
+        return not exists(join(self.ctx.javaclass_dir, 'org', 'jnius'))
 
     def prebuild_arch(self, arch):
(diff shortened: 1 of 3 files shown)


thopiekar:
Ran into a lot of breakages. Will need more work :disappointed:

thopiekar:
Well, there are some places in the toolchain.py code, which I didn't get. There is much more work that is needed to be done. However, merging this code could be a good beginning :grinning:

thopiekar:
Btw. can revert 72f2466c57202464af1ff72e7d4978dd0cb6df85 if you like. I personally didn't see a huge difference after adding this line, but I'm sure by looking at postbuild_arch that it should skip the build when the java classes are installed/copied.

inclement:
The javaclass check isn't necessary anyway because Pyjnius is a PythonRecipe, so the default check of seeing if it's installed in the python install ation should already work.

thopiekar:
Ok, it just seemed to me that pyjnius is rebuilding everytime as long as the distribution build wasn't complete. Will revert it as soon as possible and ping you here again.

inclement:
Also I have to ask, what's the actual use case for multiple recipes in one file? Right now it seems like it just makes it more complicated to access recipes as they are currently distributed. I had in mind some ways to load recipes in other fashions (so that get_recipe is really a name index that happens to be looking in named folders for builtin recipes), but didn't think about this.

thopiekar:
Well, the my idea was to merge all python recipes (python2, python3, hostpython2, hostpython3) into one file (eg. recipes/python/__init__.py) or as a package (eg. recipes/python/{__init__.py, hostpython.py, python.py}). I guess it would create a better overview about all recipes, while sorting them related to a topic/group.

inclement:
I have some reservations about this. On the technical side it is simply a very different way to do things  that also needs some other changes - you need to not only support multiple recipes in one file, but give them all a name manually and probably load them in a key-value store when toolchain.py is run, because otherwise you don't know what recipes are available and can't access a specific one - for instance, your `hostpython_recipe = Recipe.get_recipes('hostpython2', self.ctx)[0]` won't work once there are multiple recipes in one file, because you can't hardcode the [0].

The current way of doing things is motivated largely by the idea that actually there aren't very many related recipes (and even e.g. the python recipes aren't that interrelated), that they are rarely modified, and that they really are recipes philosophically rather than python objects. That is, they will largely be interacted with by quickly looking at individual recipes via the filesystem, not by loading them in different ways with code, so making the overall set obvious is more valuable than creating conceptual relationships for developers. At least, these are my motivations for keeping it simple like this, I can't speak for sure why tito chose this structure.

I'm not necessarily averse to going your way and constructing a big list of recipes when toolchain.py is run (this would also make it obvious how to support user-loaded recipes, whereas right now at least one new thing is necessary to do that), but I'm not convinced right now it's actually a better way. I guess  I suspect I'm also instinctively conservative about it, so I'm open to changing my mind.

PR #442, opened 2015-09-06
[revamp] toolchain: Update DEFAULT_ANDROID_API to 15

thopiekar:
ANDROID_API = 14 is obsolete according to the Android SDK Manager.
Thats why I would suggest to update the DEFAULT_ANDROID_API at least to the next supported version.


inclement:
Sounds good to me, thanks.

PR #443, opened 2015-09-06
[revamp] Adding direct git support to toolchain.py

thopiekar:
While working on the Qt5 recipe, I wanted to download the latest code from git. I know there is the possibility to download a tar or zip archive, but Qt5 uses submodules for their branch.
These won't be packaged by GitHub, so I started to add git support.

While coding I found out that a function, which should extract data from an archive isn't used anymore. But I can be wrong: https://github.com/kivy/python-for-android/compare/kivy:revamp...thopiekar:toolchain_git_support?expand=1#diff-9386f079c34547ba8b06a93309463a72L1365

### DIFF
--- a/pythonforandroid/toolchain.py
+++ b/pythonforandroid/toolchain.py
@@ -1390,8 +1390,10 @@ class Recipe(object):
             return target
         elif parsed_url.scheme in ('git',):
             if os.path.isdir(target):
-                with current_directory(self.get_build_dir(arch.arch)):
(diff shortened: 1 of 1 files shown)


inclement:
This seems good, is it ready to merge?

thopiekar:
Yes, it is ready, but only supports git:// URLs. Will add git+http:// when it is needed to support the usual http://[...].git URLs.

inclement:
Great, merged and thanks again.

PR #446, opened 2015-09-11
[revamp] Generate project.properties dynamically for bootstraps

### LINKED ISSUE #425: [revamp] --android_api is ignored on SDL2 bootstrap
*.properties files needed to be set manually for now. Defaults to android-12

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25168143-revamp-android_api-is-ignored-on-sdl2-bootstrap?utm_campaign=plugin&utm_content=tracker%2F11371&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F11371&utm_medium=issues&utm_source=github).
</bountysource-plugin>

str4d:
Closes #425.


inclement:
Thanks for the contribution! I thought I tried just deleting the files and had some problem in the past, but I'll test it and it will be great if it works.

str4d:
A `project.properties` containing `target=android-#` is still necessary, but this PR generates it now instead. But the `default.properties` is unnecessary, it is an older file that was deprecated and renamed to `project.properties`.

inclement:
Great, I had to make a small fix for the pygame bootstrap to copy the project.properties from the right place, but it seems to work nicely now. Thanks!

PR #447, opened 2015-09-11
[revamp] Undefine HAVE_FTIME for Android API > 19

### LINKED ISSUE #261: compile error: sys/timeb.h not found
It seems that in the android-ndk-r10, the file timeb.h is removed, which leads to a compiling failure.
Has no idea of how to fix it besides downloading a r9 version of ndk...

P.S.: And in r10, the API dir has been changed into Android-L from Android-14 or so, which I made a link to solve the problem.

Thanks for any suggestion.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/3936418-compile-error-sys-timeb-h-not-found?utm_campaign=plugin&utm_content=tracker%2F11371&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F11371&utm_medium=issues&utm_source=github).
</bountysource-plugin>

str4d:
The `sys/timeb.h` header was removed in Android 21, because this header and the
`ftime` function were removed from the POSIX 2008 specification.
(Info reference: https://rbcommons.com/s/OpenH264/r/683/diff/1/ )

Closes #261, and replaces #410. The only difference is that in #410 the `#undef` is surrounded by an `#ifdef ANDROID`, which I didn't think was necessary (because this is only ever going to be patched for Android, because... `python-for-android`), but I can add it if desired.

### DIFF
--- a/pythonforandroid/recipes/python2/__init__.py
+++ b/pythonforandroid/recipes/python2/__init__.py
@@ -11,7 +11,7 @@ class Python2Recipe(Recipe):
     url = 'http://python.org/ftp/python/{version}/Python-{version}.tar.bz2'
     name = 'python2'
 
-    depends = ['hostpython2']  
+    depends = ['hostpython2']
     conflicts = ['python3']
 
     def prebuild_armeabi(self):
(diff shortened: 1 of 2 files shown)


inclement:
Super, thanks.

Issue #450, opened 2015-09-12
[revamp] Unpack archives that don't list their root directory


str4d:
The code in `Recipe.unpack()` gets the first item on an archive's list as the root directory, and renames it if necessary. Archives like http://python.org/ftp/python/2.7.2/Python-2.7.2.tar.bz2 show their root directory as the first entry of `tar tjf`. But others like https://pypi.python.org/packages/source/T/Twisted/Twisted-15.4.0.tar.bz2 do not, and so instead the first file in the list gets renamed, causing the later build stage to fail (because the expected directory doesn't exist).

inclement:
python-for-android uses a recipe with http://twistedmatrix.com/Releases/Twisted/15.2/Twisted-15.2.1.tar.bz2 as the source, which is a more politely packaged archive. I recognise that the archive unpacking should probably be made better and some code changes for that would be fine, but does using this source solve your problem?

str4d:
No, it has the same issue here :(

inclement:
Are you sure? The top level of the archive is a directory, how does it fail now?

str4d:
So was the PyPI one. But what is happening, is that this top level is not shown in the list. I don't understand why, but it has to be something in the way the archive was produced.

```shell
$ tar tjf packages/python2/Python-2.7.2.tar.bz2
Python-2.7.2/
Python-2.7.2/PC/
Python-2.7.2/PC/py.ico
Python-2.7.2/PC/testpy.py
Python-2.7.2/PC/pyconfig.h
Python-2.7.2/PC/_msi.c
Python-2.7.2/PC/pyc.ico
Python-2.7.2/PC/icons.mak
Python-2.7.2/PC/frozen_dllmain.c
Python-2.7.2/PC/winsound.c
Python-2.7.2/PC/os2emx/
Python-2.7.2/PC/os2emx/pyconfig.h
...
$ tar tjf packages/twisted/Twisted-15.2.1.tar.bz2
Twisted-15.2.1/bin/_preamble.py
Twisted-15.2.1/bin/conch/
Twisted-15.2.1/bin/conch/cftp
Twisted-15.2.1/bin/conch/ckeygen
Twisted-15.2.1/bin/conch/conch
Twisted-15.2.1/bin/conch/tkconch
Twisted-15.2.1/bin/lore/
Twisted-15.2.1/bin/lore/lore
Twisted-15.2.1/bin/mail/
Twisted-15.2.1/bin/mail/mailmail
[...]

PR #457, opened 2015-09-14
Recipes for setuptools, zope.interface and twisted

### LINKED ISSUE #454: [revamp] Twisted
I'm creating this issue because my meta-discussion doesn't belong in the other tickets I have opened. If it wasn't already obvious, my motivation is to get Twisted working with the revamp, so I can use it in a new project.

In #453 it was suggested that Twisted will require a recipe. I just tried the same process as #453 again, but using `--requirements setuptools,zope.interface,twisted` and with a Twisted `CompiledComponentsPythonRecipe` (but no recipes for `setuptools` or `zope.interface`). This fails (even after manually fixing the build env after #450 occurs) because the Python modules are installed after the recipes, and of course Twisted depends on the Python modules.

So I am currently stuck between #453 (preventing use of Python modules only), #449 (preventing use of a recipe depending on Python modules) and #451 (preventing use of recipes only). I'm not sure which is the "right way" to be doing this, which doesn't help either. Given that the deprecated Twisted recipe required compilation, I agree that Twisted probably requires a recipe, which implies that #453 is not the right way forward, but that still leaves me stuck :tired_face:

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/26650664-revamp-twisted?utm_campaign=plugin&utm_content=tracker%2F11371&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F11371&utm_medium=issues&utm_source=github).
</bountysource-plugin>

str4d:
Closes #454.

The build will fail until #450 is closed, but you can manually fix the filepath:

```
$ cd ~/.local/share/python-for-android/build/other_builds/twisted/armeabi
$ mv twisted Twisted-15.4.0/bin/_preamble.py
$ mv Twisted-15.4.0/ twisted
```

and then run the build again.

### DIFF
--- a/pythonforandroid/recipes/setuptools/__init__.py
+++ b/pythonforandroid/recipes/setuptools/__init__.py
@@ -1,0 +1,45 @@
+
+from pythonforandroid.toolchain import (
+    PythonRecipe,
+    Recipe,
+    current_directory,
+    info,
+    shprint,
+)
+from os.path import join
+import sh
+
+
+class SetuptoolsRecipe(PythonRecipe):
+    version = '18.3.1'
+    url = 'https://pypi.python.org/packages/source/s/setuptools/setuptools-{version}.tar.gz'
+
+    depends = ['python2']
+
+    def install_python_package(self, name=None, env=None, is_dir=True):
+        arch = self.filtered_archs[0]
+        if name is None:
+            name = self.name
+        if env is None:
+            env = self.get_recipe_env(arch)
+
+        info('Installing {} into site-packages'.format(self.name))
+
+        with current_directory(self.get_build_dir(arch.arch)):
+            # Here we do *not* use the normal hostpython binary in the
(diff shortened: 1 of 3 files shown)


inclement:
I've added these recipes manually in a slightly different way, abstracting away the way hostpython is called to a new attribute of the PythonRecipe (though it has a terrible name that I will change), and an option to install in the hostpython build dir. This should hopefully make it easier to do the same thing with other recipes.

I didn't include your commit because of big merge conflicts thanks to my big rebase to get the revamp into the master branch, but I think I've realised how to do so and will cherry-pick it manually to keep it in the history.

Are you able to test if I got it right, and the new twisted etc recipes work? I checked that they build but don't have examples to test with. The problem with folder unpacking may also be solved thanks to another commit.

PR #458, opened 2015-09-14
Add --permission to SDL2 bootstraps

str4d:
A more complete setup might try to factor out the commonalities between the three bootstraps (`pygame`, `sdl2` and `sdl2python3`), but for now this at least allows permissions to be specified for SDL2 bootstraps.

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/build/build.py
+++ b/pythonforandroid/bootstraps/sdl2/build/build.py
@@ -248,7 +248,7 @@ def make_package(args):
 
     default_icon = 'templates/kivy-icon.png'
     shutil.copy(args.icon or default_icon, 'res/drawable/icon.png')
-    
+
     versioned_name = (args.name.replace(' ', '').replace('\'', '') +
                       '-' + args.version)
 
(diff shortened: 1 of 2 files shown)


inclement:
At one point I started out to factor out the common stuff using a build.py jinja2 template, but eventually decided it wasn't actually worth it since it seemed either very fiddly (factoring out in a fine grained way) or not really doing much (just factoring it into large blocks that were mostly duplicated anyway). I'm happy to do it if someone wants to, but it made me think it's not a priority, unless maybe we somehow get several very similar bootstraps.

Anyway, seems good, I'll check it later.

inclement:
It seems you don't actually use the --permissions optio for anything, so they can be specified but are ignored. Is this correct?

str4d:
Incorrect. The `--permissions` options are [rendered](https://github.com/str4d/python-for-android/blob/permissions/pythonforandroid/bootstraps/sdl2/build/build.py#L262) in [`AndroidManifest.xml.tmpl`](https://github.com/str4d/python-for-android/blob/permissions/pythonforandroid/bootstraps/sdl2/build/templates/AndroidManifest.xml.tmpl), and must be specified if any Android permissions are required. I made this PR because, to get Twisted to run, the `INTERNET` permission was required.

inclement:
Oh, okay, I didn't remember copying it. Thanks.

str4d:
Yeah, the code to use permissions is all there, all that was missing from SDL2 was the command-line arguments.

inclement:
Super then, thanks.

PR #471, opened 2015-10-06
Update toolchain.py

ibobalo:
tar.bz2 unpacking, twisted recipe faults with "OSError: [Errno 20] Not a directory" fixed


inclement:
Crude, but effective, I guess. Thanks!

PR #472, opened 2015-10-10
add flask recipe and flask depend recipes.

lc-soft:


inclement:
Looks good, thanks. Just to check, I assume you checked that openssl is working with the new version (just paranoid about this)?

Also, how are you using flask, is it as part of a kivy app? I'm thinking about making a standalone flask/django + webview bootstrap with the new toolchain and I'm interested to hear if it works on the python side with no issues.

lc-soft:
my app is used by kivy、flask and webview, like this:
simple

inclement:
Okay, great. Thanks for the contribution and sorry for the delay in merging.

PR #474, opened 2015-10-14
added .aar bundle (binary distribution of an Android Library Project)…

ibobalo:
… support in recepies


inclement:
Looks interesting. Can you give an example of using this, I'm not familiar with it.

ibobalo:
Hi. In my next commit i'l give vlc library videoplayer recipe. Testing now.
In general:
def build_arch...:
     shprint(sh.cp, '-a', 'libvlc-3.0.0.aar', self.ctx.aars_dir)
14 жовт. 2015 16:02 "Alexander Taylor" <notifications@github.com> пише:

inclement:
Super cool, I'll look forward to it.

I'll have to check the details, but as an initial comment how about abstracting the aar extraction to a method of Bootstrap, that way it doesn't have to be duplicated in each bootstrap class?

PR #480, opened 2015-10-27
VLC recepie, sdl2 bootstaps improvements

ibobalo:
added vlc videoplayer recipe (for both pygame and sdl2 bootstraps)
added sdl2 support for native widget integration (see kivy vlc example)
added sdl2 bootstrap metadata
bootstraps common code abstractions
improved sdl2 bootstrap command line compatibility with pygame bootstrap
toolchain support for aar libs added
toolchain logging output cosmetic


inclement:
Looks super cool from a quick glance, I'll hopefully have time to take a proper look at the weekend. Thanks!

inclement:
I see you've added some more commits, are you still working on this or is it ready to check and merge?

ibobalo:
now (8ae6758) it is ready to merge. from this point i start other task (usb/bt hardware joysticks/mouse/gamepads etc support for android)

ibobalo:
existing conflicts resolved. pls make a decision about this pull request to avoid future conflicts.

inclement:
I'm sorry for not getting back to you sooner. This looks good, and I expect to merge it, but I haven't been through it thoroughly yet (because it touches a few different things) - it's a priority to do in the next few days though. After my checks so far, I think I'll have a couple of small changes to suggest, but probably nothing major, and I'll confirm with details as soon as possible.

Thanks for the nice contribution :)

pythonforandroid/bootstraps/pygame/__init__.py
```diff
@@ -40,21 +43,14 @@ def run_distribute(self):
             info('Copying python distribution')
             hostpython = sh.Command(self.ctx.hostpython)
             # AND: This *doesn't* need to be in arm env?
-            shprint(hostpython, '-OO', '-m', 'compileall', self.ctx.get_python_install_dir())
+            shprint(hostpython, '-OO', '-m', 'compileall', self.ctx.get_python_install_dir(),
+                    _tail=10, _filterout="^Listing", _critical=True)
```
inclement: What do these extra arguments do? I thought they would be sh arguments, but couldn't find them.

ibobalo: its new args of shprint. for pretty log optputs.
_tail mean show only N last lines of stdout. stderr will out full. like
command args | tail 10
_filterout mean do not print stdout lines matching regexp. like command
args | grep -v -e "re"
also _filter available
and _critical mean stop processing immediate after error, and do not
re-raise exception.
all this changes tot effect commands execution (except critical), just
butify its printout

2015-11-27 0:40 GMT+02:00 Alexander Taylor <notifications@github.com>:

pythonforandroid/toolchain.py
```diff
@@ -1477,15 +1601,30 @@ def extract_source(self, source, cwd):
...
         build directory.
         """
         info("Applying patch {}".format(filename))
         filename = join(self.recipe_dir, filename)
-        # AND: get_build_dir shouldn't need to hardcode armeabi
-        sh.patch("-t", "-d", self.get_build_dir('armeabi'), "-p1", "-i", filename)
+        shprint(sh.patch, "-t", "-d", self.get_build_dir(arch), "-p1",
+                   "-i", filename, _tail=10)
+
+    def apply_all_patches(self, wildcard=join('patches','*.patch'), arch='armeabi'):
```
inclement: I understand the idea of this one, but I'm not sure about it - it could lead to fragile situations where things break if a new patch is added in the directory (e.g. during debug), and also only supports using the recipe dir. How would you feel about just dropping this function?

I can remove it on merge if you agree, there's no need for you to do it yourself if you don't want to mess with it.

inclement:
Would you mind removing the kivy-presplash image? I don't think we want a kivy/sdl2 logo like that.

pythonforandroid/bootstraps/pygame/__init__.py
inclement: Do you have a reason for limiting to the last 10 lines? If we're passing the output through to the user, don't we want to pass all of it (or, if debug is not enabled, none of it)?

ibobalo: 1 its ok to drop patchall
2 presplash it was just experiment, actually no presplash code for sdl2 so
it is my fault to commit unused image
3. some commands produce really long outputs > 1000 lines. it was a big
problem for me to find error reason.
actually, in current state of p4a there no stdout listing at all (even id
debug) - just exception traceback and shortened last line. so, 10 lines is
much better then zero ?
 anyway feel free to increase it to any reasonable number of lines you want
:)

2015-11-27 0:56 GMT+02:00 Alexander Taylor <notifications@github.com>:

pythonforandroid/toolchain.py
```diff
@@ -137,38 +149,67 @@ def shprint(command, *args, **kwargs):
...
+                msg = line.replace('\n', ' ').replace('\t', ' ').replace('\b', ' ').rstrip()
+                if msg:
+#                    if len(msg) > msg_width: msg = msg[:(msg_width - 3)] + '...'
+                    sys.stdout.write('{}\r{}{:<{width}}'.format(Err_Style.RESET_ALL, msg_hdr, shorten_string(msg, msg_width), width=msg_width))
+                    sys.stdout.flush()
+                    need_closing_newline = True
+            else:
+                logger.debug(''.join(['\t', line.rstrip()]))
+        if need_closing_newline: sys.stdout.write('{}\r{:>{width}}\r'.format(Style.RESET_ALL, ' ', width=(columns - 1)))
+    except sh.ErrorReturnCode, err:
```
inclement: Minor comment: this would be better as `except sh.ErrorReturnCode as err' - especially since I don't think python3 supports the comma syntax any more.

pythonforandroid/bootstraps/pygame/__init__.py
inclement: Okay, ignore my last reply, I understand better now and it seems good (though is there a reason for picking this tail number in particular?).

inclement:
Okay, I think I'm fine with everything. Let me do a couple more tests, but then (unless something goes wrong) I'll merge it.

ibobalo:
good luck. btw i almost done with next - support for hw joysticks (BT,
UsbHost etc)

2015-11-27 1:28 GMT+02:00 Alexander Taylor <notifications@github.com>:

inclement:
All done, thanks again!

PR #482, opened 2015-10-29
added the new patches and applied them in recipe.sh

squarrel:
Not sure if this has already been done by someone else. With this I can package an application including the numpy module successfully.


inclement:
I'm afraid I was mistaken about the previous PR, these patches were actually included in it, and I've now merged them as described at https://github.com/kivy/python-for-android/pull/414, so I'll close this one. Thanks for the contribution though.

PR #483, opened 2015-10-31
Fix keyboard height issue

SapienTech:
Use the print lines to debug. It seems that mActivity.getWindow()... changes rctx.top to a non-zero number, which causes the app to permanently shift upwards . To replicate this error, make an app that calls the keyboard. The hotfix I added was to set rctx.top to zero.

### DIFF
--- a/pythonforandroid/recipes/android/src/android/_android.pyx
+++ b/pythonforandroid/recipes/android/src/android/_android.pyx
@@ -190,8 +190,14 @@ if mActivity:
         @java_method('()V')
         def onGlobalLayout(self):
             rctx = Rect()
+            # print('rctx_bottom: {0}, top: {1}'.format(rctx.bottom, rctx.top))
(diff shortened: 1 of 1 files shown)


akshayaurora:
where is the issue you are trying to solve? detail the version of kivy and py4a being used in the issue.

inclement:
@akshayaurora As i remember, @SapienTech was finding that the keyboard code permanently offset the kivy window. I didn't look into the issue, but does the solution seem reasonable to you? I'm not sure if forcing top to 0 is correct on all devices.

SapienTech:
@akshayaurora @inclement kivy is v1.9.1-dev , Buildozer is 0.31dev  with the py4a it pulls by default. I agree with inclement that I'm not sure if this fix will work on all devices, but I have not heard any issues from the users of my app.

rafalo1333:
I can confirm that #544 can be solved with #483 merge. The fix by SapienTech is working very well, if the app is fullscreen or if the statusbar is shown.

PR #495, opened 2015-11-20
send notifications to stderr. use colorization on stdout or stderr only if a tty.

denys-duchier:
This patch sends log messages to stderr.  See e.g. the url below for why that's preferable:

http://www.jstorimer.com/blogs/workingwithcode/7766119-when-to-use-stderr-instead-of-stdout

basically stdout is for primary data, and stderr for ancillary data (erros, notifications, ...), and this separation plays well with combining processes using pipes.  It improves on my previous patch in that log info when calling p4a from buildozer will no longer be thrown away.

I also modified colorization, whether on stdout or stderr, to happen only when the corresponding stream is a tty.

### DIFF
--- a/pythonforandroid/toolchain.py
+++ b/pythonforandroid/toolchain.py
@@ -9,7 +9,7 @@ This tool intend to replace all the previous tools/ in shell script.
 from __future__ import print_function
 
 import sys
-from sys import stdout, platform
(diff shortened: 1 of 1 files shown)


inclement:
Great, thanks :)

PR #499, opened 2015-11-26
remove the print of BrokenRecipe.__init__ and turn it into a Recipe.info attribute

denys-duchier:
instead of unexpected pollution on stdout, we now get a nicely formatted recipe attribute.

### DIFF
--- a/pythonforandroid/recipes/brokenrecipe/__init__.py
+++ b/pythonforandroid/recipes/brokenrecipe/__init__.py
@@ -1,8 +1,6 @@
 from pythonforandroid.toolchain import Recipe
 
 class BrokenRecipe(Recipe):
-    def __init__(self):
-        print('This is a broken recipe, not a real one!')
-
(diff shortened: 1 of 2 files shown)


inclement:
This seems like overengineering - this is just a passing comment that a single recipe happens to include, and I don't think there's much general use for recipe information displayed at `__init__` time. Unless this would be widely useful, I think it would be better to just make the BrokenRecipe log a warning instead, or potentially remove it altogether.

inclement:
I'll close this for now since you didn't disagree, but feel free to raise it again if you want to discuss if this solution is better (or want to implement the alternative of deleting the recipe).

denys-duchier:
sorry, I completely missed your comment.  I agree with you.  I don't understand why there is a broken recipe in the first place.  I'd be much happier if it was simply removed.  my PR just tried to preserve what there was (in case there was a reason for that recipe that I could not see) while making it not misbehave.  should I submit a simpler PR to remove the broken recipe?

PR #503, opened 2015-11-29
Fixing just some more typos

dl1ksv:
I found just some more typos. The patch fixes those. But I'm not sure if this are all

### DIFF
--- a/pythonforandroid/toolchain.py
+++ b/pythonforandroid/toolchain.py
@@ -78,7 +78,7 @@ toolchain_dir = dirname(__file__)
 sys.path.insert(0, join(toolchain_dir, "tools", "external"))
 
 
-DEFAULT_ANDROID_API = 15
+DEFAULT_ANDROID_API = 16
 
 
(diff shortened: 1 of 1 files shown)


pythonforandroid/toolchain.py
```diff
@@ -78,7 +78,7 @@ def __getattr__(self, key):
 sys.path.insert(0, join(toolchain_dir, "tools", "external"))
 
 
-DEFAULT_ANDROID_API = 15
+DEFAULT_ANDROID_API = 16
```
inclement: I think 15 was chosen as the default because it was properly supported (the old option, 14, wasn't) - is this not the case any more? Why choose 16?

dl1ksv: Sorry this has nothing to do with the typo erros. It's just an adaption to my development situation.

inclement:
I think all of these changes have now been included in other commits. I'm sorry about that, and thank you anyway for the contribution.

PR #508, opened 2015-12-05
Added audiostream recipe

hottwaj:
I've included code to allow either sdl or sdl2 dependency.

However the library itself only links against SDL1 at the moment - the library needs to be changed to know when it is being linked against SDL2 (library names to link are slightly different)

### DIFF
--- a/pythonforandroid/recipes/audiostream/__init__.py
+++ b/pythonforandroid/recipes/audiostream/__init__.py
@@ -18,11 +18,12 @@ class AudiostreamRecipe(CythonRecipe):
         elif 'sdl2' in self.ctx.recipe_build_order:
             sdl_include = 'SDL'
             sdl_mixer_include = 'SDL2_mixer'
+            
+            #note: audiostream library is not yet able to judge whether it is being used with sdl or with sdl2.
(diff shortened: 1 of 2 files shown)


inclement:
Nice, thanks. Could you make the recipe actually fail (I guess print a message and throw an exception or just exit()) if it's built with sdl2, to make it clear to the user what the problem is? I think that would be the best way for now.

hottwaj:
Sure - just added that.  Do you know of a way of passing an option to setup.pu (distutils) so that we can tell audiostream whether to do an SDL or SDL2 link?  If there is a mechanism for passing that info I can get the SDL2 route finished off.  Thanks!

inclement:
Thanks, merged.

I'm not sure what's the best way to pass things to setup.py. One option is an environment variable (and to modify the setup.py to check it) - this is what e.g. kivy does. A cruder option is what I did with pyjnius, to add a patch that changes the behaviour, but this is less desirable.

hottwaj:
OK thanks, an environment variable would pretty easy to set up.

I'll add that in when I get back to testing what I'm working on with the sdl2 bootstrap. 

Cheers!

PR #513, opened 2015-12-08
add new error message and improve existing

kived:
Shows the path to the `ndk_platform` when it is not found, and shows a final error if `prepare_build_environment` is not `ok`.


inclement:
Thanks

PR #517, opened 2015-12-09
fix hard coded python2 path in sdl2 Android.mk

kived:
Makes things compile when you have openssl in your requirements, making the python2 directory `python2-openssl` instead of `python2`.


inclement:
Thanks

Issue #515, opened 2015-12-08
Inconsistent dependency graph behaviour


hottwaj:
Seems to fail when there is no actual dependency issue?  See a handful of test cases below...

e.g. here I start with bootstrap sdl2 and requirements sdl2,python2,kivy,audiostream

```
python -m pythonforandroid.toolchain create --bootstrap=sdl2 --requirements=sdl2,python2,kivy,audiostream --dist_name=sdl2_kivy

...snip...
[INFO]:    Bootstrap requires recipes ['sdl2']
[INFO]:    Loaded recipe sdl2 (depends on ['python2', 'sdl2_image', 'sdl2_mixer', 'sdl2_ttf'], conflicts ['sdl', 'pygame', 'pygame_bootstrap_components'])
[INFO]:    Loaded recipe kivy (depends on [('sdl2', 'pygame'), 'pyjnius'])
[INFO]:    Loaded recipe audiostream (depends on ['python2', ('sdl', 'sdl2'), 'pyjnius'])
[INFO]:    Loaded recipe python2 (depends on ['hostpython2'], conflicts ['python3'])
[INFO]:    Loaded recipe sdl2_image (depends on [])
[INFO]:    Loaded recipe sdl2_mixer (depends on [])
[INFO]:    Loaded recipe sdl2_ttf (depends on [])
[INFO]:    Loaded recipe pyjnius (depends on ['python2', ('sdl2', 'sdl'), 'six'])
[INFO]:    Loaded recipe hostpython2 (depends on [], conflicts ['hostpython3'])
[INFO]:    Loaded recipe six (depends on ['python2'])
[WARNING]: Didn't find any valid dependency graphs, exiting.
```

However that set of packages should be valid?

If I remove "audiostream" graph checking succeeds and compilation begins.

If instead I remove "kivy" graph checking succeeds and compilation begins 

If instead I change the order of requirements to this (i.e. swap order of sdl2 and python2): 
python2,sdl2,kivy,audiostream

then I get:

```
[INFO]:    Bootstrap requires recipes ['sdl2']
[INFO]:    Loaded recipe python2 (depends on ['hostpython2'], conflicts ['python3'])
[INFO]:    Loaded recipe kivy (depends on [('sdl2', 'pygame'), 'pyjnius'])
[INFO]:    Loaded recipe audiostream (depends on ['python2', ('sdl', 'sdl2'), 'pyjnius'])
[...]

PR #521, opened 2015-12-10
Fix for #515 - dependency graph issue

hottwaj:
Current code for removing redundant dependency graphs has a bug in it where graph[i] was compared against itself and then removed (as it appears equal to itself).  This caused #515.  In this PR I've rearranged the loop in the function for removing redundant graphs to prevent this issue


inclement:
Thanks!

PR #522, opened 2015-12-10
easier patching for recipes

kived:
Simplifies patching for recipes by removing boilerplate.

### DIFF
--- a/pythonforandroid/patching.py
+++ b/pythonforandroid/patching.py
@@ -1,0 +1,65 @@
+from os import uname
+
+
+def check_all(*callables):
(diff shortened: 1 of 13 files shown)


kived:
Note that the recipe for vispy is broken - it attempts to apply two patches which do not exist. I kept it this way when converting the recipe, as I don't know what the patches do, or whether they aren't supposed to be in the recipe or the patch files just weren't added to a commit.

Also, I didn't touch any of the python3-related recipes.

inclement:
Thanks (and also for the others).

PR #537, opened 2015-12-14
add custom recipe support [needs py3 testing]

kived:
Adds two additional locations to find recipes.

First, a local recipes folder is checked. This defaults to `./p4a-recipes`, and can be set via the command line option `--local-recipes`/`--local_recipes`.

Next, a user-local folder is checked in the user-local storage (i.e. ~/.local/share/python-for-android/recipes).

Finally, the standard p4a recipes will be checked.

Note that while this works on py2, it will need to be tested on 3.3/3.4 AND 3.5. 3.5 will use the newer `spec_from_file_location` and `module_from_spec` methods in `importlib.util`, while 3.3/3.4 will use the older (deprecated?) `SourceFileLoader`. py2 just uses the simple `imp.load_source`.

### DIFF
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -411,6 +411,8 @@ class Context(object):
         self.toolchain_prefix = None
         self.toolchain_version = None
 
+        self.local_recipes = None
+
         # root of the toolchain
         self.setup_dirs()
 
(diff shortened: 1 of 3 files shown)


inclement:
I didn't actually test on python3, but since it seems python2 should work fine I guess let's not dawdle and fix it if it turns out wrong.

PR #546, opened 2016-01-04
Add python3 support

inclement:
Not for merging yet, but the PR is at a working state with the SDL2 bootstrap, so people may be interested in seeing how it works.

You need to use the CrystaX NDK 10.3.0+ to use python3.


inclement:
I think this is basically ready to merge (though still doing some last tests and doc stuff) - if anyone wants to try it, now would be a good time.

If it seems to work, I'd like to merge it sooner rather than later even though features are missing, so that people can start trying it. It will be clearly marked as experimental.

PR #555, opened 2016-01-08
add copy libs option

kived:
Adds copy libs support back to revamp, similar to old toolchain. This copies libraries and shared objects to the APK rather than using biglink to add them to libpymodules.so, which is required for some use cases where C extension modules have naming conflicts.

### DIFF
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -1,6 +1,8 @@
+from __future__ import print_function
+
 from os.path import (join, realpath, dirname, expanduser, exists,
-                     split)
-from os import environ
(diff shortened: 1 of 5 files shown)


inclement:
I've allowed merge conflicts to arise here, but I can do the merge manually.

Is it necessary to avoid the biglink this way? With python3 I just skipped the biglink entirely, and copying the .so files in their in-place relative positions seemed to work (with python modules), but I didn't fully follow your changes and I probably don't understand the subtlety of the linking anyway.

On that note, do you think the python3 method should also be switched to something like this?

PR #557, opened 2016-01-08
ignore deps when installing via pip

kived:
This avoids pip trying to overwrite packages which were installed via recipes (and sometimes throwing an error due to the package already existing in the destination, depending on pip version). This does mean you must specify every dependency package.

### DIFF
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -580,7 +580,7 @@ def run_pymodules_install(ctx, modules):
         # It works but should be replaced with something better
         shprint(sh.bash, '-c', (
             "source venv/bin/activate && env CC=/bin/false CXX=/bin/false "
-            "PYTHONPATH={0} pip install --target '{0}' -r requirements.txt"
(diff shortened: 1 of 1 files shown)


inclement:
I was originally worried that having to work out the deps could itself be annoying for users. I'll merge it, but do you think there's a nice way to somehow make it toggleable, or if that's a good idea?

kived:
It's difficult, cause pip just kinda sucks. :( No API (well, at least no stable documented API, I'm sure it's possible to just import stuff), you're stuck with the commands they give you. Ubuntu also ships an OLD version of pip which has additional issues with the `--target` parameter, and this is likely what many people will have. If it would simply look in the right place to determine if a dep is present it would work.

I tried a number of things, including installing pip into hostpython and invoking it via targetpython etc. I even have a `install_into_targetpython` option in my code to allow for host-only recipes just for this. (I'll probably make a PR with it at some point, but it's unused now. Could be useful if we wanted to install cython in hostpython, which I think is a pretty damn good idea.)

Issue #544, opened 2016-01-01
Window.softinput_mode/TextInput - Window moves up/down when switching softinput_mode to 'below_target'/'resize'


rafalo1333:
When the Window's softinput_mode is set to 'below_target' or 'resize', the Window size and position is not correct after requesting keyboard with textinput/restartting the app after first launch.

Example app code:

https://gist.github.com/rafalo1333/5484ebdba49203074114

App UI on initial launch:

![screenshot_2016-01-01-19-33-37](https://cloud.githubusercontent.com/assets/7304986/12071805/b0c2cb5a-b0bf-11e5-9736-2a9f2050828c.png)

App UI when keyboard was requested (look at the bar under the TextInput widget:

![screenshot_2016-01-01-19-33-47](https://cloud.githubusercontent.com/assets/7304986/12071811/c7f2febc-b0bf-11e5-9218-4aba8b905416.png)

App UI after hiding keyboard/resetting the app (look at the bar under the TextInput):

![screenshot_2016-01-01-19-33-53](https://cloud.githubusercontent.com/assets/7304986/12071814/e18858d6-b0bf-11e5-99f7-ed35b1dcd548.png)

With the Window.softinput_mode set to 'resize', things are going even uglier.

It could be related to https://github.com/kivy/python-for-android/pull/483 PR.

It should be fixed as right now Kivy TextInputs are unusable on Android.

inclement:
According to the other issues linked, this should be fixed in master.

PR #572, opened 2016-01-10
Fixed keyboard issues for old_toolchain.

rafalo1333:
#483 by SapienTech for old_toolchain.

### DIFF
--- a/recipes/android/src/android/_android.pyx
+++ b/recipes/android/src/android/_android.pyx
@@ -190,14 +190,10 @@ if mActivity:
         @java_method('()V')
         def onGlobalLayout(self):
             rctx = Rect()
-            print('rctx_bottom: {0}, top: {1}'.format(rctx.bottom, rctx.top))
             mActivity.getWindow().getDecorView().getWindowVisibleDisplayFrame(rctx)
-            print('rctx_bottom: {0}, top: {1}'.format(rctx.bottom, rctx.top))
-            print('activity height: {0}'.format(mActivity.getWindowManager().getDefaultDisplay().getHeight()))
             # NOTE top should always be zero
             rctx.top = 0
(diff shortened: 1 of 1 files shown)


inclement:
The deleted comment is mildly concerning, but presumably you've tested this and it worked? Maybe the new toolchain had some intermediate changes.

akshayaurora:
This has nothing to do with #483 which is merged,  the comment is there to avoid this specifically,  this will lead to a crash when resuming the app we specifically added the comment in there for this reason.

rafalo1333:
@akshayaurora I have tested it and I havent Got any crash, after resuming the app too. For me everything is working. 

But if you have a better solution everyone will be glad to see it ;)

akshayaurora:
There is nothing to solve here,  #483 is already merged,  this just duplicates code, we already have the listener.

rafalo1333:
It is merged in master, but until master will be officially out programmers need to use buildozer with the old_toolchain. Closing it you're leaving Kivy with TextInputs UNUSABLE on Android. 

If #483 was successfully merged, there should be a way to solve this problem also for old toolchain.

Closing this PR wont close the issue. And if youre so sure about "there is nothing to solve here" just look at #544, damn.

akshayaurora:
@rafalo1333 #483 works for new toolchain however same code will not work for old_toolchain, 

old_toolchain implements the same code in SDLSurfaceview as outlined by the comment you removed https://github.com/kivy/python-for-android/blob/old_toolchain/src/src/org/renpy/android/SDLSurfaceView.java#L1210

this is where you want to include the change, not add another listener. There is a difference in the old_toolchain and the new one that's why the code will not work as is.

akshayaurora:
original issue closed by 11db161

PR #577, opened 2016-01-12
add option to not include cflags in cc

kived:
Allows a recipe to specify that CFLAGS should not be present in CC. Some packages will complain about this behavior (and they're technically right).


inclement:
Do some packages actually need them there? I noticed complaints before but wasn't sure why it's like this in the first place.

PR #586, opened 2016-01-12
add pil/png/jpeg recipes

kived:


pythonforandroid/recipes/jpeg/__init__.py
```diff
@@ -0,0 +1,28 @@
+from pythonforandroid.recipe import NDKRecipe
+from pythonforandroid.logger import shprint
+from os.path import join, exists
+import sh
+
+
+class JpegRecipe(NDKRecipe):
+	name = 'jpeg'
+	version = 'linaro-android'
```
AndreMiras: I know this is an old one, but I came across it.
Is this `version` ever being picked up?
Also I had some connection refused issues recently while cloning `git://git.linaro.org/people/tomgall/libjpeg-turbo/libjpeg-turbo.git` perhaps we should use a release tar.gz archive, what do you guys think?

PR #608, opened 2016-01-22
sdl2: add support to start multiple services

tito:
The current service approach is automatic, and supports only one service if there is a `service/main.py` file in the application.

This approach is complementary, and allow the user to declare multiple others services from name + entrypoint:
1. `build.py` have a new `--service name:entrypoint.py` option that can be used multiple time
1.1 it will generate appropriate service file out of the `Service.tmpl.java` into `your.package.ServiceNAME`
1.2 it will declare the service in the `AndroidManifest.xml`
2. `PythonActivity.start_service` and `PythonActivtiy` changed to pass the default entrypoint (either main.pyo or service/main.pyo).
3. `start.c` is changed to use the entrypoint that we passed along. It as mangling to either use pyo in priority, or fallback to py if pyo is not found.

Note that if you declare a name "S1", the java service will be named "ServiceS1", and your name will be capitalized (to have a correct java classname)

They are declared to run as separated process from the application. So it's up to you to came with a communication layer between your application and services. The stop method kill the process, without any warning. Use your own communication layer first to soft kill before doing stop.
Any call to start without stop will have no effect.

## Usage example:

Building:
    
    p4a apk ... --package org.domain --name testservices --service S1:service_s1.py --service S2:service_s2.py`

Pure python usage:
```python
from jnius import autoclass
context = autoclass("org.kivy.android.PythonActivity").mActivity
ServiceS1 = autoclass("org.domain.testservices.ServiceS1")
ServiceS2 = autoclass("org.domain.testservices.ServiceS2")

ServiceS1.start(context, "")
ServiceS2.start(context, "")
ServiceS2.stop(context)
ServiceS1.stop(context)
```

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/build/build.py
+++ b/pythonforandroid/bootstraps/sdl2/build/build.py
@@ -3,6 +3,7 @@
 from __future__ import print_function
 
 from os.path import dirname, join, isfile, realpath, relpath, split, exists
+from os import makedirs
 import os
 import tarfile
 import time
@@ -59,6 +60,10 @@ def render(template, dest, **kwargs):
     keyword arguments as template parameters.
     '''
 
+    dest_dir = dirname(dest)
+    if dest_dir and not exists(dest_dir):
+        makedirs(dest_dir)
+
     template = environment.get_template(template)
     text = template.render(**kwargs)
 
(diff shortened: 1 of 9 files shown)


kived:
Looks mostly good so far, but I'm having an issue with one of my services not being able to import some modules (not sure why, as the main app imports the modules fine, as do the other services).

I did add some changes to use different log tags for the services.

tito:
Looking good for the logtag!

On my side, i did make a library more "official" and clean to use them:
https://github.com/tito/servicelib

Maybe that would be good to put it under kivy umbrella?

kived:
I think that servicelib could be a good addition. It won't handle all cases, but it will work great when you just need a cross-platform multiprocessing type solution.

kived:
:+1:

PR #634, opened 2016-02-11
Fixed missing else statement in for loop, allowing systems where cyth…

mixedCase:
…on2 isn't symlinked to cython to build with the right cython binary

I have not tested this on systems other than mine (where cython isn't symlinked to cython2), therefore testing is advised.


inclement:
Thanks, this is definitely correct and fixes a silly bug I introduced. I've also added an extra commit to improve it further.

PR #636, opened 2016-02-12
add boost recipe

brussee:


inclement:
It looks like your libtorrent PR includes this recipe, am I right that the files are the same? If so, don't worry about it, I'll close this PR and merge the other one.

brussee:
Yes, libtorrent uses boost libraries and the recipe actually depends on it.

PR #637, opened 2016-02-12
add libtorrent recipe

brussee:
Uses boost recipe #636

### DIFF
--- a/pythonforandroid/recipes/boost/__init__.py
+++ b/pythonforandroid/recipes/boost/__init__.py
@@ -1,0 +1,65 @@
+from pythonforandroid.toolchain import Recipe, shprint, shutil, current_directory
+from os.path import join, exists
+import sh
+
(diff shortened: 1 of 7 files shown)


inclement:
Looks good, thanks! Just one thing, does this build only Boost.build, rather than all of it? If so, perhaps the recipe should be 'boost_build' or something, to leave space for a full boost recipe later, but I'm not familiar with boost's divisions.

brussee:
It bootstraps Boost.Build (b2, bjam) to be used by other recipes to (only) build the required libraries on demand: in case of libtorrent just libboost_system and libboost_python.
So yes, you could rename it but you would not build everything, rather just the libs you need (configured with .jam files in projects like libtorrent).

inclement:
Thanks for the contribution. I might look at renaming later, but it should be fine for now.

PR #643, opened 2016-02-15
add sticky services, auto-restart services, fix foreground option

kived:
Allows services to be started sticky (`sticky` option with `--service`, i.e. `--service=name:file.py:sticky`).

Adds option to restart services when they exit by calling `<ServiceObject>.setAutoRestartService(true)`. When the Python interpreter exits, it will be restarted automatically. This is done programmatically, rather than via a build option, so that a service can actually shut down if necessary (by setting auto-restart to false).

Fixes issue with the foreground option which caused it to be applied to all services, regardless of whether the option was actually specified.


inclement:
Neato

PR #644, opened 2016-02-16
Add ndg.httpsclient recipe

kmonson:
requests with ssl needs ngd.httpsclient. Unfortunately it when it is installed in the build environment with pip it grabs it's dependencies from pip which messes up the resulting APK. Thus this recipe is needed.


inclement:
Righto, thanks for the contribution.

PR #648, opened 2016-02-18
ngd.httpsclient for real this time.

kmonson:
I completely messed up the last pull request. It's what I get for trying to do it all through the github interface.


kived:
I'm curious, why is this recipe even needed? The package appears to be pure-Python, and you're not patching it or making any other changes to the build.

kived:
I just checked your original PR. You are using an outdated version of p4a - pip does not install dependencies since Jan 8.

kmonson:
Really? I was running master pull about a week ago and it was installing
dependencies with gusto.

kmonson:
I'll try it again but my version of p4a was cloned less than 10 days ago.

kmonson:
Also, wouldn't it specifying requirements that wouldn't get installed
otherwise be enough justification to make a recipe?

kived:
I turned off pip deps because it was breaking things in my builds (in particular, alembic was installing SQLAlchemy which needs a recipe to use the C extension): https://github.com/kivy/python-for-android/blob/master/pythonforandroid/build.py#L620

Though I can see how it would be a convenience to not need to specify the deps manually. I just worry that going down this route we could end up with a lot of recipe clutter.

kmonson:
Fair enough. There might be a good case for "package recipes" or whatever
you would call them. Something like requests_with_ssl which doesn't
actually install anything but requires requests and everything needed to
make requests work with ssl support. That would have saved me about a week
of spare time googling and digging through code to figure out why it wasn't
working.

I still wonder why it was installing deps in my case. I'll double check,
but I could swear I made my local clone much more recently then that. I'll
give it a shot again here soon.

inclement:
Is this PR now unnecessary? It doesn't seem to contain any real change anyway. I'll close it for now, but feel free to raise it again if I'm wrong.

PR #652, opened 2016-02-26
Added M2Crypto recipe

brussee:
Tested with this script on Android 6.0.1: https://e1ven.com/2011/04/06/how-to-use-m2crypto-tutorial/

### DIFF
--- a/pythonforandroid/recipes/m2crypto/__init__.py
+++ b/pythonforandroid/recipes/m2crypto/__init__.py
@@ -1,0 +1,40 @@
+from pythonforandroid.toolchain import PythonRecipe, shprint, shutil, current_directory
+from os.path import join, exists
+import sh
+
+class M2CryptoRecipe(PythonRecipe):
+    version = '0.23.0'
+    url = 'https://pypi.python.org/packages/source/M/M2Crypto/M2Crypto-{version}.tar.gz'
+    #md5sum = '89557730e245294a6cab06de8ad4fb42'
+    depends = ['openssl', 'hostpython2', 'python2', 'setuptools']
+    site_packages_name = 'M2Crypto'
+    call_hostpython_via_targetpython = False
+
+    def build_arch(self, arch):
+        env = self.get_recipe_env(arch)
+        with current_directory(self.get_build_dir(arch.arch)):
+            # Fix missing build dir
+            shprint(sh.mkdir, '-p', 'build/lib.' + 'linux-x86_64' + '-2.7/M2Crypto')
+            # Build M2Crypto
+            hostpython = sh.Command(self.hostpython_location)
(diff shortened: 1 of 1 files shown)


inclement:
Thanks. Do you think there is a case for a new recipe class setting these env vars for python recipes that need them (like CythonRecipe but for things other than cython)? If these things are general enough, it would make it easier to support e.g. multiple python versions.

brussee:
I agree that there is a lot of similarity between the M2Crypto and pyLevelDB recipes setting the python linker and including the headers. Why is this not needed by _every_ python recipe is what I am wondering?
This part seems to be a bug btw, because without it the setup.py uses the default gcc for linking whereas the compilation step _does_ use the right gcc for the architecture.
```
# Set linker to use the correct gcc
        env['LDSHARED'] = env['CC'] 
```

Also these settings are recurring for all libraries that want to use the regular gnu c stl:
```
env['CFLAGS'] += ' -I' + self.ctx.ndk_dir + '/platforms/android-' + str(self.ctx.android_api) + '/arch-' + arch.arch.replace('eabi', '') + '/usr/include' + \
                         ' -I' + self.ctx.ndk_dir + '/sources/cxx-stl/gnu-libstdc++/' + self.ctx.toolchain_version + '/include' + \
                         ' -I' + self.ctx.ndk_dir + '/sources/cxx-stl/gnu-libstdc++/' + self.ctx.toolchain_version + '/libs/' + arch.arch + '/include'
        env['CXXFLAGS'] = env['CFLAGS']
        env['CXXFLAGS'] += ' -frtti'
        env['CXXFLAGS'] += ' -fexceptions'
        env['LDFLAGS'] += ' -L' + self.ctx.ndk_dir + '/sources/cxx-stl/gnu-libstdc++/' + self.ctx.toolchain_version + '/libs/' + arch.arch + \
                          ' -lgnustl_shared'
```
Perhaps a special recipe class could be made for that as well?

inclement:
This is not necessary in every python recipe because they don't normally
do any compilation. This is by design.

Actually, there's already a recipe for pythonrecipes with compiled
components. Maybe you can use that one, though I don't remember exactly
what conditions it targets. I'll take a look at this next time I have time.

inclement:
For the LDSHARED thing, where is that line of code? Right now it should use liblink.sh, or with crystax does use env['CC'] + '-shared', but this seems to work (maybe the gcc arch doesn't matter for the linking?). I don't know much about linking though, and could easily have missed something.

brussee:
It is on line 36 of this recipe.

I tried refactoring it into a CompiledComponentsPythonRecipe but if I use the setup_extra_args it gives them not only tot the build call of setup.py but also to the install setup.py call and that is not what I want because that takes different arguments.

PR #654, opened 2016-02-26
fix android webbrowser module to work with both sdl2 and pygame

akshayaurora:


pythonforandroid/recipes/android/src/android/_android.pyx
```diff
@@ -318,23 +318,28 @@ IF IS_PYGAME:
...
-        android_open_url(url)
+# -------------------------------------------------------------------
+# URL Opening.
+def open_url(url):
+    Intent = autoclass('android.content.Intent')
+    Uri = autoclass('android.net.Uri')
+    browserIntent = Intent()
+    browserIntent.setAction(Intent.ACTION_VIEW)
+    browserIntent.setData(Uri.parse(url))
+    currentActivity = cast('android.app.Activity', mActivity)
```
kived: Why cast?

PR #664, opened 2016-03-02
Added sqlite3 recipe

brussee:


inclement:
Have you tested this (maybe on its own, not via python)? In the old toolchain, the python recipe also needs some additions to build with sqlite3.

brussee:
I only tried to load the library using Apsw with the following script:
```
#!/usr/bin/env python

import apsw
import sys

write=sys.stdout.write

write("                Python "+sys.executable+" "+str(sys.version_info)+"\n")
write("Testing with APSW file "+apsw.__file__+"\n")
write("          APSW version "+apsw.apswversion()+"\n")
write("    SQLite lib version "+apsw.sqlitelibversion()+"\n")
write("SQLite headers version "+str(apsw.SQLITE_VERSION_NUMBER)+"\n")
write("    Using amalgamation "+str(apsw.using_amalgamation)+"\n")

if [int(x) for x in apsw.sqlitelibversion().split(".")]<[3,7,8]:
    write("You are using an earlier version of SQLite than recommended\n")

sys.stdout.flush()
```

brussee:
Tested with Apsw code: https://github.com/rogerbinns/apsw/blob/master/example-code.py

kived:
Because that's a sqlite wrapper. What @inclement is referring to is the builtin sqlite3 module, which will not be built if Python itself doesn't know about sqlite.

kived:
Just like pyopenssl will work if you have openssl, but Python standard libs like urllib won't have SSL support unless Python was also compiled against openssl.

brussee:
It does not work without Apsw at the moment. See #672

brussee:
Fixed in #672 !

inclement:
Checking the commits, am I right in understanding that this PR can be closed, while #665 and #672 should be merged (and include all the content from this one)?

brussee:
That's right.

PR #675, opened 2016-03-05
Update openssl to 1.0.2g

brussee:
Lots of (security) issues fixed: https://www.openssl.org/news/changelog.html#x1

PR #672, opened 2016-03-05
sqlite3 for python2

brussee:
Integrating sqlite3 with python2 recipe.

Issue #668, opened 2016-03-04
LDFlags missing ' -lpython2.7'


brussee:
In case of the PyLevelDB recipe if I comment this line:

```
env['LDFLAGS'] += ' -lpython2.7'
```

it breaks on runtime with this error:

```
I/python(24256): ImportError: dlopen failed: cannot locate symbol "_Py_NoneStruct" referenced by "/data/data/org.tribler.android/files/lib/python2.7/site-packages/leveldb.so"...
```

It somehow is not added automatically apparently...
##

inclement:
From a quick check, it looks like none of the recipe classes automatically add this (although I thought they did) - does that look the same to you? I don't think there's a reason not to link it automatically in the compiled ones, so if that's right then let's add it in their get_recipe_env.

AndreMiras:
Yes I also noticed python linking flags seem to be missing in parent recipes.
As of today I only see it in [CppCompiledComponentsPythonRecipe.get_recipe_env()](https://github.com/kivy/python-for-android/blob/764b59e/pythonforandroid/recipe.py#L902) and [CythonRecipe.get_recipe_env()](https://github.com/kivy/python-for-android/blob/764b59e/pythonforandroid/recipe.py#L1058).
This pull request seems to tacle part of it https://github.com/kivy/python-for-android/pull/793.

I think it would greatly improve code base quality and recipes reliability to tackle that one.

PR #615, opened 2016-01-28
Update OpenSSL recipe & patch

brussee:
Fixes  #614

PR #678, opened 2016-03-07
Python 2 updated to version 2.7.9

opacam:
This is my first contribution to github... so be gentle with me if I do something wrong (I hope not)

This recipe has been tested with my android application, in fact, this is a migrated version from the python-for-android old toolchain, so... this works for me (it has been tested, with the new toolchain, as a local recipe and installed on a device running android version 4.0.4)

**Notes:**
- Works with pygame and sdl2 (bootstraps), using the Android's NDK version r10e-rc4 and Api 14 (other configurations must be tested)

- The new python2 recipe supports optional dependencies: sqlite3 (unless we use pygame bootstrap, because if that is the case, sqlite3 will be automatically included), openssl and libffi

- Some recipes has been modified to adapt to the new python's version, specially I would like to mention the openssl recipe. The resulting openssl libraries will be renamed to libsslx.so and libcryptox.so, in order  to avoid conflicts with the distributed android libs, so...all the recipes that depend on openssl shared  libraries must be linked differently (Instead of seting ldflags to "-lcrypto -lssl" we must use "-lcryptox -lsslx")

- The libtorrent recipe has been modified to adapt to new openssl build (as described above)

- The sdl recipe has been modified to avoid error not finding python (patch applied at build time)

- The sqlite3's recipe is from user brussee, with a little modification, so all credits must go to him

- It is recommended to run command "python-for-android clean_all" before changing python2's optional dependencies (sqlite3, openssl or libffi)...otherwise it is most likely to fail, because the patches are applied at python's build time

- Apologies for my english


inclement:
Thanks for the contribution! I will be away for a few days, and this is a big PR, so we may not have time to look at it for a little while. At a quick glance it looks good though, and we'll take a full look as soon as possible.

brussee:
> The new python2 recipe supports optional dependencies: sqlite3

Great work! I haven't got this to work #672.

> The resulting openssl libraries will be renamed to libsslx.so and libcryptox.so, in order to avoid conflicts with the distributed android libs.

I found this to be not necessary if using p4a (tested here https://github.com/brussee/SkeletonApp/tree/test2-torrent).
I guess p4a figures this out during loading?

> The sqlite3's recipe is from user brussee, with a little modification, so all credits must go to him.

Thanks! I'd like to share the credits with everyone who helped me :)

brussee:
I tested this PR for my app and noticed #688 and #689 no longer occur (#690 is still there).

However, similar to #668, I had to add this to the (python) recipes: numpy, pil, pyasn1, six, zope, android, netifaces, pyjnius, kivy, libtribler, zope_interface.
```

        env = super(NetifacesRecipe, self).get_recipe_env(arch)
        # Set linker to use the correct gcc
        env['LDSHARED'] = env['CC'] + ' -pthread -shared -Wl,-O1 -Wl,-Bsymbolic-functions'
                          ' -lpython2.7'
        return env
```
Otherwise an ELF32 error or Python.h not found occur.

kived:
> Some recipes has been modified to adapt to the new python's version, specially I would like to mention the openssl recipe. The resulting openssl libraries will be renamed to libsslx.so and libcryptox.so, in order to avoid conflicts with the distributed android libs, so...all the recipes that depend on openssl shared libraries must be linked differently (Instead of seting ldflags to "-lcrypto -lssl" we must use "-lcryptox -lsslx")

This isn't a good idea. It makes it harder to use anything which depends on these libs, as everything must now be patched, and it doesn't solve any real problem. The app's libs will be used before system libs.

brussee:
@opacam feel free to checkout this commit to undo the changes @kived is talking about https://github.com/brussee/python-for-android/commit/72afe0fd477dc3cc4876a57570025b5fda7d580e

Also feel free to checkout that branch to see what I was talking about regarding the recipes.

pythonforandroid/recipes/hostpython2/__init__.py
```diff
@@ -1,17 +1,22 @@
...
-from os import chdir
+from os import environ
 import sh
 
 
 class Hostpython2Recipe(Recipe):
-    version = '2.7.2'
-    url = 'http://python.org/ftp/python/{version}/Python-{version}.tar.bz2'
+    version = '2.7.9'
+    url = 'http://python.org/ftp/python/{version}/Python-{version}.tgz'  # tar.bz2'
```
kived: Why is this changed to `.tgz`? Are the bzip2 archives not distributed anymore? bzip2 should be preferred as it is a smaller download at the expense of slower extraction - but people are far more likely to have download caps than execution caps.

pythonforandroid/recipes/hostpython2/__init__.py
```diff
@@ -1,17 +1,22 @@
...
     name = 'hostpython2'
 
     conflicts = ['hostpython3']
 
+    def get_recipe_env(self, arch):
+        env = super(Hostpython2Recipe, self).get_recipe_env(arch)
+        env['CFLAGS'] = ' '.join([env['CFLAGS'], '-Wformat'])
```
kived: This is not necessary at all.

pythonforandroid/recipes/hostpython2/__init__.py
```diff
@@ -24,10 +29,25 @@ def prebuild_arch(self, arch):
...
+        # recipe_dir = self.get_recipe(self.name, arch.arch).recipe_dir
+        # shprint(sh.cp, join(recipe_dir, 'Setup'), join(self.get_build_dir(), 'Modules', 'Setup'))
+
 
     def build_arch(self, arch):
+
         with current_directory(self.get_build_dir()):
+            if exists('hostpython'):
+                info('Setting ctx hostpython from previous build...')
+                self.ctx.hostpython = join(self.get_build_dir(), 'hostpython')
```
kived: While this is indeed an issue, this is the wrong place to fix it.

pythonforandroid/recipes/openssl/__init__.py
```diff
@@ -5,7 +5,7 @@
 
 
 class OpenSSLRecipe(Recipe):
-    version = '1.0.2f'
+    version = '1.0.2g'
```
kived: This change has already been made in master, and required a change to the patch as well - but you haven't modified the patch. This can't actually be working as the patch will fail to apply to 1.0.2g and therefore the build will fail.

pythonforandroid/recipes/python2/__init__.py
```diff
@@ -5,41 +5,101 @@
...
                'patches/fix-dynamic-lookup.patch',
                'patches/fix-dlfcn.patch',
-               'patches/parsetuple.patch',
                'patches/ctypes-find-library-updated.patch',
-               ('patches/fix-configure-darwin.patch', is_darwin),
-               ('patches/fix-distutils-darwin.patch', is_darwin),
-               ('patches/fix-ftime-removal.patch', is_api_gt(19)),
-               ('patches/disable-openpty.patch', check_all(is_api_lt(21), is_ndk('crystax')))]
+
+               # Todo: Regular Patches from 2.7.2, Not needed on 2.7.9, must be removed?
```
kived: This requires more explanation. Why are these patches no longer necessary? Some of these are absolutely necessary, BTW.

pythonforandroid/recipes/python2/__init__.py
```diff
@@ -5,41 +5,101 @@
...
+            patch_xcompile_mod = join(recipe_dir, 'patches/Python-{0}-xcompile_mod.patch'.format(self.version))
+            shutil.copy(patch_xcompile_base, patch_xcompile_mod)
 
+            # Creates a copy of basic-android.patch...to be adapted to our build
+            patch_android_base = join(recipe_dir, 'patches/basic-android-{0}.patch'.format(self.version))
+            patch_android_mod = join(recipe_dir, 'patches/basic-android-{0}_mod.patch'.format(self.version))
+            shutil.copy(patch_android_base, patch_android_mod)
+
+            # REWRITES PATCH TO MATCH SSL DIRS
+            if 'openssl' in self.ctx.recipe_build_order:
```
kived: Don't do this. Apply a separate patch if necessary, but this is ugly and convoluted.

pythonforandroid/recipes/python2/__init__.py
```diff
@@ -56,89 +116,122 @@ def build_arch(self, arch):
...
-            # AND: Should probably move these to get_recipe_env for
-            # neatness, but the whole recipe needs tidying along these
-            # lines
-            env['HOSTARCH'] = 'arm-eabi'
-            env['BUILDARCH'] = shprint(sh.gcc, '-dumpmachine').stdout.decode('utf-8').split('\n')[0]
-            env['CFLAGS'] = ' '.join([env['CFLAGS'], '-DNO_MALLINFO'])
+            env.update({
+                'RFS': "{0}/platforms/android-{1}/arch-arm".format(self.ctx.ndk_dir, self.ctx.android_api),
+                })
+            # info('\t-> RFS:               {}'.format(env['RFS']))
```
kived: Please remove extraneous commented lines before submitting a PR.

pythonforandroid/recipes/python2/__init__.py
```diff
@@ -56,89 +116,122 @@ def build_arch(self, arch):
...
 
             env = arch.get_env()
-
-            # AND: Should probably move these to get_recipe_env for
-            # neatness, but the whole recipe needs tidying along these
-            # lines
-            env['HOSTARCH'] = 'arm-eabi'
-            env['BUILDARCH'] = shprint(sh.gcc, '-dumpmachine').stdout.decode('utf-8').split('\n')[0]
-            env['CFLAGS'] = ' '.join([env['CFLAGS'], '-DNO_MALLINFO'])
+            env.update({
```
kived: Don't call `.update()` to add a single value. Or to add two values, as done below.

pythonforandroid/recipes/python2/__init__.py
```diff
@@ -56,89 +116,122 @@ def build_arch(self, arch):
...
-
-            print('Second install (expected to work)')
-            shprint(sh.touch, 'python.exe', 'python')
-            shprint(make, '-j5', 'install', 'HOSTPYTHON={}'.format(hostpython),
-                    'HOSTPGEN={}'.format(hostpgen),
+                    '--with-system-ffi',
+                    '--disable-ipv6',
+                    # '--disable-toolbox-glue',
+                    # '--disable-framework',
+                    'ac_cv_file__dev_ptmx=no',
```
kived: Is there a way to do this *without* directly setting autoconf variables?

pythonforandroid/recipes/openssl/__init__.py
brussee: The patch needed to be changed from e->f #615 
f->g was ok #675

pythonforandroid/recipes/hostpython2/__init__.py
opacam: Because the package no longer exist in *.bz2 compression for this version of python, only exist in *.tgz format...

pythonforandroid/recipes/openssl/__init__.py
opacam: You are totally right, I forgot to update the patch, now the changes has been made. Many thanks to mention it !!!

dessant:
Why not go for 2.7.11 directly?

opacam:
Oooo...that's a point!!! but I made this recipe a long time ago (with the old toolchain) and I recently migrated to the new toolchain...going to the new python 2.7.11 will require more work to do and that is not an easy task (despite I think it must be done at some point) but for now this version is enough to solve some security issues related with the requests module that can not be solved with the old python version (at least I cannot solve) and this recipe is working (it's building and running fine)

Also notice that the current python version (2.7.2) distributed with the python-for android is running since I can remember (at least 2 years...or more) and it is really outdated, and I know that some user tried to update the python version for the old toolchain without success (sorry...I don't remember who), so... I think that this is a first step into the future.

pythonforandroid/recipes/python2/__init__.py
opacam: Some of the changes has been applied with other new patches, others are not needed because changes has been made into the new python version, some of them has not been tested (as noticed in the comment) and must be tested but i don't have the tools to test it (different api, platform...), maybe I can look into the code...this will take some time, I will be happy If you can help me with that.

pythonforandroid/recipes/python2/__init__.py
opacam: Ok, you are right...already removed from my local source code and I will push changes later

pythonforandroid/recipes/hostpython2/__init__.py
opacam: Thanks!!!

pythonforandroid/recipes/python2/__init__.py
opacam: Ok, I took your advice,now I have made a different approach in my local system (now I rewrite directly the affected  files, after the patch has been applied)...I would like to try the changes, before I push changes into master branch.

pythonforandroid/recipes/python2/__init__.py
opacam: Already changed in my local repository... changes will be pushed into master branch later, when it has been tested.

pythonforandroid/recipes/python2/__init__.py
opacam: Maybe setting as CPPFLAGS variable instead of passing as arguments into the configure script...what do you think?

PD: I Changed my source code with this approach but now I don't have time to test it, maybe tonight or tomorrow....

opacam:
Ei Kived and Brussee...many thanks to take your time to review my code and help me to improve it. I will try to make a build with all the changes when I will have enough time...maybe tonight or tomorrow.

opacam:
Tested build with command:

python-for-android create --dist_name=test_sdl2 --bootstrap=sdl2 --requirements=hostpython2,libffi,openssl,sqlite3,python2,pil,kivy

Notice that hostpython, and python must be build with the right order when using sdl2  bootstrap and using sqlite3 recipe, otherwise it will fail to build because the build order is not correct (this is not happening when using the pygame bootstrap)...Maybe is a bug when checking optional dependencies?

pythonforandroid/recipes/python2/__init__.py
opacam: Now, with the latest commit,  I think that no more explanation is needed

opacam:
Ok... now we are on the road...I think that all the proposed changes are been done

Note: Tested with sdl2 bootstrap

opacam:
Now I'm doing some test with the last python version (2.7.11) ... and it's working...so this will be closed and I will make a new Pull Request with the new version when I have all in order.

Thanks to all who take his time to see this pull request...hope I see you soon when i make the new pull request for the last version of python

PR #683, opened 2016-03-13
Fixed : alignment in instructions

miohtama:
The output did not look good in rendered Markdown. An extra space was added before :  which one should not do in English spelling.


inclement:
Thanks.

PR #694, opened 2016-03-29
fix platform fail on p4a

xxnet:
python-for-android executable is not a file so can't be opened.
platform.platform()  will fail because get the system arch by parse the executable.

### DIFF
--- a/pythonforandroid/recipes/python2/__init__.py
+++ b/pythonforandroid/recipes/python2/__init__.py
@@ -23,6 +23,7 @@ class Python2Recipe(TargetPythonRecipe):
                'patches/fix-gethostbyaddr.patch',
                'patches/fix-setup-flags.patch',
                'patches/fix-filesystemdefaultencoding.patch',
+               'patches/fix-platform.patch',
                'patches/fix-termios.patch',
                'patches/custom-loader.patch',
                'patches/verbose-compilation.patch',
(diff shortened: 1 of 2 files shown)


inclement:
So this occurs because python is being set up and run via the C script? If that's the case, I'm not sure if we should patch the behaviour out, since it's the normal behaviour of the python interpreter. That said, it seems fine to do so if it's causing problems - is there somewhere in particular where this comes up and should be fixed?

I also wonder if python's internal behaviour is the same in newer versions?

xxnet:
The reason that raise Exception is because python-for-android implement python interpreter as JNI(Java Native Interface), not a standard alone executable file. 
platform.platform() get the system libc version by parse the executable file.

I think platform.platform() should return a string on any platform, not raising an Exception.
It same on python 2.7.11 and python 3.4 for this point.

We can use:
try:
  platform = platform.platform()
catch:
  platform = "anything"

but I don't think this is a good idea.

We can return "Android" and more detail like android version and manufacturer in feature.
May be we can submit this patch to python project.

inclement:
Right, my concern is that this is not (apparently) a bug in python-for-android, but in the python interpreter itself. I'm not sure philosophically if we should fix these things with hacky patches, rather than working around them and contributing upstream if necessary.

xxnet:
This error will occur on embedded python interpreter like p4a.
Normal python will not encounter this error.

inclement:
I'm going to close this without merging because I still don't think it's the right choice to patch python's natural behaviour when it's not essential. Please do raise it again if you feel strongly about it, I'm happy to discuss it, but when I brought it up on irc people seemed to agree with me.

PR #701, opened 2016-04-08
Add spine-cython recipe.

pythonic64:
Recipe for [spine-cython](https://github.com/tileworks/spine-cython).


inclement:
Great, thanks!

PR #702, opened 2016-04-09
Bump colorama requirement to 0.3.3

kahowell:
The EX color variants weren't added until 0.3.3. I found out because
Fedora 23 packages colorama 0.3.2, which satisfies the requirement, but
causes the error: `AttributeError: 'AnsiCodes' object has no attribute
'LIGHTBLUE_EX'`...


inclement:
Thanks!

Issue #705, opened 2016-04-14
Enabling SSL for python3.5 using crystax


emillynge:
Hi

I have just spend a full day solving a problem with accessing HTTPS sites on android. I finally figured out the problem:

Almost all http libraries in python relies on the ssl.py module in the standard library. This module is uses a C-extension `_ssl.so`  that must be "baked in" with the cpython interpreter at compilation. 
The crystax android ndk however does _not_ do this by default however, and so reaching any https site is not possible. For the novice programmer it is, in no way obvious why ssl.py cannot be imported, so I would propose that somewhere in the documentation, it is described how to recompile the crystax python interpreter _with_ ssl support from the crystax-ndk.

I would gladly write this, but where would it be appropriate to put it in the docs?

Regards
Emil Lynge
##

inclement:
There is a python3 section somewhere in the docs that explains the need to use crystax. You could add further instructions here.

Do you know how to add openssl, using the crystax build scripts to recompile python but compiling openssl first? If not, I can give some information about this (another user pointed it out to me recently), but it needs some manual hacking of p4a builds right now.

I think the real problem is that crystax's build process currently doesn't build openssl binaries (see https://groups.google.com/forum/#!topic/crystax-ndk/x2LXh3egRuM), but apparently will do so in the next release. I'd guess they'll also compile python against it at that point, so I've been leaving this problem partly to wait for that, though the openssl recipe will need modifying to load openssl differently in this case (rather than compiling it internally).

emillynge:
Well their `build-target-python` script actually does a check to see if openssl has been compiled and builds `_ssl.so` if found.
So I would guess that the problem is resolved when they start building openssl binaries by default.
I have no idea how far out in the future that is, so I don't whether it is worth putting in the effort to fix it here and now..

I guess that it would be possible to make a `openssl-crystax` recipe that checks if the library has been compiled in the ndk, and recompiles if not. For the `build-target-python` to work the openssl binaries _must_ be placed in the ndk, so it's difficult to just tweak the existing openssl recipe.

Maybe it is possible to just tweak the python-crystax  recipe to copy over, and recompile if openssl is in the build order...  hmmmmm..
I may look into that first.

mgc8:
[...]

PR #707, opened 2016-04-15
crystaxpython with openssl support

emillynge:
Fix to #705 

This PR makes crystax-python recompile with `_ssl.so` module if openssl is in the build order.

tested for arm64-v8a

There is a bug that I don't know how to fix for armeabi-v7a similar to this issue:
https://github.com/openframeworks/openFrameworks/issues/4171

Building openssl shared lobraries with the crystax ndk script `build-target-openssl.sh` however, instead of copying over the ones from the openssl recipe, solves this problem. I'm not nearly good enough at examing make files to know what the difference between the produced .so files are. But crystax produces slightly smaller .so files, and uses make install rather than make build_libs.

Maybe the recipe should ignore recompilation if  armeabi-v7a and throw a warning that ssl.py _will not work_ until such time that this bug is fixed.

I could also make the openssl recipe use the crystax build script, but then it's hard to have proper bookeeping on what has been built, and what is missing :-/
Additionally I have no idea whether the crystax script produces openssl libs that will work on the phone, or whether they just serve to be linked against by the python compilation...

NB: I had to implement a realpath function as `os.path.realpath` is buggy. maybe that func should go into utils...

Regards
Emil

### DIFF
--- a/pythonforandroid/recipes/python3crystax/__init__.py
+++ b/pythonforandroid/recipes/python3crystax/__init__.py
@@ -1,3 +1,4 @@
+import os
 
 from pythonforandroid.recipe import TargetPythonRecipe
 from pythonforandroid.toolchain import shprint, current_directory, ArchARM
@@ -7,10 +8,64 @@ from os.path import exists, join
 from os import uname
 import glob
 import sh
+from sh import Command
+
+
+# This is the content of opensslconf.h taken from
+# ndkdir/build/tools/build-target-openssl.sh
+OPENSSLCONF = """#if defined(__ARM_ARCH_5TE__)
+#include "opensslconf_armeabi.h"
+#elif defined(__ARM_ARCH_7A__) && !defined(__ARM_PCS_VFP)
+#include "opensslconf_armeabi_v7a.h"
+#elif defined(__ARM_ARCH_7A__) && defined(__ARM_PCS_VFP)
+#include "opensslconf_armeabi_v7a_hard.h"
+#elif defined(__aarch64__)
+#include "opensslconf_arm64_v8a.h"
+#elif defined(__i386__)
+#include "opensslconf_x86.h"
+#elif defined(__x86_64__)
+#include "opensslconf_x86_64.h"
+#elif defined(__mips__) && !defined(__mips64)
+#include "opensslconf_mips.h"
+#elif defined(__mips__) && defined(__mips64)
+#include "opensslconf_mips64.h"
+#else
+#error "Unsupported ABI"
+#endif
+"""
+
+
+def realpath(fname):
+    """
+    Own implementation of os.realpath which may be broken in some python versions
+    Returns: the absolute path o
+
+    """
+
+    if not os.path.islink(fname):
+        return os.path.abspath(fname)
+
+    abs_path = os.path.abspath(fname).split(os.sep)[:-1]
+    rel_path = os.readlink(fname)
+
+    if os.path.abspath(rel_path) == rel_path:
+        return rel_path
+
+    rel_path = rel_path.split(os.sep)
+    for folder in rel_path:
+        if folder == '..':
+            abs_path.pop()
+        else:
+            abs_path.append(folder)
+    return os.sep.join(abs_path)
+
+
 
 class Python3Recipe(TargetPythonRecipe):
     version = '3.5'
-    url = ''
+    url = ('https://github.com/crystax/android-vendor-python-{}-{}'
+           '/archive/master.tar.gz'.format(*version.split('.')))
+
     name = 'python3crystax'
 
     depends = ['hostpython3crystax']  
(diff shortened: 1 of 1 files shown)


inclement:
Nice. I've a small concern about the build directories - where are python and openssl built? It would be best to do it in p4a's build dirs.

I'm also not sure about some of the code structure, but if it works then it will be great to make some changes and merge it. Maybe the openssl build should really go in the openssl recipe (or a new one, but it can just use an if statement) and the python recipe should have a more generic python build option rather than just copying the crystax libs.

emillynge:
In this implementation the crystax makes it's own temporary builddir, and relevant files are extracted into p4a after build is complete.

When using the crystax build scripts, it is possible to specify a build dir. This could be pointed to the p4a build dir I guess. This would be a rewrite of openssl recipe which would be very different if the crystax ndk is used.

>  ...python recipe should have a more generic python build option rather than just copying the crystax libs
Not sure what you mean?

inclement:
> When using the crystax build scripts, it is possible to specify a build dir. 

If it uses a relative directory by default, the command can just be called from `with current_directory(self.get_build_dir(...))`.

> This would be a rewrite of openssl recipe which would be very different if the crystax ndk is used.

This would be an alternative, not a replacement of the existing recipe.

> Not sure what you mean?

The python3crystax recipe should have the option to compile python, regardless of any of these concerns.

emillynge:
Quite a lot of the dirs and options are set via the `dev-defaults.sh` script which makes it very hard to tweak anything other than the options that is explicitly set at input arguments :/

The program description states:

> Rebuild python libraries for the CrystaX NDK.
> 
> This requires a temporary NDK installation containing
> toolchain binaries for all target architectures.
> 
> By default, this will try with the current NDK directory, unless
> you use the --ndk-dir=<path> option.
> 
> The output will be placed in appropriate sub-directories of
> \<ndk\>/$PYTHON_SUBDIR, but you can override this with the --out-dir=<path>
> option.

So i think it's easiest just to use the "--out-dir" option

Thing is, that I cannot make the build scripts look for the openssl libraries any other place than in the NDK itself :(
As seen by this snippet from `build-target-python.sh`

> OPENSSL_HOME=''
> if [ -n "$DEFAULT_OPENSSL_VERSION" ]; then
>     echo "version found: $DEFAULT_OPENSSL_VERSION"
>     if [ -f "$NDK_DIR/$OPENSSL_SUBDIR/$DEFAULT_OPENSSL_VERSION/Android.mk" \
>          -a -f "$NDK_DIR/$OPENSSL_SUBDIR/$DEFAULT_OPENSSL_VERSION/include/openssl/opensslconf.h" ]; then
>         echo "home found"
>         OPENSSL_HOME="openssl/$DEFAULT_OPENSSL_VERSION"
>     fi
> fi

`DEFAULT_OPENSSL_VERSION` and `OPENSSL_SUBDIR` is set by `dev-defaults.sh` 

> This would be an alternative, not a replacement of the existing recipe.

Ok, so i can just make an `openssl-crystax` recipe? Although I guess some dependencies should be fixed to accept either `(openssl, openssl-crystax)`

> The python3crystax recipe should have the option to compile python, regardless of any of these concerns.

Can I tap into the `--force-build` argument from the recipe methods?

inclement:
I see what you mean about the crystax build scripts, but let's just not worry about it.

Feel free to leave it for now, I'll take a proper look at the changes later.

As an aside, an alternative to all of this would be to build openssl manually (ignoring the crystax scripts), as is done with python2. This would avoid all these problems, but would be more complexI'm not suggesting you should do it, just mentioning the possibility.

Feel free to leave it for now, I'll take a proper look at the changes later.

FeralBytes:
So when I first run this with a clean P4A; I get a traceback; because it appears to have a dependency on the original Python3Crystax recipe having been run once.
```
  File "setup_SDL2_crystax_py3.py", line 82, in <module>
    package_data=package_data
  File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/bdistapk.py", line 83, in run
    main()
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/toolchain.py", line 900, in main
    ToolchainCL()
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/toolchain.py", line 508, in __init__
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/toolchain.py", line 147, in wrapper_func
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/toolchain.py", line 190, in build_dist_from_args
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/build.py", line 550, in build_recipes
    recipe.prepare_build_dir(arch.arch)
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/recipe.py", line 550, in prepare_build_dir
    self.unpack(arch)
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/recipe.py", line 472, in unpack
    .format(extraction_filename))
Exception: Given path is neither a file nor a directory: /home/wolfrage/.local/share/python-for-android/packages/python3crystax/master.tar.gz
```

professor-python:
Some info about this feature please. Tell me if I right when using p4a with buildozer. CrystaxNdk is downloaded as binaries (10.3.2 now). All works without ssl. And now we just set p4a.source_dir to p4a source with fixed python3crystax recipe (?). Need I download CrystaxNdk sources for this recipe if it will try to compile it?

inclement:
@professor-python I'm not sure I understand the question - are you trying to test this PR?

professor-python:
@inclement yes, I'm trying to add openssl support with crystax by this PR

ecdsa:
@emillynge could you please update your PR?

emillynge:
@ecdsa update how? as in rebase?

I am no longer actively using this project, so I'n not sure whether I am at all capable of rebasing without causing problems :/ I can barely remember what I did at all.

ecdsa:
yes, make it compatible with current p4a
I had to copy .so files manually in order to compile an APK with python3 and SSL

fyookball:
I am also having the same problem as @ecdsa.  I see that @frmdstryr put some changes on their branch but I get the same error.

JonasT:
@AndreMiras since I assume crystax will be left behind now, maybe it would make sense to close this?

@emillynge thanks so much for your work! We got openssl/TLS working on p4a master with "python3" (no crystax!) now so this is probably no longer required, but it helped a lot as an intermediate solution

PR #718, opened 2016-04-20
rewrite recipe and update cryptography to 1.3.1

brussee:
Dynamically linking works just fine.


brussee:
Changelog:

1.3.1 - 2016-03-21

    Fixed a bug that caused an AttributeError when using mock to patch some cryptography modules.

1.3 - 2016-03-18

    Added support for padding ANSI X.923 with ANSIX923.
    Deprecated support for OpenSSL 0.9.8. Support will be removed in cryptography 1.4.
    Added support for the PolicyConstraints X.509 extension including both parsing and generation using CertificateBuilder and CertificateSigningRequestBuilder.
    Added is_signature_valid to CertificateSigningRequest.
    Fixed an intermittent AssertionError when performing an RSA decryption on an invalid ciphertext, ValueError is now correctly raised in all cases.
    Added from_issuer_subject_key_identifier().

PR #726, opened 2016-04-27
add libtribler recipe

brussee:
Privacy with BitTorrent and resilient to shut down http://www.tribler.org

Recipe uses the latest development version from git.
Supersedes PR #405


inclement:
Some of the dependencies here don't have recipes, should they be in python_depends instead?

inclement:
Actually it looks like I just missed your PR that adds these recipes. If that's the case for all of them, never mind (and thanks for all the contributions).

Issue #700, opened 2016-04-07
Webview bootstrap. Where to start? [$100]


bobatsar:
I would like to have a python4android application which just starts a simple webview and accesses a local python webframework, like flask.

What is the recommended way to start.
Should I take the sdl2 bootstrap and generate a new stripped down version or should I start with the empty bootstrap?

Is there anything already done in that direction?

How can the webview be started? Should it be initialized by a java class or is it ok to use pyjnius like in the following gist? 
https://gist.github.com/tito/6118172

inclement:
I'd take the SDL2 bootstrap as a reference for the python unpacking etc, but strip out the SDL2 stuff and replace it with a webview. I'm not sure how threading will need to work with the python interpreter (maybe it will already), so I might start by running flask in a service.

> Is there anything already done in that direction?

I'll probably try this sometime if nobody else does first.

kollivier:
Actually, I've been making a project that is a wrapper above Kivy, wxPython, etc. specifically for the purpose of creating cross-platform apps that use Web UI. You can check it out here:

http://pyeverywhere.org/

Follow the Getting Started section and let me know if you see any issues. Even if it doesn't work for you, the pew/kivy_pew/webview.py module can show you a start on how to do it yourself.

bobatsar:
I think Pyeverywhere is already too much for our purpose. But thanks for the link. Perhaps it can help in the future.

I had a look to strip out the sdl stuff but I really don't know whats still needed in the java classes and so on.

@inclement, could you try that? 
I think the webview activity should be started from Java and the main python thread should be available for anything else (like the webframework)

The only thing which is needed as a connection between the two components is to provide a URL for the Webview to display.

thanks a lot

kollivier:
Sorry, I wasn't suggesting just pulling the whole project into Kivy. Feel free to just integrate the code I wrote to get the Android WebView loaded into Kivy if it works for your purposes. It's mostly just some pyjnius code. There shouldn't be any licensing issues between the projects but if there are, let me know and I'll give permission to license it under Kivy's license. The main part of it is here:

https://github.com/kollivier/pyeverywhere/blob/master/src/pew/kivy_pew/webview.py

[...]

PR #728, opened 2016-05-01
Add webview bootstrap

inclement:
Not quite for merging yet, but it's basically ready. I intend to finish and merge it shortly.

Edit: resolves https://github.com/kivy/python-for-android/issues/700


tito:
This is just awesome!

inclement:
All the core stuff I wanted to do is now finished, so testing would be welcome and it should be ready to merge (though it would be good if someone could give a confirmation that I haven't missed anything).

Issue #741, opened 2016-05-12
Webview - back button bug


jligeza:
My webview app displays ,,hello world'' text.

I open it, and close shortly after with back button (without deleting from recently opened apps): http://pastebin.com/bAhqqYPS

Then, I try to open it again (from ,,desktop'', or how to call it), and it crashes with segmentation error: http://pastebin.com/uH22kSUV

I use python 2 for this app.

bobatsar:
The problem is, that the python_thread is not killed when the app is closed.
I have a workaround in pull request #750. Not nice but at least working for now.

inclement:
I'll close this as I've merged @bobatsar's PR, but a better solution would be welcome.

PR #750, opened 2016-05-20
Webview improvements

bobatsar:
Hi there,

here are some improvements to the webview bootstrap:
- load an url in the webview from python code
- as in #741 described, a second start of the app crashes. At the moment I kill the whole application to make sure the python thread stops. I don't know if this is the best approach but at least it is working.
- I implemented a handler for the back button to scroll through the browser history and finally close the app. Perhaps this should be made configurable to choose one of the options.
I am also currently trying to somehow get the back button event via a callback to the python program to handle it there. Do you have a suggestion how to do that? How do you do that in SDL2?


inclement:
I'm still not sure about killing the child threads this way, but maybe I should just merge it and leave it to be fixed later.

inclement:
I'm not certain if some of this behaviour might need adding to for flexibility, but anyway, great contribution and I'm sorry for the delay merging it.

PR #754, opened 2016-05-24
Port storm and psycopg2 recipes from the old toolchain to the revamp.

felipenehmi:
Port psycopg2 (such as it's dependency 'libpq') and storm recipes from the old toolchain to the revamp.


inclement:
Great, thanks!

PR #763, opened 2016-06-04
[WIP] add experimental wheel support

kived:
Recipes with `use_pip = True` will attempt to install via pip by default. Adding `--build-wheels` on the command line will cause p4a to build the wheels and then install rather than downloading via pip. Currently uses `http://localhost:8080/simple/` as the extra index url (for testing with pypi-server).

Wheel platform is set by Android API and ARM EABI - i.e. `android19-armeabi`.


mhsmith:
For current work in this area, see #3013.

PR #764, opened 2016-06-05
Add distutils command for 'python setup.py apk'

inclement:
Not for merging yet (needs doc + cleanup), but basically works.

This adds a distutils command so that you can run `python setup.py apk` to call p4a. It will use the normal setuptools information (name, version, package_data) unless otherwise specified, and you can also use all the normal p4a arguments.

This works by papering over all the normal distutils behaviour (e.g. manually adding all the user arguments to the allowed list, since we want distutils to ignore them and let p4a handle it). I don't know if this could be done better.

I took the approach of using a normal distutils Command, but we could also provide our own setup() that would take extra arguments. This would be neater than the current `register_args`, but I didn't do it initially because I haven't seen other packages do this, and it's a little awkward.

(Edit: cx_freeze [does provide its own setup](http://cx-freeze.readthedocs.io/en/latest/distutils.html), but I don't think we need to)


inclement:
This is now probably ready to merge, testing or comments welcome.

PR #772, opened 2016-06-11
Sqlite3 for Python 2.7.11 (mandatory dependency of Python 2.7.11)

opacam:
Actually this recipe is from @brussee (PR #672) but without changing any other files, only adds the recipe for the new python 2.7.11 (PR #775). So... all credits should go to @brussee

**Merge Note**: This could be merged safely, but I will recommend to merge, only if the PR of Python 2.7.11 is accepted, otherwise it's better to merge the @brussee 's PR


opacam:
This is not needed anymore, because the sqlite3's recipe from @brussee has been merged

PR #773, opened 2016-06-11
Libffi for Python 2.7.11 (mandatory dependency of Python 2.7.11)

opacam:
This recipe is needed by PR #775 (Python 2.7.11). It adds the function "get_lib_dir" needed by the new python2 recipe and fixes the error "sh.CommandNotFound" on unix platforms (at least on my debian system)

**Merge Note**: I think it could be merged safely without merging the PR of Python 2.7.11


opacam:

PR #774, opened 2016-06-11
Openssl for Python 2.7.11 (mandatory dependency of Python 2.7.11)

opacam:
This recipe is needed by PR #775 (Python 2.7.11).

This libraries are problematic because android already distributes an older version and we compile all the stuff against the latest version. This causes some troubles when we try to load the openssl libs from the app, because the android system loads first his own libs... and it is almost sure that the app will fail at some point when trying to call some function that not exist anymore, or maybe, is a newer function that not exist in the old libs.

Is for this that I decide to compile this libs as static instead of shared (I know that it increase a little bit the final size of the app but this way we make sure that all the calls to openssl libs will work as expected). Cause building as static libs maybe is not necessary for some recipes or bootstraps, I doubt so, if we need to add this libs into the final app,  I decide to add the ability to build as static or shared (if the new python2 recipe is used it will be build as static, unless the user uses the variable "force_shared" set to True, otherwise it will be build as shared as always has been).

**Merge Note**: This could break the old python2 recipe...so, only should be merged if the Python 2.7.11 PR is accepted

### DIFF
--- a/pythonforandroid/recipes/openssl/__init__.py
+++ b/pythonforandroid/recipes/openssl/__init__.py
@@ -1,15 +1,32 @@
 from functools import partial
-
-from pythonforandroid.toolchain import Recipe, shprint, current_directory
+from pythonforandroid.toolchain import Recipe, current_directory
+from pythonforandroid.logger import shprint
+from os.path import join, exists
 import sh
-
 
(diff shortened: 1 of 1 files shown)


opacam:

PR #775, opened 2016-06-11
Python 2.7.11

opacam:
The new python2 recipe supports optional dependencies: sqlite3, openssl and libffi

This PR updates the python2's recipe to version 2.7.11. Libffi (#773) and Openssl (#774)

If this is finally merged, all PRS with "mandatory for Python 2.7.11" in his title should be merged, otherwise it will not work as expected (it will fail to build if the affected recipe is used). I will like to mention that it is possible that we have to create a conditional statement for all this affected recipes to make it work with Python 3.x, I really hope that this has not to be necessary, but I really don't know ( I don't have any code written in python 3), this new recipe is closer to the python3 recipe than the old one ...so, probably, all the changes made with the affected recipes, will also work on python 3.x.

**Notes for testers:** I will recommend to programmers who want to test the new python 2.7.11 against the old PR(#693) because has all the mandatory and optional changes incorporated.

_(For more details read the following message)_

Thanks to:

- @brussee he found the right direction for almost all of the updated recipes!!!

- @kived who gave to me his wisdom with his advises and forced me to order and polish all the code.

- @dessant to encourage me to try to update from my initial pull request (pytthon 2.7.9)

- users who give me positive feedback in the old Python 2.7.11 & Updates (PR #693): @dessant, @tito and @Tungsteno74 It means a lot to me :)))

- all the developers who make possible the kivy project and all the related sub projects!!!

PR #776, opened 2016-06-11
 kivy's recipe (mandatory for Python 2.7.11)

opacam:


opacam:

Issue #669, opened 2016-03-04
Python compiled components recipe: "bad gcc/glibc config?"


brussee:
In case of the PyLevelDB recipe if I comment this line:

```
env['PYTHON_ROOT'] = self.ctx.get_python_install_dir()
#env['CFLAGS'] += ' -I' + env['PYTHON_ROOT'] + '/include/python2.7'
```

it breaks like this: https://github.com/rogerbinns/apsw/issues/206
##

inclement:
Is this still a problem, or resolved in your later work with apsw etc?

brussee:
This was resolved already: https://github.com/rogerbinns/apsw/issues/206#issuecomment-191744968

PR #778, opened 2016-06-11
Pyjnius's recipe (mandatory for Python 2.7.11)

opacam:


brussee:
I see you are copying code to set the right include and link paths that obviously only works for python 2.7 while the recipe also accepts python 3.5 from crystax.
```
        env = super(PyjniusRecipe, self).get_recipe_env(arch)
                          ' -lpython2.7'
```
Issues #668 and #669 point out the reason why I added this to my recipes, but I consider this a workaround and not good practice.

opacam:
A quick look at the affected recipes make me notice that all of them has the variable "call_hostpython_via_targetpython" set to False... so I guess, that modifying The function "PythonRecipe.get_recipe_env" (with an "if else" for python2), will make it work for all of them (and some others that has not been tested by me, or you)... I will make some tests with this when I have enough time...this will simplify things for all the recipes you mention and, possibly, for other recipes not tested

Thanks for letting me notice !!!

opacam:

PR #781, opened 2016-06-11
Pyasn1's recipe (mandatory for Python 2.7.11)

opacam:


inclement:
Does this recipe not already work? Why the change?

opacam:
As the numpy recipe (PR #782) without set the variable "call_hostpython_via_targetpython" to false, it fails to build when trying to import the _io module...so this is also needed to be merged if the PR #775 (Python 2.7.11) is finally accepted...otherwise is probably useless...I don't know... this is for the new pthon-2.7.11

opacam:

PR #782, opened 2016-06-11
Numpy's recipe (mandatory for Python 2.7.11)

opacam:


inclement:
Why is this change necessary or desirable? Are these linker args important?

opacam:
call_hostpython_via_targetpython: for the new python-2.7.11 throws ImportError: at build time when trying to import the _io module....wrong ELF class: ELFCLASS32

without the env['LDSHARED'] part: throws error when compiling C sources

So... for the new python 2.7.11... those changes are definitively needed for successfully build the numpy recipe, If the python 2-7-11, PR #775 is accepted, this PR should be merged, otherwise is probably useless....I don't know...this is for the new pthon-2.7.11

opacam:

PR #784, opened 2016-06-12
Fixes cyclic dependency between freetype and harfbuzz (optional for Python 2.7.11)

opacam:
The changes made in this PR should fix the issues with cyclic dependencies between freetype and harfbuzz. This PR has the words "optional for Python 2.7.11" to keep reference to the original PR #775 (Python 2.7.11), but actually it should work as an independent PR so... I think it could be merged safely without merging all the other stuff related with the "Python 2.7.11" update


opacam:

PR #693, opened 2016-03-27
Python 2.7.11 & Updates (Only for testing purposes)

opacam:
## Python 2.7.11 & Updates
[This is only for testing purposes, because, as pointed by @kived, it modifies a lot of stuff, so very soon it will be created some small PR's that will be easier to review and test, so... THIS PR IS NOT TO MERGE and it will be closed soon]

The new python2 recipe supports optional dependencies: sqlite3, openssl and libffi
- Updates the sqlite3's version for pygame's bootstrap
- Updates the jpeg's version for pygame's bootstrap
- Updates the freetype's version for pygame's bootstrap
- Updates the pil's version, adding freetype's support and solves jpeg's issues
- Solves conflicts with the distributed android libraries (openssl, freetype, harfbuzz), the trick is to link against the static libraries instead of linking with the shared.
- Fixes the pygame build (solves issue #687)
- Solves PIL does not compile with freetype2 support (issue #413) 
- The sdl recipe has been modified to avoid error not finding python (patch applied at build time)

Important notes:
- The last commits, changes the build for pygame bootstrap. The integrated libraries inside pygame bootstrap components,  jpeg, freetype and sqlite3, will not be used anymore... instead we will use the recipes from python-for-android because they are more updated, allow us to have more control, simplifies the code used in some recipes and also solves some issues related with the mentioned libraries.

My Build Specs:
- Tested with my app (with a lot of dependencies), using pygame and sdl2 bootstraps
- Android's NDK version: r10e-rc4
- Api version: 14
- Build System: Debian Stretch (amd64)

Thank to:
- Many Thanks to users: @brussee and @kived for his comments to helped me to polish things and improve the code.

- Also thanks to @dessant to encourage me to try to update from my previous pull request...actually it was quite easy starting from my previous python recipe (2.7.9).

- Also thanks all the developers who make possible this great project

-Update:
After the latest commits and some testing, I can say that the PR for python-2.7.11 works as expected for pygame and sdl2 and it should compile and run for everyone (I hope so...)

Note: Would like to mention that the issues that i found with sdl2 were caused by my old kivy's config file and some button calls triggered by "on_press" event (this two caused that i receive duplicated events for almost all events in my app). After modifying the  entry "input" for my kivy's config file and changing my calls "on_press" to "on_release"....all works as expected :))

PR #785, opened 2016-06-12
Adds functions to find png files when pygame is used (optional for Python 2.7.11)

opacam:
This PR has the words "optional for Python 2.7.11" to keep reference to the original PR #775 (Python 2.7.11).

### DIFF
--- a/pythonforandroid/recipes/png/__init__.py
+++ b/pythonforandroid/recipes/png/__init__.py
@@ -1,12 +1,36 @@
 from pythonforandroid.recipe import NDKRecipe
+from pythonforandroid.logger import shprint
+from pythonforandroid.toolchain import current_directory
+from os.path import join, exists
+import sh
 
 
 class PngRecipe(NDKRecipe):
-	name = 'png'
-	version = '1.6.15'
-	url = 'https://github.com/julienr/libpng-android/archive/{version}.zip'
+    name = 'png'
+    version = '1.6.15'
+    url = 'https://github.com/julienr/libpng-android/archive/{version}.zip'
 
-	generated_libraries = ['libpng.a']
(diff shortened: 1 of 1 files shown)


inclement:
I'm not clear what's going on with the `if 'pygame' in recipe_build_order' stuff - is this to do with a bootstrap/jni vs recipe build?

opacam:
Yes, You are right, let me clarify this stuff: all the recipes that are dependent on png (I think that those are recipes: pil and pygame) are linked to the build of png's recipe, but, when using pygame bootstrap, the library inside the app is the one builded inside jni (that has different version from the one from recipe), that fact causes errors at runtime, because the distributed png's library is different from the one who we linked at build time (for pygame and pil). With this PR I force to link the mentioned recipes (pil, pygame) against the final library used in the app, this way we avoids errors at runtime.

This is a dependency for PRs: #786 and #787 (I mention in those posts but not here, sorry about that)

PD: At some point I tried to force the build of png's recipe instead of the one distributed inside the "pygame_boostrap_components", as  I do in PR #787, but unfortunately, the newer version missed some function that prevents the build...so... we must stick with the old version distributed with "pygame_boostrap_components"

NOTE: @inclement...be careful with this PR (and all the PRs with python-2.7.11 in his title) because is not tested against the old recipes...all my test are against the new pytyhon 2.7.11 (PR #775) and all his manadatory and optional PRs, but maybe it will work with the current recipes: python2, pil pygame,pygame_boostrap_components....
I thought about a big PR #693, and at some point, the user @kived, notices that it was very difficult to review all the changes so I tried to split all the initial PR into little pieces to make it easy to review/test.

opacam:

PR #787, opened 2016-06-12
Pygame's recipe updated to use latest libjpeg-turbo and libfreetype (optional for Python 2.7.11)

opacam:
This PR has the words "optional for Python 2.7.11" to keep reference to the original PR #775 (Python 2.7.11). This depends on PRs: #784,  #785 and #793.

Here I force pygame to build against the p4a's libjpeg-turbo and p4a's libfreeetype instead of the ones distributed with the pygame_bootstrap_components. This way we avoid building all the stuff against different versions of libraries and the ones from the p4a are more updated. Unfortunately I cannot manage to use the same formula with libpng's recipe.

Merge note: Be careful with this because the `pygame_bootstrap_components/__init__.py` file also has been modified in the Python 2.7.11 (PR #775) and it is possible that a merge conflict happens, if this is finally merged.

### DIFF
--- a/pythonforandroid/recipes/pygame/__init__.py
+++ b/pythonforandroid/recipes/pygame/__init__.py
@@ -1,12 +1,11 @@
 
-from pythonforandroid.recipe import Recipe
-from pythonforandroid.util import current_directory, ensure_dir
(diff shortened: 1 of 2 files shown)


opacam:

PR #788, opened 2016-06-12
Sdl's recipe updated to use latest libjpeg-turbo and libfreetype (optional for Python 2.7.11)

opacam:
This PR has the words "optional for Python 2.7.11" to keep reference to the original PR #775 (Python 2.7.11). This depends on PR #784.

Here I force sdl to build against the p4a's libjpeg-turbo and p4a's libfreeetype instead of the ones distributed with the sdl package. Unfortunately I cannot manage to use the same formula with libpng's recipe.


opacam:

PR #792, opened 2016-06-16
Fixes the include and link paths for python2 (mandatory for Python 2.7.11)

opacam:
Fixes the right include and link paths for python2's recipes when variable "call_hostpython_via_targetpython" set to False. This will fix issues #668 and #669 with the new python2's (Python 2.7.11) but should work with the old python2's recipe. Thanks to @brussee who pointed this  in PR #778.


opacam:
Buuuuf, I cloned from the wrong branch...sorry, those days I'm a little tired cause of my job and I make silly mistakes...now I will do it again from the master branch

PR #794, opened 2016-06-16
Added python modules check

inclement:
I'm not sure if this is a good idea, but can't hurt to ask and it would help with the regular problem of users not understanding error messages when modules aren't present (e.g. when buildozer hasn't installed them).

The code just checks if appropriate versions of each dependency are present, and prints clear errors if not. Currently if these are missing the user just gets an ImportError traceback.


inclement:
Per @tito on irc, he'd prefer this as a more generic check that could be applied in other projects too. I'm tempted to merge this anyway, since nothing will ever depend on it and it helps in the short term.

PR #798, opened 2016-06-20
Android Activities can resume with some or all of the UI object state…

### LINKED ISSUE #797: On resume crash in SDL2 bootstrap
This problem actually was noted by another reviewer in ticket #646, but I suspect they stopped seeing it because only certain situations seem to trigger it. Specifically, it appears that rotation and pause / resume can cause the PythonActivity to be destroyed and re-created, even if the app is still running in the background. The behavior is documented in the Android docs here:

https://developer.android.com/training/basics/activity-lifecycle/recreating.html

To avoid having a complete app restart, it appears that Android caches the app state using the Bundle instance and tries to restore the app state automatically on resume. This is why onCreate gets a Bundle object that sometimes is empty (new start), and sometimes is not (rotation change or resume).

Unfortunately, it appears the state saving behavior is "under the hood" and it would take a deep look into Android's code to know what state it saves and when. It does seem to differ from app to app, which is why the problem is not persistent and easily reproduced. 

When it does happen, you get the crash noted in #646, and tracking that down, you find that the problem is that when onCreate is called with a non-empty Bundle, it causes mLayout and mImageView to be restored from cache, so the attempt in showLoadingScreen to add mImageView to mLayout gives a 'duplicate view' error. 

I'm going to submit a pull request with a fix for this. I didn't discover a way to check if a view exists, and I wasn't sure if doing a try block with removeView would have any side-effects, so I went with the solution of only setting mImageView to mLayout if we don't have an existing mImageView. The potential side-effect is that resume may not show the splash screen, but resume startup won't have the potential unzip slowdown, and the Bundle resuming should reduce the loading time on resume at least somewhat. 

Anyway, I'm open to a different or better approach if anyone knows of one. :) 

Also, I think #730 and #732 are affected by recreating the PythonActivity based on their problem descriptions, but I do not know this for certain.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/35332265-on-resume-crash-in-sdl2-bootstrap?utm_campaign=plugin&utm_content=tracker%2F11371&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F11371&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kollivier:
… cached in the Bundle instance, so it's possible for showLoadingScreen to be called when mLayout is already created and has the mImageView attached. Only attempt to assign to mLayout if there isn't already an mImageView. This fixes #797. Also, don't try to create the loading screen twice on startup.

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonActivity.java
+++ b/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonActivity.java
@@ -61,7 +61,6 @@ public class PythonActivity extends SDLActivity {
         super.onCreate(savedInstanceState);
         Log.v(TAG, "Did super onCreate");
 
-        this.showLoadingScreen();
         this.mActivity = this;
 
         String mFilesDirectory = mActivity.getFilesDir().getAbsolutePath();
@@ -325,6 +324,8 @@ public class PythonActivity extends SDLActivity {
       // 1. if the image is valid and we don't have layout yet, assign this bitmap
       // as main view.
       // 2. if we have a layout, just set it in the layout.
+      // 3. If we have an mImageView already, then do nothing because it will have
+      // already been made the content view or added to the layout.
       if (mImageView == null) {
         int presplashId = this.resourceManager.getIdentifier("presplash", "drawable");
         InputStream is = this.getResources().openRawResource(presplashId);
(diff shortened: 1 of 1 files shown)


inclement:
Thanks :+1:

PR #799, opened 2016-06-20
Copylibs fix

kollivier:
BSD variants of CP, like on OS X, do not support the -t argument, so just pass the target directory as the final argument instead.


inclement:
Makes sense, thanks.

PR #805, opened 2016-06-26
Add generic recipe for C++ extensions using STL and recipes for Enaml, Atom, Kiwisolver

frmdstryr:


inclement:
Great, thanks. It might be better to add the cxxstl stuff to the JNI build rather than manually copying the .so, but if it works for now that doesn't need to hold up merging.

PR #815, opened 2016-07-03
Enable launcher for pygame

KeyWeeUsr:
If there's no `main.py/main.pyo` in the command, the script will exit. However, for launcher you _can't_ put any of that into the command and therefore with the current state will exit. This will let you go and launcher is built.


inclement:
Great, thanks!

PR #816, opened 2016-07-03
Add docutils recipe

KeyWeeUsr:
<s>Not working yet.</s>

### DIFF
--- a/pythonforandroid/recipes/docutils/__init__.py
+++ b/pythonforandroid/recipes/docutils/__init__.py
@@ -1,0 +1,57 @@
+from pythonforandroid.toolchain import Recipe, shprint, current_directory
+from os.path import exists, join
+import sh
+
+
+class DocutilsRecipe(Recipe):
+
+    url = ('http://prdownloads.sourceforge.net/docutils/'
+           'docutils-{version}.tar.gz')
+    # md5sum = '4f3dc9a9d857734a488bcbefd9cd64ed'    add later
+    site_packages_name = 'docutils'
+    depends = ['pil', ]
+    version = '0.9.1'
+    conflicts = []
+
+    def should_build(self, arch):
+        name = self.site_packages_name
+        if name is None:
+            name = self.name
+        if self.ctx.has_package(name):
+            shprint(sh.echo, 'Python package already exists in site-packages')
+            return False
+        shprint(sh.echo,
+                '{} apparently isn\'t already in site-packages'.format(name))
+        return True
+
+    def prebuild_arch(self, arch):
+        super(DocutilsRecipe, self).prebuild_arch(arch)
+        return True
+
+    def build_arch(self, arch):
+        super(DocutilsRecipe, self).build_arch(arch)
+
+        with current_directory(self.get_build_dir(arch.arch)):
+            env = super(DocutilsRecipe, self).get_recipe_env(arch)
+            env['LDFLAGS'] = ' '.join([env['LDFLAGS'],
+                                      '-L'+self.ctx.get_libs_dir(arch.arch)])
+            hostpython = sh.Command(self.ctx.hostpython)
+
+            if self.ctx.python_recipe.from_crystax:
+                env['LDSHARED'] = env['CC'] + ' -shared'
+            else:
+                env['LDSHARED'] = join(self.ctx.root_dir,
+                                       'tools', 'liblink.sh')
+
+            shprint(hostpython, 'setup.py', 'build_ext')
+            shprint(sh.Command('find'), '.', '-name', "'*.pyx'",
+                    '-exec', 'cython', '{}', ';')
+            shprint(hostpython, 'setup.py', 'build_ext', '-v')
+            shprint(hostpython, 'setup.py', 'install', '-O2')
+            env['LDSHARED'] = None
+
+    def postbuild_arch(self, arch):
+        super(DocutilsRecipe, self).build_arch(arch)
+        return True
+
+recipe = DocutilsRecipe()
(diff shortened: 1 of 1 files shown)


akshayaurora:
docutils is pure python? Pure python libs don't need a recipe.

KeyWeeUsr:
This one had in old toolchain and I had some funny outputs trying to use it with the new one, but I think I almost got it done.

akshayaurora:
it was in old toolchain, but we don't need any recipe for it anymore. It should just work like any other pure python library.

KeyWeeUsr:
http://pastebin.com/NLw9Rtzq docutils & lxml were the issues

inclement:
In the new toolchain, this functionality is largely abstracted to CythonRecipe (so you can inherit from that and then don't need to manually change the docutils one in particular). Would you be able to try that instead?

KeyWeeUsr:
I don't think this will be a good idea though it'd definitelly look better, because of this:
```
  File "/usr/local/bin/p4a", line 9, in <module>
    load_entry_point('python-for-android==0.4', 'console_scripts', 'p4a')()
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/toolchain.py", line 775, in main
    ToolchainCL()
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/toolchain.py", line 450, in __init__
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/toolchain.py", line 142, in wrapper_func
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/toolchain.py", line 185, in build_dist_from_args
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/build.py", line 561, in build_recipes
    recipe.build_arch(arch)
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/recipe.py", line 958, in build_arch
    self.build_cython_components(arch)
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/recipe.py", line 1004, in build_cython_components
    shprint(sh.find, build_lib[0], '-name', '*.o', '-exec',
IndexError: list index out of range
```

Which would lead to either putting some empty `.pyx` file there or rewriting the build function, which is what I did manually (I think?). But as @akshayaurora said, pure python should work with the new toolchain, so even if the recipe make it work, maybe better approach would be to find what's wrong with pure python installation? If you feel so, for now it might be good to fix it temporarily with recipe and then revert it.

inclement:
If there really aren't any cython files, it doesn't need the cython stuff at all.

KeyWeeUsr:
Should I also remove parts where it only contains `super()` and `return True`, or are those necessary to activate the recipe?

inclement:
You can remove that part too.

Does docutils actually have external components to build? If so, it doesn't look like you're passing the android env, which would mean they wouldn't work on Android.

KeyWeeUsr:
Docutils work(s) with `pil`, `pygments`(code_analyzer.py), `roman`, and `lxml`(for odt), but only lxml support is missing now in p4a and remaining stuff either has a recipe or is pure python. For some reason however, it wouldn't work without the recipe and throwed an error before. Now I reinstalled it again from git and it builds with docutils and also runs(just switch Showcase -> RstDocument, there are imports which crash the app without docutils). Were there some changes recently? O.o

Could someone confirm that it works without the recipe, so I could close this PR then? If doesn't, I push edited recipe anyway.

inclement:
Since you aren't passing the android env to the shprint calls, I suspect the build_ext isn't actually important (if it were, it would need the android env to compile properly). Does it work if you use a PythonRecipe and remove both methods (just letting the defaults be used), or failing that a CompiledComponentsPythonRecipe?

Sorry to keep querying what's ultimately a fairly simple thing, it's just that I've been making an effort avoid recipes becoming overcomplicated as has happened in the past.

KeyWeeUsr:
I'm not sure if I understood correctly, but if you meant this:
```
from pythonforandroid.toolchain import Recipe, shprint, current_directory
from os.path import exists, join
import sh

class DocutilsRecipe(Recipe):

    url = ('http://prdownloads.sourceforge.net/docutils/'
           'docutils-{version}.tar.gz')
    # md5sum = '4622263b62c5c771c03502afa3157768'
    site_packages_name = 'docutils'
    depends = ['pil', ]  # 'lxml', ]

    version = '0.12'
    conflicts = []

recipe = DocutilsRecipe()
```

then no, the recipe only prints to console this:

    [INFO]:    Building docutils for armeabi

and jumps right to the next recipe.

inclement:
By using a PythonRecipe or CompiledComponentsPythonRecipe, I mean inherit from these classes instead of Recipe. You can see the class definitions at https://github.com/kivy/python-for-android/blob/master/pythonforandroid/recipe.py#L701 . These recipe types incorporate the commands for normal Python recipes (or with compiled components, calling build_ext), whereas Recipe just doesn't do anything.

KeyWeeUsr:
Ah, my bad. Changing the `Recipe` to `PythonRecipe` works as expected, docutils package is available and works with RstDocument. Does it mean that PythonRecipe == installing in through pip but "manually"? As I mentioned sooner, it built even without recipe after update, so... does it work for you without any recipe?

inclement:
You can read the source of PythonRecipe in the above link. It calls setup.py install.

KeyWeeUsr:
I updated the recipe, but I don't think it'll be necessary as it works for me even without it. If there were someone that wouldn't be able to build with `docutils`, this might help (and be reopened).

inclement:
Okay, thanks for you efforts!

PR #817, opened 2016-07-04
Add docs for launcher

KeyWeeUsr:
ref https://github.com/kivy/kivy/issues/2706

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -32,6 +32,7 @@ Contents
    recipes
    bootstraps
    services
+   launcher
    apis
    troubleshooting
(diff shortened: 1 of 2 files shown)


inclement:
Thanks again. I made a few changes before merging, feel free to submit another PR proposing something else if you don't like them.

KeyWeeUsr:
Nah, seems good to me. I shouldn't probably include that "advantages" lines as it has nothing with building the launcher. However, I still think listing the packages might be useful, but if it will be always in the descriptioin of stable launcher, then it sounds good to me too. :+1:

PR #818, opened 2016-07-05
Fix build errors in docs

KeyWeeUsr:
~31 errors purged to 2:
```
p4a\doc\source\commands.rst:17: WARNING: autodoc: failed to import class u'ToolchainCL' from module u'toolchain'; the module executes module level statement and it might call sys.exit().
p4a\doc\source\recipes.rst:488: WARNING: autodoc: failed to import class u'Recipe' from module u'toolchain'; the module executes module level statement and it might call sys.exit().
```
 Most of the errors came with `ext/sphinx_rtd_theme` because it used non-existing `math` extension. Then some broken links, underlines, missing `_static` etc.

I took out `ext/*` part of the docs out because it's just an example from sphinx and shouldn't be available in the final documentation. It isn't included in pdf or in the website, but if you build a local docs, then it'll be there.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -82,7 +82,7 @@ language = None
 
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
-exclude_patterns = []
+exclude_patterns = ['ext/*', ]
 
 # The reST default role (used for this markup: `text`) to use for all
 # documents.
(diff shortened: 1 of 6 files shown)


inclement:
Great, thanks. I've gone further and removed the 'related' doc page, since it wasn't really worth the space.

KeyWeeUsr:
I wasn't really sure about removing it, so I left it there, hehe.

PR #838, opened 2016-07-14
introduce versioning for openssl

brussee:
On Android versions lower than 6.0 the openssl library of the system is being loaded instead of the one provided with the app.
Adding the version to the name of the shared library file avoids the problem of the linker at runtime giving the wrong version of the library and provides a clear indication of what version you are using (good ensuring you use the latest version for security).


brussee:
Without versioning m2crypto failed to load.

inclement:
Looks good, thanks. Do you know why this wasn't happening before, is the system openssl a new addition?

brussee:
It appears that unknowingly the system openssl was being used and cyanogenmod had a newer version installed than stock. I discovered this when the specific missing symbol turned out to be a new addition going from 1.0.1.l to 1.0.2.
So apparently stock Android 6.0.1 has openssl 1.0.2 and Android 5.1.1 and lower have openssl 1.0.1.

PR #845, opened 2016-07-19
Improve service_only bootstrap

brussee:
* Add option to have a different icon for the python service.
 * Bugfix bootstrap Gradle build script
 * Binder interface

Because it will be shown in the notification bar in black and white you likely want to use a different icon than the colored app one, since Android forces that one to black and white otherwise (ugly).

### DIFF
--- a/pythonforandroid/bootstraps/service_only/build/src/org/kivy/android/PythonService.java
+++ b/pythonforandroid/bootstraps/service_only/build/src/org/kivy/android/PythonService.java
@@ -1,18 +1,16 @@
 package org.kivy.android;
 
-import android.app.Activity;
 import android.app.PendingIntent;
 import android.app.Service;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.ApplicationInfo;
 import android.os.Bundle;
-import android.os.IBinder;
 import android.os.Process;
 import android.support.v4.app.NotificationCompat;
 import android.util.Log;
 
-public class PythonService extends Service implements Runnable {
+public abstract class PythonService extends Service implements Runnable {
     private static String TAG = PythonService.class.getSimpleName();
 
     /**
@@ -41,14 +39,6 @@ public class PythonService extends Service implements Runnable {
 
     public boolean getAutoRestart() {
         return false;
-    }
-
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public IBinder onBind(Intent intent) {
-        return null;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


brussee:
Now also contains a bugfix to install the *.gradle build scripts included with the service_only bootstrap: https://github.com/kivy/python-for-android/pull/845/commits/db26263b04a83d70fe4aac1c6c9ebcededbb7382

pythonforandroid/bootstraps/service_only/build/src/org/kivy/android/PythonService.java
```diff
@@ -136,7 +131,6 @@ public void onDestroy() {
     @Override
     public void run() {
         PythonUtil.loadLibraries(getFilesDir());
-        mService = this;
```
inclement: Why remove the mService stuff? Is this because it no longer makes sense when multiple services can run at once? (Honest question, I haven't used the services).

Looks good otherwise, thanks!

brussee: Exactly. Unless you need something to be a singleton, using static variables to hold instances of classes is not what you want.

inclement: I'm concerned that some users may be using a single service and relying on this method of accessing it, as that's the way the old toolchain worked. Would you be terribly disappointed if I merged this but added mService back in for now (at least until after it's clearly deprecated)?

brussee: I understand your concern because of the change of behavior with regards to the old tool-chain.
The solution to this is a binder: https://developer.android.com/guide/components/bound-services.html

Example from http://stackoverflow.com/a/24541584
```
import android.app.Service;
import android.content.Intent;
import android.os.IBinder;

public class MyService extends Service {
    private Binder binder;  

    @Override
    public void onCreate() {
        super.onCreate();
        binder = new Binder();
    }

    @Override
    public IBinder onBind(Intent intent) {
        return binder;
    }

    public class Binder extends android.os.Binder {
        public MyService getService() {
            return MyService.this;
        }
    }
}
```

brussee:
> I will add this to my service_only bootstrap, ...

Done! Enjoy!

inclement:
Yes, that's fine. On that note, would you be interested in adding some doc for it in the bootstraps page?

brussee:
Yes, in about three weeks I'll be delivering my MSc thesis of which my contributions to p4a are a part.

brussee:
Sorry for the delay. I'll try to write documentation about how use this bootstrap at some point, especially how to work with the .gradle build template.

PR #350, opened 2015-03-26
unpack into subdirectories __kivy_root__ of private and external storage...

denys-duchier:
....

Since unpacking performs first a recursiveDelete, everything stored in an
unpack root is lost after an app update.  In order for e.g. a database on
private storage to persist across app updates it was necessary to make
the private unpack root a subdirectory of private storage rather than
the top level.

PR #852, opened 2016-07-25
move app dir to `<private_dir>/app` allowing for re-installation/upda…

akshayaurora:
…tion to not delete the user data.

This is only working for sdl2 bootstrap, still need to test with service and other bootstraps.

This is based on https://github.com/kivy/python-for-android/pull/350 

We need to have the following fix in kivy with this pr. https://github.com/kivy/kivy/pull/4490

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/build/jni/src/start.c
+++ b/pythonforandroid/bootstraps/sdl2/build/jni/src/start.c
@@ -156,7 +156,7 @@ int main(int argc, char *argv[]) {
   if (dir_exists("lib")) {
     /* If we built our own python, set up the paths correctly */
     LOGP("Setting up python from ANDROID_PRIVATE");
-    PyRun_SimpleString("private = posix.environ['ANDROID_PRIVATE']\n"
+    PyRun_SimpleString("private = posix.environ['ANDROID_PRIVATE'] + '/app'\n"
                        "argument = posix.environ['ANDROID_ARGUMENT']\n"
                        "sys.path[:] = [ \n"
                        "    private + '/lib/python27.zip', \n"
--- a/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonActivity.java
+++ b/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonActivity.java
@@ -48,14 +48,21 @@ public class PythonActivity extends SDLActivity {
     private Bundle mMetaData = null;
     private PowerManager.WakeLock mWakeLock = null;
 
+    public String getKivyRoot() {
+        String kivy_root =  getFilesDir().getAbsolutePath() + "/app";
+        return kivy_root;
+    }
+
+
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         Log.v(TAG, "My oncreate running");
         resourceManager = new ResourceManager(this);
         this.showLoadingScreen();
+        File kivy_root_file = new File(getKivyRoot());
 
         Log.v(TAG, "Ready to unpack");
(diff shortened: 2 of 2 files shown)


inclement:
Great. I'll check it at the weekend if it can wait a couple more days.

pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonActivity.java
```diff
@@ -48,29 +48,37 @@
     private Bundle mMetaData = null;
     private PowerManager.WakeLock mWakeLock = null;
 
+    public String getKivyRoot() {
```
inclement: Might be nice to use getAppRoot or something instead of Kivy, but it doesn't really matter.

ecdsa:
Thank you!

brussee:
I solved this by putting the user data in a different folder.
I agree putting the p4a files into a different folder would be better. However I think this should not be a subfolder of /files because the user can clear the files and essentially the p4a files are app src code, not data files.
The same goes for the libs. There is a separate data folder on Android to store app libs, also private afaik.

akshayaurora:
> However I think this should not be a subfolder of /files because the user can clear the files 

The default behavior should be to allow the user the private data if he want's to.
If the app developer want's the data to survive, then they are free to use a custom dir just like you did, no?

brussee:
> The default behavior should be to allow the user the private data if he want's to.

Yes, I agree, but I consider the p4a private files _not_ to be user data, so I would put those in a different folder.
So, in my case I should swap the folders: put the user data back in the files directory and the p4a private files in another, outside of the files directory that is currently in use by the p4a private files.

akshayaurora:
> So, in my case I should swap the folders: put the user data back in the files directory and the p4a private files in another, outside of the files directory that is currently in use by the p4a private files.

So currently with this pr,  <private_dir>/files is where user can put their data (create a <private_dir>/files/data folder if they want to.) The app sources and files goto <private_dir>/files/app libs reside in <private_dir>/files/app/libs

Only the files/apps folder is deleted when we need to update the app.

p4a app files are not with the data.

brussee:
I just tested it and everything in the private dir is deleted if the user presses the "Clear data" button. So my assumption was incorrect: I thought only the /files contents would be deleted by that.
The conclusion is that it really doesn't matter where the p4a private dir and libs are located, they all are lost.

akshayaurora:
> The conclusion is that it really doesn't matter where the p4a private dir and libs are located, they all are lost.

Yup, and that's how it should be,  if the dev wants to prevent this behavior then they should choose a separate dir manually like /sdcard/<xyz> but issue with that would be it is not protected/private by the os the developer may have to manage that themselves..

pythonforandroid/bootstraps/sdl2/build/jni/src/start.c
```diff
@@ -156,7 +156,7 @@ int main(int argc, char *argv[]) {
   if (dir_exists("lib")) {
     /* If we built our own python, set up the paths correctly */
     LOGP("Setting up python from ANDROID_PRIVATE");
-    PyRun_SimpleString("private = posix.environ['ANDROID_PRIVATE']\n"
+    PyRun_SimpleString("private = posix.environ['ANDROID_PRIVATE'] + '/app'\n"
```
inclement: This should actually use ANDROID_APP_PATH rather than guessing /app, right? If that's right, just let me know and I'll make the change manually and merge.

akshayaurora: 👍

PR #857, opened 2016-07-31
Check if bdist_dir exists before removing

nocarryr:
First call to `setup.py apk` fails since the build dir would not have been previously created:
```
error: [Errno 2] No such file or directory: 'build/bdist.android-armeabi'
```


inclement:
Good catch, thanks!

PR #862, opened 2016-08-10
enable recipe package top-level directory names to be the same as the…

cidermole:
… package name

For my own custom package `package-name`, I created a package zip file with a top-level directory `package-name`. I got an error in `File "pythonforandroid/recipe.py", line 437, in unpack` because a move of the same directory from a relative to an absolute path is attempted. Taking `basename()` first checks if the directory name is already the same as it should be.


inclement:
Thanks.

PR #863, opened 2016-08-12
libzmq recipe: do not hardcode g++ version

chaubold:
The latest android NDK (r12b) doesn't ship gcc 4.8 any more for cross compilation but switched to 4.9, and thus building the `libzmq` recipe fails: it cannot locate `libgnustd_shared.so`.

_(On a sidenote: I am facing a similar bug to #234 in `_clock.so`, but that shouldn't be related to my changes to the recipe.)_


inclement:
Great, thanks!

PR #869, opened 2016-08-26
fix issue with loading_screen

akshayaurora:
Used to get a blank loading screen without this.


ecdsa:
I still get a blank (well, black) loading screen after this commit

PR #872, opened 2016-08-29
fix bug preventing pure python dependencies to be installed under pyt…

loulouxiv:
…hon3


loulouxiv:
I did not mean to add the second commit to the pull request. Please note that while the java modification compiles, I am not 100% sure that it does what the previous code meant.

inclement:
Nice catch, thanks for the PR. I'd rather not merge these unrelated changes with it, but I'll just cherry-pick the first commit if you don't mind.

loulouxiv:
No I don't mind, I will simply add these to another PR

inclement:
Thanks, I've merged just the fix commit.

Issue #864, opened 2016-08-14
p4a generates deprecated code under Android API 23


hackalog:
p4a is generating a deprecated call - `setLatestEventInfo()` in PythonService.java - when used with API 23:

```
-compile:  
    [javac] Compiling 23 source files to /home/hackalog/.local/share/python-for-android/dists/unnamed_dist_1/bin/classes 
    [javac] /home/hackalog/.local/share/python-for-android/dists/unnamed_dist_1/tmp-src/org/kivy/android/PythonService.java:96: error: cannot find symbol    
    [javac]         notification.setLatestEventInfo(context, serviceTitle, serviceDescription, pIntent);    
    [javac]                     ^   
    [javac]   symbol:   method setLatestEventInfo(Context,String,String,PendingIntent)  
    [javac]   location: variable notification of type Notification  
    [javac] Note: Some input files use or override a deprecated API.    
    [javac] Note: Recompile with -Xlint:deprecation for details.    
    [javac] 1 error 
```
##

brussee:
This was already fixed in the service_only bootstrap: 6b30ca35b79e6fb7c0109738c733ebcba7141df9

PR #873, opened 2016-08-31
replace calls to setLatestEventInfo, which was removed in android23 API

loulouxiv:


inclement:
Looks great, thanks. I'd like to check it before merging but should be able to at the weekend.

Also just to note for bookkeeping: fixes https://github.com/kivy/python-for-android/issues/864

brussee:
You can also copy the solution from the service_only bootstrap: https://github.com/kivy/python-for-android/blob/master/pythonforandroid/bootstraps/service_only/build/src/org/kivy/android/PythonService.java#L94

PR #886, opened 2016-09-15
Fix clean_download_cache, add deleting of separate packages

KeyWeeUsr:


pythonforandroid/toolchain.py
```diff
@@ -578,13 +578,30 @@ def clean_recipe_build(self, args):
...
 
         This does *not* delete the build caches or final distributions.
         '''
         ctx = self.ctx
-        if exists(ctx.packages_path):
-            shutil.rmtree(ctx.packages_path)
+        msg = "Download cache removed!"
+
+        if args.unknown_args:
+            for package in args.unknown_args[0].split(','):
```
inclement: Could you replace this with a `parser_clean_download_cache.add_argument('recipes')` when the parser is defined, instead? That will mean that the user gets a helpful argparse error message if they don't explicitly pass the intended argument. You could also just allow any number of arguments (I think for argparse something like `nargs='*'`) instead of requiring the comma separation.

PR #889, opened 2016-09-22
Enable SDL2 MP3 support for Android

AndreMiras:
We had to copy the external/smpeg2-2.0.0/ folder
to the jni/ folder to fix linker errors.
See https://groups.google.com/forum/#!topic/kivy-users/N4isIjQ1Ja8

### DIFF
--- a/pythonforandroid/recipes/sdl2_mixer/__init__.py
+++ b/pythonforandroid/recipes/sdl2_mixer/__init__.py
@@ -38,8 +38,6 @@ class LibSDL2Mixer(BootstrapNDKRecipe):
         Copies the smpeg2 external library to the jni/ folder.
         This is required by the linker.
         """
-        jni_dir = self.get_jni_dir()
-        build_dir = self.get_build_dir(arch)
         self._copy_smpeg2_to_jni()
         super(LibSDL2Mixer, self).prebuild_arch(arch)
 
--- a/pythonforandroid/recipes/sdl2_mixer/__init__.py
+++ b/pythonforandroid/recipes/sdl2_mixer/__init__.py
@@ -1,3 +1,8 @@
+from os.path import join
+
+import sh
+
+from pythonforandroid.logger import shprint
 from pythonforandroid.toolchain import BootstrapNDKRecipe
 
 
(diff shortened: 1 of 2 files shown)


AndreMiras:
I wrote "SDK2" in the commit but I meant "SDL2".
It adds MP3 support to Android. It was a lot of investigations because I'm new to the Android toolchain and all, but it doesn't actually do much.
It seems the "external" library must be copied to the jni/ folder if we want it to be linked. This is what it does.

pythonforandroid/recipes/sdl2_mixer/__init__.py
```diff
@@ -6,6 +11,34 @@ class LibSDL2Mixer(BootstrapNDKRecipe):
     url = 'https://www.libsdl.org/projects/SDL_mixer/release/SDL2_mixer-{version}.tar.gz'
     dir_name = 'SDL2_mixer'
 
-    patches = ['toggle_modplug_mikmod_smpeg_ogg.patch']
+    patches = ['toggle_modplug_mikmod_ogg.patch']
+
+    def _external_dir(self):
```
Fak3: Why do you split it into 4 methods? I would be more clear if you put it right in prebuild_arch()

AndreMiras: Hi @Fak3, thanks for your feedback.
I used "inline methods" because I find it more clear to read, but this is my point of view.
For example when I read the `prebuild_arch()` method. On first sight to the `self._copy_smpeg2_to_jni()` line, I can quickly say, OK, the `prebuild_arch()` just copies the smpeg2 lib to the jni directory. And if I want to see in more details how this was achieved, I can step deeper into the stack.
On the other hand, if I had written:
```
external_dir = join(self.get_build_dir(None), 'external')
smpeg2_external_dir = join(self._external_dir(), 'smpeg2-2.0.0')
jni_dir = self.get_jni_dir()
shprint(sh.cp, '-r', smpeg2_external_dir, jni_dir)
super(LibSDL2Mixer, self).prebuild_arch(arch)
```
I fell like it takes more time to understand what the method just does, event though it's a very simple method. It's a little bit like the [import antigravity](https://xkcd.com/353/) thing.
It's also a way to avoid commenting each single line knowing that comments are future lies.
But again this my point of view and if it's an issue I can do another pull request, merging all these lines into a single method with some comments.
Does the fix work for you by the way?

Issue #890, opened 2016-09-24
onResume deadlock with pyjnius/pygame/sdl on android


hottwaj:
It seems that interaction between android.pyx, pyjnius and app pause mode can cause deadlock when going into pause mode or when resuming from pause mode.  I am using the pygame bootstrap.

A few logs and things below to show what's going on

Here's the typical sequence - first I press the home button and the app should pause.

```
I/python  ( 4288): [INFO   ] [Android     ] Must go into sleep mode, check the app
I/python  ( 4288): [INFO   ] [Android     ] App paused, now wait for resume.
I/System.out( 4288): python pyjnius java:invoke(<proxy>, public abstract void android.view.ViewTreeObserver$OnGlobalLayoutListener.onGlobalLayout(), [Ljava.lang.Object;@33d6fa5f)
```

Note that *after* the "app paused" message, the Java interface of pyjnius receives a call which it tries to push to Python.  The log message is from line 15 of NativeInvocationHandler.java of pyjnius

The above happens *before* I try to resume my app.

Once I try to resume, I get a black screen and nothing happens.  I have logging setup on the pygame bootstrap and I can see that the app's onResume() method in PythonActivity.java is never called.

I get the following log output

```
W/ActivityManager( 1005): Activity stop timeout for ActivityRecord{20c40984 u0 org.hottwaj.hottwajtest/org.renpy.android.PythonActivity t8207}
I/ActivityManager( 1005): Activity reported stop, but no longer stopping: ActivityRecord{20c40984 u0 org.hottwaj.hottwajtest/org.renpy.android.PythonActivity t8207}
W/ActivityManager( 1005): Activity pause timeout for ActivityRecord{20c40984 u0 org.hottwaj.hottwajtest/org.renpy.android.PythonActivity t8207}
W/ActivityManager( 1005): Activity stop timeout for ActivityRecord{20c40984 u0 org.hottwaj.hottwajtest/org.renpy.android.PythonActivity t8207}
I/ActivityManager( 1005): Killing 4288:org.hottwaj.hottwajtest:python/u0a281 (adj 11): remove task
I/WindowState( 1005): WIN DEATH: Window{2a469087 u0 org.hottwaj.hottwajtest/org.renpy.android.PythonActivity}
W/WindowManager( 1005): Force-removing child win Window{200ab265 u0 SurfaceView} from container Window{2a469087 u0 org.hottwaj.hottwajtest/org.renpy.android.PythonActivity}

```

[...]

PR #892, opened 2016-09-25
fixes for #890 (deadlock on pause/resume with pyjnius/pygame)

hottwaj:
Moved java LayoutListener logic out of 'android' package into pygame bootstrap

This avoids deadlock due to LayoutListener locking python thread with pyjnius when app onPause/onResume occurs

### DIFF
--- a/pythonforandroid/bootstraps/pygame/build/src/org/renpy/android/PythonActivity.java
+++ b/pythonforandroid/bootstraps/pygame/build/src/org/renpy/android/PythonActivity.java
@@ -8,10 +8,12 @@ import android.os.Environment;
 import android.view.KeyEvent;
 import android.view.Window;
 import android.view.WindowManager;
+import android.view.ViewTreeObserver;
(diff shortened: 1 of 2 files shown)


inclement:
Looks good I think, but I'm just making sure I understand the change.

hottwaj:
Hey @akshayaurora, that does look like it will solve the deadlock issue as it avoids a callback to python from the java thread (I haven't tested) that could be called when the app is paused.

It looks like your method does more work on each call to get_keyboard_height(), whereas the layout listener i moved into Java code in dbccfe6 is reactive - it only does work when the global layout changes

I assumed that the "reactive"-ness property was why the code in android.pyx was set up with the LayoutListener logic in the first place, though I don't know how often get_keyboard_height() is called and whether performance is actually an issue

In summary, as long as the deadlock issue is solved and performance is not affected I'm happy with either solution!

Cheers & thanks for taking a look at this.

PR #896, opened 2016-10-08
Fix bdistapk for launcher

KeyWeeUsr:
This allows `--launcher` switch to work with the setup.py thing for creating apps.


inclement:
Thanks, and sorry for the delay.

PR #901, opened 2016-10-09
Made clean-recipe-build delete dists

inclement:


KeyWeeUsr:
This is very nice addition :) Maybe even including a check before cleaning dists could be good i.e. remove only those dists that use that particular recipe, not all of them.

inclement:
@KeyWeeUsr I thought about that, but then I thought...nah, I'll just merge it. It would be a great feature addition though.

Issue #819, opened 2016-07-06
Can't install on Windows using pip


NudeCaveMan:
When trying to install on Windows, i get:

Collecting sh>=1.10 (from python-for-android)
  Using cached sh-1.11.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 20, in <module>
      File "C:\Users\CaveMan\AppData\Local\Temp\pip-build-11cs8l6t\sh\setup.py",
 line 4, in <module>
        import sh
      File "C:\Users\CaveMan\AppData\Local\Temp\pip-build-11cs8l6t\sh\sh.py", li
ne 37, in <module>
        support." % __version__)
    ImportError: sh 1.11 is currently only supported on linux and osx. please in
stall pbs 0.110 (http://pypi.python.org/pypi/pbs) for windows support.

Install pbs instead sh does nothing.

KeyWeeUsr:
Even if you managed to install it on Windows, it won't work, because it's set up for linux environment. All recipes that have additional console script have `.sh`(bash) file. I'm not sure about this, but you might be able to run them with console included in "GitHub for Windows" but that's pretty much it what you can do on Windows with it right now.

For packaging use the `.vdi` file from [kivy website](https://kivy.org/#download)(Virtual Machine) which has at least the required environment (haven't seen it for a while, don't know what is and what isn't set up).

inclement:
I'd like to be installable on windows even if p4a can't be run beyond the trivial. This will be a good starting point for any binary distribution stuff, like the possible PR at https://github.com/kivy/python-for-android/pull/804

PR #902, opened 2016-10-10
Allow installation on Windows

ethanhs:
**Note**: this does _not_ allow one to use the full Python for Android toolchain. It simply removes the requirement for the sh module on Windows so that it is pip installable.

This addresses #819. I am not sure if a warning about the limited abilities is required, but this does allow one to pip install or use setup.py install. When I run p4a (for the curious), I get:

```
C:\Users\ethanhs>p4a
ERROR: The sh Python module could not be found, please install version 1.10 or higher
python-for-android is exiting due to the errors above.
```

Something that I could modify for this PR is to change [`check_python_dependencies`](https://github.com/kivy/python-for-android/blob/master/pythonforandroid/toolchain.py#L12) to not error on Windows, and instead add a message to wait for binary redistributable recipes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,6 +14,14 @@ package_data = {'': ['*.tmpl',
 
 data_files = []
 
+
+if os.name == 'nt':
+    install_reqs = ['appdirs', 'colorama>=0.3.3', 'jinja2',
+                        'six']
+else:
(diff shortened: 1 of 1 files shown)


inclement:
Thanks for the PR, I guess it's harmless to include it even though p4a doesn't run.

PR #904, opened 2016-10-12
Easy way to set the presplash background color

mrhdias:
Easy way to set the presplash background color of Loading Screen.

--presplash-color=#AARRGGBB, --presplash-color=#RRGGBB or --presplash-color=red

Default color: #FFFFFFFF (white)

More information:
https://developer.android.com/reference/android/graphics/Color.html
Parse the color string, and return the corresponding color-int. If the string cannot be parsed, throws an IllegalArgumentException exception. Supported formats are: #RRGGBB #AARRGGBB or one of the following names: 'red', 'blue', 'green', 'black', 'white', 'gray', 'cyan', 'magenta', 'yellow', 'lightgray', 'darkgray', 'grey', 'lightgrey', 'darkgrey', 'aqua', 'fuchsia', 'lime', 'maroon', 'navy', 'olive', 'purple', 'silver', 'teal'.

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonActivity.java
+++ b/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonActivity.java
@@ -30,6 +30,7 @@ import android.widget.ImageView;
 import java.io.InputStream;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
+import android.graphics.Color;
 
 import org.libsdl.app.SDLActivity;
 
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
There are `.xml.tpl` (or something like that) template files. If you edit them to fetch `{{ variable }}` and put it into `build.py` file of a desired bootstrap, it'll do the same. Hint: it's the `render()` function.

mrhdias:
Thanks for the help, I make it work as you said.

inclement:
Sorry for the delay testing this, but it looks good, thanks! I've switched the default to black to match the old behaviour though.

PR #908, opened 2016-10-16
fixed recipe zope_interface

goffi-contrib:
zope_interface was not installing "common" hierarchy because it
couldn't find setuptools and was using distutils instead.

This commit fix this by using hostpython own build dir, and also
removed unused imports (shprint and join).


inclement:
Thanks!

PR #909, opened 2016-10-16
Fixed app path for SDL2 services

inclement:
Fixes SDL2 services by setting the correct path to the app, which was broken following the recent move to the app subdir.

This is more hardcoded than I'd like, I'll try to check out if we can use the PythonActivity method instead (maybe it should be a class method) to avoid the same breaking problems in the future.


akshayaurora:
@inclement 🍶 🍷

PR #912, opened 2016-10-23
Empty: Adding .gitkeep in bootstraps/empty/build

thopiekar:
When experimenting around new recipes for python3, I personally needed empty to get something built.
The reason is that almost every boostrap needs python2, which conflicts with python3.
While using empty I noticed that the build directory is missing.

Using .gitkeep is not documented in git, but widely used as far as I read on the internet.


inclement:
Sounds good. I've been meaning to add a build recipe option to avoid the need for this anyway.

Issue #913, opened 2016-10-23
Python3: armeabi used to copy, but armeabi-v7a choosen


thopiekar:
```
thopiekar@home:/media/hdd/home/thopiekar/Projekte/GIT/thopiekar/python-for-android_fork$ LANG=C PYTHONPATH=. python -m pythonforandroid.toolchain create --dist-name sip-test --requirements sip --arch=armeabi-v7a --bootstrap empty --debug
[INFO]:    Will compile for the following archs: armeabi-v7a
[INFO]:    Found Android API target in $ANDROIDAPI
[INFO]:    Available Android APIs are (2, 3, 4, 5, 6, 7, 9, 11, 12, 13, 14, 23, 24, 25)
[INFO]:    Requested API target 23 is available, continuing.
[INFO]:    Found NDK dir in $ANDROIDNDK
[INFO]:    Got NDK version from $ANDROIDNDKVER
[INFO]:    Using Google NDK r10e
[INFO]:    Found virtualenv at /usr/bin/virtualenv
[INFO]:    Found the following toolchain versions: ['4.9']
[INFO]:    Picking the latest gcc toolchain, here 4.9
[INFO]:    No existing dists meet the given requirements!
[INFO]:    No dist exists that meets your requirements, so one will be built.
[INFO]:    Loaded recipe sip (depends on ['python3'])
[INFO]:    Loaded recipe python3 (depends on ['hostpython3'], conflicts ['python2'])
[INFO]:    Loaded recipe hostpython3 (depends on [], conflicts ['hostpython2'])
[INFO]:    Found a single valid recipe set (this is good)
[INFO]:    The selected bootstrap is empty
[INFO]:    # Creating dist with empty bootstrap
[INFO]:    Dist will have name sip-test and recipes (sip)
[DEBUG]:   -> running cp -r /media/hdd/home/thopiekar/Projekte/GIT/thopiekar/python-for-android_fork/pythonforandroid/bootstraps/empty/build /home/thopiekar/.local/share/python-for-android/build/bootstrap_builds/empty
[INFO]:    -> directory context /home/thopiekar/.local/share/python-for-android/build/bootstrap_builds/empty
[INFO]:    <- directory context /media/hdd/home/thopiekar/Projekte/GIT/thopiekar/python-for-android_fork
[INFO]:    Recipe build order is ['hostpython3', 'python3', u'sip']
[INFO]:    # Downloading recipes 
[INFO]:    Downloading hostpython3
[...]

PR #914, opened 2016-10-23
hostpython3: Removing hardcoded armeabi

thopiekar:
Compared the recipes for hostpython[2,3] and saw that hostpython2 doesn't use the 'armeabi' inside the function.
Probably just old code from the days of armeabi.

Possible fix for #913

### DIFF
--- a/pythonforandroid/recipes/hostpython3/__init__.py
+++ b/pythonforandroid/recipes/hostpython3/__init__.py
@@ -28,9 +28,9 @@ class Hostpython3Recipe(Recipe):
 
             if exists('hostpython'):
                 info('hostpython already exists, skipping build')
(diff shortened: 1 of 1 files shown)


thopiekar:
Any questions/feedback left?

inclement:
Sorry for the delay replying, I missed this PR.

It looks like the hostpython recipe doesn't have a get_build_dir that can take no arguments, so this would break it. The hostpython2 recipe defines it in the `__init__.py`. I think you're right that it causes your error though, although the hostpython3 recipe shouldn't actually be used and I realise now what's going on. I'll reply at #913.

inclement:
Closing per discussion at #913

PR #916, opened 2016-10-24
Make boost recipe compatible with Google NDK 13.0

brussee:
The ``ndk-dir`` argument of the ``make-standalone-toolchain.sh`` script is no longer supported and appears unnecessary because the script is located within the ndk-dir.


inclement:
Looks good then, thanks.

brussee:
You're welcome.

PR #923, opened 2016-11-10
Complete closure of the application with SDL2.

dzmitrykrukouski:
Added `System.exit(0); ` in  _SDLActivite.java_ _onDestroy()_ class to fix this issue kivy/kivy#4264, but I still think that is not right way and need to find more 'Android' method.


inclement:
I'm not sure about this either, although qua-non supports it so maybe should just merge. I haven't had any problems with sdl2 after trying some relatively involved things with a service, do you know what I'd need to do to specifically reproduce the problem?

dzmitrykrukouski:
Try as soon as the application is run, click the back button, and then try to run the application again.

akshayaurora:
@dzmitrykrukouski thanks.

Issue #928, opened 2016-11-16
Presplash delay


KeyWeeUsr:
The current presplash on sdl2 has some kind of weird delay, because it is still visible even after Kivy app was clearly started (you can see widgets through the presplash). For example the test app (`testapp_setup`) when run from apk or from launcher has this behavior.

I haven't looked at the code of presplash, but it seems like there's a time in seconds or something, not the actual fetching for ping from app, to kill the presplash or similar thing at the right time, that is probably in the pygame bootstrap.

inclement:
This is a known issue arising from a limitation of the new display method. I think tito just wasn't able to find a better way. Improvements are welcome, but this is here to stay for now.

You don't have to put up with it though, you can make it leave at the right time by calling the function with pyjnius when the app actually started (but because p4a doesn't know how long the app will take, it can't do this automatically). I have an example [here](https://github.com/inclement/Pyonic-interpreter/blob/master/pyonic/main.py#L196).

KeyWeeUsr:
Thanks for the nice workaround ^^ I'll put it here if something happened to the repo.

    from jnius import autoclass
    activity = autoclass('org.kivy.android.PythonActivity').mActivity
    activity.removeLoadingScreen()

I think I'll still leave the issue open, so that people can find it quickly. Hopefully I or someone else will get to the code soon.

hackalog:
I've seen the presplash occasionally stick around, too. (like, never disappear), app is running in the background.

Another +1 for the workaround

inclement:
I've added some documentation for this [here](http://python-for-android.readthedocs.io/en/latest/apis/#dismissing-the-splash-screen).

PR #930, opened 2016-11-19
Add remove_presplash

KeyWeeUsr:
Remove SDL2 presplash function in `android` package.

ref #928

### DIFF
--- a/pythonforandroid/recipes/android/src/android/_android.pyx
+++ b/pythonforandroid/recipes/android/src/android/_android.pyx
@@ -212,6 +212,14 @@ TYPE_TEXT_VARIATION_POSTAL_ADDRESS = 112
 TYPE_TEXT_VARIATION_URI = 16
 TYPE_CLASS_PHONE = 3
 
+# SDL2 presplash remove
+def remove_presplash():
+    '''Remove android presplash in SDL2 bootstrap.'''
+    try:
+        mActivity.removeLoadingScreen()
+    except:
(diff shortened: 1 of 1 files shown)


inclement:
I wonder if it would be preferable to just put this in an `if BOOTSTRAP == 'sdl2'` block rather than using the catchall try/except - this would be more in line with how the android module is currently structured, and I think fits fine with its somewhat ad-hoc support for things.

KeyWeeUsr:
So it's just a [casual Python](https://github.com/KeyWeeUsr/python-for-android/blob/2094aea9a3a7a01840eab6ee972b7004590b9e0a/pythonforandroid/recipes/android/src/android/_android.pyx#L5)? It really doesn't seem like that, C macro rather and even that probably not. I can't find the `config.pxi` where I should probably look for the thing.

I mean, I can check for `'sdl2'` but if I don't know if it really returns such thing with SDL2 on android, the code most likely won't be executed. Maybe it's just copied from the old code e.g. no `'sdl2'` will be returned, because there was pygame only. Or I'm just confused, hmmm...

inclement:
I actually just guessed that 'sdl2' would work, you're right that we should understand it first.

inclement:
Looks like the config files are all written by the android recipe, and BOOTSTRAP is indeed just the bootstrap name, so checking for `'sdl2'` should be fine.

inclement:
👍

Issue #934, opened 2016-11-20
ImportError for ssl


Polynomdivision:
When I try to import ssl using a simple import statement, e.g. `import ssl`, I get the error thrown that `_ssl` could not be found. Even when I add `openssl` to the requirements, the import error's still there.

I also tried changing the recipe for hostpython2 (In order to compile with SSL and crypto), but I get errors during the compiling process (Undefined references).

I'm wondering if that's an error of the recipe or a mistake I made

Polynomdivision:
I found a solution:
In the recipe for python2, the patch `enable-ssl.patch` is missing.

inclement:
Is this broken in master? Would you like to make a PR to apply your fix?

Polynomdivision:
It appears so; I will push my changes to my fork and then create a PR

Fak3:
@Polynomdivision do you have the same traceback as in https://github.com/kivy/python-for-android/issues/850 ?

Polynomdivision:
If I remember correctly (because I deleted everything after I fixed this issue for myself) then it was not an issue with loading libraries. It was just that `_ssl` was not found. And after some digging around, I found out that no "compiled" python file for (I guess it was called like this; I don't remember anymore) `_ssl.py`.

PR #903, opened 2016-10-10
load `ssl` and `crypto` libs before loading python, for (4.1.2 or ear…

akshayaurora:
…lier devices)

closes #866

Issue #850, opened 2016-07-21
apk, built with openssl launch error: "libssl1.0.2h.so" not found


Fak3:
Steps to reproduce:

Build distro with openssl:

```
p4a create --dist_name=testpygame --bootstrap=pygame --requirements=pygame,openssl,python2,kivy,pil,requests,plyer
```

Then build some apk with this distro.

Try to launch apk on the device. It fails to launch, and `adb logcat python:* *:E` reveals linker error:

```
E/Trace   (15822): error opening trace file: No such file or directory (2)
V/python  (15822): metadata fullscreen is1
E/linker  (15822): load_library(linker.cpp:759): library "libssl1.0.2h.so" not found
E/linker  (15822): soinfo_link_image(linker.cpp:1652): could not load library "libssl1.0.2h.so" needed by "libpython2.7.so"; caused by load_library(linker.cpp:759): library "libssl1.0.2h.so" not found
E/dalvikvm(15822): dlopen("/data/app-lib/org.spbelect.monarchy-2/libpython2.7.so") failed: Cannot load library: soinfo_link_image(linker.cpp:1652): could not load library "libssl1.0.2h.so" needed by "libpython2.7.so"; caused by load_library(linker.cpp:759): library "libssl1.0.2h.so" not found
E/AndroidRuntime(15822): FATAL EXCEPTION: Thread-5955
E/AndroidRuntime(15822): java.lang.UnsatisfiedLinkError: Cannot load library: soinfo_link_image(linker.cpp:1652): could not load library "libssl1.0.2h.so" needed by "libpython2.7.so"; caused by load_library(linker.cpp:759): library "libssl1.0.2h.so" not found
```

I am using p4a git master.
##

Fak3:
I checked the `/data/app-lib/org.spbelect.monarchy-2` dir on my android phone, and i have that library installed there
```
1|u1@android:/data/app-lib/org.spbelect.monarchy-2 # ls -l                                                                                                                                                         
-rwxr-xr-x system   system      13464 2016-08-21 20:53 libapplication.so
-rwxr-xr-x system   system    1540604 2016-08-21 20:53 libcrypto1.0.2h.so
-rwxr-xr-x system   system    2518340 2016-08-21 20:53 libpython2.7.so
-rwxr-xr-x system   system     206364 2016-08-21 20:53 libsdl.so
-rwxr-xr-x system   system     206160 2016-08-21 20:53 libsdl_image.so
[...]

PR #935, opened 2016-11-20
Add 'enable-ssl.patch' to Python2Recipe.patches

Polynomdivision:
This PR is supposed to fix the issue that ssl cannot be imported in python2 
See issue #934

### DIFF
--- a/pythonforandroid/recipes/python2/__init__.py
+++ b/pythonforandroid/recipes/python2/__init__.py
@@ -31,6 +31,7 @@ class Python2Recipe(TargetPythonRecipe):
                'patches/fix-dlfcn.patch',
                'patches/parsetuple.patch',
                'patches/ctypes-find-library-updated.patch',
+               'patches/enable-ssl.patch',
                ('patches/fix-configure-darwin.patch', is_darwin),
                ('patches/fix-distutils-darwin.patch', is_darwin),
                ('patches/fix-ftime-removal.patch', is_api_gt(19)),


inclement:
I was looking into this to understand it better, and I think this patch is supposed to be unnecessary following https://github.com/kivy/python-for-android/pull/838/files , but I'm not really sure. I've never used openssl directly myself, so I'll try to test it just to be sure.

Polynomdivision:
Maybe this helps to reproduce it: If I try to import (python2) ssl or use a package that imports ssl, I always got the error that `_ssl` cannot be found (See #934, like in the original PR description).
But I'm not sure it it's a bug/erorr in p4a or just my incredible ability to break every piece of software I touch.

brussee:
Wether you can use openssl directly is not changed by #838. Packages that link to openssl, as you can see in #838, do so dynamically. If you find a package that fails to import ssl you need to link it at compile time or use the 'enable-ssl.patch', I think.

Fak3:
This maybe a fix for https://github.com/kivy/python-for-android/issues/850 I will try to check when i have time

frankgould:
I tried this patch and got the same error (below) after the patch.

E/dalvikvm( 3869): dlopen("/data/app-lib/com.masterpics.slideshow-1/libpython2.7.so") failed: Cannot load library: soinfo_link_image(linker.cpp:1635): could not load library "libssl1.0.2h.so" needed by "libpython2.7.so"; caused by load_library(linker.cpp:745): library "libssl1.0.2h.so" not found

frankgould:
Are there any suggestions to fix this problem? I'm still getting the same error and can't find a solution online.

KeyWeeUsr:
@frankgould see [this answer](http://stackoverflow.com/a/28817942/5994041). Basically it means that 4.2 isn't able to load libs automatically, so we do it [here](https://github.com/kivy/python-for-android/blob/master/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonUtil.java#L31) instead.

frankgould:
Hi KeyWeeUsr, thanks for your quick response. I tried several different compiles with PythonUtil.java edited with the additional "load" code but these all errored out by not finding the names.
                System.loadLibrary(openssl);
                System.loadLibrary(pyopenssl);
                System.loadLibrary(libssl);

So, I grep searched for libpythonXXX.so and found 5 copies of PythonUtil.java, some did not have the System.loadLibrary(lib), which I added and rebuilt. Now the error is "Cannot load library: find_library(linker.cpp:889): "/data/app-lib/com.masterpics.slideshow-1/libpython2.7.so" failed to load previously."

I wasn't sure if it is helpful but below are recipes I am using.
includes recipes (hostpython2, openssl, sdl2_image, sdl2_mixer, sdl2_ttf, python2, numpy, sdl2, setuptools, six, android, opencv, pyjnius, pyopenssl, kivy), built for archs (armeabi-v7a)

My buildozer.spec requirements are:
requirements = kivy,openssl,pyopenssl,hostpython2,opencv,android

Thanks again for your help.

frankgould:
Please let me know if there is anything else I can provide to help with this.  Thanks.

frankgould:
Is there somewhere I can find more information to solve this problem? I am waiting patiently for some guidance.

frankgould:
I was able to solve this using the information from the following link:
https://github.com/kivy/python-for-android/pull/903

Bakterija:
I just used this to enable ssl in gevent . Otherwise it complains about missing _ssl

Zen-CODE:
This change did not fix the issue 850 for me.

https://github.com/kivy/python-for-android/issues/850

AndreMiras:
This is rather old and conflicting with recent p4a refactoring, e.g. https://github.com/kivy/python-for-android/pull/1537
Feel free to open a new one if needed

PR #948, opened 2016-12-04
Made the sdl2 bootstrap strip unneeded symbols with python3

inclement:
Not for merging yet.


inclement:
Well, this seems to work, reduces python3 apk size by ~25%.

PR #956, opened 2016-12-13
Add ``screenSize`` to android:configChanges in AndroidManifest.xml if API >= 13

rnixx:
According to https://developer.android.com/guide/topics/manifest/activity-element.html#config -> orientation:

If your application targets API level 13 or higher (as declared by the minSdkVersion and targetSdkVersion attributes), then you should also declare the "screenSize" configuration, because it also changes when a device switches between portrait and landscape orientations.

If ``screenSize`` not given app freezes on orientation change.


akshayaurora:
👍

PR #965, opened 2016-12-19
Add permission groups

### LINKED ISSUE #843: Feature request: access to all permissions
Don't have an option to write a PR(debugged), so I rather ask beforehand. When I use custom launcher, I put there every permission I know could be at least a little bit useful. As the launcher isn't intended for final distribution, e.g. `__init__.py` could be a nice place for a list of all available permissions.

Then `--permission ALL` would be possible as it'd only need to have a simple loop in each `build.py` that can build the launcher(or even in all builds). However, I can see ALL not available in arguments as a safety switch too for the final dist.

Is it a good idea to include ALL into p4a?  pros could be simple process of adding permissions, maybe future check if permission is typed correctly(in a list of permissions) and maybe prevent crashes this way(don't know if there is some check for this), then make updating _custom_(I don't want you to go for stable launcher with all permissions) one less frequent. This could make the launcher slowly become "ide" on android for Kivy.

Two critical cons are basically disabling permission security at all with this(app would have ALL) and a risk that some beginner would go for this craziness in final dist.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/36092049-feature-request-access-to-all-permissions?utm_campaign=plugin&utm_content=tracker%2F11371&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F11371&utm_medium=issues&utm_source=github).
</bountysource-plugin>

KeyWeeUsr:
( extracted from https://android.googlesource.com/platform/frameworks/base/+/master/core/res/AndroidManifest.xml )

This might be later even included in documentation after autodoc works (again?).

closes #843

### DIFF
--- a/pythonforandroid/permissions.py
+++ b/pythonforandroid/permissions.py
@@ -1,0 +1,543 @@
+"""
+.. _perm_groups:
+
+Permission groups for Android
+=============================
+
+A collection of permissions fetched directly from the Android source code,
+grouped by the categories presented in the ``AndroidManifest.xml``
+
+* CONTACTS - Permissions for accessing user's contacts including personal
+  profile
+
+* CALENDAR - Permissions for accessing user's calendar
+
+* SMS - Permissions for accessing and modifying user's SMS messages
+
+* STORAGE - Permissions for accessing external storage
+
+* LOCATION - Permissions for accessing the device location
+
+* PHONE - Permissions for accessing the device telephony
+
+* MICROPHONE - Permissions for accessing the device microphone
+
+* UCE - Permissions for accessing the UCE Service
+
+* CAMERA - Permissions for accessing the device camera
+
+* SENSORS - Permissions for accessing the device sensors
+
+* REMOVED - Permissions removed in the latest source kept for backwards
+  compatibility
+
+* ALARM - Permissions for setting the device alarm
+
+* VOICEMAIL  - Permissions for accessing the user voicemail
+
+* LOCATION - Permissions for accessing location info
+
+* NETWORKS - Permissions for accessing Internet, WiFi and short range,
+  peripheral networks
+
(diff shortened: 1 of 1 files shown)


inclement:
Looks nice. Are there any licensing implications if these are extracted directly from the Android source?

KeyWeeUsr:
I'm not sure, at all, however it's also in [Manifest.permission_group](https://developer.android.com/reference/android/Manifest.permission_group.html) and also in casual Manifest documentation (docs is from source like Kivy does it with sphinx).

The documentation seems to be under [CC Attribution 2.5](https://creativecommons.org/licenses/by/2.5/) and the code I fetched it from is under [Apache 2.0](https://tldrlegal.com/license/apache-license-2.0-%28apache-2.0%29), however there's no code that is runnable, usable or actually related to the Android as a part of the project, this PR contains lists of names, so I'd treat it as plaintext / documentation, but with an attribution to the original license.

For example:

    Permission names and permission groups are licensed under Apache 2.0 License,
    Copyright 2006, The Android Open Source Project

KeyWeeUsr:
@inclement I'm not sure how to handle permissions similar to [mocking](http://stackoverflow.com/a/12644377/5994041) permissions. It's in that state as I have it here even in the Manifest, which brings me to a thought that _user_ is responsible for filtering the permissions and also it seems more reasonable for me to _force_ users to choose only specific permissions only for their app and use groups for more comprehensive testing of their app.

On the other hand, a user can make an idiot from him/herself quite easily this way judging from that status, so it might be maybe more useful to have a group `MOCK`, which would have such permissions, or we could put it directly into `DEVELOPMENT`, but this collides with Manifest.

PR #981, opened 2017-01-16
Fix python2 for webview

KeyWeeUsr:
new env var ANDROID_UNPACK to target a folder `/data/data/<app>/files` and for sdl2 that + `/app` to get libpymodules.so


KeyWeeUsr:
@brussee If you can, please review the new variable in the `service_only` bootstrap. I'm not sure if I got it right.

inclement:
I think this may have some problems with the python3 build, so I've temporarily reverted the change to custom-loader.patch to use ANDROID_APP_PATH instead (so that everything except the launcher works).

kollivier:
Has anyone tested the webview backend since the revert? I've cleaned everything, made sure the reverted code is in my tree, and rebuilt, but I'm still getting the 'module is 0 bytes' error when building a pretty simple app with webview. (I'm building with pyjnius and android modules)

inclement:
All merged, with a couple of other fixes! Thanks again KeyWeeUsr, I'm ashamed for the long time taken to pull it all together :<

These changes now mean that all the primary bootstraps seem to work for me.

PR #987, opened 2017-01-18
Add recipe for google protobuf cpp implementation

bakwc:
Protobuf with c++ implementation is much faster than native-python (10-60 times). Here is a recipe for building protobuf c++ extension. So **instead** of adding `protobuf` requirement, add `protobuf_cpp` one.
To ensure that you are using cpp implementation you can add folowing before importing protobuf:
```
import os
os.environ['PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION'] = 'cpp'
```
It will raise exception if cpp implementation is not available.


inclement:
Looks good, thanks!

Issue #983, opened 2017-01-16
Unpacking and copying app contents causes app to appear hung


kollivier:
We have an app that has a lot of data files, and we support preloading our app on devices without internet access, so the Play restrictions do not apply and furthermore we cannot rely on downloadable asset packs. When the app is first loaded, the unpack and copy of all the app's data files to the /data partition causes the app to appear hung for upwards of 1 minute because the unpack happens in onCreate. Having an app display a blank screen for any real length of time, much less a minute or more, gives a pretty bad first impression.

My understanding is that the technical reason for this process is that fopen and co. will not directly open files on Android because of the need to use AssetManager, so this is worked around by moving all the files to a place where files are directly accessible. I've found another cross-platform app developer who addressed this by devising a drop-in replacement for the fopen series of functions that use the asset manager under the hood. 

http://www.50ply.com/blog/2013/01/19/loading-compressed-android-assets-with-file-pointer/#comment-1850768990 

Would something like this be an option? Or perhaps a slightly more advanced version that tries the asset manager only for certain paths, or falls back to fopen when that fails?  I've spoken with @inclement and others on kivy-dev about this, but I wanted to open a ticket to track this issue as we will really need to get this fixed to continue using p4a.

inclement:
As far as I know, something like that could work but needs some hacking to test it out.

kollivier:
Cool, I'll probably start playing with adding an experimental --nozip option for p4a that just puts all the files in the assets folder. Once I get to the point where it's giving file access errors, I'll start playing around with  this and seeing what might work.

 

For patched software like Python 2, do you have a repo where you maintain the patches or do you basically just download the tarball, apply the patches, and then just make new ones each time there's a change like this?

inclement:
> For patched software like Python 2, do you have a repo where you maintain the patches or do you basically just download the tarball, apply the patches, and then just make new ones each time there's a change like this?

Just the latter one, so you can make new patches by downloading the tarball, adding the relevant files to a repo, committing any changes and getting the patch with `git diff`.

inclement:
Closing as resolved by #990

PR #990, opened 2017-01-20
Move the unpackFiles step into an AsyncTask

kollivier:
Right now the loading screen does not show up until loading is actually pretty much complete because all the loading tasks block the UI thread. As a result, you will see a dark grey gradient background instead of the loading screen. For apps with more data to unpack, this can appear for longer and give the impression the app has hung. This is a fix for Issue #983.

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonActivity.java
+++ b/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonActivity.java
@@ -18,6 +18,7 @@ import android.app.Activity;
 import android.content.Intent;
 import android.util.Log;
 import android.widget.Toast;
+import android.os.AsyncTask;
 import android.os.Bundle;
 import android.os.PowerManager;
 import android.graphics.PixelFormat;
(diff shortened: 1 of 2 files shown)


kollivier:
Taking a second look at the code, I think I see why we get the black screen. I was assuming that once the SDL view was loaded, we no longer needed the loading screen, but apparently there's a delay between setting the view and the first actual render. Typically in my own apps I load up my WebView at this point so I didn't notice it. I'm going to run some tests but I bet adding another call to this.setLoadingScreen() right after mActivity.finishLoad() is called should fix it.

inclement:
I'm just trying some tests, but this seems to be working now, which is great! Just to check, is it ready to merge as far as you are concerned?

pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonActivity.java
```diff
@@ -178,6 +111,100 @@ public void run() {
...
+            //
+            // Otherwise, we use the public data, if we have it, or the
+            // private data if we do not.
+            mActivity.finishLoad();
+
+            // finishLoad called setContentView with the SDL view, which
+            // removed the loading screen. However, we still need it to
+            // show until the app is ready to render, so pop it back up
+            // on top of the SDL view.
+            this.showLoadingScreen();
```
inclement: Actually, I think this should be  mActivity.showLoadingScreen() - I don't know why it worked for me before, but it's crashing now.

kollivier:
Yeah, I've been using it in my app and so far no issues, just a much nicer initial boot experience. :) In addition, I've gotten an --assets-dir argument working as well (though it's not quite ready for pushing yet), and with the combination there is virtually no hang or startup delay at all! :)

kollivier:
Oh sorry, yes, that's correct. It should be mActivity, I thought I had pushed that already...

inclement:
Okay, I fixed it locally to test with so I just pushed that. Thanks!

kollivier:
Cool, thanks!

Issue #988, opened 2017-01-18
Create a space for common bootstrap code along with a base class for all bootstraps


kollivier:
Currently, each backend has a completely separate tree of Java classes for initialzing the bootstrap, although a significant amount of the code is sharable among the backends. We should create a single place for this common code and refactor the bootstraps to use them rather than duplicate them.

AndreMiras:
Thanks for tracking the issue and thanks for the pull request. I was actually discussing this in the group recently https://groups.google.com/forum/#!topic/kivy-dev/MvAzPilialw
We'll take a look at he PR :+1:

inclement:
Closing as work along these lines has been merged thanks to various contributors :)

PR #991, opened 2017-01-23
Start on a common area for bootstraps, so that different bootstraps

kollivier:
can share common code. This is a first step to fixing issue #988. Note that this new approach  copies over bootstrap files on every build, so I think it may remove the need for --symlink-java-src, though I've left that in just in case I'm misunderstanding something. I can remove it if needed. BTW I did test performance and the impact is minimal.

Next step will be to build a BasePythonActivity that has any common functions that are used by, or could be useful to, multiple bootstraps.


kollivier:
`BasePythonActivity` is more like an additional step to help with sharing code, not a separate approach. This PR is about not keeping multiple copies of the same files in each separate backend, which would still be a problem if we added `BasePythonActivity`. 

I think we can deal with backend-specific differences in shared files on a case-by-case basis. Worst-case, we can patch them, but I don't think it would come to that.

I'm interested in taking a stab at BasePythonActivity, but I probably won't have time in the near future to work on it. I think landing this patch is a good first step to start moving that process forward, and maybe once there's a shared space for backends, someone will feel it's easier to hack on a `BasePythonActivity` class. :) Love to hear your thoughts on my approach so far or if you guys see any issues with it!

AndreMiras:
OK I agree with the approach indeed, thanks!
I'm curious why your PR doesn't have any deleted files, we should have plenty of them right?
Also this is a bit cumbersome to test, since we need to test all bootstraps for regression :confused: 
Perhaps we should improve the test suite first to see if at least it compiles and next step would be to check for run time errors, but this is more tricky to achieve.

I've merged the merge conflicts and now it seems that your `git mv` got changed to `git add` ahah nice online github editor :stuck_out_tongue: ...
If you don't have time for it give me write access to your branch and I'll try to fix and squash.

kollivier:
I've checked "Allow edits from maintainers", do I need to do something else in addition to that?

AndreMiras:
I think I actually need access to the branch of your fork rather than the pull request.
https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/

That would probably allow me to do my rebasing/squashing magic :smile:

AndreMiras:
That's perfect, thanks for sharing. I've started the rebasing/squashing.
So basically now we have your changes in one single commit on top of the git log.
Next up is to make it works, then to make it right. Feel free to continue helping in these topics if you think there's something to test/add/fix and I can deal with the version control part :smile:

Some of the things I'd like to address before finishing with that pull requests are https://github.com/kivy/python-for-android/issues/1382 and https://github.com/kivy/python-for-android/issues/1380.
Basically in #1382 I'd like to add the different bootstrap builds in our continuous integration. And in #1380 I'd like cover more code in unit test fashion (not yet sure how for bootstraps).

By the way, I made [a back up of the branch](https://github.com/kollivier/python-for-android/compare/bootstrap_refactor_backup) in case I messed it up.

JonasT:
Just as a note, I highly depend on this to be able to redo #1433 in a better, comprehensive and non-horrific way - and #1433 itself is highly relevant for proper Python 3 support (at least without kivy, not sure if that has something built-in that handles this). So until some functionality like this one is merged in first, it will be difficult for me to advance

JonasT:
What's the progress / state of this pull request?

AndreMiras:
I think it still make sense, @kollivier already made his part, now it's on our side. We need to discuss if we want to integrate it this way. If we do I would fix the conflicts and merge, but I was busy on other things :confused:

JonasT:
I haven't looked at the details, but simply having a `common` folder where at least *most* of the files get copied out into all the separate bootstraps sounds like a rather good idea to me. (But then again, I have barely done work on the bootstraps so I'm not the best person to judge this) So thumbs up for the basic approach from me :+1:

JonasT:
@tito @inclement what's your input on this one? It would be lovely to get this in or something similar, so I can pick up work on https://github.com/kivy/python-for-android/pull/1433 and finally fix `ctypes.util.find_library` crashing for Python 3

JonasT:
Weird, not sure what GitHub is on about, but I checked out the branch and rebased it onto master with no issue or no not-auto-resovled conflicts: https://github.com/JonasT/python-for-android/tree/bootstrap_refactor

I have yet to test it though. I will try to rebase my ctypes fix on top of this, and see if I can get things to build & run. This of course would be a huge waste of time if there wasn't some sort of plan to eventually merge this, so if that *isn't* the case, please let me know *now*

JonasT:
Trying a reworked thing based on this here: #1496

AndreMiras:
I'm sorry we didn't make yours to the tree @kollivier. Thanks a mil for original initiative and the hard work you put in there :beer: 
I do think however your point is still valid since we didn't yet kill the repeated code. But we might address it in a second round since we need to investigate what can be merged and what can't.

JonasT:
@kollivier I put your name into the commit message, I hope that's fine with you! You did great work, very cool we could finally get this in

kollivier:
Yup, no worries! 👍 Thanks to you both for getting this past the finish line! 

There is definitely still more work to be done, I have some customizations I want to merge upstream and this work makes doing so much easier. Specifically, one thing is that I made the python zip extraction into an AsyncTask in PythonActivity, so that it doesn't block the UI and we can show the load image sooner. I'd like to actually have the base code for that (along with some other common functionality) as a start on BasePythonActivity!

AndreMiras:
That sounds like a very great idea!

PR #994, opened 2017-01-25
Add experimental mingw support.

matham:
First, don't let the size of the PR scare you. Most of the changes are the additon for patches for hostpyhton2/python2. Also the addition of the sh module and the replacement of everything that used sh. The rest of the changes are quite small.

Overall notes
---------------
* This only supports python2.
* ant has issues with spaces in paths. Try to have everything in a place without spaces. As a result, the temp directory may fail to work if your user name has spaces. As a result I set my temp directory to msys' temp directory which is at  `C:\msys64\tmp` for me. See the example bashrc file for how to set temp variables. All of them need to be set because it's not clear which of the versions are used.
* hostpython2 cannot be built when the path is too deep because of a max path issue. The solution is to build it in a temp directory which is close to the root and then move it to the expected location. This will happen automatically if you define `PY4A_BUILD_TEMPDIR` to point to a temp directory in the env.
* I added patched and added `XCOMPILE_BUILD_PYTHONHOME` and `XCOMPILE_BUILD_PYTHONHOME_EXEC` to hospython2 (I used 2.7.12 btw, but target is still 2.7.3). When set they are read first thing and determine where python looks for python modules, and compiled python modules respectively. This is because otherwise the copying of hostpython didn't work very well as it looked for the modules in different places. This way we can set the exec version to the hostpython's path and the other one to the target python's path so it uses its modules. But this patch is only applied on msys for now.

How to maintain msys support
-----------------------------------

* I added a reduced version of the sh module at `pythonforandroid.sh`. It works like normal (didn't test on linux yet), however it doesn't support some of the more fancy things like piping or some `kwargs` which it ignores. You can see the accepted list in the file. E.g. `_tail=blah` will just be ignored.
* So please always use `from pythonforandroid import sh`.
* While the msys env is fully linux (e.g. `C:\prog\file` is stored as e.g. the env variable MYPATH as `/c/prog/file`), when a windows program is called it converts these paths to windows format and when bash is called it converts the env paths to linux style. 

This generally makes things work. However, if you pass a path as a variable (e.g. `CFLAGS`) or save it to file (e.g. `local.properties`) or whatever it won't go through this conversion process. Also, not all variables seem to be converted (maybe just the path type ones). So when passing paths to e.g. configure, if you use `join`, it'll use a `\` which will not be converted automatically to a `/` which means that the configure or makefile will have a backslash in it which will break it.

In any case, to be safe for paths being passed from python code we should always use this style `C:/prog/file` as this is understood both by linux and msys exes. There's a function called `mpath` in `pythonforandroid.utils` that will do this conversion. I had to add calls to that in many places. It's best to use this everywhere we pass args to e.g. configure or to the env which is not a normal path type variable. E.g. the `PATH` variable should not (not sure if it breaks it if it does) undergo this change as it should happen automatically when python calls bash. This part may have to happen to existing recipes if there are issues.

Install guide, brief
----------------------
* Download [msys2 ](https://msys2.github.io/) 64-bit (32 bit will probably also work, but I only tested with 64 bit). Install it to a root drive, e.g. `C:\msys64` (we'll assume this path for now).
* **Very important**: msys comes with `msys`, `mingw64`, and `mingw32` installed at C:\msys64. These executables start different environments using different compilers. For this we must run `mingw64`. The following assumes you run it using `mingw64.exe` which will launch a bash shell environment. You can read more about msys [here](https://sourceforge.net/p/msys2/wiki/Home/).
* msys uses pacman for package management so update as described on that download page. You'll need quite a few packages for the build. After the update, at the end, is a list of all the packages I installed. I'm not sure if all of them are needed, but many of them are. So please install all until we figure out which are not needed. To install, save the list to a file and from bash run `pacman -S $(cat filename)`.
*Next you'll need to set up the mingw64 environment to use python and its dependencies. We're using the normal windows python, nothing special. You need to add the python root as well as it's scripts folder to the path (see example at the end). You can do it using `export PATH=$PATH:...` or add it to the bashrc file.
* Install `cython`, `virtualenv`, `python-for-android` with pip.
* Install Android sdk, you can do it from android studio, which has a sdk manager and install the api level you want to use from there (I'm sorry to say @KeyWeeUsr but you'll need to get a new harddrive to fit all of these). Set the `ANDROIDSDK` variable to the install path as well as `ANDROIDAPI` to the api level.
* Install ndk and set the `ANDROIDNDK` to its paths. Add `ANDROIDNDKVER` with the version as well to the environment variables.
* Install ant and set `ANT_HOME` to it's path. Add `$ANT_HOME/bin` to the path.
* Install the java jdk and add it's path to `JAVA_HOME`.

An environment variable can be set using `export NAME=value`, or to make it more permanent add it to the `.bashrc` file. That file can be found at ``. Here's how the end of mine looks like:

```
export PATH=$PATH:/e/apache-ant-1.10.0/bin
export ANT_HOME=/e/apache-ant-1.10.0
export JAVA_HOME="/c/Program Files/Java/jdk1.8.0_112"
export ANDROIDSDK="/e/Android/sdk"
export ANDROIDNDK="/e/Android/android-ndk-r13b"
export ANDROIDAPI="22"
export ANDROIDNDKVER="r13b"
export PY4A_BUILD_TEMPDIR="/tmp"
export temp="/tmp"
export tmp="/tmp"
export TMPDIR="/tmp"
export TMP="/tmp"
export TEMP="/tmp"
```

Once all set, just run py4a like normal. E.g. my command was `time p4a apk --private /g/Python/py4a/code/testapp --package=org.matham.myapp --name "mingw_win" --version 1.1 --bootstrap=sdl2 --requirements=sdl2,pyjnius,kivy,python2 --storage-dir=/g/python/py4a/builds/storage --arch=armeabi-v7a 2>&1
| tee log` and took 72 minutes.

```
autoconf
automake-wrapper
bash
bash-completion
binutils
bsdcpio
bsdtar
bzip2
catgets
ccache
coreutils
crypt
curl
dash
doxygen
file
filesystem
findutils
flex
gawk
gcc
gcc-libs
git
glib2
grep
gzip
inetutils
info
less
libevent-devel
libffi-devel
libgdbm-devel
libreadline-devel
libsqlite-devel
libtool
libunistring
libunistring-devel
lndir
make
mingw-w64-cross-binutils
mingw-w64-cross-zlib
mingw-w64-i686-gcc
mingw-w64-i686-ncurses
mingw-w64-x86_64-ccache
mingw-w64-x86_64-gcc
mingw-w64-x86_64-gdbm
mingw-w64-x86_64-glib2
mingw-w64-x86_64-gtk2
mingw-w64-x86_64-gtk3
mingw-w64-x86_64-gtkglext
mingw-w64-x86_64-libevent
mingw-w64-x86_64-libffi
mingw-w64-x86_64-libunistring
mingw-w64-x86_64-ncurses
mingw-w64-x86_64-openssl
mingw-w64-x86_64-pkg-config
mingw-w64-x86_64-readline
mingw-w64-x86_64-sqlite3
mingw-w64-x86_64-tcl
mingw-w64-x86_64-tk
mingw-w64-x86_64-tklib
mintty
msys2-keyring
msys2-launcher-git
msys2-runtime
msys2-runtime-devel
nano
ncurses
ncurses-devel
pacman
pacman-mirrors
pactoys-git
patch
pax-git
pkgfile
rebase
sed
tar
tcl
tftp-hpa
time
tmux-git
ttyrec
tzcode
unzip
util-linux
which
```


inclement:
Following the discussion on irc I thought I should summarise the situation here.

First: nice!

I'm not really sure how to proceed with this - as discussed, it's probably not a great idea to actually merge it unless we can commit to supporting windows and the issues that will inevitably arise there. As far as I'm concerned it's fine to leave the PR open to see if it gets any interest, and it's already a good reference resource.

yaki29:
@matham It looks interesting I learned a lot while reading this and searching on net to actually understand what you are talking about. I don't know if it is right to ask you that but what is the need to do all that stuff?Correct me if I'm wrong somewhere.

matham:
I'm not sure what you're asking. Are you asking why these patches are
needed or why we should support mingw?

For the latter it's because that's the only way to use py4a on windows
(other than using the Ubuntu win10 subsystem). And these patches are needed
to support mingw.

On Jan 30, 2017 12:14 PM, "Yash Jain" <notifications@github.com> wrote:

@matham <https://github.com/matham> It looks interesting I learned a lot
while reading this and searching on net to actually understand what you are
talking about. I don't know if it is right to ask you that but what is the
need to do all that stuff?Correct me if I'm wrong somewhere.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/kivy/python-for-android/pull/994#issuecomment-276124362>,
or mute the thread

yaki29:
@matham I don't use windows much that's why I didn't know about that. Thanks for replying.

AndreMiras:
What an impressive pull request, well done @matham!
I'm sorry we could not make it to the tree :disappointed: 
Would it make sense if you created a dedicated p4a for mingw just like we have https://github.com/kivy/kivy-ios ? But it could be built on top of p4a and inherit/override some classes.
I'm not sure how feasible it would be. But if so that would be awesome. If you start investigating, please share your repo.

matham:
@AndreMiras thanks for your kind words. Part of the reason I never perused this PR further is like @inclement said, it's not clear that there's community interest, especially given that's it's super easy to fire up a ubuntu vm, or even use the windows susbsytem for linux. 

The changes are quite invasive and it's not clear that (a) it's really needed and (b) we would have people to maintain it. Personally, I have since lost interest - I don't and never have made android apps. I just mainly worked on this out of curiosity, so I wouldn't undertake to maintain it.

Regarding doing something like kivy-ios, I'm afraid that the changes are so invasive that it couldn't happen on top of py4a but it would need to happen in it.

AndreMiras:
Fair point for the now built-in Windows/Linux subsystem. Thank you for your reply and your efforts!
Feel free to close that pull request then. If we hear about somebody willing to take over we can point to it.

PR #999, opened 2017-01-27
recipe for libxml2 

yaki29:
This is a recipe for libxml2. If there are still some changes to be done I would be happy to do them.


inclement:
How about adding the full suite of libxml2, libxslt and lxml to the one PR? Since they're probably mostly going to be used together, this would make testing easier.

pythonforandroid/recipes/libxml2/__init__.py
```diff
@@ -0,0 +1,31 @@
+from pythonforandroid.toolchain import (CythonRecipe, shprint,
+                                        current_directory, info, Recipe)
+from pythonforandroid.patching import will_build, check_any
+import sh
+from os.path import join
+
+
+class Libxml2Recipe(CythonRecipe):
+    version = '2.9.4'
```
inclement: Presumably this version is much newer than the one the old toolchain used? It's hopefully fine and definitely should be targeted, but just to note it could be a source of problems.

pythonforandroid/recipes/libxml2/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+    url = 'ftp://xmlsoft.org/libxml2/libxml2-{version}.tar.gz'
```
inclement: I'm not sure if p4a will work with ftp urls, have you tested it?

pythonforandroid/recipes/libxml2/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+    name = 'libxml2'
+    depends = [('python2', 'python3crystax')]
+
+    def get_recipe_env(self, arch):
+        env = super(Libxml2Recipe, self).get_recipe_env(arch)
+
+        try:
+            sh.sed('runtest\$(EXEEXT) \\/ \\/', 'Makefile')
+        except:
```
inclement: What are these try/except blocks for? If you expect a specific error, catch it explicitly, catching everything is bad style since you'll find it hard to debug problems that you didn't expect as they're swallowed the same way as the one you did anticipate. Since the old toolchain didn't expect any errors, you probably just don't want the try/except at all - if it fails, it's not for an expected reason so you can let p4a crash.

pythonforandroid/recipes/libxml2/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+    version = '2.9.4'
+    url = 'ftp://xmlsoft.org/libxml2/libxml2-{version}.tar.gz'
+    name = 'libxml2'
+    depends = [('python2', 'python3crystax')]
+
+    def get_recipe_env(self, arch):
+        env = super(Libxml2Recipe, self).get_recipe_env(arch)
+
+        try:
+            sh.sed('runtest\$(EXEEXT) \\/ \\/', 'Makefile')
```
inclement: You might not need to escape the $ character since this is a python string rather than shell code.

pythonforandroid/recipes/libxml2/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+        except:
+            pass
+        try:
+            sh.sed('testrecurse\$(EXEEXT)$//', 'Makefile')
+        except:
+            pass
+        try:
+            sh.make('-j', '$MAKE_JOBS')
```
inclement: `$MAKE_JOBS` was an environment variable in the old toolchain, but this won't be picked up here, and in fact I don't think it exists any more. For now you could just set `-j5` (possibly to be improved later).

pythonforandroid/recipes/libxml2/__init__.py
```diff
@@ -0,0 +1,31 @@
+from pythonforandroid.toolchain import (CythonRecipe, shprint,
+                                        current_directory, info, Recipe)
+from pythonforandroid.patching import will_build, check_any
+import sh
+from os.path import join
+
+
+class Libxml2Recipe(CythonRecipe):
```
inclement: This isn't a Cython module, so you'll want to just use `Recipe` as the base class.

yaki29:
@inclement I'm working on libxslt, I'll send a pr when all are ready and also add BeautifulSoup4 recipe with it which was my real target. I'll also make changes to this recipe.

pythonforandroid/recipes/libxml2/__init__.py
yaki29: It used to work with the old_toolchain ... so it would work with the new one. Wouldn't it?

pythonforandroid/recipes/libxml2/__init__.py
yaki29: @inclement You are right it would be difficult debug problems with all try and except statements.
I'll remove them.

pythonforandroid/recipes/libxml2/__init__.py
inclement: It definitely can work, I'm just not sure if ftp:// is handled properly right now. It shouldn't be a real problem.

pythonforandroid/recipes/libxslt/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+
+class LibxsltRecipe(CythonRecipe):
+    version = '1.1.29'
+    url = 'ftp://xmlsoft.org/libxslt/libxslt-{version}.tar.gz'
+    name = 'libxslt'
+    depends = [('python2', 'python3crystax'), 'libxml2']
+    patches = ['fix-dlopen.patch']
+
+    def should_build(self, arch):
+        return not exists(join(self.get_build_container_dir(arch.arch),"libxslt/.libs/libxslt.a"))
```
inclement: I think you can use `self.get_build_dir(arch.arch)` and remove the `libxslt` folder from the path - this should work the same, but will also work if the extracted libxslt folder is something unexpected. This is a small thing though, it probably isn't important to having the recipe actually work.

yaki29: @inclement Yes, it can cause problems. I have changed it.

inclement:
@yaki29 This all looks good on a quick check, nice work. But, have you tried it, does it work?

dessant:
"Robots and Screen Scraping: You may not use data mining, robots, screen scraping, or similar data gathering and extraction tools on this site, except with our express written consent as noted below. "

http://www.imdb.com/conditions

We could use a site which does not forbid scraping. There are some other cleanups to do in the example, i'm not sure what is happening with those globals, pep8 method names etc. Though it's also not clear if the example is even necessary, there is nothing kivy/p4a specific needed for bs4 to work in terms of code.

inclement:
++ to @dessant's comments, although I'll note that I asked for this example as these are very useful for testing p4a - I don't know anything about beautifulsoup or its compiled dependencies, so it would be difficult to test without an example.

testapps/BeautifulSoup4/main.py
```diff
@@ -0,0 +1,128 @@
...
+			
+			for tag in soup.find_all("span", {"itemprop":"actors"}):
+			
+				actor += tag.text 
+				
+				
+
+			self.Pop(movie,imdb,actor,director,writer,url)
+
+		except Exception:
```
inclement: When catching exceptions, work out what exception you expect and catch that one specifically. Catching Exception itself will make the code ignore just about anything, which can hide all manner of problems.

testapps/BeautifulSoup4/main.py
```diff
@@ -0,0 +1,128 @@
...
+	TextInput:
+		id: name
+		font_size: 40
+		multiline: False
+	Button:
+		text: "show details"
+		on_press: root.detail(name.text)
+
+''')
+global writer
```
inclement: You don't need to declare these as global.

testapps/BeautifulSoup4/main.py
```diff
@@ -0,0 +1,128 @@
...
+global actor
+actor = "Actors:"
+global director
+director = "Directors:"
+class PopupScreen(BoxLayout):
+	def error(self):
+		popup = Popup(title="ERROR",content=Label(text="Bro, Check your movie name"),size_hint=(0.8,0.8))
+		popup.open()
+
+	def Pop(self,movie,imdb,actor,director,writer,url):
```
inclement: It would be preferable to use kv to set up these labels.

yaki29:
@dessant @inclement I didn't know that imdb doesn't let you scrap it's data. Sorry for that. I'll correct those mistakes pointed by you.

KeyWeeUsr:
@yaki29 I think you [could](https://archive.org/about/terms.php) use [OpenLibrary](https://openlibrary.org) or e.g. a static website that has no big database in the background.

dessant:
Scraping the content of the `h1` tag form https://httpbin.org/html could be enough, and keeping the example simple.

yaki29:
Thanks @dessant @KeyWeeUsr . I was about to commit the changes. I was little busy for last two days because of some college work. I'll send a commit.

pythonforandroid/recipes/libxml2/__init__.py
```diff
@@ -5,27 +5,18 @@
...
-            sh.sed('testrecurse\$(EXEEXT)$//', 'Makefile')
-        except:
-            pass
-        try:
-            sh.make('-j', '$MAKE_JOBS')
-        except:
-            pass
+        sh.sed('runtest$(EXEEXT) \/ \/', 'Makefile')
+        sh.sed('testrecurse$(EXEEXT)$//', 'Makefile')
+        sh.make('-j', '5')
```
inclement: These lines should be placed in the `build_arch` method, and you need to navigate to the right directory, something like

    def build_arch(self, arch):
        with current_directory(self.get_build_dir(arch)):
            sh.sed(....)

pythonforandroid/recipes/beautifulsoup4/__init__.py
```diff
@@ -0,0 +1,25 @@
+ 
+from pythonforandroid.toolchain import (PythonRecipe, shprint,
+                                        current_directory, info, Recipe)
+from pythonforandroid.patching import will_build, check_any
+import sh
+from os.path import join
+
+class BSoup4Recipe(PythonRecipe):
```
inclement: The fact that you need to set the compiler stuff seems to suggest that this is not a pure python recipe, so maybe you need to use `CythonRecipe` or `CompiledComponentsPythonRecipe` instead.

testapps/BeautifulSoup4/main.py
```diff
@@ -1,128 +1,67 @@
...
-		actor = "Actors:"
-		global director
-		director = "Directors:"
 
-		s="+".join(title.split())
 
+class PopupScreen(BoxLayout):
+    def error(self):
+        popup = Popup(title="ERROR",
+            content=Label(text="Check Your Network Connection"),
```
dessant: If you must catch exceptions in this example, please show the exception message instead of assuming what the error is.

yaki29: @dessant  Will do the correction.

inclement:
What's the current status of this, is it working for you?

pythonforandroid/recipes/libxml2/__init__.py
```diff
@@ -13,10 +13,15 @@ class Libxml2Recipe(Recipe):
 
     def get_recipe_env(self, arch):
         env = super(Libxml2Recipe, self).get_recipe_env(arch)
+        return env
 
-        sh.sed('runtest$(EXEEXT) \/ \/', 'Makefile')
-        sh.sed('testrecurse$(EXEEXT)$//', 'Makefile')
```
yaki29: @inclement it is having error here(with this recipe alone others are working fine ( I have not tested bs4 recipe as it depends on this one)).

yaki29:
@inclement I'll look into it and try to make it work.

FilipDem:
Hi, is there any idea when this could be merged? I use lxml and want to migrate to the new toolchain to solve the problem with the old OpenSSL (on Google Play). Thanks for an update...

yaki29:
@FilipDem I'm right now working on kvCompiler for gsoc 2017. But I assure you I'll spare some time and complete it.

FilipDem:
Hi Yaki29,
I cannot find yet the lxml recipe in the master branch. Is this normal?
Thanks for your help!

FilipDem:
Is there any idea when the lxml will be available with the new toolchain? I am using this intensively and porting to cElementTree give really problems. Thanks for the info.

inclement:
@FilipDem This isn't something I'm likely to attack directly any time soon, so it probably won't get into p4a unless someone else submits a PR.

FilipDem:
Thanks for the information, Inclement. Is there any alternative? I tried using cElementTree, but it was not successful.

inclement:
@FilipDem I'm afraid I'm not familiar with these libraries and their alternatives.

If you are able, it would help to provide a simple runnable example using lxml functionality, so that anyone who does try to make a recipe has something to test to check if it's actually working.

Scefing:
I just added #1166 which should implement an `lxml` recipe. Needs testing, tho.

goffi-contrib:
Hello everybody, a PR based on #1166 has been merged, so this one can now be closed (ping @AndreMiras). Thanks to everybody who worked on that.

AndreMiras:
Fixed in https://github.com/kivy/python-for-android/pull/1428 thanks for the heads-up

PR #1004, opened 2017-02-01
BeautifulSoup4 testapp added

yaki29:
@inclement Now you can check. I have added testapp for bs4(beautifulsoup4). It shows movie details and needs internet to work.


inclement:
Great, but could you add this commit to your other PR at https://github.com/kivy/python-for-android/pull/999?

PR #1005, opened 2017-02-01
Documentation mistake corrected

yaki29:
Small documentation mistake "is" and "it" are used twice.
Sorry for three commits  I made some mistake.


inclement:
Thanks!

PR #1007, opened 2017-02-04
Log every python library load error

andrewm41:
Fix the situation when something happens while loading python2.7 library. Since it is loading first it's error is discarding and the error from only last python3 is shown.

This patch show every error with a note that there are some cases when it is normal.

Allows to debug python2 load library errors.


inclement:
I've added a slightly different check and improved error message as part of supporting different Python3 versions, which hopefully should resolve this problem.

PR #1011, opened 2017-02-12
Added support for Python 3.6

inclement:
~~Not for merging yet.~~

This PR removes some final references to python3.5 to make everything generic, so that python3.6 can be used by adding `python3crystax=3.6` in the requirements (the default remains 3.5 for now).

This won't actually do anything yet as CrystaX doesn't come with a prebuilt Python 3.6, but I've got this working as well. I'll add support for downloading a prebuilt one that I've generated, or for building a new one locally, shortly.


inclement:
This seems to be working now - Python 3.6 will only work for the archs armeabi and armeabi-v7a that I precompiled, but other that it seems to run fine.

I'll be looking at improving the python build to let python be built locally in the near future. In the meantime, it would be useful if anyone were able to test this before I merge it.

inclement:
I haven't had any python 3.6 related error reports after deploying an app with this, so I'll merge it.

PR #1014, opened 2017-02-15
Remove the excessive ccache in the command

MaChengxin:
It is the same as the one at the end of the command.


inclement:
Thanks.

PR #1021, opened 2017-03-18
Fixed protobuf cpp

Deniskore:
Android application crashes without STL shared library, so we need copy STL library for protobuff cpp.


inclement:
Thanks

PR #1034, opened 2017-04-11
Recipe updates and small fixes to build process

bobatsar:
New dateutil recipe and updated icu recipe.

Small addition to build process:
- build status threw an exception if no bootstrap was
  available e.g. after clean_bootstrap_builds
- remove duplicates in list of python packages to install


inclement:
Looks good, thanks!

PR #1037, opened 2017-04-12
Support ruamel.yaml, improved yaml lib.

mixmastamyk:
Details at:  https://pypi.python.org/pypi/ruamel.yaml

Bug necessitating patch: 
https://bitbucket.org/ruamel/yaml/issues/114/

### DIFF
--- a/pythonforandroid/recipes/ruamel.yaml/__init__.py
+++ b/pythonforandroid/recipes/ruamel.yaml/__init__.py
@@ -1,0 +1,14 @@
+from pythonforandroid.toolchain import PythonRecipe
+
+
+class RuamelYamlRecipe(PythonRecipe):
+    version = '0.14.5'
+    url = 'https://pypi.python.org/packages/5c/13/c120a06b3add0f9763ca9190e5f6edb9faf9d34b158dd3cff7cc9097be03/ruamel.yaml-{version}.tar.gz'
+
+    depends = [ ('python2', 'python3crystax') ]
+    site_packages_name = 'ruamel'
+    call_hostpython_via_targetpython = False
+
+    patches = ['disable-pip-req.patch']
+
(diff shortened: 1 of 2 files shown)


inclement:
It looks like this issue was resolved in ruamel.yaml, would it be better now to just set the env var that seems to have been added?

mixmastamyk:
Yes in general, however at the time I was using this, the fix had not made it to the release version on PyPI.  

Looks like it probably has by now.  I'm working on backend stuff now, but hope to get back to the kivy client in a week or two, and will take a look at a newer version of a ruamel.yaml  that does not need a patch.

inclement:
Okay, sorry about the delay merging this, if this hasn't made it to pypi yet just let me know and I'll merge this version.

AndreMiras:
I gave it a try and as in last pypi version `ruamel.yaml==0.15.77` the problem is still there.
I made a couple of fixes to the recipe and will merge if Travis is happy.

AndreMiras:
Conditional build is green, even though I'm not too sure why it decided to fallback to python2. Because to me all the dependencies set is also compatible with `python3crystax`, here is the relevant part of the log:
```
[INFO]:    Found 4 acceptable bootstraps: ['webview', 'service_only', 'sdl2', 'pygame']
[INFO]:    Found a single valid recipe set: ['hostpython2', 'python2', 'setuptools', u'ruamel.yaml']
[INFO]:    -> directory context testapps/
('recipes modified:', set([u'ruamel.yaml']))
('recipes to build:', set([u'ruamel.yaml']))
falling back to python2
('recipes to build (no broken):', set([u'ruamel.yaml']))
('requirements:', u'ruamel.yaml,python2')
```
More details: https://api.travis-ci.org/v3/job/453630101/log.txt

PR #1042, opened 2017-05-06
SDL2/Gradle bootstrap

tito:
This is a new SDL2 bootstrap, using gradle for building instead of ant. It can handle natively .aars (Android archive fragments), and external dependencies (fetch & compilation).

Because gradle is something new to everyone here (and me), let's just have a note about how the project structure is different:

Files removed:
```
ant.properties
build.xml
custom_rules.xml
```
Files/directory moved:
```
AndroidManifest.xml -> src/main/AndroidManifest.xml
assets -> src/main/assets
libs [.jar] -> src/main/libs
libs [.so] -> src/main/jniLibs
res -> src/main/res
src -> src/main/java
```

The jni directory in the root folder should stay for the moment (it should be in src/main/jni), but i get ndkBuild works with Android.mk within jni directory (aka, i can't pass any options)).


tito:
Well, it works with my app that includes aars + specific gradle dependencies.
Woot woot!

Someone care to test if standard build are not broken?
If not, then we could merge this as it is for now :)

pythonforandroid/bootstrap.py
```diff
@@ -271,6 +272,15 @@ def fry_eggs(self, sitepackages):
                     shprint(sh.mv, '-t', sitepackages, *files)
                 shprint(sh.rm, '-rf', d)
 
+    def apk(self, build_mode, env):
```
inclement: Does this function actually get called from somewhere? I see the ant/gradle build stuff is handled elsewhere.

inclement:
I'm closing this in favour of https://github.com/kivy/python-for-android/pull/1071, which combines everything into the existing sdl2 bootstrap in such a way that both ant and gradle can work.

PR #1045, opened 2017-05-10
[Core] Get the latest version of requests.

hobbestigrou:


KeyWeeUsr:
This might even work if it doesn't require a newer interpreter as it's just a standalone recipe. Try it without modifying the interpreter recipe and see if it works correctly on your device.

inclement:
This is probably fine, but I'll leave merging it until after the release unless someone can confirm it.

adrienbrunet:
I do confirm it works for me and our app  :)

inclement:
Thanks!

PR #1046, opened 2017-05-10
H[Core] Get the latest version of requests.

hobbestigrou:


KeyWeeUsr:
Closing, because this won't be so easy afaik as it needs patches for the whole interpreter. See #775 for reference :)

Issue #1050, opened 2017-05-14
python-for-android doesn't work with current Android SDK


inclement:
The rearranged SDK internals mean that python-for-android uses the wrong path to call the android binary, which also no longer should be used. p4a needs to know how to call sdkmanager instead.

hemna:
I just noticed this today too.  I just tried and the new Android SDK = Android studio.  I can't find the older SDK anywhere to try and get p4a to work.

inclement:
Fixed by https://github.com/kivy/python-for-android/pull/1057

PR #1051, opened 2017-05-14
Made p4a call `sdkmanager` instead of `android` in new Android SDK versions

inclement:


nocarryr:
In working through the same errors as in #1050 I've found that my android-sdk layout places `avdmanager` under `./tools/bin/avdmanager` instead of `./bin/avdmanager` (as shown in L232-L233).

I also have a depreciated version of the `./tools/android` callable, so in testing I had to move the `avdmanager` check to the top of the stack.

Not sure if this is what everyone is experiencing or not, but this is on a freshly installed VM (16.04) and fresh download of the sdk

inclement:
@nocarryr Yeah, this PR is not quite right as I downloaded the SDK from scratch and misunderstood the directory structure. I have an improved version that isn't pushed yet.

nocarryr:
@inclement :+1:  Thanks, just thought I'd share the discovery.  Have you also noticed any issues with a non-existent `build.xml` expected under `./tools/ant/` (within the sdk root)?

This may be a separate issue, but I thought I'd ask here in case you've come across it

inclement:
@nocarryr Thanks for reporting it, it's always useful!

The build.xml problem is because the build tools no longer ship with ant, it is also deprecated. I think you can install an older version for now, or tito has done the work to make p4a use gradle instead (PR at https://github.com/kivy/python-for-android/pull/1042), which we're working to review and merge.

nocarryr:
Thanks for the info, I haven't checked in on p4a progress in awhile so I'm​ catching up. Great work btw

PR #1063, opened 2017-06-02
Use Kivy setup.py flag instead of rmtree

KeyWeeUsr:
Instead of removing it manually we should use the [`KIVY_SPLIT_EXAMPLES` flag](https://github.com/kivy/kivy/blob/master/setup.py#L1022) set for it and prevent failing the build maybe if `rmtree` would be able to do the task.

ref #1062


inclement:
Thanks, much better :p

Issue #826, opened 2016-07-07
Recipes are still not resolved properly sometimes


inclement:
It's possible to end up with python2 but not hostpython2.

Here's a log demonstrating it, provided by frmdstryr on irc:

```
# Check that aidl can be executed
# Search for Aidl
# Run '/home/jrm/.buildozer/android/platform/android-sdk-20/build-tools/24.0.0/aidl'
# Cwd None
# Android packages installation done.
# Check application requirements
# Check garden requirements
# Compile platform
# Run 'python -m pythonforandroid.toolchain create --dist_name=gpsexample --bootstrap=sdl2 --requirements=plyer,kivy --arch armeabi-v7a --copy-libs'
# Cwd /home/jrm/workspace/plyer/examples/gps/.buildozer/android/platform/python-for-android-master
[INFO]:    Will compile for the following archs: armeabi-v7a
[INFO]:    Found Android API target in $ANDROIDAPI
[INFO]:    Available Android APIs are (19)
[INFO]:    Requested API target 19 is available, continuing.
[INFO]:    Found NDK dir in $ANDROIDNDK
[INFO]:    Got NDK version from $ANDROIDNDKVER
[INFO]:    Using Google NDK r9c
[INFO]:    Found virtualenv at /usr/local/bin/virtualenv
[INFO]:    Found the following toolchain versions: ['4.6', '4.8', 'clang3.3']
[INFO]:    Picking the latest gcc toolchain, here 4.8
[WARNING]: Missing executable: autoconf is not installed
[WARNING]: Missing executable: automake is not installed
[WARNING]: Missing executable: libtoolize is not installed
[INFO]:    No existing dists meet the given requirements!
[INFO]:    No dist exists that meets your requirements, so one will be built.
[INFO]:    Bootstrap requires recipes ['sdl2', ('python2', 'python3crystax')]
[INFO]:    No recipe named plyer; will attempt to install with pip
[INFO]:    Loaded recipe sdl2 (depends on [('python2', 'python3crystax'), 'sdl2_image', 'sdl2_mixer', 'sdl2_ttf'], conflicts ['sdl', 'pygame', 'pygame_bootstrap_components'])
[INFO]:    Loaded recipe kivy (depends on ['pyjnius', ('sdl2', 'pygame'), ('python2', 'python3crystax')])
[INFO]:    Loaded recipe sdl2_image (depends on [])
[INFO]:    Loaded recipe sdl2_mixer (depends on [])
[...]

PR #1064, opened 2017-06-04
Rewrite recipe graph

inclement:
This PR rewrites most of the recipe graph calculation, because the old version was extremely hacky and I couldn't actually work out how to fix it. The new version is much simpler but seems to give either the same or better results.

Fixes https://github.com/kivy/python-for-android/issues/826 and https://github.com/kivy/python-for-android/issues/1056 (and probably others).

~~Not for merging quite yet.~~ Seems to work.


inclement:
Merging because it seems to work, and I don't want to see this bug any more!

PR #1066, opened 2017-06-05
implement wakelock for sdl2

### LINKED ISSUE #1061: wakelock == 1 not preventing screen from locking on sdl2
After migrating to Kivy 1.10 / latest p4a I'm seeing the case where android device is not staying awake where it should be. 

wakelock == 1  (android.wakelock = True in buildozer.spec) and android permissions include WAKE_LOCK. 

our buildozer.spec file has not changed in this area when it was built for 1.9.1, and wakelock was working as expected. 

I hacked PythonActivity.java to output some debug information if it chose the branch to set wakelock, and I was able to see it setting the wakelock in action.  See screenshots:

image

image

Running latest buildozer / kivy / py4a

Let me know how else I can help debug.

brentpicasso:
resolves #1061. 

Not sure if this is the best implementation, but seems to work based on my limited testing.

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonActivity.java
+++ b/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonActivity.java
@@ -183,7 +183,8 @@ public class PythonActivity extends SDLActivity {
 
                 PowerManager pm = (PowerManager) mActivity.getSystemService(Context.POWER_SERVICE);
                 if ( mActivity.mMetaData.getInt("wakelock") == 1 ) {
-                    mActivity.mWakeLock = pm.newWakeLock(PowerManager.SCREEN_BRIGHT_WAKE_LOCK, "Screen On");
+                	mActivity.mWakeLock = pm.newWakeLock(PowerManager.SCREEN_BRIGHT_WAKE_LOCK, "Screen On");
+                	mActivity.mWakeLock.acquire();
                 }
                 if ( mActivity.mMetaData.getInt("surface.transparent") != 0 ) {
                     Log.v(TAG, "Surface will be transparent.");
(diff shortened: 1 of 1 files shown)


brentpicasso:
Note, I tried to implement the approach in the pygrame implementation, but I didn't get the initial onResume() when the app first fired up.  So, I acquired the lock immediately if the flag was set; seemed like there was no reason for it to not work that way.

The pygame implementation seemed cleaner with only one acquire and one release present, however.  

Feedback appreciated; interested in getting this merged asap as my app needs wakelock.

inclement:
Thanks for the PR, it looks okay but I haven't had time to test it yet.

There's actually an alternative way to do this with SDL2, which in Kivy you can access with from kivy.core.window import Window` and `Window.allow_screensaver = False`. I think this uses [FLAG_KEEP_SCREEN_ON](https://developer.android.com/reference/android/view/WindowManager.LayoutParams.html#FLAG_KEEP_SCREEN_ON) rather than a wakelock.

brentpicasso:
thanks for checking it out. Happy to continue using the official mechanism if it's not being deprecated.

inclement:
Seems good, thanks.

pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonActivity.java
```diff
@@ -452,5 +453,27 @@ protected void showLoadingScreen() {
     }
 
     }
+    
+    @Override
+    protected void onPause() {
+    	// fooabc
+        if ( this.mWakeLock != null &&  mWakeLock.isHeld()){
```
bionoid: Not ``this.mWakeLock.isHeld()``?

PR #1071, opened 2017-07-02
SDL2/Gradle bootstrap with fixes

inclement:
This PR is an extended version of tito's https://github.com/kivy/python-for-android/pull/1042 , which did not work for me without the new changes.

The main changes are:

- Moved .so files from `src/main/jniLibs` back to `libs`, and added this location explicitly to the `build.gradle`. I couldn't get the libs to be loaded properly without this change (although it might also work to move all of them, not just the libpython, and again add the location explicitly to `build.gradle`).
- Removed the NDK build stuff from the `build.gradle` file. This seemed to run the whole build a second time, and isn't necessary as p4a already does the ndk-build step as part of the SDL2 recipe. This also seemed to make some of the changes to Android.mk unnecessary.

Can anyone test if this new version works for them?


inclement:
@tito Would you be able to check if this PR works for you? I'm not sure if I've missed something.

inclement:
The sdl2_gradle bootstrap now supports both ant and gradle, either via a command line parameter or by automatically detecting if gradle will work by checking if it is present and that the android-sdk build-tools are new enough.

I intend to finally merge these changes back into the sdl2 bootstrap, so that the one bootstrap can take care of both build systems.

inclement:
I've now moved everything back to the sdl2 bootstrap, which supports both gradle and ant as available. I intend to merge it right after releasing, since it may break things and need a little time to fix any problems that come up.

KeyWeeUsr:
One of the changes fixed the license issue I think, because I don't experience it anymore. I got further with the build, but still couldn't build an app. I've never experienced [this error](https://gist.github.com/KeyWeeUsr/95e60ebf4ab392abba9c559349a513ac), so I don't know what to look for.

Also, I'm not sure, but think some recent changes also removed `--sdk` argument compatibility, so I needed to run the building command from buildozer myself (w/o the SDK argument). Then I got the linked error.

inclement:
Interesting error, not sure what's going on there. https://stackoverflow.com/questions/43916513/gradle-build-exception-aapt-process-not-ready-to-receive-commands seems to indicate that it's something to do with gradle versions, which seems understandable but probably needs some research to fix properly.

jb1123:
For me it has failed. I have downloaded pull request with:
git fetch upstream pull/1071/head:sdl2_gradle_inclement
git checkout sdl2_gradle_inclement
installed it from source
pip install --user -e /home/jb/python/ve27/python-for-android
Tried to build by apk and got error message suggesting it tried to access ant file no longer available in Android Sdk

BUILD FAILED	
/home/jb/.local/share/python-for-android/dists/nmeter/build.xml:93: Cannot find /home/jb/Android/Sdk/tools/ant/build.xml imported from /home/jb/.local/share/python-for-android/dists/nmeter/build.xml	

[error_log.txt](https://github.com/kivy/python-for-android/files/1194479/error_log.txt)

inclement:
@jb1123 Thanks, this is useful to know.

inclement:
@KeyWeeUsr Do you still get the error you posted at https://gist.github.com/KeyWeeUsr/95e60ebf4ab392abba9c559349a513ac ?

inclement:
I had no problems getting this working in an Ubuntu VM, so I'm inclined to merge it after doing some final tests. I won't be surprised if there are still issues, but at this point I think it's better to make people come across them in master since they won't get found otherwise.

Issue #1072, opened 2017-07-07
requests module not compiling in buildozer when used with Python3crystax


VineethRaghavan:
The recipe for the requests module gives a conflict error with regard to hostpython2 when I try to compile it as a dependency of kivy with python3crystax on buildozer. Please fix it.

Here are the logs:

```
kivy@kivyvm:~/Desktop$ python3.6 -m buildozer -v android debug
# Check configuration tokens
# Ensure build layout
# Check configuration tokens
# Preparing build
# Check requirements for android
# Run 'dpkg --version'
# Cwd None
Debian 'dpkg' package management program version 1.18.10 (amd64).
This is free software; see the GNU General Public License version 2 or
later for copying conditions. There is NO warranty.
# Search for Git (git)
#  -> found at /usr/bin/git
# Search for Cython (cython)
#  -> found at /usr/local/bin/cython
# Search for Java compiler (javac)
#  -> found at /usr/lib/jvm/java-8-openjdk-amd64/bin/javac
# Search for Java keytool (keytool)
#  -> found at /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/keytool
# Install platform
# Run 'pip install -q --user "appdirs" "colorama>=0.3.3" "sh>=1.10" "jinja2" "six"'
# Cwd None
# Apache ANT found at /home/kivy/.buildozer/android/platform/apache-ant-1.9.4
# Android SDK found at /home/kivy/.buildozer/android/platform/android-sdk-20
# Android NDK found at /media/sf_windows_desktop/crystax-ndk-10.3.2
# Check application requirements
# Check garden requirements
# Compile platform
# Run '/usr/bin/python3.6 -m pythonforandroid.toolchain create --dist_name=app --bootstrap=sdl2 --requirements=python3crystax==3.6,requests,pyrebase,kivy==master,git+https://gitlab.com/kivymd/KivyMD.git,pyrebase --arch armeabi-v7a --copy-libs --color=always --storage-dir=/home/kivy/Desktop/.buildozer/android/platform/build'
# Cwd /home/kivy/Desktop/.buildozer/android/platform/python-for-android-master
[INFO]:    Recipe python3crystax: version "3.6" requested
[INFO]:    Recipe kivy: version "master" requested
[INFO]:    Will compile for the following archs: armeabi-v7a
[INFO]:    Found Android API target in $ANDROIDAPI
[INFO]:    Available Android APIs are (19)
[...]

PR #1075, opened 2017-07-09
Removed conflict with python3crystax for requests

VineethRaghavan:
Remove python2 dependencies in requests recipe to remove conflict with python3crystax

### DIFF
--- a/pythonforandroid/recipes/requests/__init__.py
+++ b/pythonforandroid/recipes/requests/__init__.py
@@ -3,7 +3,7 @@ from pythonforandroid.toolchain import PythonRecipe
 class RequestsRecipe(PythonRecipe):
     version = '2.13.0'
     url = 'https://github.com/kennethreitz/requests/archive/v{version}.tar.gz'
-    depends = ['hostpython2', 'setuptools']
+    depends = []
     site_packages_name = 'requests'
     call_hostpython_via_targetpython = False
 


inclement:
It looks like this allows requests to be included (so the openssl dependency was apparently erroneous), but that it won't actually run on the device without openssl, which our python3 build doesn't yet include. Is this experience the same as yours?

inclement:
(For reference): Would be fixed by https://github.com/kivy/python-for-android/issues/1072

VineethRaghavan:
I didn't get an opportunity to try it out, so I don't know. But I'll believe what you say.

Regarding the openssl dependency, if requests does require openssl can I expect that dependency to be fixed by python for android 0.5?

VineethRaghavan:
BTW, why hadn't you added openssl to the python 3 build considering the fact that it is a dependency of a library like requests?

inclement:
I don't expect openssl for python3 to be added before the next release
(which should be quite soon), but it will be a priority afterwards.

inclement:
> BTW, why hadn't you added openssl to the python 3 build considering the fact that it is a dependency of a library like requests?

Because I haven't had time to do everything I'd like to, and this is something that has lost out so far.

VineethRaghavan:
Thanks for the reply. I'll be looking forward to the next release (hopefully with openssl support for python 3!).
I'll close this issue now.
Once again, thanks!

inclement:
I'll leave this open, the fix seems good, it's just that the lack of openssl prevents the request import from actually working.

AndreMiras:
I think [current version of requests recipe](https://github.com/kivy/python-for-android/blob/ccd5b59/pythonforandroid/recipes/requests/__init__.py) compiles just fine with python3crystax. I'm using it for a couple of projects.
Feel free to reopen otherwise.

PR #1092, opened 2017-08-21
glob.so - glob is not in Android's libc

dotnetslash:
This recipe creates a libglob.so allowing porting of C libs and programs that use glob. Patch created using sources from https://github.com/white-gecko/TokyoCabinet


dotnetslash:
oops. still learning git. Ignore this one.

PR #1094, opened 2017-08-21
recipe for libglob.so

dotnetslash:
Creates a -lglob for programs that use libc globbing that's absent in android.

### DIFF
--- a/pythonforandroid/recipes/libglob/__init__.py
+++ b/pythonforandroid/recipes/libglob/__init__.py
@@ -1,0 +1,74 @@
+"""
+    android libglob
+    available via '-lglob' LDFLAG
+"""
+from os.path import exists, join, dirname
+from pythonforandroid.toolchain import (CompiledComponentsPythonRecipe,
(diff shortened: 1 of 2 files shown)


KeyWeeUsr:
Just making a small backup about that this part of the project [isn't licensed under LGPL](https://github.com/white-gecko/TokyoCabinet/commit/83b26cdc6de64d2726968fe4026a1b23fb918305#diff-87dfdce25224bab6e555146630739a0d) like the whole [project](https://github.com/white-gecko/TokyoCabinet):

image

If you can, please try this PR without https://github.com/white-gecko/TokyoCabinet/commit/0086f173d120c166fd16e10d22e4b4c8e3c75a29, so that we could safely fall back to the BSD 3c instead of LGPL. :) For reason why you can look e.g. [here](https://fosswire.com/post/2007/04/the-differences-between-the-gpl-lgpl-and-the-bsd/).

dotnetslash:
issetguid is also missing on android. gotta fix that

PR #1100, opened 2017-08-27
Added "regex" module recipe

germn:
[regex](https://pypi.python.org/pypi/regex/) - alternative regular expression module, offers additional functionality.


KeyWeeUsr:
Thanks! :)

PR #1095, opened 2017-08-21
Ndk 15

dotnetslash:
Android NDK introduced optional unified headers in version 14. In 15, they are no longer optional. This request corrects flags as specifies an https://android.googlesource.com/platform/ndk/+/ndk-r15-release/docs/UnifiedHeaders.md

The changes broke the libffi and python2 in different ways. Fixes are included.

I suspect there's going to be impact on other recipes as well. Is there a test mechanism to test all available recipes?

PR #1103, opened 2017-08-29
Fixes for NDK 15+

inclement:
This includes the main changes of @dotnetslash from #1095, plus some further fixes that I needed to get builds working.

~~Not for merging yet, still checking at least one thing.~~


inclement:
This seems ready for merging to me, but testing would be appreciated. I've tried with the NDK r11c, r15c, and current CrystaX NDK.

inclement:
It looks like this breaks sqlite3, the same way as described at https://github.com/android-ndk/ndk/issues/332 . Looking into it.

inclement:
I think I'll also test this with NDK 16 (currently beta1 is available) before merging, as this removes the old platform headers. That should expose any issues with still using the old headers.

inclement:
Everything seems to be working now, including with the NDK 16 beta1. I've run quite a few tests using NDK r11c, r15c, r16-beta1 and crystax, and they all seem to work, so let's go ahead and merge.

astha-garg123:
How Adding -D_FILE_OFFSET_BITS=32 solves the issue ?

inclement:
@astha-garg123 I found information about this at https://github.com/android-ndk/ndk/issues/332.

PR #1106, opened 2017-09-01
Fixed ssl, sqlite and crystax library loads on Android versions before ~4.4

### LINKED ISSUE #866: "Couldn't load python3.5m: findLibrary returned null" on older versions of Android
When I try to run my APK on older versions of Android (<4.x) I'm getting a crash with this error: "Couldn't load python3.5m: findLibrary returned null"

I'm using the new python-for-android toolchain checked out a few days ago.

I get a popup on the device that says it's an "SDL Error"

logcat: http://pastebin.com/A0EJ8dQG
buildozer: http://pastebin.com/WCbU96TM (I did add python2 to the list of requirements but it did not help)

I also changed the Kivy recipe to use stable as opposed to master. I didn't think that was the problem.

Tested on a Kindle Fire HD (3.x fails), LG Volt (4.4 fails, but for a different reason much later in the startup process after kivy started).

Uninstalling and reinstalling the app did not help.

It looks python-for-android/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonUtil.java in the function loadLibraries it is failing to find either python2.7 or python3.5m.

If this all sounds familiar it's a similar problem to #645. In fact before I updated p4a I was at revision 084ddec which @kived had just checked in to fix #645. I was able to build a working version of my app with that revision. After I pulled the latest revision  (9d457e3636d3d55ee2ba781d99faed98c93f5b8f) I immediately started seeing this problem.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/36942982-couldn-t-load-python3-5m-findlibrary-returned-null-on-older-versions-of-android?utm_campaign=plugin&utm_content=tracker%2F11371&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F11371&utm_medium=issues&utm_source=github).
</bountysource-plugin>

inclement:
This PR rebases and restructures https://github.com/kivy/python-for-android/pull/903.

I've tested this only in the emulator, but I was able to reproduce the original issue, and this rebased PR seems to fix everything.

Fixes #866.

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonUtil.java
+++ b/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonUtil.java
@@ -5,38 +5,48 @@ import java.io.File;
 import android.util.Log;
 import java.util.ArrayList;
 import java.io.FilenameFilter;
+// import android.os.PatternMatcher;
+import java.util.regex.Pattern;
 
 public class PythonUtil {
 	private static final String TAG = "pythonutil";
 
+    protected static void addLibraryIfExists(ArrayList<String> libsList, String pattern, File libsDir) {
+        // pattern should be the name of the lib file, without the
+        // preceding "lib" or suffix ".so", for instance "ssl.*" will
+        // match files of the form "libssl.*.so".
+        File [] files = libsDir.listFiles();
+
+        pattern = "lib" + pattern + "\\.so";
+        Pattern p = Pattern.compile(pattern);
+        for (int i = 0; i < files.length; ++i) {
+            File file = files[i];
+            String name = file.getName();
+            Log.v(TAG, "Checking pattern " + pattern + " against " + name);
+            if (p.matcher(name).matches()) {
+                Log.v(TAG, "Pattern " + pattern + " matched file " + name);
+                libsList.add(name.substring(3, name.length() - 3));
+            }
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


KeyWeeUsr:
What about listing the `libDir` and fetching all `.so` files? It'd solve this issue even for other recipes. Also, is there a possibility that the NDK would generate versioned libs? If yes, then the approach will probably fail: `libmy.so.1`.

rnixx:
@KeyWeeUsr There has been complaints about auto loading all libs here https://github.com/kivy/python-for-android/pull/903#issuecomment-265801834

rnixx:
Nevertheless. It would be great if there was a way to explicitly define libs to be loaded in the specs file or so.

inclement:
@KeyWeeUsr I thought about that, and I don't actually agree with @akshayaurora's objection (it seems to me that anyone who can place .so files for automatic loading can probably also just modify an existing one), but went for the extra explicit option instead. Automatic loading would also have the problem that we'd still need to impose some ordering on what is loaded when, so it wouldn't be possible to be completely naive about it.

I'm thinking to add a Recipe attribute that specifies .so files which will be produced, then to try to load them in that order according to the recipe build order. I guess this will normally be correct. But, if this works, I think it's mergeable as it is to fix the issue for now.

PR #1109, opened 2017-09-05
Try system python when performing compileall

inclement:
This is kind of messy, but there doesn't seem to be a very neat way to do it.

The point of doing this is that the compileall can fail for some files if it tries to load .so files (which hostpython doesn't have available). In particular, this occurs for files which try to access unicode characters with `'\N{...}'`. For instance, several files in sympy do this, and currently crash when sympy is imported on the device. This PR would make the compilation work in such cases.

For now, I think discussion would be better than straight merging, but having the change documented could help someone.


inclement:
I suppose another thing we could do is detect files for which compilation has failed, and automatically whitelist them. This might be a good general solution, along with also flagging this problem to the user.

PR #1121, opened 2017-09-23
Fix errors caused by python_recipe being None

kdheepak:


inclement:
Thanks for the contribution. Is this to make the build work if you don't actually include python in the built dist?

kdheepak:
This was because I was getting error similar to #620 and I had to check for if `python_recipe` was not None to proceed. I still wasn't able to build an apk because I was getting some other error, so perhaps I may have been missing something fundamental.

JonasT:
This seems to be pretty old and lots of things changed regarding the affected code specifically of how the python recipes interact. Is this still issue still present and in need of fixing for latest p4a?

kdheepak:
I haven't used this in a while, so we can close this. If I remember correctly, I was trying some odd combination of inputs for iOS or Android, and that's why that variable was ending up as `None`.

JonasT:
thanks for the quick feedback :+1:

Issue #1130, opened 2017-10-01
Error on building FFPYPLAYER for VideoPlayer Widget


mfaessel:
Hi,

Need help…
I’m trying to build an Android APK with Kivy VideoPlayer and Video Widgets
But get a ffplayer error

Here is my buildozer.spec file
https://pastebin.com/xeEJX9PB

I have tried on OSX Sierra and it's the same error appears on the Buildozer VM build.

Some comments :

- the videos work fine in Kivy Launcher
- if i build without ffmepg or ffpyplayer the app is build but VideoPlayer don't work
- I even try with different Cython version downgrade but no luck

Any idea?

****
```

[INFO]:    -> running configure --disable-everything --enable-parser=h264,aac --enab...(and 679 more)
           working: See ./configure --help for available options.                                              Exception in thread background thread for pid 51407:
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/Users/martial/Library/Python/2.7/lib/python/site-packages/sh.py", line 2170, in background_thread
    handle_exit_code(exit_code)
  File "/Users/martial/Library/Python/2.7/lib/python/site-packages/sh.py", line 1929, in fn
    return self.command.handle_command_exit_code(exit_code)
  File "/Users/martial/Library/Python/2.7/lib/python/site-packages/sh.py", line 672, in handle_command_exit_code
    raise exc
ErrorReturnCode_1: 

[...]

PR #1134, opened 2017-10-05
Update ffpyplayer recipe and it's dependencies

germn:
- update versions
- partly fixes https://github.com/kivy/python-for-android/issues/1130
- compile ffpyplayer with SDL2 mixer
- tested with gradle/r15b

**Recipes don't work with NDK r15c+** yet due to [this](https://github.com/android-ndk/ndk/issues/477) problem.


inclement:
Thanks! Would you be interested in doing a PR to ffpyplayer instead of a patch? Since it's a kivy project, that should be fine, unless I missed something.

germn:
Done!

inclement:
Thanks!

PR #1139, opened 2017-10-15
Two humble changes

wexi:
1) For a local recipe it is useful to checkout code from a local git
2) Have to step back cryptography to V1.3 as V1.4 is not compatible with pyOpenSSL V0.14


inclement:
Thanks.

PR #1140, opened 2017-10-16
This current Twisted version actually runs!

wexi:
Had to update Twisted to the latest version as addTimeout to a deferred is not available in V16.0.
Could we develop a more aggressive recipe upgrades policy so we don't fall behind Linux so much?
Thanks, Enoch.

### DIFF
--- a/pythonforandroid/recipes/twisted/__init__.py
+++ b/pythonforandroid/recipes/twisted/__init__.py
@@ -12,10 +12,10 @@ import sh
 
 
 class TwistedRecipe(CythonRecipe):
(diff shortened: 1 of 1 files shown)


inclement:
Thanks for the fix!

I'd be very happy to improve our ongoing support for recipes, but it's difficult to keep updating them becuase no one person knows much about very many of the different ones. I'd like to get some automated testing working for all of them, but that's already tricky and wouldn't solve the update problem.

For what it's worth, fixes like this one where you verify the bug and the fix can generally be merged very quickly.

PR #1150, opened 2017-10-23
Add note about NDK version on 32-bit OS

bionoid:
Not sure about the markdown formatting, but it seems like  NDK's past r10e are all 64-bit only. Trying to use them on a 32-bit linux gives a "c compiler cannot produce executables" error

### DIFF
--- a/doc/source/quickstart.rst
+++ b/doc/source/quickstart.rst
@@ -90,7 +90,13 @@ You need to download and unpack the Android SDK and NDK to a directory (let's sa
 For the Android SDK, you can download 'just the command line
 tools'. When you have extracted these you'll see only a directory
 named ``tools``, and you will need to run extra commands to install
-the SDK packages needed.
+the SDK packages needed. 
+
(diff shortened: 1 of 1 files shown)


inclement:
Thanks.

bionoid:
Actually I had planned to update this issue but didn't get around to it apparently.

I do think the information is correct, ie that it is the latest 32-bit NDK, but it's not clear that it actually works. At the time I made the PR, I had built a successful APK on the 32-bit setup (I think -- it succeeded, did not test the APK).

However since then, I could not build another APK (it fails with a missing mp3 error iirc). I was mucking about with a lot of stuff, eventually reinstalled the VM from scratch. Had the same experience, built one APK then broke, couldn't fix it. 

Eventually I reinstalled a third time, meticulously snapshotting the VM at every stage during installation. Of course, this time, it didn't work the first build, probably because I screwed up something, and I couldn't get it working. 

So I think there are some issues with 32-bit..

PR #1155, opened 2017-11-02
Accept pypi fragmented URLs (#md5...)

wexi:
pypi.python.org is the most convenient place to obtain source tar.gz archives.
Thus, to update our stale recipes, we better adapt to their new URL scheme.
Thanks, Enoch.


inclement:
Looks good, thanks!

PR #1145, opened 2017-10-19
Adds zbar (and dependencies) support, refs #854

AndreMiras:
- zbar Python extension
  - zbar library
  - iconv library

PR #1156, opened 2017-11-03
Adds iconv's recipe

opacam:
Adds iconv's recipe (migrated from the old toolchain for the new toolchain)


AndreMiras:
Hi @opacam, could you give that iconv recipe a try?
https://github.com/kivy/python-for-android/pull/1145

opacam:
Of course, no problem...I didn't see that, sorry!!

AndreMiras:
Well it may also not work, for example I think I saw you're referencing to a `libiconv.a` static library. while I only deploy the dynamic one. But it's worth look at it.
Thank you very much for Pillow by the way :)

PR #1157, opened 2017-11-03
The pypi python return http 403 error on http

khertan:
The pypi python server return : error 403 : SSL is required. So build failed with an empty archive that can t be extracted.


khertan:
Fix #1158

inclement:
Thanks.

PR #1166, opened 2017-11-13
Add lxml recipe and other recipes from ZachGoldberg fork

Scefing:
This is a merge-compatible fork of [https://github.com/ZachGoldberg/python-for-android](url) meant to implement an `lxml` recipes and other recipes.

Please look it over and point out any changes that need to be made so it can be merged. I'd like to use `lxml`!


inclement:
Looks good, I'll try to test it. Have you tested it yourself?

pythonforandroid/recipes/levenshtein/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+    name="levenshtein"
+    version = '0.12.0'
+    url = 'https://pypi.python.org/packages/source/p/python-Levenshtein/python-Levenshtein-{version}.tar.gz'
+    depends = [('python2', 'python3'), 'setuptools']
+
+    call_hostpython_via_targetpython = False
+
+    def get_recipe_env(self, arch):
+        env = super(LevenshteinRecipe, self).get_recipe_env(arch)
+        libxslt_recipe = Recipe.get_recipe('libxslt', self.ctx)
```
Scefing: My testing application compiles until these lines get referenced, which I believe may be because the class `Recipe` is called instead of the instance.  But something seems to have accidentally filled my VM up to the brim so I need to clean it out before I can continue testing, so I haven't been able to test the change into an instance.

goffi-contrib:
Hello @Scefing and @inclement , any update on this PR? I need lxml in new tool chain, and would be happy to see this PR move forward. I've tried it it's not working at the moment, there are at least the `from pythonforandroid.toolchain import CompiledComponentsPythonRecipe` to replace by `from pythonforandroid.recipe import CompiledComponentsPythonRecipe`. Then the `CC` env doesn't seems correct, but I'm not familiar enough yet with the build environment to fix it right now.
Also it seems that this recipe it only for python 2.x, isn't it?

Thanks in advance

inclement:
@goffi-contrib No update, it isn't something I can prioritise compared to other important changes that are waiting, but testing is welcome in general.

About the python 2 part, that may just be a matter of adding python3crystax as a build option, i.e. `depends = [(python2, python3crystax), ...]`.

I haven't thought about the CC options, have you tried the current ones?

goffi-contrib:
Hello @Scefing and @inclement I've made a PR with fixes to this one at https://github.com/kivy/python-for-android/pull/1428, this has been tested working. I was not sure how to add the modification, but if you want to stay on this PR, the fixing patch can be found at https://github.com/kivy/python-for-android/pull/1428/commits/4fd0a6ddba4742f04cdd35b429fa1529f00bfab2 .

Thanks a lot for your work!

goffi-contrib:
Hi @AndreMiras I think this PR can be closed now.

AndreMiras:
Yep true, fixed in https://github.com/kivy/python-for-android/pull/1428

PR #1173, opened 2017-11-20
add recipes for pyrxp & reportlab

replabrobin:
Just adds recipes for reportlab pdf generation package & pyrxp rxp based python xml parser


inclement:
Thanks, looks good, I'll merge it shortly (before doing the 0.6 release) unless something comes up.

PR #1178, opened 2017-11-23
Update __init__.py to support SDL2

snuq:
I managed to get audiostream compiled and working using sdl2 on android.
I had to change the setup.py for audiostream, but I need help making my edits be system-independent, and how to make it able to use sdl OR sdl2....
My modified setup.py is over at: https://github.com/snuq/audiostream/blob/master/setup.py
Mostly i changed instances of 'SDL' into 'SDL2', and included '/home/kivy/.buildozer/android/platform/android-ndk-r9c/sources/android/support/include'

### DIFF
--- a/pythonforandroid/recipes/audiostream/__init__.py
+++ b/pythonforandroid/recipes/audiostream/__init__.py
@@ -16,13 +16,13 @@ class AudiostreamRecipe(CythonRecipe):
             sdl_include = 'sdl'
             sdl_mixer_include = 'sdl_mixer'
         elif 'sdl2' in self.ctx.recipe_build_order:
-            sdl_include = 'SDL'
+            sdl_include = 'SDL2'
             sdl_mixer_include = 'SDL2_mixer'
             
             #note: audiostream library is not yet able to judge whether it is being used with sdl or with sdl2.
             #this causes linking to fail against SDL2 (compiling against SDL2 works)
             #need to find a way to fix this in audiostream's setup.py
-            raise RuntimeError('Audiostream library is not yet able to configure itself to link against SDL2.  Patch on audiostream library needed - any help much appreciated!')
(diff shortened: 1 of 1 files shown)


inclement:
It looks like you only included one small commit, but you mention some other changes. Are some things missing from this PR?

snuq:
inclement: yeah, I mostly changed the audiostream setup.py, but im not sure how to include that into this since its a different project?

inclement:
@snuq In general we can include that kind of change by adding a patch to python-for-android, but actually in this case since audiostream is a Kivy project we can just update it. Could you open a PR for audiostream with the changes that you've made? If you mention the url of this issue, github should generate a link between them.

inclement:
Great, thanks!

snuq:
dont we still need to fix the include path? for one, if the user is not named 'kivy', it will fail...

pythonforandroid/recipes/audiostream/__init__.py
```diff
@@ -12,19 +12,16 @@ class AudiostreamRecipe(CythonRecipe):
...
-            sdl_include = 'SDL'
+            sdl_include = 'SDL2'
             sdl_mixer_include = 'SDL2_mixer'
-            
-            #note: audiostream library is not yet able to judge whether it is being used with sdl or with sdl2.
-            #this causes linking to fail against SDL2 (compiling against SDL2 works)
-            #need to find a way to fix this in audiostream's setup.py
-            raise RuntimeError('Audiostream library is not yet able to configure itself to link against SDL2.  Patch on audiostream library needed - any help much appreciated!')
+            env['USE_SDL2'] = 'True'
+            env['SDL2_INCLUDE_DIR'] = '/home/kivy/.buildozer/android/platform/android-ndk-r9c/sources/android/support/include'
```
AndreMiras: :cry:

AndreMiras:
> dont we still need to fix the include path? for one, if the user is not named 'kivy', it will fail...

Indeed, I just came across it, that should be a fix for a next pull request then :smile:

PR #1182, opened 2017-11-25
reportlab, preppy & pyrxp should depend on hostpython

replabrobin:
These all work for python2.7 >=3.3 so I changed to make depend on hostpython. This also fixes problems caused by installing preppy too early. Sorry for double pull. Am still a newby to buildozer etc etc


inclement:
The depends should probably be `python2`, not `hostpython`. Does that work for you?

replabrobin:
It works for me, but is python2 not restricting just to python2? reportlab, pyrxp & preppy will all work in python>=3.3. I thought hostpython was a kind of meta recipe. I normally use python 2.7 so will be happy with that. The lack of python2/hostpython in preppy caused the python2 build to fail because the python2 installation failed because the lib dir was created too early.

inclement:
Ah, for both python2 and python3 you can use ('python2', 
'python3crystax') (i.e. a tuple of strings) instead.

hostpython isn't a meta recipe, and you specifically want to install 
these modules after the python distribution for the device has been 
prepared. The above recipe names are the recipes that build this python 
distribution in each case.

replabrobin:
OK I change to use python2,python3crystax and check they build at least for python2.

AndreMiras:
Hi @replabrobin thanks for the PR, looking good to me, sorry for the delay.
Could you simply fix the conflicts and squash the commits, then we're good to merge :+1:

replabrobin:
Hi, I'm no longer working actively with Kivy, but I can try to do what you request. I am not a git person so I'm unsure how to do what you want in the context of the existing PR.  I did the following to my fork
1) merge upstream/master
2)  fix conflict in __init__.py
3) commit the merge
4) push

so my repository is allegedly up to date and has my changes.

AndreMiras:
Yes you could make a new one of course it's simpler when you're not used to rebasing/squashing, but it creates some noise on the issue tracker. I'm fine with it, but I would also be more than happy to assist you on discord to do it http://chat.kivy.org ping me when you're there

replabrobin:
I am really sorry, but I have no time to learn github workflows; if you want to apply the changes your self I attach a patch.
[p4a-rl-dependency.patch.txt](https://github.com/kivy/python-for-android/files/2397117/p4a-rl-dependency.patch.txt)

AndreMiras:
I can understand you don't have time for it, no worries. You can also just give me write access to your branch/PR and probably I should be able to do it.
https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

AndreMiras:
OK no worries it actually is to the fork that i need access if I want to rebase.
https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/

I've created a follow up pull request here https://github.com/kivy/python-for-android/pull/1377 (commit author is still you) so we can close that one.
Thank for your contributions!

PR #1189, opened 2017-12-05
 vlc recipe improvements

ibobalo:
vlc recipe compilation updated to match recent changes in vlc port android
vlc recipe improved precompiled aar processing


inclement:
Thanks!

PR #1197, opened 2017-12-15
Add support for minimum Android API.

tito:
This allows the compilation to use a different version for compiling while targetting a higher API version.

Typically, you may need to use API 23 for Java, but want to run on devices with API 19. If you compile with API 23, when running on older device (let's say 4.4.2), you'll end up with a symbol not found "srand", "signal", etc.

It is important to set the Android API minimum version when compiling in order to avoid theses symbol issues.

CF: https://stackoverflow.com/a/41079462/69877

### DIFF
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -160,7 +160,8 @@ class Context(object):
         self._ndk_dir = value
 
     def prepare_build_environment(self, user_sdk_dir, user_ndk_dir,
-                                  user_android_api, user_ndk_ver):
+                                  user_android_api, user_android_min_api,
+                                  user_ndk_ver):
         '''Checks that build dependencies exist and sets internal variables
         for the Android SDK etc.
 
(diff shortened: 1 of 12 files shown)


QuadTriangle:
I also had to change this line to successfully compile python2
https://github.com/kivy/python-for-android/blob/2b3d52bacd33992e040b39121221fa5dc55b86dd/pythonforandroid/archs.py#L38
 still,  getting 'srand' missing error from 'libSDL2.so'

inclement:
@QuadTriangle I just did a quick test and it worked okay, do you know what needs to happen to duplicate your problem?

AndreMiras:
Hi @tito and @inclement was this covered during the recent refactoring in #1412 ? If so feel free to close, otherwise maybe we want to rebase and fix the conflicts.

inclement:
Yep, the python3 changes implement the same thing (plus some fixes beyond this).
I don't think I ever saw this PR :/

PR #1201, opened 2017-12-19
FFpyplayer fix for NDK 15+

germn:
This PR:

1. Fixes recipe for NDK 15+ (tested on ver. 16) by disabling `-D_FILE_OFFSET_BITS=64` ([see here](https://github.com/android-ndk/ndk/issues/332#issuecomment-344560630)).
2. Updates dependencies versions.
3. Stabilizes `ffpyplayer` recipe version to latest tested commit: if we would would use master we can't guarantee same requirements will be compiled successfully every time. On the other hand someone who needs newest ffpyplayer can specify `ffpyplayer==master` directly in buildozer.


inclement:
Thanks!

PR #1205, opened 2017-12-21
sdl2: build.tmpl.gradle: sourceCompatibility  1.7

QuadTriangle:


inclement:
Thanks! Sorry about the delay in accepting this, I was away from working on p4a for a little while.

Issue #1141, opened 2017-10-18
Can't build numpy with current master, python 2, NDK 15


germn:
Output:
https://pastebin.com/2sH9aKgx

Related part:

```
compile options: '-Inumpy/core/src/private -Inumpy/core/src -Inumpy/core -Inumpy/core/src/npymath -Inumpy/core/src/multiarray -Inumpy/core/src/umath -Inumpy/core/src/npysort -Inumpy/core/include -I/home/p4a/test_kivy/.buildozer/android/platform/build/build/python-installs/test_kivy/include/python2.7 -c'
ccache: _configtest.c
/usr/bin/ccache arm-linux-androideabi-gcc -DANDROID -mandroid -fomit-frame-pointer -D__ANDROID_API__=19 -isystem /home/p4a/android-ndk-r15c/sysroot/usr/include/arm-linux-androideabi -isysroot /home/p4a/android-ndk-r15c/sysroot -I/home/p4a/test_kivy/.buildozer/android/platform/build/build/python-installs/test_kivy/include/python2.7 _configtest.o -o _configtest
/home/p4a/android-ndk-r15c/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/../lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: cannot open crtbegin_dynamic.o: No such file or directory
/home/p4a/android-ndk-r15c/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/../lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: cannot open crtend_android.o: No such file or directory
/home/p4a/android-ndk-r15c/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/../lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: cannot find -lc
/home/p4a/android-ndk-r15c/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/../lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: cannot find -ldl
collect2: error: ld returned 1 exit status
[...]

PR #1209, opened 2017-12-26
fix for #1141 - added ldflags to numpy

mahomahomaho:
Fix for #1141 , tested in 2.7 only, not sure if will work with py3 crystax, however it's copied from recipe built for crystax, so there is chance.


inclement:
Super, thanks!

Issue #1208, opened 2017-12-25
add-jar doesn't work


bauerj:
I'm trying to add classes from a jar file to the generated apk so I can use them with pyjnius.

In the buildozer spec file I added

    android.add_jars = gui/kivy/tools/jars/*.jar

Buildozer seems to pass that correctly to pythonforandroid:

```bash
/usr/bin/python -m pythonforandroid.toolchain apk --debug --bootstrap=sdl2 --dist_name Electrum --name Electrum --version 3.0.3 --package org.electrum.electrum --android_api 19 --sdk 19 --minsdk 9 --private /home/bauerj/electrum/.buildozer/android/app --permission INTERNET --permission WRITE_EXTERNAL_STORAGE --permission READ_EXTERNAL_STORAGE --add-jar /home/bauerj/electrum/gui/kivy/jars/core-3.3.1.jar --presplash /home/bauerj/electrum/./icons/electrum_presplash.png --icon /home/bauerj/electrum/./icons/electrum_launcher.png --orientation portrait --window --intent-filters /home/bauerj/electrum/gui/kivy/tools/bitcoin_intent.xml --copy-libs --arch armeabi-v7a --color=always --storage-dir=/home/bauerj/electrum/.buildozer/android/platform/build
```

However, the generated apk file contains no classes from `core-3.3.1.jar`. Even if I pass a JPEG image to `android.add_jars`, pythonforandroid does not complain.

I'm using the current master (8829312fd1871213849) for python-for-android.
Spec file: [buildozer.spec](https://github.com/kivy/python-for-android/files/1585839/buildozer.spec.txt)

Let me know if I can provide more information.

PR #1211, opened 2017-12-27
Include jar files in build.gradle

bauerj:
Jars from `android.add_jar` were copied to a directory but never referenced in the gradle build script.

Closes: #1208


inclement:
Thanks!

PR #1223, opened 2018-01-25
Python 3 fix for android package

wolfgangp:
This should work with Python 2 and 3. It's tested on Python 3.5 only. I did test `.decode` on unicode and regular str in Python 2.7 and it's fine.
Without this change, a buildozer requirements line like this
`requirements = python3crystax,kivy==master,pyjnius,android`
will result in a python error stating that a string cannot be concatenated to a bytes object, i.e. JAVA_NAMESPACE is a bytes object.

### DIFF
--- a/pythonforandroid/recipes/android/src/android/_android.pyx
+++ b/pythonforandroid/recipes/android/src/android/_android.pyx
@@ -175,7 +175,7 @@ api_version = autoclass('android.os.Build$VERSION').SDK_INT
 version_codes = autoclass('android.os.Build$VERSION_CODES')
 
 
-python_act = autoclass(JAVA_NAMESPACE + '.PythonActivity')
+python_act = autoclass(JAVA_NAMESPACE.decode("utf-8") + '.PythonActivity')
 Rect = autoclass('android.graphics.Rect')
 mActivity = python_act.mActivity
 if mActivity:


inclement:
Thanks, looks good but I'll test under python2 before merging.

jleaders:
`(JAVA_NAMESPACE.decode() if type(JAVA_NAMESPACE) is bytes and type('') is not bytes else JAVA_NAMESPACE)` is a ternary that would return the base string type regardless of interpreter, but I'm sure his patch is fine (I couldn't test either patch because when I put this in my buildozer.spec it didn't work, did I do something wrong?):
`requirements = python3crystax,kivy,android==git+git://github.com/wolfgangp/python-for-android,click`
I also tried:
`requirements = python3crystax,kivy,git+git://github.com/wolfgangp/python-for-android,click`

AndreMiras:
Just to make sure I understand before testing/merging, this was a problem when running on host Python3 only right?
So if I run buildozer/p4a with these requirements from host python3 I should see the issue, correct?

wolfgangp:
Yes, only running Python 3 will elicit the problem p4a had. I think tests confirming this works in Python 2 are not in.

wolfgangp:
I'm fine with closing this, as KeyWeeUsr's fix is more to the point.

Issue #1234, opened 2018-02-10
Cannot import name 'uname' on Windows


MisterY:
os.uname is not supported on Windows.

AndreMiras:
We still seem to rely on `os.uname()` rather than `platform.uname()` actually. I don't have a Windows platform to play with, but the fix should be pretty straight forward and pull requests are welcome @MisterY :smiley:

MisterY:
I submitted a fix a while back and it was rejected. Also, as you can see above, there was not much will to fix it on Windows due to a host of other issues that pop up afterwards, unfortunately.

SwagatRanjit84:
@inclement wont it work on windoiws?

Issue #1232, opened 2018-02-09
The sh Python module could not be found


MisterY:
Running on Windows 10, Python 3.6.4, installed the current (2018-02-09) version of p4a from GitHub with `pip install -e`.
When running p4a, it exits with 

>ERROR: The sh Python module could not be found, please install version 1.10 or higher

I have installed sh-1.12.14 but nothing changed.
Also tried installing with `pip install git+https://github.com/kivy/python-for-android.git` and `pip install python-for-android` with no change. The same error still shows up.
Any hints? Thanks!

MisterY:
If I try to import sh module in a new Python script, I get
>ImportError: sh 1.12.14 is currently only supported on linux and osx. please install pbs 0.110 (http://pypi.python.org/pypi/pbs) for windows support.

MisterY:
Considering that only .ln(), .cp(), and .Command() functions are used, there's gotta be a universal solution.

inclement:
python-for-android is not supported on Windows. This is just the first small example of many issues you'll come across if you try it. There's a PR that does a lot of the work, but it's considered just a proof of concept unless someone wants to polish it and maintain it.

MisterY:
I wonder what is the reason for that. Both Python and Android SDK work fine on Windows. 
Clearly I've missed any such statement in the texts I've read about p4a.
Pity

haywirephoenix:
I made the same mistake :(

SwagatRanjit84:
@inclement Won't it work on windows?

PR #1235, opened 2018-02-10
adjusting for windows.

MisterY:
Python 3.6
The only (known) item left to solve is how to check for required libs in toolchain.py.

This solves #1232 and #1234. 
p4a runs and shows the argument info. 

Requires further testing.

### DIFF
--- a/pythonforandroid/archs.py
+++ b/pythonforandroid/archs.py
@@ -1,5 +1,6 @@
 from os.path import (exists, join, dirname)
-from os import environ, uname
+from os import environ
+from platform import uname
 import sys
 from distutils.spawn import find_executable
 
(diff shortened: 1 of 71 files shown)


MisterY:
Now I'm getting `[WARNING]: Asked to compile for no Archs, so failing.`

inclement:
Instead of copying the sh import error catching everywhere, I think it would be methodologically better to put it in one place (a new p4a file) and then every other file can import it from there.

That said, I don't think there's much point in doing this unless p4a is changed to actually work on windows, doing something like https://github.com/kivy/python-for-android/pull/994 (filling in all the gaps, polishing it, and actually supporting it post-merge). Barring that, I think it would be better to detect the windows environment and exit with an error message before even trying the imports.

Thanks for the PR, but for these reasons I won't merge these changes.

MisterY:
Thanks for the feedback. Having that clearly stated somewhere in the ReadMe would have saved me (and perhaps others) significant time and effort which I could have contributed to other projects.

PR #1236, opened 2018-02-13
Added sympy recipe

inclement:


inclement:
I made this recipe ages ago, but I think I found a better way to fix the unicode thing, looking into it.

AndreMiras:
Looking good to me, I've just added a minor comment

AndreMiras:
Looks like this is not compiling, Travis fails with:
```
Please install the mpmath package with a version >= 0.19
```

Ohadforman:
How can I add Please install the mpmath package with a version >= 0.19 to ignore list?

PR #1252, opened 2018-04-03
compile numpy for x86

ilyaiqoqo:
different architectures have different include location, this PR allows to specify different include directory per architecture, so python x86 can be built.

numpy now uses the AR binary corresponding to architecture.

### DIFF
--- a/pythonforandroid/archs.py
+++ b/pythonforandroid/archs.py
@@ -9,6 +9,9 @@ from pythonforandroid.recipe import Recipe
 
 class Arch(object):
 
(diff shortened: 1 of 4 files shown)


inclement:
Thanks very much, sorry about the delay merging.

inclement:
I've reverted this merge as it broke building with the crystax ndk (or other NDKs without the sysroot folder). I've opened https://github.com/kivy/python-for-android/issues/1274 to remind about the need to fix it.

PR #1256, opened 2018-04-10
allow setting launchMode for the main activity in the manifest

SomberNight:
As in title, this allows setting the [activity launch mode](https://developer.android.com/guide/topics/manifest/activity-element.html#lmode) in the AndroidManifest from buildozer.spec.

It is cumbersome to manually (or via custom scripts) modify the manifest at `[app project]/.buildozer/android/platform/python-for-android/dist/[app]/templates/AndroidManifest.tmpl.xml`.

related: https://github.com/kivy/buildozer/pull/650


akshayaurora:
@SomberNight thanks

PR #1276, opened 2018-05-19
get_recipe_dir() also handle local recipes, fixes #1185

### LINKED ISSUE #1185: Local recipes can not be patched any longer
It seems since commit 9242bea826a2bdc58915d4954343aee5165d9ad4 local recipes which try to patch files can not run any longer. The call to get_recipe_dir() always returns a path in the ctx.root_dir and that is wrong for local recipes.

I tried to revert that commit here locally and now the patching works again.

AndreMiras:
First tries to return local recipe path or defaults to core one.

### DIFF
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -293,6 +293,14 @@ class Recipe(with_metaclass(RecipeMeta)):
         return join(self.get_build_container_dir(arch), self.name)
 
     def get_recipe_dir(self):
+        """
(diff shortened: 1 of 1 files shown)


AndreMiras:
@inclement here's a fix for #1185, I'm not sure it's the way you imagined it.
Jenkins build is green on my fork https://travis-ci.org/AndreMiras/python-for-android/builds/381028336

inclement:
Great, thanks!

Issue #1263, opened 2018-05-07
python-for-android recipe tests


inclement:
As far as I'm concerned, the core recipes that need testing are:

- python2
- python3crystax
- openssl
- sqlite
- numpy
- pyjnius
- kivy
- setuptools
- sdl2
- kivent
- ffpyplayer
- android
- plyer

This is the core set that I try to make sure will always build, and which are (I think) the most commonly used by p4a projects. They're probably the ones that need testing the most. All of these should be compatible except python2 with python3crystax, and openssl with python3crystax.

That said, I haven't put much deep thought into this, I'm happy to consider other recipes equally important if people suggest them.

LarsDu:
Do you have a working kivent recipe? It's the last recipe I need for my app!

inclement:
No, of all of these kivent is the one I haven't actually tried for a long time, I mostly rely on kovak to maintain it. Adding it to the tests would go some way to avoiding the accumulation of problems, though.

AndreMiras:
Great looks good for a start. I think introducing them in the current `.travis.yml` should be fairly, even though I know some may already fail.
However at the end the `.travis.yml` file will look ugly. I already don't like the long lines that we have. So I'm trying to think of another approche, like a side helper script or a Makefile. I'm not too sure.
Or we could start simple and just add more "ugly" long lines and see later for a refactoring.
How do you guys see this?

inclement:
@AndreMiras p4a has a couple of ways to set command line arguments via files, which I guess should be sufficient to solve that problem.

AndreMiras:
Agrr I wanted to take a look at `setup_testapp_python2_sqlite_openssl.py`, but `peewee` is causing nested virtualenv issues.
```
[ERROR]:   virtualenv is needed to install pure-Python modules, but
[ERROR]:   virtualenv does not support nesting, and you are running
[ERROR]:   python-for-android in one. Please run p4a outside of a
[ERROR]:   virtualenv instead.
```
This is annoying, because this setup has a lot of interesting recipes to test, such as `openssl` and `sqlite3`. And also even though it would be easily achievable to run outside a virtualenv, I don't really like it because we loose the isolation and rely on system modules. So I'm not too sure about it, what would you guys suggest. I currently see two ways:

1. create another `setup_testapp_python2_sqlite_openssl.py` and `testapp_sqlite_openssl/main.py` with no  `peewee` dep
2. go without a virtualenv at all :cry: 

What do you guys think?

[...]

PR #1277, opened 2018-05-19
Build `testapp_python2` using `setup_testapp_python2.py`

AndreMiras:
Has the following requirements:
sdl2,pyjnius,kivy,python2
refs #1263


AndreMiras:
So this of course doesn't fix completely #1277, but it's already an improvement because:

1. it compiles more requirements than the initial `p4a` command
2. it's more concise than the initial p4a command

I'll add other build in subsequent pull requests because I had a nested virtualenv issue that I would like to invest.

inclement:
> So this of course doesn't fix completely #1277

Do you mean https://github.com/kivy/python-for-android/issues/1263?

Great stuff though.

PR #1278, opened 2018-05-19
Build `testapp_sqlite_openssl` using `setup_testapp_python2.py`

AndreMiras:
Has the following requirements:
sdl2,pyjnius,kivy,python2,openssl,requests,sqlite3

peewee had to be skipped because pure python modules fail with
nested virtualenv issue, see:
https://github.com/kivy/python-for-android/issues/1263#issuecomment-390421054

refs #1263


AndreMiras:
Ouch:
```

AndreMiras:
Build is green now! &#x2705;

Using parallel building fixed the 4M logs issue and also speed up build.
I've made two commits, one that adds `testapp_sqlite_openssl` build and one that refactor `.travis.yml` to introduce the parallel thing. I think it's better having two separate commits, because we see the first one failed with that 4M logs issue and the second one fixes it. But if you guys prefer I can squash them.

PR #1282, opened 2018-05-20
Builds setuptools with both python2 & hostpython3crystax

AndreMiras:
Covers `setuptools` case in both `python2` and `hostpython3crystax`.
The `hostpython3crystax` build is supposed to fail until branch
`feature/ticket1154_hostpython3crystax_build_dir_and_synlink` is merged.
The merge will be done in next commit to demonstrate the fix.
Also adds android recipe case.
Plus removed the `uname -a` and `p4a apk --help` simple test cases since
the maximum of parallel run seems to be 5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,12 @@ before_install:
   - docker build --tag=p4a .
 
 env:
-  - COMMAND='uname -a'
-  - COMMAND='. venv/bin/activate && p4a apk --help'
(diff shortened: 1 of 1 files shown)


AndreMiras:
So as expected only the [last build fails](https://travis-ci.org/kivy/python-for-android/builds/381368753), because current `hostpython3crystax` recipe doesn't yet provide `hostpython`.
Merging https://github.com/kivy/python-for-android/pull/1249 should fix it. Let's find out...

tshirtman:
LGTM

PR #1283, opened 2018-05-21
Introduces linting (tox/flake8)

AndreMiras:
Makes sure lint checks are ran before `docker build` so we fail fast.
Currently `flake8` errors are ignored except for `E101` and `W191`.
This will demonstrate the fail fast case.
Next commit will fix these two errors in the code base.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ services:
   - docker
 
 before_install:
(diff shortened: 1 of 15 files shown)


AndreMiras:
Build is failing as expected. Second commit fixes linter errors and (hopefully) the build.

PR #1285, opened 2018-05-22
Fixes linter F821 (undefined name)

AndreMiras:
Fixes or ignores some undefined names.
Plus deletes `get_directory()` dead code.


pythonforandroid/util.py
```diff
@@ -128,20 +123,3 @@ def is_exe(fpath):
                 return exe_file
 
     return None
-
-
-def get_directory(filename):
```
AndreMiras: dead code

PR #1217, opened 2018-01-04
Fix building ssl for python3

agilewalker:
Pull request https://github.com/kivy/python-for-android/pull/1195 commits squashed and rebased onto master.

PR #1250, opened 2018-04-01
Greenlet Crystax/Python3 support, fixes #1245

AndreMiras:
Fixes CFLAGS, LDSHARED and LDFLAGS to point to Crystax.

I've tested it with the following `main.py` on Android.
```
import gevent

def foo():
    print('Running in foo')
    gevent.sleep(0)
    print('Explicit context switch to foo again')

def bar():
    print('Explicit context to bar')
    gevent.sleep(0)
    print('Implicit context switch back to bar')


if __name__ == '__main__':
    gevent.joinall([
        gevent.spawn(foo),
        gevent.spawn(bar),
    ])
```
It worked as expected, see `adb logcat | grep python` output:
```
04-01 23:48:22.448 10853 10895 I python  : Initialize Python for Android
04-01 23:48:22.449 10853 10895 I python  : Changing directory to the one provided by ANDROID_ARGUMENT
04-01 23:48:22.449 10853 10895 I python  : /data/user/0/org.test.myapp/files/app
04-01 23:48:22.451 10853 10895 I python  : Preparing to initialize python
04-01 23:48:22.455 10853 10895 I python  : crystax_python exists
04-01 23:48:22.455 10853 10895 I python  : calculated paths to be...
04-01 23:48:22.455 10853 10895 I python  : /data/user/0/org.test.myapp/files/app/crystax_python/stdlib.zip:/data/user/0/org.test.myapp/files/app/crystax_python/modules
04-01 23:48:22.459 10853 10895 I python  : set wchar paths...
04-01 23:48:22.750 10853 10895 I python  : Initialized python
04-01 23:48:22.750 10853 10895 I python  : AND: Init threads
04-01 23:48:22.750 10853 10895 I python  : testing python print redirection
04-01 23:48:22.751 10853 10895 I python  : Android path ['.', '/data/user/0/org.test.myapp/files/app/crystax_python/stdlib.zip', '/data/user/0/org.test.myapp/files/app/crystax_python/modules', '/data/user/0/org.test.myapp/files/app/crystax_python/site-packages']
04-01 23:48:22.752 10853 10895 I python  : os.environ is environ({'PYTHONHOME': '/data/user/0/org.test.myapp/files/app', 'PATH': '/sbin:/vendor/bin:/system/sbin:/system/bin:/system/xbin:/system/vendor/bin:/vendor/xbin:/system/vendor/xbin:/product/bin:/product/xbin', 
[...]

PR #1288, opened 2018-05-27
Groestlcoin hash

tshirtman:
Adds a recipe for groestlcoin_hash, a hash algorithm used by the Groestlcoin cryptocurrency.

This PR depends on #1217 because as with any electrum fork, the app using this recipe uses this patched p4a to allow using ssl with python3crystax, so this recipe wasn't tested on p4a master.

The overriden `get_recipe_env` could maybe be removed and the change it does be backported to the `CythonRecipe.get_recipe_env` method, at the same place `-lpython{}m' is added, so other recipes could benefit from it. I don't know if it could potentially lead to issues with others though, so that's an open question for me.

### DIFF
--- a/pythonforandroid/recipes/groestlcoin_hash/__init__.py
+++ b/pythonforandroid/recipes/groestlcoin_hash/__init__.py
@@ -1,0 +1,12 @@
+from pythonforandroid.recipe import CythonRecipe
+
+
+class GroestlcoinHashRecipe(CythonRecipe):
+    version = '1.0.1'
+    url = 'https://github.com/Groestlcoin/groestlcoin-hash-python/archive/{version}.tar.gz'
+    depends = ['python3crystax']
+    call_hostpython_via_targetpython = True
+    cythonize = False
+
+
+recipe = GroestlcoinHashRecipe()


pythonforandroid/recipes/groestlcoin_hash/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+        # libpythonX.Y location to LDFLAGS to be able to link against it
+        # maybe this should be done in the super method and benefit
+        # other recipes instead?
+        env = super(groestlcoin_hashRecipe, self).get_recipe_env(arch)
+        ndk_dir = self.ctx.ndk_dir
+        python_version = '.'.join(self.ctx.python_recipe.version.split('.')[:2])
+        ndk_path = os.path.join(
+            ndk_dir, 'sources', 'python', python_version, 'libs', arch.arch
+        )
+        env['LDFLAGS'] = ' -L {}'.format(ndk_path) + env['LDFLAGS']
```
inclement: This change makes sense, but I don't understand why it wouldn't be needed by other CythonRecipes.

AndreMiras: Agree, in fact I've been adding it in all my recipes e.g. https://github.com/AndreMiras/EtherollApp/blob/v20180517/src/python-for-android/recipes/scrypt/__init__.py#L26
That and also maybe the `-I` header path like here https://github.com/kivy/python-for-android/pull/1250/files#diff-569e13021e33ced8b54385f55b49cbe6

inclement: The include dir is already included by CythonRecipe, at https://github.com/kivy/python-for-android/blob/master/pythonforandroid/recipe.py#L1045

tshirtman:
If the pr #1217 is not going to be merged, i might rebase the PR on master instead to simplify the job of the users of these PR, as they would only need one recipe to patch, not two? Since this doesn't technically depends on it, and could be used separately.

tshirtman:
So i did the rebase, and additionally, i moved the get_recipe_env patch to CythonRecipe, as the comments seemed to agree on the usefulness of that.

Didn't test, but that seems straightforward, if the tests pass, we should be fine, right? :D

AndreMiras:
Yes the recipe looks straightforward indeed, make sure tests are passing. Basically just fix the PEP8 and let's see :crossed_fingers:

tshirtman:
Maybe i missed something, i looked if there was recent changes to CythonRecipe and didn't see anything more recent, so i applied it directly, maybe there was something elsewhere?

AndreMiras:
Yes I was thinking maybe it's already covered by the recent changes to `PythonRecipe.get_recipe_env()` since it's a parent class, see this pull request/file: https://github.com/kivy/python-for-android/pull/793/files#diff-ba2a2345fbe0c6c42bd6eb8536bf4564
Anyway If it's the case I don't think it's a big deal for now to have `-L` redundancy as long as the builds are passing.
If you could simply fix the PEP8 errors raised by Travis build that would be awesome, that would be enough to verify if covered recipes still build.

AndreMiras:
Thanks for the update, we're getting closer :clinking_glasses: 
FYI you can run the linter locally, by simply calling the `tox` command.
Also I think the commit history would look way nicer if we squash one feature into one single commit.

AndreMiras:
Comments addressed and compiled fine in my Docker container.
Compiled command was:
```
python setup_testapp_python3.py apk --sdk-dir /opt/android/android-sdk --ndk-dir /opt/android/crystax-ndk --requirements python3crystax,setuptools,android,groestlcoin_hash
```

PR #1291, opened 2018-06-03
fixes find_library when deployed external storage

AndreMiras:
Fixes `find_library()` `OSError`/`FileNotFoundError` exception when app
deployed external SD storage.
Apps deployed on external SD storage have different root prefix.
External SD storage apps root dir prefix is like:
```
/mnt/expand/<sd-card-id>/user/0/<package.domain>.<package.name>/files/app
```
While internal storage apps root dir prefix is:
```
/data/data/<package.domain>.<package.name>/files/app
```
Hence the `[0:4]` trick doesn't work.

### DIFF
--- a/pythonforandroid/recipes/python2/patches/ctypes-find-library-updated.patch
+++ b/pythonforandroid/recipes/python2/patches/ctypes-find-library-updated.patch
@@ -12,7 +12,7 @@ index 52b3520..01b13a9 100644
 +if True:
 +    def find_library(name):
 +        # Check the user app lib dir
-+        app_root = os.path.abspath('./').split(os.path.sep)[0:4]
++        app_root = os.path.abspath('../../').split(os.path.sep)
 +        lib_search = os.path.sep.join(app_root) + os.path.sep + 'lib'
 +        for filename in os.listdir(lib_search):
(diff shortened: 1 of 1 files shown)


inclement:
Looks good, did you test it on a device in the end? If not I'll try to do that before merging.

AndreMiras:
Yes I'm using that fix in one of my apps with monkey patching here:
https://github.com/AndreMiras/EtherollApp/commit/604f5c632c307efbef698f84cef632f529024c41
So I do not actually use the fix from a p4a fork, but rather monkey patching with the same fix.

When the fix is not applied the app crashes when deployed on SD card. With the fix it's OK.
Basically on some device, download the APK from web browser and installing it, goes by default on SD card and then I received some Sentry reports. On some other devices it goes to the internal storage, but going to app info and changing it for external storage you can reproduce the crash.

inclement:
Great, thanks. I'm delighted that this probably fixes some of the mysterious crashes I've heard of but never been able to diagnose.

AndreMiras:
Great thanks for the merge. I hope so for the crashes. For records here are some stack trace from these crashes:
https://github.com/AndreMiras/EtherollApp/issues/96

PR #1296, opened 2018-06-10
Fix twisted and cryptography build for python 3

om26er:
Fix Twisted and Cryptography to not conflict with python3crystax

Note: Cryptography still does not successfully build, but at least it no longer complains about a conflict. https://paste.ubuntu.com/p/nMxXnP7tzH/


om26er:
I think the CI error there is unrelated to my change.

inclement:
Thanks!

PR #1307, opened 2018-06-29
Improves gevent recipe robustness

AndreMiras:
Tries to match and replace patterns to be moved from one environment variable to another rather than hard coding it.
That way unexpected flags added to parent recipes hierarchy also get handled.
This issue popped when testing https://github.com/kivy/python-for-android/pull/793

Also added some logs that produced the following output:
```
[INFO]:    Moved "-DANDROID -D__ANDROID_API__=19 -I/home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -I/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/include/python3.6 -I/home/andre/.buildozer/crystax-ndk-10.3.2/sources/python/3.6/include/python" from CFLAGS to CPPFLAGS.
[INFO]:    Moved "-lm -lcrystax -lpython3.6m" from LDFLAGS to LIBS.
```

Note that since it drops all the explicit `-l`, `-L`, `-I` it now relies on the WIP #793 and was tested with it. I deployed an app on device and it worked OK.
I'll test it again once #793 gets merged, in case things have moved.

### DIFF
--- a/pythonforandroid/recipes/gevent/__init__.py
+++ b/pythonforandroid/recipes/gevent/__init__.py
@@ -1,4 +1,5 @@
-import os
+import re
+from pythonforandroid.logger import info
 from pythonforandroid.recipe import CompiledComponentsPythonRecipe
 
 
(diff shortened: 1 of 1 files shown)


inclement:
Sorry for falling behind on this, is it still good to merge?

AndreMiras:
Thanks for asking, I think it's still valid, but I didn't get much time to double check. I'd like to have the feedback from https://github.com/kivy/python-for-android/issues/1333 so I'm waiting for his logs https://github.com/kivy/python-for-android/issues/1333#issuecomment-415991461
Basically I would like to see if he has errors regarding `LDFLAGS` or `CFLAGS` not being used properly in his log file and see if this PR fixes it for him. That the issue I had back then, but didn't get a chance to verify again since master moved.

PR #1308, opened 2018-07-04
Reduce docker image size

opacam:
The final docker image size is reduced about 2GB by cleaning up the apt files and removing unneeded compressed files.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,21 +36,24 @@ ENV ANDROID_NDK_DL_URL="https://dl.google.com/android/repository/${ANDROID_NDK_A
 
 # install system dependencies
 RUN apt update -qq && apt install -qq --yes --no-install-recommends \
-	python virtualenv python-pip wget curl lbzip2 patch bsdtar
+	python virtualenv python-pip wget curl lbzip2 patch bsdtar && \
+    rm -rf /var/lib/apt/lists/*
 
 # build dependencies
 # https://buildozer.readthedocs.io/en/latest/installation.html#android-on-ubuntu-16-04-64bit
 RUN dpkg --add-architecture i386 &&  apt update -qq && apt install -qq --yes --no-install-recommends \
 	build-essential ccache git libncurses5:i386 libstdc++6:i386 libgtk2.0-0:i386 \
 	libpangox-1.0-0:i386 libpangoxft-1.0-0:i386 libidn11:i386 python2.7 python2.7-dev \
-	openjdk-8-jdk unzip zlib1g-dev zlib1g:i386
+	openjdk-8-jdk unzip zlib1g-dev zlib1g:i386 && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -36,21 +36,24 @@ ENV ANDROID_NDK_DL_URL="https://dl.google.com/android/repository/${ANDROID_NDK_A
...
+    rm -rf /var/lib/apt/lists/*
 
 # build dependencies
 # https://buildozer.readthedocs.io/en/latest/installation.html#android-on-ubuntu-16-04-64bit
 RUN dpkg --add-architecture i386 &&  apt update -qq && apt install -qq --yes --no-install-recommends \
 	build-essential ccache git libncurses5:i386 libstdc++6:i386 libgtk2.0-0:i386 \
 	libpangox-1.0-0:i386 libpangoxft-1.0-0:i386 libidn11:i386 python2.7 python2.7-dev \
-	openjdk-8-jdk unzip zlib1g-dev zlib1g:i386
+	openjdk-8-jdk unzip zlib1g-dev zlib1g:i386 && \
+    rm -rf /var/lib/apt/lists/*
```
AndreMiras: why running it twice rather than once at the end?

AndreMiras:
Thank you very much for taking a look at this. In fact I didn't remove these files on purpose for 2 reasons:

1. it may ease debugging if needed in the interactive shell
2. the image is not cached or saved to DockerHub

So I didn't see the point reducing the size. But I'm open to discussions regarding it. Just that often when I clean my Docker image, I end up being annoyed when I want to verify/update/debug things from the interactive shell.
What's your point of view?

Dockerfile
AndreMiras: Yes I was suspecting that, wanted to make sure. Thanks!

opacam:
I think that once we unzip/untar/install all the ndk/sdk/apt files there is no need to have those in our system...I think that are useless for debugging purposes...unless we could reuse it when we do another image....but i don't thing that docker use that (unless we create another image based on that image...but if we do that we already have again the uncompressed files...again useless).

I found an answer in stackoverflow that talks about the docker caching system  that could clarify a little how the caching system of docker works: https://stackoverflow.com/questions/25102272/why-doesnt-docker-hub-cache-automated-build-repositories-as-the-images-are-bein

The point of reducing docker image size will benefit all of those who work with ssd disks, who have limited disk space.

Saying all that, maybe i miss some point about interactive shell debugging and the use of those compressed files...I don't see why we should keep them...Maybe I miss something?

AndreMiras:
An example where I had to use the archive was to play with the `--exclude` to see if more/less directories should be excluded for instance. But I'm fine deleting it. I'll just remove this temporary line if I ever need to debug things like that.
Thanks for the PR :+1:

PR #1332, opened 2018-08-18
fix android import

akshayaurora:


inclement:
Sorry @AndreMiras, that was sloppy of me. I actually thought it was more obvious than it really is what was wrong. It would have been most appropriate to have added the noqa, in lieu of whatever improvements people might want to make to the android module itself in the future.

AndreMiras:
Thanks for the #1340 pull request that fixes to tox.

Issue #1326, opened 2018-08-10
No "pil" or "pillow" avaliable for Python 3


JonasT:
I renamed this issue and contents a couple of times because I'm silly and confused about PIL & pillow. Anyway, the state is that none of them are available with `python3crystax` since `PIL` is hard-wired to python 2 and `--requirements=pillow` doesn't work, since without a recipe it packages the wrong `.so` library (incorrect architecture). That means, effectively, no PIL/pillow for any python 3 app which is unfortunate :cry:

---
old issue text:

The "pil" recipe is misguidedly hardwired to python 2:

```
 depends = ['python2', 'png', 'jpeg']
```

This breaks the build with Python 3:
```
# p4a apk --debug --private /my-app/ --orientation user --window --package=org.example.myapp --name "My application" --version 0.1 --bootstrap=sdl2 --requirements=python3crystax,pil,sdl2,pysdl2,pyjnius --arch=armeabi-v7a
[INFO]:    Will compile for the following archs: armeabi-v7a
[INFO]:    Found Android API target in $ANDROIDAPI
[INFO]:    Available Android APIs are (19)
[INFO]:    Requested API target 19 is available, continuing.
[INFO]:    Found NDK dir in $ANDROIDNDK
[INFO]:    Got NDK version from $ANDROIDNDKVER
[INFO]:    Using Crystax NDK 10.3.2
[INFO]:    Found virtualenv at /usr/bin/virtualenv
[INFO]:    ccache is missing, the build will not be optimized in the future.
[INFO]:    Found the following toolchain versions: ['4.9', '5', 'clang3.6', 'clang3.7']
[INFO]:    Picking the latest gcc toolchain, here 5
[INFO]:    No existing dists meet the given requirements!
[INFO]:    No dist exists that meets your requirements, so one will be built.
[ERROR]:   Didn't find any valid dependency graphs.
[ERROR]:   This means that some of your requirements pull in conflicting dependencies.
[ERROR]:   Exiting.
```
I suggest this should be fixed.

JonasT:
Based on a conversation in IRC chat, it seems like this is actually a fault in the `pil` recipe being hardcoded for Python 2 instead of working with both. I updated the ticket description accordingly.

AndreMiras:
[...]

Issue #1114, opened 2017-09-13
pil and pillow modules for python3


sherifawad:
## pil doesn't support python 3
pillow has no recipe
cv depend on numpy and there is problem with crystax

*****error*****
[INFO]:    # Prebuilding recipes
[INFO]:    Prebuilding hostpython3crystax for armeabi-v7a
[INFO]:    hostpython3crystax has no prebuild_armeabi_v7a, skipping
[INFO]:    Prebuilding jpeg for armeabi-v7a
[INFO]:    jpeg has no prebuild_armeabi_v7a, skipping
[INFO]:    Applying patches for jpeg[armeabi-v7a]
[INFO]:    jpeg already patched, skipping
[INFO]:    Prebuilding png for armeabi-v7a
[INFO]:    png has no prebuild_armeabi_v7a, skipping
[INFO]:    Prebuilding sdl2_image for armeabi-v7a
[INFO]:    sdl2_image has no prebuild_armeabi_v7a, skipping
[INFO]:    Applying patches for sdl2_image[armeabi-v7a]
[INFO]:    sdl2_image already patched, skipping
[INFO]:    Prebuilding sdl2_mixer for armeabi-v7a
[INFO]:    sdl2_mixer has no prebuild_armeabi_v7a, skipping
[INFO]:    Applying patches for sdl2_mixer[armeabi-v7a]
[INFO]:    sdl2_mixer already patched, skipping
[INFO]:    Prebuilding sdl2_ttf for armeabi-v7a
[INFO]:    sdl2_ttf has no prebuild_armeabi_v7a, skipping
[INFO]:    Prebuilding python3crystax for armeabi-v7a
[INFO]:    python3crystax has no prebuild_armeabi_v7a, skipping
[INFO]:    Prebuilding pil for armeabi-v7a
[INFO]:    pil has no prebuild_armeabi_v7a, skipping
[INFO]:    Applying patches for pil[armeabi-v7a]
[INFO]:    pil already patched, skipping
[INFO]:    Prebuilding sdl2 for armeabi-v7a
[INFO]:    sdl2 has no prebuild_armeabi_v7a, skipping
[INFO]:    Applying patches for sdl2[armeabi-v7a]
[INFO]:    sdl2 already patched, skipping
[INFO]:    Prebuilding six for armeabi-v7a
[INFO]:    six has no prebuild_armeabi_v7a, skipping
[INFO]:    Prebuilding pyjnius for armeabi-v7a
[...]

PR #1339, opened 2018-08-26
Add Pillow Recipe

krinnewitz:
This patch adds a new recipe for Pillow. There already is a recipe for PIL, but PIL does not support python3.

This new recipe should work with `python3crystax` as well as with `python2`. However, with `python2`, there are the following issues:

- The existing recipe for `setuptools` provides an older version of `setuptools`, which is too old for Pillow. This can be circumvented by specifying `setuptools==40.0.0` when building an APK. I am not sure if it is possible to specify a minimum version for dependencies inside recipes.
- There seems to be an issue with the `jpeg` recipe which requires `libcutils`. I resolved it locally by removing `libcutils` from  `pythonforandroid/recipes/jpeg/build-static.patch`.

With `python3crystax`, there should be no issues.

This PR addresses #1114 and #1326.

### DIFF
--- a/pythonforandroid/recipes/Pillow/__init__.py
+++ b/pythonforandroid/recipes/Pillow/__init__.py
@@ -1,0 +1,72 @@
+from pythonforandroid.recipe import CompiledComponentsPythonRecipe
+from pythonforandroid.toolchain import shprint
+from os.path import join, dirname
+import sh
+
+
+class PillowRecipe(CompiledComponentsPythonRecipe):
+
+    version = '5.2.0'
+    url = 'https://github.com/python-pillow/Pillow/archive/{version}.tar.gz'
+    site_packages_name = 'Pillow'
+    depends = [
+        ('python2', 'python3crystax'),
+        'png',
+        'jpeg',
+        'freetype',
+        'setuptools'
+    ]
+    patches = [
+        join('patches', 'fix-docstring.patch'),
+        join('patches', 'fix-setup.patch')
+    ]
+
+    call_hostpython_via_targetpython = False
+
+    def get_recipe_env(self, arch=None):
+        env = super(PillowRecipe, self).get_recipe_env(arch)
+        py_ver = self.ctx.python_recipe.version[0:3]
+
+        ndk_dir = self.ctx.ndk_platform
+        ndk_lib_dir = join(ndk_dir, 'usr', 'lib')
+        ndk_include_dir = (join(self.ctx.ndk_dir, 'sysroot', 'usr', 'include')
+            if py_ver == '2.7' else join(ndk_dir, 'usr', 'include'))
+
+        png = self.get_recipe('png', self.ctx)
+        png_lib_dir = png.get_lib_dir(arch)
+        png_jni_dir = png.get_jni_dir(arch)
+
+        jpeg = self.get_recipe('jpeg', self.ctx)
+        jpeg_lib_dir = jpeg.get_lib_dir(arch)
+        jpeg_jni_dir = jpeg.get_jni_dir(arch)
+
+        env['JPEG_ROOT'] = '{}|{}'.format(jpeg_lib_dir, jpeg_jni_dir)
+        env['ZLIB_ROOT'] = '{}|{}'.format(ndk_lib_dir, ndk_include_dir)
+
+        cflags = ' -nostdinc'
+        cflags += ' -I{} -L{}'.format(png_jni_dir, png_lib_dir)
+        cflags += ' -I{} -L{}'.format(jpeg_jni_dir, jpeg_lib_dir)
+        cflags += ' -I{} -L{}'.format(ndk_include_dir, ndk_lib_dir)
+
+        gcc_lib = (shprint(sh.gcc, '-print-libgcc-file-name')
+            .stdout.decode('utf-8').split('\n')[0])
+        gcc_include = join(dirname(gcc_lib), 'include')
+        cflags += ' -I{}'.format(gcc_include)
+
+        if self.ctx.ndk == 'crystax':
+            py_inc_dir = join(self.ctx.ndk_dir, 'sources', 'python', py_ver,
+                'include', 'python')
+            py_lib_dir = join(self.ctx.ndk_dir, 'sources', 'python', py_ver,
+                'libs', arch.arch)
+            cflags += ' -I{}'.format(py_inc_dir)
+            env['LDFLAGS'] += ' -L{} -lpython{}m'.format(py_lib_dir, py_ver)
+
+        env['LDFLAGS'] += ' {} -L{}'.format(env['CFLAGS'], self.ctx.libs_dir)
+        if cflags not in env['CFLAGS']:
+            env['CFLAGS'] += cflags
+        env['LDSHARED'] = '{} {}'.format(env['CC'],
+            '-pthread -shared -Wl,-O1 -Wl,-Bsymbolic-functions')
+        return env
+
+
+recipe = PillowRecipe()
--- a/pythonforandroid/recipes/Pillow/patches/fix-docstring.patch
+++ b/pythonforandroid/recipes/Pillow/patches/fix-docstring.patch
@@ -1,0 +1,13 @@
+diff --git a/src/PIL/__init__.py b/src/PIL/__init__.py
+index a07280e..6b9fe99 100644
+--- a/src/PIL/__init__.py
++++ b/src/PIL/__init__.py
+@@ -24,7 +24,7 @@ PILLOW_VERSION = __version__ = _version.__version__
+ 
+ del _version
+ 
+-__doc__ = __doc__.format(__version__)  # include version in docstring
++__doc__ = ''
+ 
+ 
+ _plugins = ['BlpImagePlugin',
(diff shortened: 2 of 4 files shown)


inclement:
Looks good to me, but I'll let @AndreMiras check it since I think he's followed the development more.

About setuptools, it should be fine to update the version in the recipe, as long as it does work (which it sounds like it does).

About libcutils, I don't know what this is used for, do things seem to work fine without it?

AndreMiras:
Thank you very much for the pull request @krinnewitz it's looking good so far. I'll give it a try with https://github.com/kivy-garden/garden.zbarcam later tonight and keep you posted.

krinnewitz:
Thanks guys. I have resolved the problems detected by flake8.

Regarding `setuptools`: Never mind. I cannot reproduce the problem anymore. Maybe a mistake on my side.

Regarding libjpeg: I was able to load a jpeg image without `libcutils`. So I guess it is not needed. However, I am not entirely sure about that since I have not tested excessively.

inclement:
> Regarding setuptools: Never mind. I cannot reproduce the problem anymore. Maybe a mistake on my side.

Great. It would still be fine to update the recipe version if you tested the newer one and it worked, there recipe exists more for internal reasons than because the version should matter.

krinnewitz:
> Great. It would still be fine to update the recipe version if you tested the newer one and it worked, there recipe exists more for internal reasons than because the version should matter.

Ok, I have added another commit which upgrades it to 40.0.0.

> About libcutils, I guess let's leave it for now but maybe leave a comment about it in the recipe. If someone really wants to use pillow with python2, they can look into it.

Actually, the problem should occur for any app using `jpeg` and `python2`. I have just created #1341 for that. Hope that's okay.

JonasT:
Once you want to have me test this with my `python3crystax` app, drop me a note. I'd be happy to give it a try

pythonforandroid/recipes/Pillow/patches/fix-docstring.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/src/PIL/__init__.py b/src/PIL/__init__.py
```
AndreMiras: Minor, FWIW I would reference the upstream bug https://github.com/python-pillow/Pillow/issues/3231

pythonforandroid/recipes/Pillow/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+    call_hostpython_via_targetpython = False
+
+    def get_recipe_env(self, arch=None):
+        env = super(PillowRecipe, self).get_recipe_env(arch)
+        py_ver = self.ctx.python_recipe.version[0:3]
+
+        ndk_dir = self.ctx.ndk_platform
+        ndk_lib_dir = join(ndk_dir, 'usr', 'lib')
+        ndk_include_dir = (join(self.ctx.ndk_dir, 'sysroot', 'usr', 'include')
+            if py_ver == '2.7' else join(ndk_dir, 'usr', 'include'))
```
AndreMiras: Was the `zlib.h` include actually deployed here on your Python 2.7 installation? Because on mine it's seems that `join(ndk_dir, 'usr', 'include'))` doesn't exist, while the former does `(join(self.ctx.ndk_dir, 'sysroot', 'usr', 'include')`.
So basically when running recipe on Python2, the `ZLIB_ROOT` would contain the following:
```
/home/andre/.buildozer/android/platform/android-ndk-r9c/platforms/android-19/arch-arm/usr/lib|/home/andre/.buildozer/android/platform/android-ndk-r9c/sysroot/usr/include
```
The lib directory seems fine:
```
ls /home/andre/.buildozer/android/platform/android-ndk-r9c/platforms/android-19/arch-arm/usr/lib/
crtbegin_dynamic.o  crtbegin_so.o  crtbegin_static.o  crtend_android.o  crtend_so.o  libandroid.so  libc.a  libc.so  libdl.so  libEGL.so  libGLESv1_CM.so  libGLESv2.so  libGLESv3.so  libjnigraphics.so  liblog.so  libm.a  libm_hard.a  libm.so  libOpenMAXAL.so  libOpenSLES.so  libstdc++.a  libstdc++.so  libthread_db.so  libz.so  rs
```
However the include is not:
```
ls /home/andre/.buildozer/android/platform/android-ndk-r9c/sysroot/usr/include
ls: cannot access '/home/andre/.buildozer/android/platform/android-ndk-r9c/sysroot/usr/include': No such file or directory
```
But if I skip line 33 and set the include as `ndk_include_dir = (join(self.ctx.ndk_dir, 'sysroot', 'usr', 'include')` then it contains the includes:
```
ls /home/andre/.buildozer/android/platform/android-ndk-r9c/platforms/android-19/arch-arm/usr/include/
alloca.h  asm          byteswap.h  dlfcn.h  endian.h  fcntl.h     fnmatch.h  GLES   grp.h       KHR        limits.h  locale.h  math.h    mtd      netinet     OMXAL       poll.h     regex.h   sched.h      sgtty.h   SLES      stdlib.h   sys       termios.h    time.h    utime.h  wctype.h
android   asm-generic  ctype.h     EGL      err.h     features.h  fts.h      GLES2  inttypes.h  lastlog.h  link.h    machine   memory.h  net      netpacket   pathconf.h  pthread.h  resolv.h  semaphore.h  sha1.h    stdint.h  string.h   syslog.h  thread_db.h  unistd.h  utmp.h   zconf.h
arpa      assert.h     dirent.h    elf.h    errno.h   fenv.h      getopt.h   GLES3  jni.h       libgen.h   linux     malloc.h  mntent.h  netdb.h  nsswitch.h  paths.h     pwd.h      rs        setjmp.h     signal.h  stdio.h   strings.h  termio.h  time64.h     util.h    wchar.h  zlib.h
```

inclement:
@AndreMiras Good catch about the case sensitivity. I think this recipe should definitely use lower case, both for consistency with the other recipes and because if that's what the pypi entry uses then we should probably match anyway.

Maybe p4a should always ignore case for recipes, that would probably be a good idea (although usually not that important).

JonasT:
The official PyPI package appears to be named `Pillow`, not `pillow`: https://pypi.org/project/Pillow/

However, it seems `pip` allows both variants, so if the recipes don't work with that, maybe whatever code finds the recipes needs to be fixed?

inclement:
In that case it seems clear that p4a should not be case sensitive when looking for recipes, but that's a separate issue. I'd encourage this recipe to use lower case for consistency with p4a in general, but it isn't a blocking issue as far as I'm concerned.

AndreMiras:
Yes that's probably something we should fix in p4a to ignore the case and then have all recipe to be lowercase in the file system. But that's something we could probably tackle later.

So far I was able to compile Pillow for a Python2 project after fixing the include path. However garden.zbarcam doesn't seem to catch qrcodes anymore. That might be for some unrelated reasons I'll dig later.
I'm now taking a look at Crystax Python3, but I got some runtime crashes, I'll dig a bit further.

AndreMiras:
I'm facing xcamera issues to play with garden.zbarcam, but the recipe looks good overall @krinnewitz !
There's probably just the Python2 include path thing to look at, but it don't mind if this is done in a dedicated pull request if other people also get the issue on Python2. Who's using Python2 now anyway :wink: 

@JonasT actually the idea is just to ignore the case, not to decide whether it should be lower or upper. The thing is currently if you put `pillow` in the requirements, we will have run time errors because it doesn't match with the current recipe. If we were to just change the case the error would be the same but in the opposite fashion (e.g. if people used `Pillow` in requirements). By enforcing everything lowercase and then ignoring the case completely it's a way to document it as a feature: "we ignore the case". But that again can be decided later and addressed in a dedicated PR.

AndreMiras:
@inclement the PEP8 comment you made https://github.com/kivy/python-for-android/pull/1339#pullrequestreview-149533875 was addressed, can you approve and merge? The merging is currently being blocked with that comment.

AndreMiras:
Thanks for the merge @inclement and thanks for your work @krinnewitz 

@JonasT yes please give it a try https://github.com/kivy/python-for-android/pull/1339#issuecomment-416049154
> Once you want to have me test this with my python3crystax app, drop me a note. I'd be happy to give it a try

JonasT:
Yea, I was just suggesting, your ignoring case thing seems artificial. `--requirements` is exactly named like the pip option, so people will most likely paste the package name from the official docs I think. I don't think it's a good idea, but of course it's still a minor nitpick

JonasT:
Also, amazing work! :tada: :tada: :balloon: :+1: :heart: I'll go test

AndreMiras:
> Yea, I was just suggesting, your ignoring case thing seems artificial. --requirements is exactly named like the pip option, so people will most likely paste the package name from the official docs I think. I don't think it's a good idea, but of course it's still a minor nitpick

But what if people put `pillow` in their requirements, just like I did before? `pillow` all lowercase it's still valid and would be pip installed and then it would crash with runtime issues, resulting with people spamming the bug tracker with "pillow doesn't work" kinda issues. Maybe you have a more elegant way in mind to avoid that?

JonasT:
> But what if people put pillow in their requirements, just like I did before?

Well, what if they have `Pillow`? I was just suggesting if people go by the official docs like me, they're gonna have `Pillow`, so the number of people may be larger. Maybe not, it was just a note to consider

AndreMiras:
That's exactly why we suggest to totally ignore the case in p4a code base so one case or the other would never fail :smile:

JonasT:
Oh, right. I was just talking about the transition up to then. Sorry for the misunderstanding!

JonasT:
I just tested the recipe, runs fine for me! Builds and all images show up! :+1:

krinnewitz:
Thank you all for reviewing, merging, and testing so fast!

PR #1347, opened 2018-08-30
 Fix Outdated SDLActivity 

krinnewitz:
While updating the SDL2 recipe to version 2.0.4 in PR #713, the
SDLActivity was not updated accordingly.

For example, SDL 2.0.4 expects the Activity to have a method called
"openAPKExpansionInputStream" while this method was called
"openAPKExtensionInputStream" in older versions. SDL2 is not able to
read files when the method is missing.

This can result in multiple problems, e.g. segfaults. For example,
without this fix, you get NULL when loading a font using SDL2_ttf,
because SDL_RWFromFile needs openAPKExpansionInputStream.

Also, text input is partly broken as it is not possible to enter a
"space" character.

This PR replaces the SDLActivity with the one from SDL2-2.0.4 preserving the
adjustments that seem to be introduced by the python-for-android developers.

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/libsdl/app/SDLActivity.java
+++ b/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/libsdl/app/SDLActivity.java
@@ -11,6 +11,7 @@ import java.lang.reflect.Method;
 
 import android.app.*;
 import android.content.*;
+import android.text.InputType;
 import android.view.*;
 import android.view.inputmethod.BaseInputConnection;
 import android.view.inputmethod.EditorInfo;
@@ -27,6 +28,7 @@ import android.graphics.*;
 import android.graphics.drawable.Drawable;
 import android.media.*;
 import android.hardware.*;
+import android.content.pm.ActivityInfo;
 
 /**
     SDL Activity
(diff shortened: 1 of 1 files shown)


JonasT:
I am seeing spurious segfaults on app resume. I wonder if those are related? It would be interesting to see this merged and see if it makes those go away

pythonforandroid/bootstraps/sdl2/build/src/main/java/org/libsdl/app/SDLActivity.java
```diff
@@ -1063,7 +1111,43 @@ public void surfaceChanged(SurfaceHolder holder,
         mWidth = width;
         mHeight = height;
         SDLActivity.onNativeResize(width, height, sdlFormat, mDisplay.getRefreshRate());
-        Log.v("SDL", "Window size:" + width + "x"+height);
+        Log.v("SDL", "Window size: " + width + "x" + height);
```
JonasT: You used Log.v(TAG, ...) in most places now, but not here. Is that an oversight maybe?

krinnewitz: What I did in this PR is just replacing the SDLActivity with the one that comes with SDL 2.0.4 and then merging some changes by the python-for-android team back in. 

Here, the SDL guys dediced to not use the `TAG` constant and to just keep the raw `"SDL"` string. Don't ask me why :P

JonasT: Haha okay, that explains it. Well it's no biggie, it's not like the output prefix is vital. Just something that stumped me

JonasT:
This should be superseded by #1528 (which updates both SDL2 and the SDLActivity + related .java classes to the latest 2.0.9 version)

JonasT:
@krinnewitz is it ok if I close this for now? I'm pretty sure the more comprehensive SDL 2.0.9 version bump covers this and gets us to an even more recent `SDLActivity.java` that matches the native library version

krinnewitz:
Yes, perfectly fine. Having SDL 2.0.9 is even better!

PR #1351, opened 2018-09-04
Fix segfault on hostpython2 when compiling for x86_64

### LINKED ISSUE #1297: Python2 Build fails with make: *** [Makefile:426: sharedmods] Error 139
# Python version: 3.6
# OS: Arch Linux 
# python-for-android version: 0.6.0 

The command I use to build is:

`  
    p4a  apk --private ~/Projects/Python/Mobile_Apps/BeerApp/ --package=org.drink.recommendations --name "Drink Recommendations" --version 0.2 --bootstrap=sdl2 --requirements=python2,kivy --ndk_version r9c
`

The error is:

`
    make: *** [Makefile:426: sharedmods] Error 139
`

The build logs are in the following file.
[p4a_errors.txt](https://github.com/kivy/python-for-android/files/2091833/p4a_errors.txt)

Initally I thought that this was a buildozer issue, as I attempted it that way first. So, I opened an issue on their github page and multiple users pointed out that they too were experiencing this issue. I've tried with both python3 and python2, the out come is the same. There is absolutely no unicode in any of my source files, I've also attempted the build with pygame instead of sdl2 (for python 2). There are also multiple simillar SO threads open about this. 

Does anyone have any sugesstions or ideas as to why this is happening and how to go about fixing it?

It's also worth noting that if I use the kivy buildozer vm, I can use buildozer to carry out a successful build. Just not on any other machine using either buildozer or p4a, using the same source and build commands.

The buildozer issue is here: https://github.com/kivy/buildozer/issues/673

The output from the dump file is:

`
    Reading symbols from /home/suroh/.local/share/python-for-android/build/other_builds/hostpython2/desktop/hostpython2/python...done.
    [New LWP 28854]
    [Thread debugging using libthread_db enabled]
    Using host libthread_db library "/usr/lib/libthread_db.so.1".
    Core was generated by ./python -E ./setup.py -q build.
    Program terminated with signal SIGSEGV, Segmentation fault.
    #0  0x000055731803eb2a in PyInstance_NewRaw (klass=klass@entry=0x7f7cbf1d1c18, dict=0x557319325210, dict@entry=0x0) at Objects/classobject.c:534
    534         inst->in_dict = dict;
    File "/home/suroh/.local/share/python-for-android/build/other_builds/hostpython2/desktop/hostpython2/python-gdb.py", line 55
        Py_TPFLAGS_HEAPTYPE = (1L << 9)
                                ^
    SyntaxError: invalid syntax
`

tito:
This is a fixes for #1297 - on some platform, Python < 2.7.15 crashes due to a bug highlighted only with recent compilers.

This one liner came from https://github.com/python/cpython/commit/e348c8d154cf6342c79d627ebfe89dfe9de23817#diff-fb41bdaf12f733cf6ab8a82677d03adc that contains a extensive explaination about it. It could be done on Python 2 for ARM, but the bug isn't triggered at the moment.

I compiled a fresh app with this patch, it doesn't segfault during the compilation of hostpython2, and the app on Android runs.

(I talked about moving to Python 2.7.15, but it would take more time that just accepting this patch, as we would need to update hostpython2 and python2 recipes, and all cross compilation patches. Plus more testing as ssl module changed somewhere, unsure if we don't need more patching to make it work.)

Closes #1297

### DIFF
--- a/pythonforandroid/recipes/hostpython2/__init__.py
+++ b/pythonforandroid/recipes/hostpython2/__init__.py
@@ -10,6 +10,7 @@ class Hostpython2Recipe(Recipe):
     version = '2.7.2'
     url = 'https://python.org/ftp/python/{version}/Python-{version}.tar.bz2'
     name = 'hostpython2'
+    patches = ['fix-segfault-pygchead.patch']
 
     conflicts = ['hostpython3']
 
--- a/pythonforandroid/recipes/hostpython2/fix-segfault-pygchead.patch
+++ b/pythonforandroid/recipes/hostpython2/fix-segfault-pygchead.patch
@@ -1,0 +1,12 @@
+diff -Naur Python-2.7.2.orig/Include/objimpl.h Python-2.7.2/Include/objimpl.h
+--- Python-2.7.2.orig/Include/objimpl.h	2011-06-11 17:46:23.000000000 +0200
++++ Python-2.7.2/Include/objimpl.h	2018-09-04 17:33:09.254654565 +0200
+@@ -255,7 +255,7 @@
+         union _gc_head *gc_prev;
+         Py_ssize_t gc_refs;
+     } gc;
+-    long double dummy;  /* force worst-case alignment */
++    double dummy;  /* force worst-case alignment */
+ } PyGC_Head;
+ 
+ extern PyGC_Head *_PyGC_generation0;


AndreMiras:
I didn't test it on Fedora/Arch VM, but it looks good to me, thanks :)

tito:
Thanks you @andremiras

PR #1355, opened 2018-09-06
 Fix apk argument parsing oversight

wolfgangp:
Due to iteration until unknown_args[:-1] the very last argument could be left without its path expanded (if it looked like "--add-jar=some.jar", i.e. contained an equals sign)
This new implementation will fail when something like "--icon=" is the last element of args.unknown_args.
If "--icon=some/path.png" is passed, it's fine.
If "--icon", "some/path.png" is passed, it's fine, too.


Zen-CODE:
This looks good, but would it be possible to rebase so we can merge this as one commit? Otherwise all the commits will add a lot of  "noise" to the merge...

Thanks

KeyWeeUsr:
@Zen-CODE https://blog.github.com/2016-04-01-squash-your-commits/ :)

AndreMiras:
Sorry we forgot about merging that one, thanks for the fix

PR #1356, opened 2018-09-06
Update Docker base image Ubuntu Bionic 18.04

AndreMiras:
Stay up to date with latest Ubuntu LTS.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@
 # TODO:
 #	- delete archives to keep small the container small
 #	- setup caching (for apt, pip, ndk, sdk and p4a recipes downloads)
-FROM ubuntu:16.04
+FROM ubuntu:18.04
 
 
 # get the latest version from https://developer.android.com/ndk/downloads/index.html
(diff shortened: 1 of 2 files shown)


pythonforandroid/recipes/python3crystax/__init__.py
```diff
@@ -12,7 +12,7 @@
 
 
 class Python3Recipe(TargetPythonRecipe):
-    version = '3.5'
+    version = '3.6'
```
AndreMiras: I've bumped the default to 3.6 because Python 3.5 is getting older and 3.6 is the default Python3 version of various distributions including Ubuntu 18.04 LTS.
It's still possible to force specific version with `python3crystax==3.6` if needed.

Dockerfile
```diff
@@ -11,7 +11,7 @@
 # TODO:
 #	- delete archives to keep small the container small
 #	- setup caching (for apt, pip, ndk, sdk and p4a recipes downloads)
-FROM ubuntu:16.04
+FROM ubuntu:18.04
```
AndreMiras: I think we should be up to date, mainstream, but not too cutting edge, hence the choice for the last Ubuntu LTS.

PR #1358, opened 2018-09-06
Fixes example service name example consistency

AndreMiras:
It makes it easier to understand by using the initial example with
`myservice` as service name.

### DIFF
--- a/doc/source/services.rst
+++ b/doc/source/services.rst
@@ -70,9 +70,9 @@ python-for-android creates for each one, as follows::
 
 Here, ``your.package.name`` refers to the package identifier of your
 APK as set by the ``--package`` argument to python-for-android, and
-the name of the service is ``ServiceYourservicename``, in which
-``Yourservicename`` is the identifier passed to the ``--service``
-argument with the first letter upper case. You must also pass the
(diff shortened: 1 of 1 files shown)


AndreMiras:
Thanks for the review. Actually the option is `--service=myservice:/path/to/myservice.py` as stated in the current version of the document so name in the param is `myservice`. Then that gets "turned to" a dynamic class that way (pseudo Python code):
```python
service = 'myservice' # comes from the `--service` param
f'Service{service.title()}' # then the dynamic class is the following -> ServiceMyservice
```
Not sure if it's clear?

JonasT:
Ah. I think I was confused because I misread the text, I kind of thought I should specify `--service` again with the first letter uppercase. Maybe a minor change like this would work better:

*Previously:*
> the name of the service is ServiceMyservice, in which Myservice is the identifier passed to the --service argument with the first letter upper case.

*Suggestion:*
> the name of the service is ServiceMyservice, in which Myservice is the identifier that was previously passed to the --service argument, but with the first letter upper case.

But of course that's kind of unrelated to the change in the pull request, which is definitely an improvement in any case

AndreMiras:
Thank you @JonasT I've adopted your suggestion :beers:

Issue #1360, opened 2018-09-06
Fix/clean-up LDSHARED 


AndreMiras:
`LDSHARED` doesn't seem to be set in `PythonRecipe` leading to failing compilation in some scenarios.
The compilation may be failing because p4a would then pick up the system linker rather than the cross-compiler one.
For this reasons some recipes added that change in their `get_recipe_env()` method, like below:
```python
env['LDSHARED'] = env['CC'] + ' -pthread -shared -Wl,-O1 -Wl,-Bsymbolic-functions'
```
As in last master commit 0dc7664 we have at least 17 of them:
```
$ grep -r "env\['LDSHARED'\] = env\['CC'\]" pythonforandroid/recipes/ | wc -l
17
```
 
In his PR JonasT tries to address it via `pythonforandroid/archs.py`, see https://github.com/kivy/python-for-android/pull/1357/files#diff-569e13021e33ced8b54385f55b49cbe6R115
This probably requires investigation to know where is the proper place to put it and then the pull request fixing it should also clean recipes overriding it in their `get_recipe_env()` method.

JonasT:
By the way, similar issue: ```env['CFLAGS'] += ' -I{}/sources/python/{}/include/python/'.format(self.ctx.ndk_dir, self.ctx.python_recipe.version[0:3])```

Your greenlet recipe pull request does that, mine does, I've seen yet another do it - maybe this should be handled centrally as well?

JonasT:
I added a pull request here: https://github.com/kivy/python-for-android/pull/1361 (still WIP, I'm building my app with it right now to see if things blow up)

AndreMiras:
Even though https://github.com/kivy/python-for-android/pull/1361 is now merged, we still need to clean existing recipes to remove redundant `env['LDSHARED']` override before closing the issue.
I think it would be great to also verify everything is going OK at unit test level e.g. checking all recipes `get_recipe_env()` behaviour.

PR #1361, opened 2018-09-06
Fix LDSHARED and missing CFLAGS include path for python3crystax sources

JonasT:
As proposed in #1360 this change fixes `env["LDSHARED"]` not being set which can confuse distutils / `python setup.py build_ext` to use the incorrect host compiler, and it adds the `-I` folder to the python3 crystax sources when using python 3 crystax. This should allow a lot of recipes to be cleaned up later which currently have the same repeated hacks to work around this. (Uncleaned recipes should just duplicate include paths which should do no harm, so this shouldn't require any recipes to be changed to be safely merged)

**Testing state**:
- tested with python3crystax *without kivy* (but with `Pillow` recipe)
- **not tested** with python 2
- **not tested** with other recipes

### DIFF
--- a/pythonforandroid/archs.py
+++ b/pythonforandroid/archs.py
@@ -112,8 +112,12 @@ class Arch(object):
         env['AR'] = '{}-ar'.format(command_prefix)
         env['RANLIB'] = '{}-ranlib'.format(command_prefix)
         env['LD'] = '{}-ld'.format(command_prefix)
-        # env['LDSHARED'] = join(self.ctx.root_dir, 'tools', 'liblink')
-        # env['LDSHARED'] = env['LD']
+        env['LDSHARED'] = env["CC"] + " -pthread -shared " +\
+            "-Wl,-O1 -Wl,-Bsymbolic-functions "
+        if self.ctx.python_recipe and self.ctx.python_recipe.from_crystax:
+            # For crystax python, we can't use the host python headers:
+            env["CFLAGS"] += ' -I{}/sources/python/{}/include/python/'.\
+                format(self.ctx.ndk_dir, self.ctx.python_recipe.version[0:3])
         env['STRIP'] = '{}-strip --strip-unneeded'.format(command_prefix)
         env['MAKE'] = 'make -j5'
         env['READELF'] = '{}-readelf'.format(command_prefix)


AndreMiras:
Thanks, you're a machine! :joy: 
I'd like to test this change myself, hence killing this flag from the 17 recipes before and after your change and see if the compilation first fails and then passes.
Of course if we had a 100% integration coverage that would be way easier, but as discussed we don't have that yet for various reasons :confused:

JonasT:
@AndreMiras I really need python-for-android for my app, so it is worth it for me to sink some time into getting things fixed. :+1: 

I just tested this with my actual real world app which is **python3crystax** and uses the recipes **Pillow** and **pysdl2**, and some other pure python dependencies. For this app, this pull request builds & runs fine.

AndreMiras:
OK good stuff. However it seems to break the Python2 builds :confused: yes I know it's a nightmare.
3 builds are failing with the same error https://travis-ci.org/kivy/python-for-android/builds/425450653
```
/usr/bin/ccache arm-linux-androideabi-gcc -DANDROID -mandroid -fomit-frame-pointer -D__ANDROID_API__=19 -isystem /opt/android/android-ndk-r16b/sysroot/usr/include/arm-linux-androideabi -isysroot /opt/android/android-ndk-r16b/sysroot -I/root/.local/share/python-for-android/build/python-installs/bdisttest_python2/include/python2.7 --sysroot /opt/android/android-ndk-r16b/platforms/android-19/arch-arm -lm -L/root/.local/share/python-for-android/build/libs_collections/bdisttest_python2/armeabi -Xlinker -export-dynamic -o python \
		Modules/python.o \
		-L. -lpython2.7 -ldl  -L/root/.local/share/python-for-android/build/other_builds/python2/armeabi/python2/python-install/lib -lz    -lm  
Creating directory /root/.local/share/python-for-android/build/other_builds/python2/armeabi/python2/python-install/bin
/bin/sh: 2: --sysroot: not found
Makefile:429: recipe for target 'sharedmods' failed
make: *** [sharedmods] Error 127
make: *** Waiting for unfinished jobs....
/usr/bin/install -c python /root/.local/share/python-for-android/build/other_builds/python2/armeabi/python2/python-install/bin/python2.7
if test -f libpython2.7.so; then \
	if test -n "" ; then \
		/usr/bin/install -c -m 555  /root/.local/share/python-for-android/build/other_builds/python2/armeabi/python2/python-install/bin; \
	else \
		/usr/bin/install -c -m 555 libpython2.7.so /root/.local/share/python-for-android/build/other_builds/python2/armeabi/python2/python-install/lib/libpython2.7.so; \
		if test libpython2.7.so != libpython2.7.so; then \
			(cd /root/.local/share/python-for-android/build/other_builds/python2/armeabi/python2/python-install/lib; ln -sf libpython2.7.so libpython2.7.so) \
		fi \
	fi; \
else	true; \
fi


  STDERR:
```
You can view the complete error clicking on the "Raw logs" button, e.g. https://api.travis-ci.org/v3/job/425450654/log.txt

Yes sometimes simple change break unsuspected things. I also don't see yet the relation between your changes and the way the build broke, but that's quite usual in p4a :stuck_out_tongue:
The way I usually debug this is by comparing the `ENV` before and after my changes, trying to understand what would break.

JonasT:
Yeah, interesting, it definitely didn't fail for me for some reason. Anyway, I fixed the trailing comma now. Let's see if that improves things

JonasT:
Is there something you have in mind I could test? Or how do we go about making a test branch with the recipes adapted accordingly so we can find out if this works for them?

Issue #1341, opened 2018-08-26
python2: jpeg recipe broken due to missing libcutils


krinnewitz:
When building an APK using `python2` and `jpeg`, the following error occurs:

```
  STDERR:


Traceback (most recent call last):                                                                                                                                                                                 
  File "/usr/local/bin/p4a", line 11, in <module>
    load_entry_point('python-for-android==0.6.0', 'console_scripts', 'p4a')()
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/toolchain.py", line 973, in main
    ToolchainCL()
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/toolchain.py", line 512, in __init__
    getattr(self, args.subparser_name.replace('-', '_'))(args)
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/toolchain.py", line 147, in wrapper_func
    build_dist_from_args(ctx, dist, args)
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/toolchain.py", line 191, in build_dist_from_args
    build_recipes(build_order, python_modules, ctx)
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/build.py", line 573, in build_recipes
    recipe.build_arch(arch)
  File "/p4a/pythonforandroid/recipes/jpeg/__init__.py", line 29, in build_arch
    super(JpegRecipe, self).build_arch(arch)
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/recipe.py", line 685, in build_arch
    shprint(sh.ndk_build, 'V=1', 'APP_ABI=' + arch.arch, *extra_args, _env=env)
  File "/usr/local/lib/python2.7/dist-packages/pythonforandroid/logger.py", line 176, in shprint
    for line in output:
  File "/usr/local/lib/python2.7/dist-packages/sh.py", line 863, in next
    self.wait()
  File "/usr/local/lib/python2.7/dist-packages/sh.py", line 792, in wait
    self.handle_command_exit_code(exit_code)
  File "/usr/local/lib/python2.7/dist-packages/sh.py", line 815, in handle_command_exit_code
    raise exc
sh.ErrorReturnCode_2: 

  RAN: /android/android-ndk-r16b/ndk-build V=1 APP_ABI=armeabi

  STDOUT:
Android NDK: APP_PLATFORM not set. Defaulting to minimum supported version android-14.    
[...]

PR #1363, opened 2018-09-07
Fix jpeg build for newer NDKs

mkg20001:
Otherwise it throws with an error message about cjpeg depending on undefined dependency cutils when using newer ndk (r15)


krinnewitz:
This resolves #1341.

PR #1370, opened 2018-09-12
Adds issue template with basic required info

AndreMiras:
The same way we did on buildozer project (https://github.com/kivy/buildozer/pull/688).
See what it looks like on my fork:
https://github.com/AndreMiras/python-for-android/issues/new


AndreMiras:
Agree, thank you for the feedback @KeyWeeUsr I've added another bullet point to make it more explicit.

PR #1371, opened 2018-09-12
Sets default service pythonName, fixes #1049

### LINKED ISSUE #1049: Cant start service app
_From @Fogapod on April 28, 2017 17:1_

### Versions

* Python: 2.7
* OS: android
* Kivy: master

### Description

Service app raises android error right after starting. No exceptions, no kivy logs. Just do not start. Faced this problem only on android building with buildozer's `android_new` toolchain

### Code

I get code from here: https://github.com/tshirtman/kivy_service_osc
(changed [line 7 and 8](https://github.com/tshirtman/kivy_service_osc/blob/master/main.py#L7) to `from kivy import platform`)

### Logs
[Part of logcat where app fails to start service](https://gist.github.com/Fogapod/bf83faad62961a7e43a4b3bf2b70ba71)

Edit: removed apk.

_Copied from original issue: kivy/kivy#5140_

AndreMiras:
In https://github.com/kivy/python-for-android/pull/609 `PYTHON_NAME` was
being used while potentially null (in the case of basic service folder
method), leading to segfault on `setenv()` call.


AndreMiras:
While I've also added the fixes for `webview` and `pygame` bootstraps, I've only tested it in `sdl2`.

PR #1374, opened 2018-09-15
Fixes and clarifies services auto-restart

AndreMiras:
1) Fixes services auto-restart introduced in #643:
`onTaskRemoved()` stops the service gracefully using `stopSelf()` when
it gets killed. That way `onDestroy()` gets triggered and service will
restart under different process if auto-restart is enabled.

2) Documents process behavior and auto-restart usage:
Emphasis that services are running in a dedicated processes, different
from the app process, which is not the Android default behavior.
Adds a code snippet.

### DIFF
--- a/doc/source/services.rst
+++ b/doc/source/services.rst
@@ -8,9 +8,14 @@ possible to use normal multiprocessing on Android. Services are also
 the only way to run code when your app is not currently opened by the user.
 
 Services must be declared when building your APK. Each one
-will have its own main.py file with the Python script to be run. You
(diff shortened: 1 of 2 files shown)


AndreMiras:
Build is green :heavy_check_mark: 
This pull request doesn't address the sticky flag. At first I also wanted to document this feature as well, but since [it still seems broken to me](https://groups.google.com/forum/#!topic/kivy-users/1-zIEkI0snY), I'd like to make sure it works first.

inclement:
Seems a clear improvement, thanks :)

PR #1377, opened 2018-09-19
Adds Python3 support to reportlab, pyrxp and preppy

AndreMiras:
Follow up for https://github.com/kivy/python-for-android/pull/1182
Squashed commit and also removed `pil` from the requirements list of the initial PR.
Edit:
Not tested, but well relatively safe change. In the worst case we have compilation error on Python3, but no regressions on Python2.


inclement:
Thanks

PR #1383, opened 2018-09-26
ReportLabRecipe fixes and unit tests, fixes #1380

### LINKED ISSUE #1380: Unit test recipes (reportlab to begin with)
The test suite is currently running full integration tests for a bunch of recipes.
While integration tests are good, we cannot currently use them for all recipes because they run for too long.
However having unit tests for all recipes should be feasible and may still cover some issues like https://github.com/kivy/python-for-android/pull/1357#issuecomment-423614116.
Unit tests were recently enabled the following pull request https://github.com/kivy/python-for-android/pull/1379. So the idea is to increase the coverage start from reportlab recipe as a use case.

AndreMiras:
- fixes `TypeError` on host Python3
- adds recipe unit test
- makes it possible to override `pytest` args via `tox`

### DIFF
--- a/pythonforandroid/recipes/reportlab/__init__.py
+++ b/pythonforandroid/recipes/reportlab/__init__.py
@@ -26,16 +26,15 @@ class ReportLabRecipe(CompiledComponentsPythonRecipe):
             info('reportlab recipe: ft_lib_dir={}'.format(ft_lib_dir))
             info('reportlab recipe: ft_inc_dir={}'.format(ft_inc_dir))
             with current_directory(recipe_dir):
-                sh.ls('-lathr')
                 ensure_dir(tmp_dir)
                 pfbfile = os.path.join(tmp_dir, "pfbfer-20070710.zip")
                 if not os.path.isfile(pfbfile):
                     sh.wget("http://www.reportlab.com/ftp/pfbfer-20070710.zip", "-O", pfbfile)
                 sh.unzip("-u", "-d", os.path.join(recipe_dir, "src", "reportlab", "fonts"), pfbfile)
                 if os.path.isfile("setup.py"):
-                    with open('setup.py', 'rb') as f:
+                    with open('setup.py', 'r') as f:
                         text = f.read().replace('_FT_LIB_', ft_lib_dir).replace('_FT_INC_', ft_inc_dir)
-                    with open('setup.py', 'wb') as f:
+                    with open('setup.py', 'w') as f:
                         f.write(text)
 
 
--- a/tests/recipes/test_reportlab.py
+++ b/tests/recipes/test_reportlab.py
@@ -1,0 +1,55 @@
+import os
+import tempfile
+import unittest
+from mock import patch
+from pythonforandroid.archs import ArchARMv7_a
+from pythonforandroid.build import Context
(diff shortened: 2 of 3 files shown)


pythonforandroid/recipes/reportlab/__init__.py
```diff
@@ -26,16 +26,15 @@ def prebuild_arch(self, arch):
             info('reportlab recipe: ft_lib_dir={}'.format(ft_lib_dir))
             info('reportlab recipe: ft_inc_dir={}'.format(ft_inc_dir))
             with current_directory(recipe_dir):
-                sh.ls('-lathr')
```
AndreMiras: Removed that "debug" line because I didn't feel like it provided any value.

tests/recipes/test_reportlab.py
```diff
@@ -0,0 +1,54 @@
...
+from pythonforandroid.archs import ArchARMv7_a
+from pythonforandroid.build import Context
+from pythonforandroid.graph import get_recipe_order_and_bootstrap
+from pythonforandroid.recipe import Recipe
+from pythonforandroid.util import ensure_dir
+
+
+class TestReportLabRecipe(unittest.TestCase):
+
+    def setUp(self):
```
AndreMiras: For now the `setUp()` looks a bit heavy, but as recipe unit tests grow we will easily move common setup things to common class for other recipe tests to use.
So the idea is not to have it perfect for the first iteration :smile:

pythonforandroid/recipes/reportlab/__init__.py
```diff
@@ -26,16 +26,15 @@ def prebuild_arch(self, arch):
...
                 ensure_dir(tmp_dir)
                 pfbfile = os.path.join(tmp_dir, "pfbfer-20070710.zip")
                 if not os.path.isfile(pfbfile):
                     sh.wget("http://www.reportlab.com/ftp/pfbfer-20070710.zip", "-O", pfbfile)
                 sh.unzip("-u", "-d", os.path.join(recipe_dir, "src", "reportlab", "fonts"), pfbfile)
                 if os.path.isfile("setup.py"):
-                    with open('setup.py', 'rb') as f:
```
AndreMiras: This is what was causing the error on host Python3. The error was:
```
TypeError: a bytes-like object is required, not 'str'
```

tests/recipes/test_reportlab.py
```diff
@@ -0,0 +1,54 @@
...
+        self.arch = ArchARMv7_a(self.context)
+        self.recipe = Recipe.get_recipe('reportlab', self.context)
+        self.recipe.ctx = self.context
+        self.bootstrap = None
+        recipe_build_order, python_modules, bootstrap = \
+            get_recipe_order_and_bootstrap(
+                self.context, [self.recipe.name], self.bootstrap)
+        self.context.recipe_build_order = recipe_build_order
+        self.context.python_modules = python_modules
+        self.context.setup_dirs('/tmp/')
```
tshirtman: Not sure how portable we need to be here, but should probably use `tempfile.gettempdir()` instead. or even tempfile.TemporaryDirectory if we want/need to be isolated from other runs.

AndreMiras: Addressed, thanks!

PR #1386, opened 2018-09-30
Linter fixes E114, E127, E128, E129

AndreMiras:
- E114 indentation is not a multiple of four (comment)
- E127 continuation line over-indented for visual indent
- E128 continuation line under-indented for visual indent
- E129 visually indented line with same indent as next logical line

### DIFF
--- a/pythonforandroid/recipes/vlc/__init__.py
+++ b/pythonforandroid/recipes/vlc/__init__.py
@@ -33,15 +33,15 @@ class VlcRecipe(Recipe):
         else:
             aar_path = join(port_dir, 'libvlc', 'build', 'outputs', 'aar')
             self.aars[arch] = aar = join(aar_path, 'libvlc-{}.aar'.format(self.version))
-            warning("HINT: set path to precompiled libvlc-<ver>.aar bundle " \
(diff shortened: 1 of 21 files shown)


pythonforandroid/recipes/vlc/__init__.py
```diff
@@ -34,14 +34,14 @@ def prebuild_arch(self, arch):
             aar_path = join(port_dir, 'libvlc', 'build', 'outputs', 'aar')
             self.aars[arch] = aar = join(aar_path, 'libvlc-{}.aar'.format(self.version))
             warning("HINT: set path to precompiled libvlc-<ver>.aar bundle " \
-                        "in {} environment!".format(self.ENV_LIBVLC_AAR))
+                    "in {} environment!".format(self.ENV_LIBVLC_AAR))
```
tshirtman: \ doesn't seem necessary here, as things are already in parentheses, but they are not harmful either.

AndreMiras: Good catch, I've addressed. I think there's a PEP for that one also by the way. Hopefully we can address all the other later

Issue #1382, opened 2018-09-25
Investigate conditional builds


AndreMiras:
# High level description
Discussed earlier today (at 13:30 UTC) with @tshirtman on discord `#dev`.
We would like to investigate how to both increase continuous integration coverage and speed by introducing conditional builds.
Basically the idea would be to build only the recipes updated by a pull request, plus a subset of critical-core recipes.

We could take advantage of both [Travis conditional builds](https://docs.travis-ci.com/user/conditional-builds-stages-jobs/), [dependency resolver](https://github.com/kivy/python-for-android/blob/b1f7033/pythonforandroid/graph.py) and `git diff --stat master` magic to only build the relevant subset of recipes.

This ticket will be used as a brainstorm/spec before implementing the feature.

### Conflicting recipes
The result of the stat command may be a list of recipe that conflict with each other.
A simple, but not so efficient way to overcome the issue would be to always build them separately rather than in batch.
So when batch is:
```bash
requirements="python3crystax,setuptools,android"
python setup_testapp.py apk --requirements $requirements
```
Separately would be:
```bash
for requirement in $(echo $requirements | tr "," " ") do
    python setup_testapp.py apk --requirements $requirement
done
```

### Don't build twice
Also we should try not building exact thing twice, e.g. we may want to always build some core critical recipes e.g. kivy, python2, etc, but if theses recipes are also updated in the pull request, they will also be built as part of the conditional build. Hence we want to make sure there's no overlap for not wasting time.

### Keep it simple and uncoupled
We don't want to have an over-complicated and unmaintainable Travis file. So things should be thought as simple as possible as as reusable as possible.
Sor for example it might be needed to have a dedicated util Bash or Python script for handling some the logic so that the Travis file just has to call it.

### Keep building in parallel
Builds currently run in parallel in Travis, currently the maximum jobs that can run in parallel is 5.
If possible we need to make the most of it.

### Address different bootstraps
Just like we want to rebuild recipes when they get updated, we also want to batch the updated boostrap (sdl2, pygame, service_only...).

### git diff --stats
Use `git diff --stat master` to see what has changed between master and current branch.
See example output:
```
git diff --stat master
 .travis.yml                                                   |  3 ++-
 pythonforandroid/bootstraps/pygame/__init__.py                |  2 +-
 pythonforandroid/bootstraps/pygame/build/build.py             |  1 -
 pythonforandroid/bootstraps/pygame/build/buildlib/argparse.py |  2 +-
[...]

Issue #173, opened 2013-11-14
Doubt about distribute.sh


alexsilva:
When trying to run the below command:
`./distribute.sh -m "requests kivy"`

I noticed the below error:
`./distribute.sh: line 93: --python=python2.7: command not found`

So as a good programmer, I sought such error line (93).
I noticed that the variable "$VIRTUALENV" was empty!

I changed the code below (line 93):
`try $VIRTUALENV --python=python2.7 venv`

For this code and it worked as expected:
`try virtualenv --python=python2.7 venv`

The variable "$VIRTUALENV" is empty by default ?

The documentation says nothing about (not having seen):
http://python-for-android.readthedocs.org/en/latest/toolchain/#usage

PR #1390, opened 2018-10-03
Test Numpy on Python 2 / 3 on all architectures

jgavris:
I'm working on getting numpy built for Python 3.6 on `x86_64` and `arm64-v8a` ABIs, I just wanted to add and or check if this passes on CI in the meantime. 

The test matrix could probably be refactored to be a bit cleaner, I think it would be great if every build command were tested against every ABI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,14 @@ env:
   # overrides requirements to skip `peewee` pure python module, see:
   # https://github.com/kivy/python-for-android/issues/1263#issuecomment-390421054
   - COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_python2_sqlite_openssl.py apk --sdk-dir /opt/android/android-sdk --ndk-dir /opt/android/android-ndk --requirements sdl2,pyjnius,kivy,python2,openssl,requests,sqlite3,setuptools'
-  - COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_python2.py apk --sdk-dir /opt/android/android-sdk --ndk-dir /opt/android/android-ndk --bootstrap sdl2 --requirements python2,numpy'
-  - COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_python3.py apk --sdk-dir /opt/android/android-sdk --ndk-dir /opt/android/crystax-ndk'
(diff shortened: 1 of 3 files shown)


AndreMiras:
Nice, thanks for the pull request.
I agree we should have a better coverage, but Jenkins would never be ending for every single PR.
So we need to first tackle this story https://github.com/kivy/python-for-android/issues/1382
Basically the idea is to dynamically build only the part that were touched by code change. That way we could cover more (e.g. different Python host/target version and architecture) the concerned recipes.
So I'm keeping your PR open, but can't merge it before #1382 because it would really cause too much overhead on the CI. Once we've done with it we will review/rework your PR to make it to the tree.
Does it make sense to you?

jgavris:
What are the expensive parts of the docker build? I would be concerned it is difficult to get 'conditional builds' done correctly (without going all-in with a fancy build system like [bazel](https://bazel.build/)). I wonder if there is any low-hanging fruit like rebuilding the docker images with up to date system packages and android SDK / NDK. That looks like it's taking a real long time, versus the actual python-for-android apk build.

AndreMiras:
Yes the `docker build` takes time since it needs to download some heavy component for instance.
But not only, also compiling p4a recipes take time.
You can take a look at the last master build for instance:
https://travis-ci.org/kivy/python-for-android/builds/435279943
And open some of the different jobs to see where the time is spent. For instance in [Job #173.1](https://travis-ci.org/kivy/python-for-android/jobs/435279944). Here `docker build --tag=p4a .` 1000 seconds (16 minutes) are spent and the rest (10 minutes) is spent on the `docker run` that builds the recipes.

So yes SDK/NDK takes ages, but it's not recommended to cache it in Travis. Yes one solution could be to use a Docker image that has SDK/NDK and is optimized, but I also like that we have the control over our Docker image.

Yes the conditional build is not an easy one, but I still imagine it's feasible.

AndreMiras:
Thank you again for your PR! I'm sorry we didn't make it to the tree.
The support for `x86_64` architecture was added recently in https://github.com/kivy/python-for-android/pull/1476.

Conditional builds are in place via:
- https://github.com/kivy/python-for-android/blob/12ac837/ci/rebuild_updated_recipes.py
- https://github.com/kivy/python-for-android/blob/0c936d1/.travis.yml#L27

It's still far from being perfect, so feel free to open a dedicated PR/issue if you want to further discuss the CI improvements. For instance yes pre-built docker image may help. Also the recent drop of CrystaX support in the CI speed up the Docker build.

PR #1392, opened 2018-10-05
jpeg: use https: URL instead of git:

goffi-contrib:
This way download can work when non HTTP ports are blocked.


AndreMiras:
Well done thanks.
Actually it would be even better to add the version string and download the archive directly rather than cloning.

tshirtman:
see tito's answer again, that repos has no such archive feature, but has different code than the github one so that's why this one was used this way.

Issue #1400, opened 2018-10-11
Investigate Azure Pipelines


AndreMiras:
We're currently using Travis for continuous integration.
While things are working we have a very annoying issue, which is Travis not providing Docker layer caching, see https://github.com/travis-ci/travis-ci/issues/5358.
This means every single build will rebuild docker images from scratch ~15 minutes when it should take less than a minute if layer caching was properly implemented Travis side.
Other CI providers seem to provide solutions for it in their free tier, e.g. CircleCI, GitLab Pipelines, Azure Pipelines...
Apparently Azure also seems to provide up to 10 parallel builds when Travis offers 5.
Find out more https://azure.microsoft.com/en-us/services/devops/pipelines/
The goal of this task is to investigate and try at least other competitors to see how they perform compared to Travis in our environment.

AndreMiras:
I already gave a try to Azure Pipeline in a branch of my fork. While the integration was pretty straight forward and parallel jobs are great (https://github.com/AndreMiras/python-for-android/pull/1), I didn't see any caching of previous builds. See build https://dev.azure.com/andremiras/andremiras/_build/results?buildId=14&view=logs
Since the agent seems to get provisioned every time from scratch cached images don't seem to be reusable.
Sources:
- https://github.com/Microsoft/vsts-tasks/issues/6439
- https://medium.com/@marcodesanctis2/a-build-and-release-pipeline-in-vsts-for-docker-and-azure-kubernetes-service-aks-41efc9a0c5c4

misl6:
I guess that we're now fine with Github Actions and we can close this issue?

AndreMiras:
Completely, thanks for the heads up, closing

PR #1403, opened 2018-10-11
Fixes Python3.6 and enum34

AndreMiras:
Module `enum34` should only be installed on Python < 3.4
Fixes host Python3 regression from ecd0bf6, the error was:
```
AttributeError: module 'enum' has no attribute 'IntFlag'
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,8 @@ package_data = {'': ['*.tmpl',
 data_files = []
 
 
-install_reqs = ['appdirs', 'colorama>=0.3.3', 'jinja2', 'six', 'enum34']
+install_reqs = ['appdirs', 'colorama>=0.3.3', 'jinja2', 'six',
+    'enum34;python_version<"3.4"']
 if os.name != 'nt':
     install_reqs.append('sh>=1.10')
 


AndreMiras:
I wish I could verify that in a unit test, but can't seem to reproduce that when running from tox. And it's bit overkill to write an integration test for it because it would require updating our CI to run also on host Python3 :confused: 
While we do want to do that at some point, it would make the suite run for even longer which we can't until we've addressed https://github.com/kivy/python-for-android/issues/1400

setup.py
```diff
@@ -16,7 +16,8 @@
 data_files = []
 
 
-install_reqs = ['appdirs', 'colorama>=0.3.3', 'jinja2', 'six', 'enum34']
+install_reqs = ['appdirs', 'colorama>=0.3.3', 'jinja2', 'six',
+    'enum34;python_version<"3.4"']
```
inclement: I would indent this further, but no need to have that hold up fixing the install.

AndreMiras: True, well the linter (yes we have one :smile: ) didn't complain, but you're right I could have at least tried something like:
```
install_reqs = [
    'appdirs', 'colorama>=0.3.3', 'jinja2', 'six', 'enum34;python_version<"3.4"'
]
```

Issue #1402, opened 2018-10-11
Comprehensive list of broken recipes


AndreMiras:
When working on https://github.com/kivy/python-for-android/pull/1401 I realised we still have some broken recipes in the tree at least for python3crystax.
Even though we don't want to have red builds for things that were already broken, we still want to have a clear status of what's broken and what's not.
Basically the idea is to try to compile every single recipes and add the broken ones in the ignore list (`BROKEN_RECIPES`) from #1401. That way we can track and fix them later on meanwhile keeping a green build.
I would like to address it in this task. Basically the output of the task should be a PR making the `BROKEN_RECIPES` list comprehensive. With bonus points for creating an issue per broken recipes :smile:

PR #1408, opened 2018-10-14
List of broken recipes for python3crystax, refs #1402

AndreMiras:
Populated the list of broken recipes and moved to a dedicated
`constants.py` file.
Broken recipes are mostly divided in 3 categories:
1) recipes not yet migrated from python2 to python3crystax
2) recipes that don't have environment var well setup
3) recipes that are not broken but require a system package

### DIFF
--- a/ci/constants.py
+++ b/ci/constants.py
@@ -1,0 +1,63 @@
+from enum import Enum
+
+
+class TargetPython(Enum):
+    python2 = 0
+    python3crystax = 1
+
+
+# recipes that currently break the build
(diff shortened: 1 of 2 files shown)


AndreMiras:
Next up is is:
1) do the same for python2
2) start fixing the broken ones to slowly decrease the list

For python3crystax the list could decrease pretty fast by simply adding `python3crystax` to the `depends` class attribute for recipes missing it. Then I guess we could pick up the ones that have a dedicated bug report.

PR #1412, opened 2018-10-16
Python 3 recipe

inclement:
I haven't had much time to work on this since my initial proof of concept, but I've got back to it now. This PR is a work in progress, I wouldn't usually make it anywhere near this early but I know people are following this closely so I thought I might as well be clear what is going on with it.

Right now this is the most minimal possible implementation, I literally just got it working and there are many hacks to be resolved. However, it should work, the whole build process is now included although it's only tested on my machine so far there could be issues in other environments (especially macOS?). The main difference to the proof of concept is that I began moving the build into p4a's system rather than the shell/Makefile system used in Python bpo-30386. There's nothing wrong with that system, but since it isn't maintained it seems safer to make full use of p4a's tools instead. 

Non-exhaustive non-ordered list of things to do (note: even if checked off it doesn't mean the implementation is final, just that the bulk of the work is done on that feature):
- [x] move Python bundling code from the bootstrap to the python recipes (makes more sense now that there are several)
- [x] trim/strip the python3 bundle so that APKs are not too big (target: no bigger than python3crystax) [edit: done for now, got down to 11MB including Kivy etc., just by trimming out obvious stuff]
- [x] move to use Python 3.7.0 instead of the bpo-30386 fork
  - [x] and also check compatibility with 3.7.1
- [x] add proper APP_PLATFORM setting functionality, including adding it to the build and dist folder disambiguation
- [ ] ~~make the python3 recipe use p4a's arch system as much as possible, including making p4a understand clang toolchains~~ [edit: won't hold up merging for this, will make a new issue for it]
- [x] get everything working with APIs 21 to 27+, ~~currently tested only with API 24 and there are definitely things that need doing to make everything work with 27+~~
- [ ] ~~check why ctypes isn't building and fix it if it's simple (it should be, I think it works in bpo-30386)~~ [edit: won't hold up merging for this, will make a new issue for it]
- [ ] ~~openssl recipe compatibility~~ [edit: won't hold up merging for this, will make a new issue for it]
- [ ] ~~sqlite recipe compatibility~~ [edit: won't hold up merging for this, will make a new issue for it]**
- [x] clean the code, resolving all TODOs and hacks that are not confined to the new recipe (which can be merged as a WIP that is known to need tidying)
- [x] check that python2 and python3crystax builds still work
  - [x] python2
  - [x] python3crystax

Minor note: as it stands this recipe doesn't support API lower than 21 due to Android limitations. There are some patches that exist to fix that, but I'm probably not going to look into that any time soon, so API 21 may effectively be the lowest supported API for this recipe.

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/__init__.py
+++ b/pythonforandroid/bootstraps/sdl2/__init__.py
@@ -107,53 +107,8 @@ class SDL2GradleBootstrap(Bootstrap):
                     shprint(sh.rm, '-rf', 'config/python.o')
 
             elif self.ctx.python_recipe.name == 'python3':
-                ndk_dir = self.ctx.ndk_dir
-                py_recipe = self.ctx.python_recipe
-
-                ## Build the python bundle:
-
-                # Bundle compiled python modules to a folder
-                modules_dir = join(python_bundle_dir, 'modules')
-                ensure_dir(modules_dir)
-                
-                modules_build_dir = join(
-                    self.ctx.python_recipe.get_build_dir(arch.arch),
-                    'android-build',
-                    'build',
-                    'lib.linux-arm-3.7')
-                module_filens = (glob.glob(join(modules_build_dir, '*.so')) +
-                                 glob.glob(join(modules_build_dir, '*.py')))
-                for filen in module_filens:
-                    shprint(sh.cp, filen, modules_dir)
-
-                # zip up the standard library
-                stdlib_zip = join(self.dist_dir, python_bundle_dir, 'stdlib.zip')
-                with current_directory(
-                        join(self.ctx.python_recipe.get_build_dir(arch.arch),
-                             'Lib')):
-                    shprint(sh.zip, '-r', stdlib_zip, *os.listdir())
-                    
-                # copy the site-packages into place
-                shprint(sh.cp, '-r', self.ctx.get_python_install_dir(),
-                        join(python_bundle_dir, 'site-packages'))
-
-                # copy the python .so files into place
-                python_build_dir = join(py_recipe.get_build_dir(arch.arch),
-                                        'android-build')
-                shprint(sh.cp, join(python_build_dir, 'libpython3.7m.so'),
-                        'libs/{}'.format(arch.arch))
-                shprint(sh.cp, join(python_build_dir, 'libpython3.7m.so.1.0'),
-                        'libs/{}'.format(arch.arch))
-                
-                info('Renaming .so files to reflect cross-compile')
-                site_packages_dir = join(python_bundle_dir, 'site-packages')
-                py_so_files = shprint(sh.find, site_packages_dir, '-iname', '*.so')
-                filens = py_so_files.stdout.decode('utf-8').split('\n')[:-1]
-                for filen in filens:
-                    parts = filen.split('.')
-                    if len(parts) <= 2:
-                        continue
-                    shprint(sh.mv, filen, parts[0] + '.so')
+                self.ctx.python_recipe.create_python_bundle(
+                    join(self.dist_dir, python_bundle_dir), arch)
                 
             elif self.ctx.python_recipe.from_crystax:  # Python *is* loaded from crystax
                 ndk_dir = self.ctx.ndk_dir
--- a/pythonforandroid/bootstraps/sdl2/__init__.py
+++ b/pythonforandroid/bootstraps/sdl2/__init__.py
@@ -61,7 +61,7 @@ class SDL2GradleBootstrap(Bootstrap):
                 python_bundle_dir = 'private'
             ensure_dir(python_bundle_dir)
 
-            self.ctx.python_recipe.create_python_bundle(
+            site_packages_dir = self.ctx.python_recipe.create_python_bundle(
                 join(self.dist_dir, python_bundle_dir), arch)
                 # TODO: Also set site_packages_dir again so fry_eggs can work
 
@@ -70,7 +70,7 @@ class SDL2GradleBootstrap(Bootstrap):
                     fileh.write('\nsqlite3/*\nlib-dynload/_sqlite3.so\n')
 
         self.strip_libraries(arch)
-        # self.fry_eggs(site_packages_dir)  # TODO uncomment this and make it work with python3
+        self.fry_eggs(site_packages_dir)  # TODO uncomment this and make it work with python3
         super(SDL2GradleBootstrap, self).run_distribute()
 
 
--- a/pythonforandroid/bootstraps/sdl2/__init__.py
+++ b/pythonforandroid/bootstraps/sdl2/__init__.py
@@ -4,11 +4,7 @@ from pythonforandroid.util import ensure_dir
 from os.path import join, exists, curdir, abspath
 from os import walk
 import os
-import glob
 import sh
-
-
-EXCLUDE_EXTS = (".py", ".pyc", ".so.o", ".so.a", ".so.libs", ".pyx")
 
 
 class SDL2GradleBootstrap(Bootstrap):
--- a/pythonforandroid/bootstraps/sdl2/__init__.py
+++ b/pythonforandroid/bootstraps/sdl2/__init__.py
@@ -110,30 +110,11 @@ class SDL2GradleBootstrap(Bootstrap):
                 self.ctx.python_recipe.create_python_bundle(
                     join(self.dist_dir, python_bundle_dir), arch)
                 
-            elif self.ctx.python_recipe.from_crystax:  # Python *is* loaded from crystax
-                ndk_dir = self.ctx.ndk_dir
-                py_recipe = self.ctx.python_recipe
-                python_dir = join(ndk_dir, 'sources', 'python',
-                                  py_recipe.version, 'libs', arch.arch)
-                shprint(sh.cp, '-r', join(python_dir,
-                                          'stdlib.zip'), crystax_python_dir)
-                shprint(sh.cp, '-r', join(python_dir,
-                                          'modules'), crystax_python_dir)
-                shprint(sh.cp, '-r', self.ctx.get_python_install_dir(),
-                        join(crystax_python_dir, 'site-packages'))
+            elif self.ctx.python_recipe.from_crystax:
+                self.ctx.python_recipe.create_python_bundle(
+                    join(self.dist_dir, python_bundle_dir), arch)
+                # TODO: Also set site_packages_dir again so fry_eggs can work
 
-                info('Renaming .so files to reflect cross-compile')
-                site_packages_dir = join(crystax_python_dir, "site-packages")
-                find_ret = shprint(
-                    sh.find, site_packages_dir, '-iname', '*.so')
-                filenames = find_ret.stdout.decode('utf-8').split('\n')[:-1]
-                for filename in filenames:
-                    parts = filename.split('.')
-                    if len(parts) <= 2:
-                        continue
-                    shprint(sh.mv, filename, filename.split('.')[0] + '.so')
-                site_packages_dir = join(abspath(curdir),
-                                         site_packages_dir)
             if 'sqlite3' not in self.ctx.recipe_build_order:
                 with open('blacklist.txt', 'a') as fileh:
                     fileh.write('\nsqlite3/*\nlib-dynload/_sqlite3.so\n')
--- a/pythonforandroid/bootstraps/sdl2/__init__.py
+++ b/pythonforandroid/bootstraps/sdl2/__init__.py
@@ -22,9 +22,6 @@ class SDL2GradleBootstrap(Bootstrap):
         arch = self.ctx.archs[0]
         python_install_dir = self.ctx.get_python_install_dir()
         from_crystax = self.ctx.python_recipe.from_crystax
-        crystax_python_dir = join("crystax_python", "crystax_python")
-
-        python_bundle_dir = join('_python_bundle', '_python_bundle')
 
         if len(self.ctx.archs) > 1:
             raise ValueError("SDL2/gradle support only one arch")
@@ -43,13 +40,6 @@ class SDL2GradleBootstrap(Bootstrap):
         with current_directory(self.dist_dir):
             info("Copying Python distribution")
 
-            if 'python2' in self.ctx.recipe_build_order:
-                ensure_dir("private")
-            elif not exists("crystax_python") and from_crystax:
-                ensure_dir(crystax_python_dir)
-            elif 'python3' in self.ctx.recipe_build_order:
-                ensure_dir(python_bundle_dir)
-
             hostpython = sh.Command(self.ctx.hostpython)
             if self.ctx.python_recipe.name == 'python2':
                 try:
@@ -66,53 +56,14 @@ class SDL2GradleBootstrap(Bootstrap):
             self.distribute_javaclasses(self.ctx.javaclass_dir,
                                         dest_dir=join("src", "main", "java"))
 
-            if self.ctx.python_recipe.name == 'python2':
-                info("Filling private directory")
-                if not exists(join("private", "lib")):
-                    info("private/lib does not exist, making")
-                    shprint(sh.cp, "-a",
-                            join("python-install", "lib"), "private")
-                shprint(sh.mkdir, "-p",
-                        join("private", "include", "python2.7"))
+            python_bundle_dir = join('_python_bundle', '_python_bundle')
+            if 'python2' in self.ctx.recipe_build_order:
+                # Python 2 is a special case with its own packaging location
+                python_bundle_dir = 'private'
+            ensure_dir(python_bundle_dir)
 
-                libpymodules_fn = join("libs", arch.arch, "libpymodules.so")
-                if exists(libpymodules_fn):
-                    shprint(sh.mv, libpymodules_fn, 'private/')
-                shprint(sh.cp,
-                        join('python-install', 'include',
-                             'python2.7', 'pyconfig.h'),
-                        join('private', 'include', 'python2.7/'))
-
-                info('Removing some unwanted files')
-                shprint(sh.rm, '-f', join('private', 'lib', 'libpython2.7.so'))
-                shprint(sh.rm, '-rf', join('private', 'lib', 'pkgconfig'))
-
-                libdir = join(self.dist_dir, 'private', 'lib', 'python2.7')
-                site_packages_dir = join(libdir, 'site-packages')
-                with current_directory(libdir):
-                    removes = []
-                    for dirname, root, filenames in walk("."):
-                        for filename in filenames:
-                            for suffix in EXCLUDE_EXTS:
-                                if filename.endswith(suffix):
-                                    removes.append(filename)
-                    shprint(sh.rm, '-f', *removes)
-
-                    info('Deleting some other stuff not used on android')
-                    # To quote the original distribute.sh, 'well...'
-                    shprint(sh.rm, '-rf', 'lib2to3')
-                    shprint(sh.rm, '-rf', 'idlelib')
-                    for filename in glob.glob('config/libpython*.a'):
-                        shprint(sh.rm, '-f', filename)
-                    shprint(sh.rm, '-rf', 'config/python.o')
-
-            elif self.ctx.python_recipe.name == 'python3':
-                self.ctx.python_recipe.create_python_bundle(
-                    join(self.dist_dir, python_bundle_dir), arch)
-                
-            elif self.ctx.python_recipe.from_crystax:
-                self.ctx.python_recipe.create_python_bundle(
-                    join(self.dist_dir, python_bundle_dir), arch)
+            self.ctx.python_recipe.create_python_bundle(
+                join(self.dist_dir, python_bundle_dir), arch)
                 # TODO: Also set site_packages_dir again so fry_eggs can work
 
             if 'sqlite3' not in self.ctx.recipe_build_order:
--- a/pythonforandroid/bootstraps/sdl2/__init__.py
+++ b/pythonforandroid/bootstraps/sdl2/__init__.py
@@ -36,7 +36,6 @@ class SDL2GradleBootstrap(Bootstrap):
             with open('local.properties', 'w') as fileh:
                 fileh.write('sdk.dir={}'.format(self.ctx.sdk_dir))
 
-        # TODO: Move the packaged python building to the python recipes
         with current_directory(self.dist_dir):
             info("Copying Python distribution")
 
--- a/pythonforandroid/bootstraps/sdl2/__init__.py
+++ b/pythonforandroid/bootstraps/sdl2/__init__.py
@@ -119,6 +119,7 @@ class SDL2GradleBootstrap(Bootstrap):
                 modules_build_dir = join(
                     self.ctx.python_recipe.get_build_dir(arch.arch),
                     'android-build',
+                    'build',
                     'lib.linux-arm-3.7')
                 module_filens = (glob.glob(join(modules_build_dir, '*.so')) +
                                  glob.glob(join(modules_build_dir, '*.py')))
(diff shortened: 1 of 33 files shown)


pmp-p:
Good job !
Waiting for it and ready to offer on the side some hacky patches to help on API 19 for all those devices who still requires kitkat ;)
btw 3.7.1rc1 is ok with the patches you know so i guess it's finally getting better for Android.

inclement:
> Waiting for it and ready to offer on the side some hacky patches to help on API 19 for all those devices who still requires kitkat ;)

Sounds good to me :)

> btw 3.7.1rc1 is ok with the patches you know so i guess it's finally getting better for Android.

Glad to hear it! I was worried there could be changes that break things.

pythonforandroid/bootstraps/sdl2/build/jni/Application.mk
```diff
@@ -5,3 +5,4 @@
 
 # APP_ABI := armeabi armeabi-v7a x86
 APP_ABI := $(ARCH)
+APP_PLATFORM := android-21
```
opacam: This causes sdl2 recipe to fail. We can comment this line to bypass the error...as a temporary hack in order to make further tests, **NOT AS A FINAL SOLUTION**.

`# APP_PLATFORM := android-21`


Note: This already has been informed by @inclement into the "to do" section : "add proper APP_PLATFORM setting functionality"


Here are the log of the failure:

```[INFO]:    -> directory context /home/opacam/.buildozer-builds/android/platform/build/build/bootstrap_builds/sdl2-python3/jni
[DEBUG]:   -> running ndk-build V=1
[DEBUG]:        Android NDK: APP_PLATFORM set to unknown platform: 21.
[DEBUG]:        /media/DEVEL/Android/android-ndk-r17b/build/core/setup-app-platform.mk:135: *** Android NDK: Aborting    .  Stop.
Exception in thread background thread for pid 31274:
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/home/opacam/.local/lib/python2.7/site-packages/sh.py", line 1540, in wrap
    fn(*args, **kwargs)
  File "/home/opacam/.local/lib/python2.7/site-packages/sh.py", line 2459, in background_thread
    handle_exit_code(exit_code)
  File "/home/opacam/.local/lib/python2.7/site-packages/sh.py", line 2157, in fn
  File "/home/opacam/.local/lib/python2.7/site-packages/sh.py", line 815, in handle_command_exit_code
    raise exc
ErrorReturnCode_2: 

  RAN: /media/DEVEL/Android/android-ndk-r17b/ndk-build V=1

  STDOUT:
Android NDK: APP_PLATFORM set to unknown platform: 21.    
/media/DEVEL/Android/android-ndk-r17b/build/core/setup-app-platform.mk:135: *** Android NDK: Aborting    .  Stop.


  STDERR:


  File "/home/opacam/.local/bin/p4a", line 11, in <module>
    load_entry_point('python-for-android', 'console_scripts', 'p4a')()
  File "/home/opacam/Devel/python-for-android/pythonforandroid/toolchain.py", line 981, in main
    ToolchainCL()
  File "/home/opacam/Devel/python-for-android/pythonforandroid/toolchain.py", line 518, in __init__
  File "/home/opacam/Devel/python-for-android/pythonforandroid/toolchain.py", line 148, in wrapper_func
  File "/home/opacam/Devel/python-for-android/pythonforandroid/toolchain.py", line 192, in build_dist_from_args
  File "/home/opacam/Devel/python-for-android/pythonforandroid/build.py", line 579, in build_recipes
    recipe.build_arch(arch)
  File "/home/opacam/Devel/python-for-android/pythonforandroid/recipes/sdl2/__init__.py", line 41, in build_arch
    shprint(sh.ndk_build, "V=1", _env=env)
  File "/home/opacam/Devel/python-for-android/pythonforandroid/logger.py", line 176, in shprint
    for line in output:
  File "/home/opacam/.local/lib/python2.7/site-packages/sh.py", line 863, in next
    self.wait()
  File "/home/opacam/.local/lib/python2.7/site-packages/sh.py", line 792, in wait
    self.handle_command_exit_code(exit_code)
  File "/home/opacam/.local/lib/python2.7/site-packages/sh.py", line 815, in handle_command_exit_code
    raise exc
sh.ErrorReturnCode_2: 

  RAN: /media/DEVEL/Android/android-ndk-r17b/ndk-build V=1

  STDOUT:
Android NDK: APP_PLATFORM set to unknown platform: 21.    
/media/DEVEL/Android/android-ndk-r17b/build/core/setup-app-platform.mk:135: *** Android NDK: Aborting    .  Stop.```

File: pythonforandroid/bootstraps/sdl2/__init__.py [hunk 2]
```diff
@@ -100,7 +106,55 @@ def run_distribute(self):
...
+                                 glob.glob(join(modules_build_dir, '*.py')))
+                for filen in module_filens:
+                    shprint(sh.cp, filen, modules_dir)
+
+                # zip up the standard library
+                stdlib_zip = join(self.dist_dir, python_bundle_dir, 'stdlib.zip')
+                with current_directory(
+                        join(self.ctx.python_recipe.get_build_dir(arch.arch),
+                             'Lib')):
+                    shprint(sh.zip, '-r', stdlib_zip, *os.listdir())
```
opacam: Fix missing argument for command os.listdir, we can fix it like this:

```
                stdlib_dir = join(
                    self.ctx.python_recipe.get_build_dir(arch.arch), 'Lib')
                with current_directory(stdlib_dir):
                    shprint(sh.zip, '-r', stdlib_zip, *os.listdir(stdlib_dir))```

File: pythonforandroid/bootstraps/sdl2/build/jni/Application.mk (hunk 1 above)
Comment (inclement) [reply pythonforandroid/bootstraps/sdl2/build/jni/Application.mk]: I don't know what caused this for you, does your NDK not have platform-21?

File: pythonforandroid/bootstraps/sdl2/__init__.py (hunk 2 above)
Comment (inclement) [reply pythonforandroid/bootstraps/sdl2/__init__.py]: Alternative fix: drop python2 support :<

Comment (opacam) [reply pythonforandroid/bootstraps/sdl2/__init__.py]: If you maintain the affected code as it is you will also drop support for python 3 prior to version 3.2 and it would be great that was documented in somewhere: https://docs.python.org/3/library/os.html?highlight=listdir#os.listdir

**Comment by inclement** [comment]
I spoke with @tito on discord about the question above, but for the benefit of others:
- So far I've tested with NDK 14 and API 24 (and ndk_api 21). This is because these are the specific versions that bpo-30386 was designed for. I will move to NDK 17c and API 26+ soon, but others have confirmed this seems to work.
- My compilation line is `python3 setup_testapp_python3.py apk --ndk-dir=/home/sandy/android/android-ndk-r14b --android-api=24`, where setup_testapp_python3.py is in python-for-android/testapps. This is the main build I test during development, it is expected to work.

**Comment by inclement** [comment]
For reference, @tito found some more places where we need to be careful to use the right ndk api target, see https://gist.github.com/tito/73a2d35c236bb34f6441b64be44c5683 . These will be fixed in this PR soon.

**Comment by opacam** [comment]
OOoo, this is great, I'm testing the @tito's patch over the last published @inclement's work, we are getting closer but unfortunately the openssl libs causes troubles, linkages error when generating python 3 lib, but this can bypassed adding a new argument into the perl's configure (if we don't do that, the openssl build system will use the higher api we have available and  if it don't match with the one we use with python 3 we have the mentioned linkage errors), so here are a patch to apply after @tito's patch:

- https://github.com/opacam/python-for-android/commit/944e47da87eed8c6fd379f78daf989025dbfc997.patch

And another patch to add testapp for python3 with openssl and sqlite3

- https://github.com/opacam/python-for-android/commit/351bcc946741d89b6394f65586afa9a3fe73377c.patch

This will solve the linkage problem, however no_ssl generated. The good news are the sqlite3 module is generated.

Note: Right now I'm testing with ndk-r17b, python 3.6 interpreter and targeting api 24 and with @inclement's method but with the testapp i posted above with openssl and sqlite added.

**Comment by opacam** [comment]
Yeaah, I found the ssl problem, Tonight I will post a link pointing to the patch.

**Comment by inclement** [comment]
General update: Currently seems working fine with API 27 and newest SDK, I'm moving on to cleaning up the code and fixing some of the other things I broke along the way (like the other python recipes as noted by @opacam). I appreciate the testing and feedback.

I won't let the openssl and sqlite bullet points hold up this main PR, since others have confirmed so well that they work fine, but I expect to check through everything else in my original post before merging.

The amount of progress will depend on how much time I have to work on it, but hopefully I can get a lot of the remaining work done by next week.

**Comment by AndreMiras** [comment]
Awesome work! The overall looks good.
I can take a look/play with the `python3crystax` to check for regression.
Simply double check you're rebased with upstream and I'll hook to your fork to build/run some apps on `python3crystax`.

File: pythonforandroid/recipes/python2/__init__.py [hunk 3]
```diff
@@ -170,5 +173,52 @@ def do_python_build(self, arch):
...
+        libdir = join(dirn, 'lib', 'python2.7')
+        site_packages_dir = join(libdir, 'site-packages')
+        with current_directory(libdir):
+            removes = []
+            for dirname, root, filenames in walk("."):
+                for filename in filenames:
+                    for suffix in EXCLUDE_EXTS:
+                        if filename.endswith(suffix):
+                            removes.append(filename)
+            shprint(sh.rm, '-f', *removes)
```
AndreMiras: Out of scope since it was already in our code base.
But it hurts a little bit knowing this entire block could be a single line command using `find`:
```sh
find . -name '*.ext`' -o -name '*.ext2' -o -name '*.ext3' -delete
```
But yes that's not portable. However I'm wondering if there's not a way to do it in Python that would look nicer

pythonforandroid/recipes/python2/__init__.py
```diff
@@ -170,5 +173,52 @@ def do_python_build(self, arch):
...
+
+            info('Deleting some other stuff not used on android')
+            # To quote the original distribute.sh, 'well...'
+            shprint(sh.rm, '-rf', 'lib2to3')
+            shprint(sh.rm, '-rf', 'idlelib')
+            for filename in glob.glob('config/libpython*.a'):
+                shprint(sh.rm, '-f', filename)
```
AndreMiras: Minor, maybe this would work directly?
```
sh.rm('-f', glob.glob('config/libpython*.a'))
```

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -0,0 +1,198 @@
...
+        sys_exec_prefix = '/usr/local'
+
+        # Skipping "Ensure that nl_langinfo is broken" from the original bpo-30386
+
+        platform_name = 'android-{}'.format(self.ctx.ndk_api)
+
+        with current_directory(build_dir):
+            env = environ.copy()
+
+            # TODO: Get this information from p4a's arch system
```
AndreMiras: indeed, most* of this should be available in `arch.get_env()` right?

pythonforandroid/recipes/python2/__init__.py
inclement: I'm thinking more and more about making some better python abstractions for this. Something for the near future, perhaps.

pythonforandroid/recipes/python3/__init__.py
inclement: Yes, but it needs some modification to work seamlessly with clang, and that effort should also have an eye to making clang work for the other recipes that still use gcc. It shouldn't be hard and I originally intended to do it in this PR, but since the redundant code is confined to the new python3 recipe I don't think it's worth holding things up for it. I'd feel differently and want to do it first if it weren't for the API 26+ problem.

pythonforandroid/recipes/python2/__init__.py
tito: There is fnmatch in the standard lib that can be used.

pythonforandroid/bootstraps/sdl2/build/jni/src/start.c
```diff
@@ -76,10 +76,9 @@ int main(int argc, char *argv[]) {
   int ret = 0;
   FILE *fd;
 
-  /* AND: Several filepaths are hardcoded here, these must be made
-     configurable */
-  /* AND: P4A uses env vars...not sure what's best */
-  LOGP("Initialize Python for Android");
+  setenv("P4A_BOOTSTRAP", "SDL2", 1);  // env var to identify p4a to applications
```
tito: This is a great addition!

pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonUtil.java
```diff
@@ -66,7 +68,7 @@ public static void loadLibraries(File filesDir) {
                 // load, and it has failed, give a more
                 // general error
                 Log.v(TAG, "Library loading error: " + e.getMessage());
-                if (lib.startsWith("python3.6") && !foundPython) {
+                if (lib.startsWith("python3.7") && !foundPython) {
```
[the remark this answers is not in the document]
tito: Should even be just `python`, not `python3.`, to check for `python2` as well. Or do 2 checks.

pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonUtil.java
```diff
@@ -45,6 +45,8 @@ protected static void addLibraryIfExists(ArrayList<String> libsList, String patt
         addLibraryIfExists(libsList, "crypto.*", libsDir);
         libsList.add("python2.7");
         libsList.add("python3.5m");
+        libsList.add("python3.6m");
+        libsList.add("python3.7m");
```
tito: Shouldn't theses changes added in all others bootstraps?

testapps/setup_testapp_python3crystax.py
```diff
@@ -0,0 +1,31 @@
+
+from distutils.core import setup
+from setuptools import find_packages
+
+options = {'apk': {'debug': None,
+                   'requirements': 'sdl2,pyjnius,kivy,python3crystax',
+                   'android-api': 19,
+                   'ndk-dir': '/home/asandy/android/crystax-ndk-10.3.2',
```
tito: os.environ["ANDROIDNDK"] ?

pythonforandroid/recipes/python2/__init__.py
AndreMiras: Actually I realize `endswith` can take tuple :smile: 
So we could actually go one `for` less:
```python
if filename.endswith(EXCLUDE_EXTS):
```
But I think it's OK to tackle it outside this pull request, since we have enough to check

pythonforandroid/recipes/python3crystax/__init__.py
```diff
@@ -73,5 +73,29 @@ def build_arch(self, arch):
...
+
+    def include_root(self, arch_name):
+        return join(self.ctx.ndk_dir, 'sources', 'python', self.major_minor_version_string(),
+                    'include', 'python')
+
+    def link_root(self, arch_name):
+        return join(self.ctx.ndk_dir, 'sources', 'python', self.major_minor_version_string(),
+                    'libs', arch_name)
 
 recipe = Python3Recipe()
```
AndreMiras: Shouldn't we rename this lass `Python3CrystaxRecipe` to avoid having it clashing with the new `Python3Recipe` one?

pythonforandroid/recipes/sdl2/__init__.py
```diff
@@ -10,18 +10,28 @@ class LibSDL2Recipe(BootstrapNDKRecipe):
...
     conflicts = ['sdl', 'pygame', 'pygame_bootstrap_components']
 
     patches = ['add_nativeSetEnv.patch']
 
     def get_recipe_env(self, arch=None):
         env = super(LibSDL2Recipe, self).get_recipe_env(arch)
+
         py2 = self.get_recipe('python2', arch.ctx)
         env['PYTHON2_NAME'] = py2.get_dir_name()
+        py3 = self.get_recipe('python3', arch.ctx)
```
AndreMiras: `py3` is unused here, right?

pythonforandroid/build.py
```diff
@@ -615,7 +626,9 @@ def run_pymodules_install(ctx, modules):
 
     venv = sh.Command(ctx.virtualenv)
     with current_directory(join(ctx.build_dir)):
-        shprint(venv, '--python=python2.7', 'venv')
+        shprint(venv,
+                '--python=python{}'.format(ctx.python_recipe.major_minor_version_string()),
```
AndreMiras: `major_minor_version_string` is a property, not a method, so we should remove the brackets

pythonforandroid/recipes/python3crystax/__init__.py
```diff
@@ -73,5 +73,29 @@ def build_arch(self, arch):
...
+                            py_recipe.version, 'libs', arch.arch)
+        shprint(sh.cp, '-r', join(python_dir,
+                                    'stdlib.zip'), dirn)
+        shprint(sh.cp, '-r', join(python_dir,
+                                    'modules'), dirn)
+        shprint(sh.cp, '-r', self.ctx.get_python_install_dir(),
+                join(dirn, 'site-packages'))
+
+        info('Renaming .so files to reflect cross-compile')
+        self.reduce_object_file_names(self, join(dirn, "site-packages"))
```
AndreMiras: So I'm not the only one copy/pasting the method definitions right :stuck_out_tongue: 
You forgot a `self` in the args here :smile:

pythonforandroid/recipes/hostpython3crystax/__init__.py
```diff
@@ -27,7 +27,6 @@ def build_arch(self, arch):
         Creates expected build and symlinks system Python version.
         """
         self.ctx.hostpython = '/usr/bin/false'
-        self.ctx.hostpgen = '/usr/bin/false'
         # creates the sub buildir (used by other recipes)
         # https://github.com/kivy/python-for-android/issues/1154
         sub_build_dir = join(self.get_build_dir(), 'build')
```
AndreMiras: We should probably rename the recipe class to `HostPython3CrystaxRecipe` so it doesn't "conflict" with the new one

AndreMiras:
I could successfully build with `python3crystax` after addressing [some of the issues](https://gist.github.com/AndreMiras/f6a91222e58dae41ea09e74fbe9b0b74#file-python3crystax_fixes-patch) I reported in my recent review.
I also add to update some of my `buildozer.spec` conf:

- `android.api = 21` (before I was running the default `android.api = 19`)
- `android.minapi = 21` (before I was running the default `android.minapi = 9`)

pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonUtil.java
inclement: The other bootstraps don't actually support python3 yet, I think it's best to leave that as a separate task rather than delay merging.

inclement:
The python3 testapp is failing because something tries to call python3.7. I will investigate tomorrow.
The other tests revealed a bug with ndk_api being set to 0 if a dist already exists. I will maybe improve how this is handled, but I'd rather leave a larger refactoring to a separate PR.

inclement:
> go for a "more mainstream" Python version by default i.e. 3.6 rather than 3.7

I haven't actually tried, but I wouldn't be surprised if this doesn't work with Python 3.6.

> configure: error: python3.7 interpreter not found

What concerns me about this is that we have a hostpython3 recipe, there should be no need for the user to have their own python3.7 installed, so I'll look into it before merging.

Ham-Merhead:
Really fine work, thank you. 
FYI a user experience with actual app using p4a master, Python3, and android-28
Below my full Python3/Crystax build environment - works great.
(Though it would be nice if "pip install python-for-android" didn't require sudo)

I tried the new python3 recipe, unzipped android-ndk, made 3 changes to .p4a, new lines are:
--requirements=python3,kivy
--ndk_dir <WHERE_EVER>/android-ndk-r18b
--ndk_version 18b

But build fails during ndk-build, "Cythonizing anything necessary in pyjnius":
```
 RAN: /home/bobf/.local/share/python-for-android/build/other_builds/hostpython3/desktop/hostpython3/native-build/python setup.py build_ext -v

clang: error: unknown argument: '-mandroid'

```
I realize this is a newer ndk than discussed here, but the problem looks more like clang configuration. Google was not my friend on this one. Any suggestions? Thanks.

This works ............Python3/Crystax........

````
# Recipe for building a python3 app (written using Kivy) for Android/Crystax.
#
# ASSUMES Ubantu 18.04 CLEAN INSTALL
# Tested with Windows Subystem for Unix Ubantu 18.04 and Android Pie
#

sudo apt-get update
sudo apt-get upgrade

### Workaround Issue #1327. Java heap issue - thanks JonasT
### Also add this line to .bashrc
export GRADLE_OPTS="-Xms1724m -Xmx5048m -Dorg.gradle.jvmargs='-Xms1724m -Xmx5048m'"

### Install modified P4A Python 2.7 dependencies
# changes to documentation:
# python3-all-dev replaces python2.7-dev   
# openjdk-8-jdk   replaces openjdk-7-jdk
# apparently not required : libncurses5:i386 libstdc++6:i386 zlib1g:i386 

sudo apt-get install -y build-essential ccache git zlib1g-dev python3-all-dev openjdk-8-jdk unzip autoconf libtool

### Install more P4A dependencies
### P4A requires Python 3.6, the default Python version on Ubantu 18 is 3.6
### gradle, virtualenv, pip, pkg, libgl1
### Gradle MUST be installed AFTER openjdk-8-jdk has been installed

sudo apt-get install -y gradle virtualenv python3-pip  pkg-config libgl1-mesa-dev

### Install Python packages
### Versions I get:  Cython 0.29.1 , Kivy 1.10.1,  P4A '0.6.0'
### The P4A version is mostly meaningless it is really a newer 'master'(?) build
### P4A must be installed with sudo else p4a stub is not created
### P4A requires Cython to be installed with sudo, else it will not find cython
### In some environment with more than one Python version, these two sudo
### might be an issue?

sudo -H pip3 install cython
pip3 install kivy
sudo -H pip3 install git+https://github.com/kivy/python-for-android.git

# Download Android and Crystax Tools (ndk and sdk) to <WHERE_EVER>
# https://www.crystax.net/en/download    crystax-ndk-10.3.2-linux-x86_64.tar.xz
# https://developer.android.com/studio/  sdk-tools-linux-4333796.zip
mkdir <WHERE_EVER>
cd    <WHERE_EVER>
unzip  -d sdk  sdk-tools-linux-4333796.zip 
tar xf crystax-ndk-10.3.2-linux-x86_64.tar.xz

# Build the Android Tools
cd <WHERE_EVER>/sdk/tools/bin
./sdkmanager "platforms;android-28"
./sdkmanager "build-tools;28.0.3"

### Install Done
### Build app-version-debug.apk

cd <location of app main.py>

# Create a ".p4a" containing (replace text between and including <>)
# An app may need more options than these:
--dist_name=<XXX>
--private .
--package=com.<YYY>.<XYZ>
--name <ZZZ>
--bootstrap=sdl2
--requirements=python3crystax,kivy
--arch=armeabi-v7a
--icon <ICON>.png
--sdk_dir <WHERE_EVER>/sdk
--ndk_dir <WHERE_EVER>/crystax-ndk-10.3.2
--android_api 28
--ndk_version 10.3.2
--ndk-api 21
--version <0.0.0>

```
`

ApplySci:
> FYI a user experience with actual app using p4a master, Python3, and android-28
> Below my full Python3/Crystax build recipe - builds, installs, and runs great.

@Ham-Merhead does this produce an apk that the Google Play Store accepts? And have you tried running the apk on an Android 8.1 platform? (I've hit a wall with an apk that worked on 8.0, but fails on 8.1, failing to import _sqlite3.so)

OptimusGREEN:
I thought crystax only supported up to 21?

Ham-Merhead:
@ApplySci  My .apk works with Android 9 (I dont know about 8.1), but my app does not use sql which perhaps yours does. I have not tried submitting to the store.

Ham-Merhead:
@OptimusGREEN  My understanding is the NDK (here Crystax) defines the minimum, and the (Android) SDK the maximum. But perhaps I don't understand?!
I also have an Android-NDK (in place of Crystax) build environment that is simpler
[https://groups.google.com/forum/#!topic/kivy-users/Lf4zlYmLVPo](url)

Issue #1415, opened 2018-10-20
pyjnius really should be version pinned.


JonasT:
### Versions
latest git master of p4a, no matter what python/cython/kivy used

### Description

pyjnius isn't version pinned: 

https://github.com/kivy/python-for-android/blob/master/pythonforandroid/recipes/pyjnius/__init__.py#L10

This means even if I stay at a fixed commit of p4a master, I don't get reproducible builds and if someone breaks stuff in pyjnius like **right now** -> #1414 I might be stuck not being able to do a simple deployment of my application unless I happened to keep a previous full system image with a working combination of things.

This is bad for such a core component. If I stick on a p4a master at least all the core things should be guaranteed to work, because manually meddling around in recipes just to get back to an older version of e.g. pyjnius in an automated build chain is pretty annoying to do.

*Even if nobody ever tests the pin and the pin is updated to a broken version, this is a lot better than no pin.* (because in that case, everyone can just go back with their main p4a master clone, which is easy, instead of patching inside the recipes folder during the build to avoid broken pyjnius, which is hard and annoying.)

### buildozer.spec
not relevant
### Logs

not relevant

JonasT:
Closing since it appears to be pinned now!

KeyWeeUsr:
Ref https://github.com/kivy/python-for-android/pull/1426.

PR #1416, opened 2018-10-20
Pin pyjnius version

JonasT:
This pull request adds a version pin for pyjnius as requested here: #1415 

Please note I am proposing this as a **permanent measure** because this is such a core component - not this specific version of course, but that it is always pinned.

Even if you just randomly bump the version up in any random commit without checking, this is a huge improvement: it will prevent p4a master builds from randomly failing out of the blue *when not even changing the commit*, and people will be able to go back to an earlier p4a master commit to avoid sudden pyjnius breakages instead of patching around in the recipes folder (which depending on the build pipeline might be quite a time waster to do, especially compared to just going back to a known working p4a master commit).

Summed up, please pin this, carelessly bump it whenever, and have less unhappy users. :smile:

(And I'm not proposing pinning *everything*, I know you have way too many recipes and people would forget to bump it, I understand - but at least the core components like pyjnius, would that possibly sound feasible?)

### DIFF
--- a/pythonforandroid/recipes/pyjnius/__init__.py
+++ b/pythonforandroid/recipes/pyjnius/__init__.py
@@ -6,7 +6,7 @@ from os.path import join
 
 
 class PyjniusRecipe(CythonRecipe):
-    version = 'master'
+    version = '6233ffe6415acfef9f97ff993e12e8db064c7c75'
     url = 'https://github.com/kivy/pyjnius/archive/{version}.zip'
     name = 'pyjnius'
     depends = [('python2', 'python3crystax'), ('genericndkbuild', 'sdl2', 'sdl'), 'six']


AndreMiras:
Thanks for the pull request!
FYI I've restarted some of the failing builds because some downloads are flaky and make the build fail.
Things I generally like in pull requests:
- the [50/72 rule](https://medium.com/@preslavrachev/what-s-with-the-50-72-rule-8a906f61f09c)
- write descriptive title rather than lecturing e.g. `Pins pyjnius version, fixes #1415`
- short commit hashes are cute e.g. [6233ffe](https://github.com/kivy/pyjnius/commit/6233ffe6415acfef9f97ff993e12e8db064c7c75)

It's more a matter of personal taste than a pro tip.

rnixx:
If pinning gets done, which is actually a good idea, please make at least a tag of pyjnius.

AndreMiras:
This is now "conflicting" with https://github.com/kivy/python-for-android/pull/1417
In https://github.com/kivy/python-for-android/pull/1417 @KeyWeeUsr is updating the patch to match master. It means the pinned commit hash would now fail since the patch is now working past this hash.
In the meantime we do agree we should pin when possible. So we could wait for pyjnius to be tagged before we can eventually pin it.

JonasT:
Shouldn't #1417 just update the pin with exactly that merge? That is the point of having a pin, updating it along with all changes that make things incompatible, such that no arbitrarily incompatible versions of patch & pyjnius or whatever are combined

JonasT:
@rnixx I have been thinking about your comment, and I'm wondering, while pinning a tag seems helpful for major releases, wouldn't it be better to stick with pinning to commits for `p4a master`?

Otherwise it'll either be a huge amount of tags, or `p4a master` will be quite inflexible / it'll be effort for `p4a master` to use unusual experimental changes in pyjnius, and that seems like it wouldn't be a good idea... but that's just my first thought on this

inclement:
Summary of my opinions from discord:
- This seems like an excellent idea for releases, and for the stable branch. Perhaps it can be a part of release procedure to pin certain recipes.
- I feel like having p4a master track pyjnius master has advantages, in fact it's like this in the first place because it's usually been better than the alternative in the past as pyjnius is only updated quite rarely and usually for important reasons. It also guarantees that pyjnius gets meaningful testing without much effort, otherwise basically nobody would use pyjnius master.
That said, pinning specific commits and updating regularly probably wouldn't be a major issue for point 2, I guess I'm really saying that it might not be worth the trouble since updating the pin could easily get forgotten and lead to surprising bugs at a later date.

JonasT:
@inclement I'll sure be jumping ship from using master even if such pyjnius breakage only happens rarely. The problem is not the breakage itself, but that it's not pinned / reproducible for one commit, that's just pure acid for any actual deployment use even for the more adventurous. So if you want people to test it, I think leaving this unpinned won't necessarily do you a favor.

Also, for Python 3, the p4a stable is pretty useless right now given how much doesn't work with it. But of course that could change with upcoming releases.

AndreMiras:
I've just created https://github.com/kivy/python-for-android/pull/1427 to pin on the last freshly created tag.
I would prefer to give you the credits if you want to amend your change with tag and new commit message

JonasT:
I really couldn't care less about credits especially for such a minor change, so if it's less work, just merge it in :+1:

PR #1417, opened 2018-10-20
Fix pyjnius setup.py patches

KeyWeeUsr:
Broken due to edits in https://github.com/kivy/pyjnius/pull/356

### DIFF
--- a/pythonforandroid/recipes/pyjnius/genericndkbuild_jnienv_getter.patch
+++ b/pythonforandroid/recipes/pyjnius/genericndkbuild_jnienv_getter.patch
@@ -16,10 +16,10 @@ index 740510f..0c8e55f 100644
 +++ b/setup.py
 @@ -53,7 +53,7 @@ except ImportError:
  
- if platform == 'android':
+ if PLATFORM == 'android':
      # for android, we use SDL...
--    libraries = ['sdl', 'log']
-+    libraries = ['main', 'log']
-     library_dirs = ['libs/' + getenv('ARCH')]
- elif platform == 'darwin':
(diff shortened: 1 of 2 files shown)


KeyWeeUsr:
[`sdl2_jnienv_getter.patch`](https://github.com/kivy/python-for-android/blob/master/pythonforandroid/recipes/pyjnius/sdl2_jnienv_getter.patch) patch though seems like it should already been applied to pyjnius directly.

https://github.com/search?q=org%3Akivy+SDL_ANDROID_GetJNIEnv&type=Code
https://github.com/search?q=org%3Akivy+SDL_AndroidGetJNIEnv&type=Code SDL_ANDROID_GetJNIEnv 

is just replaced with SDL_AndroidGetJNIEnv when building SDL2... but why?

even android recipe calls the SDL2 original one [`_android_billing_jni.c#L19`](https://github.com/kivy/python-for-android/blob/master/pythonforandroid/recipes/android/src/android/_android_billing_jni.c#L19)

`SDL_AndroidGetJNIEnv` is the original `extern` one from sdl2 and that calls sdl's `Android_JNI_GetEnv`, but for pygame there is `SDL_ANDROID_GetJNIEnv`

https://github.com/kivy/p4a-pygame-bootstrap-components/blob/7d39c8d2a3f194c08041a7702419e44c042eb3b1/pygame_bootstrap_jni/sdl/src/video/android/SDL_androidvideo.c#L61 for some reason and that's not even externed :thinking: 

I don't think it should be patched, rather renamed to the correct thing hmmm.

r00tb3:
How do i fix this?

I'm unable to apply this patch.

PR #1429, opened 2018-10-23
Creates non root user in Dockerfile

AndreMiras:
Subsequent p4a command will run as unprivileged user.
This makes the setup more realistic/closer to an actual setup.

Also added .dockerignore so container and host generated files don't conflict.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,10 @@
 FROM ubuntu:18.04
 
 
+ENV USER="user"
+ENV HOME_DIR="/home/${USER}"
+ENV WORK_DIR="${HOME_DIR}" \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -36,7 +40,7 @@ ENV ANDROID_NDK_DL_URL="https://dl.google.com/android/repository/${ANDROID_NDK_A
 
 # install system dependencies
 RUN apt update -qq && apt install -qq --yes --no-install-recommends \
-	python virtualenv python-pip wget curl lbzip2 patch bsdtar && \
+	python virtualenv python-pip wget curl lbzip2 patch bsdtar sudo && \
```
AndreMiras: Added `sudo` because it could still be useful on interactive shell

PR #1430, opened 2018-10-24
removed mmap from blacklist

goffi-contrib:
mmap was blacklisted, but it is actually useful, e.g. to communicate
between a frontend and a service.
This patch remove it from blacklists.

### DIFF
--- a/pythonforandroid/bootstraps/pygame/build/blacklist.txt
+++ b/pythonforandroid/bootstraps/pygame/build/blacklist.txt
@@ -78,7 +78,6 @@ multiprocessing/dummy*
 lib-dynload/termios.so
 lib-dynload/_lsprof.so
 lib-dynload/*audioop.so
-lib-dynload/mmap.so
 lib-dynload/_hotshot.so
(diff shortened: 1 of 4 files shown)


AndreMiras:
Thanks,
I've never played with `mmap` so I can't tell, @tito you initially blacklisted a bunch including `mmap` in https://github.com/kivy/python-for-android/commit/2b02324faa32ba16353e10d279bcca70098974cb can you review/merge if it's OK?
In the meantime @goffi-contrib can you simply rebase on `master` so CI is happy? The idea is to make a full CI run to make sure the recipes compile. Here the CI stopped because of the linter.

tito:
I guess it was removed because it didn't compile for android at the time. If it works today, it can be included yep.

AndreMiras:
Thanks for the feedbak tito.
Well then @goffi-contrib if you only rebase on master so we can check Travis is happy we're good to merge

goffi-contrib:
@AndreMiras done :)

PR #1431, opened 2018-10-24
[pyOpenSSL] version update

goffi-contrib:
older version of pyOpenSSL was causing trouble ('module' object has no attribute 'OP_SINGLE_ECDH_USE'), updating it fixes it.
I didn't update to last version because OpenSSl and Cryptography are
outdated too, so this is a first step to have a working pyOpenSSL.


goffi-contrib:
@AndreMiras sure, I've just updated the PR. Thanks for your quick feedbacks :)

tox.ini
```diff
@@ -18,4 +18,5 @@ ignore =
     E123, E124, E126,
     E226,
     E402, E501, E722,
-    F812, F841, W503
+    F812, F841, W503,
+    W504
```
AndreMiras: For later reference; We ignored `W504` because the rule of `flake8==3.5.0` became the anti-pattern in `flake8==3.6.0`.

See https://github.com/LintlyCI/Flake8Rules/blob/ed09f9d/_rules/W503.md
> Note: Despite being in the best practice section, this will soon be considered an anti-pattern.

AndreMiras:
Sweet, let's see if CI is happy and merge it then

Issue #1337, opened 2018-08-24
ctypes.util.find_library doesn't work with python 3


JonasT:
We talked about this in chat before, I just wanted to make an actual issue for this: `ctypes.util.find_library` doesn't work when using `python3crystax` out of the box, because it seems to try something with a shell that doesn't work due to a missing binary.

This code makes it work when put before any ctypes.util import (it includes a PySDL2-specific hack without which PySDL2 will fail to load - this one might actually need addressing in the recipe for PySDL2):
```
def apply_hack():
    import ctypes.util
    orig_func = ctypes.util.find_library
    def android_find_library_hack(*args):
        import os
        name = args[0]

        # Truncate ending for easier comparison:
        if name.find(".so.") >= 0:
            name = name.partition(".so.")[0]
        if name.endswith(".so"):
            name = name[:-len(".so")]
        if not name.endswith("."):
            name += "."

        # Helper function to check lib name:
        def check_name(lib_name, search_name):
            if filename.endswith('.so') and (
                    filename.startswith("lib" + search_name) or
                    filename.startswith(search_name)):
                return True
            if search_name.endswith("-2.0."):  # PySDL2 hack
                search_name = search_name[:-len("-2.0.")] + "."
                return check_name(lib_name, search_name)
            return False

        # Check the user app lib dir and system dir:
        app_root = os.path.normpath(os.path.abspath(os.path.join(
            os.path.dirname(__file__), '..', '..', 'lib')))
        sys_dir = "/system/lib"
        for lib_search in [app_root, sys_dir]:
            if not os.path.exists(lib_search):
                continue
            for filename in os.listdir(lib_search):
                if check_name(filename, name):
                    return os.path.join(lib_search, filename)
        return orig_func(*args)
    import ctypes.util
    ctypes.util.find_library = android_find_library_hack
apply_hack()
```
I don't know enough to put this into python-for-android, but maybe someone with the knowledge can put this into the python launch wrapper somehow to make it magically work for everyone? Feel free to take it, consider it `CC0/public domain` (it's vaguely based on python-for-android's hack for Python 2 anyway)

AndreMiras:
Thanks for making a bug report out of it.
FWIW I also monkey patch it in a similar fashion here:
https://github.com/AndreMiras/EtherollApp/blob/v20180617/src/etheroll/utils.py#L172
[...]

PR #1433, opened 2018-10-25
Add ctypes workaround to bootstrap (fixes ctypes.util.find_library() crash)

JonasT:
This adds the missing ctypes workaround for Python 3. Since it's in `start.c` it will apply to any sort of Python 3 used, no matter if it's Crystax or any other python install that lacks a properly patched `ctypes.util.find_library()` function.

While this should probably be patched in the Python 3 itself in the long run, I suggest this fix is merged some time soon because:

- it's available & works now (at least in my test it did)
- it is rather local and can easily be removed once it is no longer necessary
- even if applied on top of a working `ctypes.util.find_library()` function, it shouldn't render it inoperable

For my app, this worked perfectly fine and allowed me to get rid of ugly import order issues related to monkey patching this in the app code. It'd be very nice to have this finally fixed upstream.

See also ticket #1337

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/build/jni/src/start.c
+++ b/pythonforandroid/bootstraps/sdl2/build/jni/src/start.c
@@ -66,6 +66,38 @@ int file_exists(const char *filename) {
   return 0;
 }
 
+/* Helper function to escape all occurrences of ' as \', and all
+ * occurances of \n and \r with \\n and \\r for inline string literals.
+ * (Unix paths can contain line breaks, so to be super safe, we need this).
+ * NOTE: return value needs to be free()'d after use by the caller!!
+ */
+char *escape_quotes_of_inline_string(const char* inline_code_arg) {
+  char *result_buf = malloc(strlen(inline_code_arg) * 2 + 1);
+  // (buffer size is large enough for worst case of every char escaped)
+  int targetoffset = 0;
+  int sourceoffset = 0;
+  while (sourceoffset <= strlen(inline_code_arg)) {
+    if (inline_code_arg[sourceoffset] == '\'') {
+      result_buf[targetoffset] = '\\';
+      targetoffset++;
+      result_buf[targetoffset] = '\'';
+    } else if (inline_code_arg[sourceoffset] == '\n') {
+      result_buf[targetoffset] = '\\';
+      targetoffset++;
+      result_buf[targetoffset] = 'n';
+    } else if (inline_code_arg[sourceoffset] == '\r') {
+      result_buf[targetoffset] = '\\';
+      targetoffset++;
+      result_buf[targetoffset] = 'r';
+    } else {
+      result_buf[targetoffset] = inline_code_arg[sourceoffset];
+    }
+    sourceoffset++;
+    targetoffset++;
+  }
+  return result_buf;
+}
+
 /* int main(int argc, char **argv) { */
 int main(int argc, char *argv[]) {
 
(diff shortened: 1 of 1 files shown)


pythonforandroid/bootstraps/sdl2/build/jni/src/start.c
```diff
@@ -66,6 +66,36 @@ int file_exists(const char *filename) {
...
+  while (sourceoffset <= strlen(inline_code_arg)) {
+    if (inline_code_arg[sourceoffset] == '\'') {
+      result_buf[targetoffset] = '\\';
+      targetoffset++;
+      result_buf[targetoffset] = '\'';
+    } else if (inline_code_arg[sourceoffset] == '\n') {
+      result_buf[targetoffset] = '\\';
+      targetoffset++;
+      result_buf[targetoffset] = 'n';
+    } else if (inline_code_arg[sourceoffset] == '\n') {
```
AndreMiras: I didn't follow the discussion, but you probably mean `'\r'` here, right?

pythonforandroid/bootstraps/sdl2/build/jni/src/start.c
```diff
@@ -66,6 +66,36 @@ int file_exists(const char *filename) {
   return 0;
 }
 
+/* Helper function to escape all occurances of ' as \', and all
+ * occurances of \n and \r with \\n and \\r for inline string literals.
+ * (Unix paths can contain line breaks, so to be super safe, we need this) */
+char *escape_quotes_of_inline_string(const char* inline_code_arg) {
+  char *result_buf = malloc(strlen(inline_code_arg) * 2 + 1);
```
AndreMiras: I don't see where this is freed?

pythonforandroid/bootstraps/sdl2/build/jni/src/start.c
JonasT: Oh right. `\r` in paths are so rare I wasn't sure whether to put it in at all, but yea that's supposed to be `\r`

pythonforandroid/bootstraps/sdl2/build/jni/src/start.c
JonasT: Good point, I should really add a note to the comment above the function that it returns something that needs to be `free()`'d

pythonforandroid/bootstraps/sdl2/build/jni/src/start.c
```diff
@@ -66,6 +66,38 @@ int file_exists(const char *filename) {
   return 0;
 }
 
+/* Helper function to escape all occurances of ' as \', and all
+ * occurances of \n and \r with \\n and \\r for inline string literals.
```
AndreMiras: s/occurances/occurrences/

tito:
Also, it has an impact on current Python 2 apps, as the patch done in the source will be applied after this one. Unsure about the implication of both running on another.

JonasT:
For what it's worth, I updated it to catch a possible import error and it has an `#if` directive to only compile for Python 3. Of course the other issues with it still apply

JonasT:
Based on @tito 's feedback, it seems to make the most to close this. I am nevertheless already using this pull request for myself because I'm not interested in moving this hack back into my program, but that's my own problem for now

JonasT:
Followup attempt is here: https://github.com/kivy/python-for-android/pull/1517

PR #1428, opened 2018-10-23
lxml recipe

goffi-contrib:
This PR fixes the small issues found in https://github.com/kivy/python-for-android/pull/1166.
It's a working recipe for lxml and its dependencies.

PR #1435, opened 2018-10-29
Conditional builds falls back on Python2 if needed

AndreMiras:
Falls back to python2 when tested recipes are python3crystax compatible.
This is for instance the case on #1428 where the recipes depend only on
python2, hence couldn't be automatically tested.

Also updated `vispy` recipe to handle the version string properly,
pin it and clean comments.
This recipe is a demo case scenario for the conditional build falling
back on python2 since `vispy` is not yet compatible with python3crystax.


AndreMiras:
Updated the pull request, removing `debug` flag from the two test apps being used by the CI as it was producing two much logs, see https://api.travis-ci.org/v3/job/448003421/log.txt

PR #1437, opened 2018-11-03
allow installing/using p4a in a virtualenv

tshirtman:
I only tried to build a distribution, not an apk from it, but it seems to work fine. The pure python libs i added to --requirements were installed fine in the python-install of the created distribution, just like when i ran p4a from outside of a virtualenv.


AndreMiras:
Nice improvement, it looks good. I just want to test it further. Will do in the coming days.

PR #1439, opened 2018-11-03
allow dumping and loading the manifest file

tshirtman:
allow tweaking it and using the tweaked version in future builds.

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/build/build.py
+++ b/pythonforandroid/bootstraps/sdl2/build/build.py
@@ -355,20 +355,29 @@ main.py that loads it.''')
                                   key=LooseVersion)
     build_tools_version = build_tools_versions[-1]
 
-    render(
-        'AndroidManifest.tmpl.xml',
-        'src/main/AndroidManifest.xml',
-        args=args,
-        service=service,
-        service_names=service_names,
-        android_api=android_api,
-        url_scheme=url_scheme)
+    manifest_path = join('src', 'main', 'AndroidManifest.xml'
+
+    if args.load_manifest:
+        shutil.copy(args.load_manifest, manifest_path)
+
+    else:
+        render(
+            'AndroidManifest.tmpl.xml',
+            manifest_path,
+            args=args,
+            service=service,
+            service_names=service_names,
+            android_api=android_api,
+            url_scheme=url_scheme)
+
+        if args.save_manifest:
+            shutil.copy(manifest_path, args.save_manifest)
 
     # Copy the AndroidManifest.xml to the dist root dir so that ant
     # can also use it
     if exists('AndroidManifest.xml'):
         remove('AndroidManifest.xml')
-    shutil.copy(join('src', 'main', 'AndroidManifest.xml'),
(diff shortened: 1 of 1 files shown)


pythonforandroid/bootstraps/sdl2/build/build.py
```diff
@@ -355,20 +355,29 @@ def make_package(args):
...
+        render(
+            'AndroidManifest.tmpl.xml',
+            manifest_path,
+            args=args,
+            service=service,
+            service_names=service_names,
+            android_api=android_api,
+            url_scheme=url_scheme)
+
+        if args.save_manifest:
```
inclement: I think that save_manifest should happen regardless of whether load_manifest was set. Even if that isn't useful (i.e. saving the one the user provided in the first place), at least it's consistent. Or they could be marked as incompatible arguments instead.

pythonforandroid/bootstraps/sdl2/build/build.py
```diff
@@ -518,6 +527,10 @@ def parse_args(args=None):
                     help='Add this Java class as an Activity to the manifest.')
     ap.add_argument('--activity-launch-mode', dest='activity_launch_mode',
                     help='Set the launch mode of the main activity in the manifest.')
+    ap.add_argument('--save-manifest', dest='save_manifest', default='',
+                    help="If set to a path, the created manifest will be saved to this path.")
+    ap.add_argument('--load-manifest', dest='load_manifest', default='',
+                    help="If set to a path, the manifest will be loaded from this path.")
```
inclement: Not sure if the paths will be messed up at this point, since the script is run from a directory that is different to the one the user ran from, so e.g. if they wrote './mymanifest.xml' that path won't make sense any more. I think there is some handling of this issue elsewhere in the p4a arguments.

pythonforandroid/bootstraps/sdl2/build/build.py
tshirtman: You mean make the render even if we overwrite it just after (moving the shutil.copy part after the else branch) ? that seems a bit wasteful, and i'm not sure how it helps with consistency, but i'm fine with it if you prefer.

pythonforandroid/bootstraps/sdl2/build/build.py
tshirtman: that's a good point, we should probably document that it should be an absolute path.

tshirtman:
About making the current way more flexible, i do agree, but there always will be things that will be hard to support without more or less giving a tree like structure instead of just flat options.

tshirtman:
would an xml "merge" be a better way, allowing to add or replace part of the xml file with determined sections, be a better plan? or even, simple diff/patch logic so the user can simply apply their change?
Maybe giving access to the xml tree through a hook system, allowing one to add/remove/update the nodes?
Because exposing all the possible xml options seems like an endless churn to me.

Fak3:
I would love to be able to extend the default template file with arbitrary blocks for my project

Fak3:
Related to #1922 #2200

PR #1440, opened 2018-11-03
[protobuf_cpp] Updated to Protobuf 3.5.1 + fixed protoc issue

goffi-contrib:
Protobuf has been updated to 3.5.1, it's not the last version (3.6.1)
because last version cause trouble during compilation and needs more
work.

This patch also fixes the issue with hardcoded local path for protoc which
could not work. Protobuf needs a version of "protoc" (its compiler, used
to produce .py file from .proto files) compiled for the host platform.
Official binaries for protoc are available, the recipe will download the
one suitable for host and use it.
If the platform is not supported, building may still work if the right "protoc"
is available in search path.

### DIFF
--- a/ci/constants.py
+++ b/ci/constants.py
@@ -44,6 +44,9 @@ BROKEN_RECIPES_PYTHON2 = set([
     'm2crypto',
     'netifaces',
     'Pillow',
+    # requires autoconf system dependency on host
+    # https://api.travis-ci.org/v3/job/450538715/log.txt
+    'protobuf_cpp',
     # https://github.com/kivy/python-for-android/issues/1405
     'psycopg2',
     'pygame',
--- a/pythonforandroid/recipes/protobuf_cpp/__init__.py
+++ b/pythonforandroid/recipes/protobuf_cpp/__init__.py
@@ -1,21 +1,60 @@
 from pythonforandroid.recipe import PythonRecipe
-from pythonforandroid.logger import shprint
+from pythonforandroid.logger import shprint, info_notify
 from pythonforandroid.util import current_directory, shutil
 from os.path import exists, join, dirname
 import sh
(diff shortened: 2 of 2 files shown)


AndreMiras:
Thanks for the PR! Could you fix the linter so the integration test makes it to the end.
```
[1mpep8 runtests: commands[0] | flake8 pythonforandroid/ tests/ ci/[0m
pythonforandroid/recipes/protobuf_cpp/__init__.py:22:5: E303 too many blank lines (2)
pythonforandroid/recipes/protobuf_cpp/__init__.py:41:25: E122 continuation line missing indentation or outdented
```
Keep in mind you can always run `tox` locally to check linting issues upfront.

goffi-contrib:
@AndreMiras done

AndreMiras:
Looking good thank you :+1:  I'm sorry that the CI is still far from being perfect and it currently fails because of a missing system dependency. Until we tacle that, can you simply add protobuf to our ignore list with something like:
```patch
diff --git a/ci/constants.py b/ci/constants.py
index 804ceffa..e7f84aca 100644
--- a/ci/constants.py
+++ b/ci/constants.py
@@ -44,6 +44,9 @@ BROKEN_RECIPES_PYTHON2 = set([
     'm2crypto',
     'netifaces',
     'Pillow',
+    # requires autoconf system dependency on host
+    # https://api.travis-ci.org/v3/job/450538715/log.txt
+    'protobuf_cpp',
     # https://github.com/kivy/python-for-android/issues/1405
     'psycopg2',
     'pygame',
```
Otherwise no worries I can do it since I'll pick up https://github.com/kivy/python-for-android/issues/1454 pretty soon.

goffi-contrib:
@AndreMiras done :)

PR #1442, opened 2018-11-03
[libsodium] Updated to last version (1.0.16)

goffi-contrib:
This patch update libsodium to last available version, and remove the
"--enable-minimal" flag which was restricting features and causing
runtime troubles in other recipes.

### DIFF
--- a/pythonforandroid/recipes/libsodium/__init__.py
+++ b/pythonforandroid/recipes/libsodium/__init__.py
@@ -4,9 +4,10 @@ import sh
 
 
 class LibsodiumRecipe(Recipe):
-    version = '1.0.8'
+    version = '1.0.16'
     url = 'https://github.com/jedisct1/libsodium/releases/download/{version}/libsodium-{version}.tar.gz'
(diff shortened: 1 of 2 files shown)


pythonforandroid/recipes/libsodium/__init__.py
```diff
@@ -17,7 +18,7 @@ def build_arch(self, arch):
         env = self.get_recipe_env(arch)
         with current_directory(self.get_build_dir(arch.arch)):
             bash = sh.Command('bash')
-            shprint(bash, 'configure', '--enable-minimal', '--disable-soname-versions', '--host=arm-linux-androideabi', '--enable-shared', _env=env)
```
AndreMiras: Just because I'm curious, why did we skip the `--enable-minimal` flag?

AndreMiras:
Thanks for the pull request.
I need to try that one manually since the CI skipped it.
```
('recipes modified:', set([u'libsodium']))
('recipes to build:', set([]))
```

PR #1441, opened 2018-11-03
[cffi] Updated recipe to last cffi version (1.11.5)

goffi-contrib:

PR #1443, opened 2018-11-03
[cryptography] Updated to last available version (2.3.1)

goffi-contrib:
This recipe update cryptography to last version.
Note that "pyasn1" is not anymore a dependency and has been replaced by
"asn1crypto".


pythonforandroid/recipes/cryptography/__init__.py
```diff
@@ -4,9 +4,9 @@
 
 class CryptographyRecipe(CompiledComponentsPythonRecipe):
     name = 'cryptography'
-    version = '1.3'
-    url = 'https://github.com/pyca/cryptography/archive/{version}.tar.gz'
-    depends = [('python2', 'python3crystax'), 'openssl', 'idna', 'pyasn1', 'six', 'setuptools', 'enum34', 'ipaddress', 'cffi']
+    version = '2.3.1'
+    url = 'https://files.pythonhosted.org/packages/22/21/233e38f74188db94e8451ef6385754a98f3cad9b59bedf3a8e8b14988be4/cryptography-2.3.1.tar.gz'
```
AndreMiras: Please don't change the `url` string and don't hardcode the version

goffi-contrib: @AndreMiras previous URL is not restored

AndreMiras:
Thanks for the pull request. CI is currently complaining with the following:
```
[1m[31m[ERROR][39m[0m:   virtualenv is needed to install pure-Python modules, but
[1m[31m[ERROR][39m[0m:   virtualenv does not support nesting, and you are running
[1m[31m[ERROR][39m[0m:   python-for-android in one. Please run p4a outside of a
[1m[31m[ERROR][39m[0m:   virtualenv instead.
```
But that should be fixed pretty soon with https://github.com/kivy/python-for-android/pull/1437
Hopefully we can merge it this week, and I'll pick yours after that.

AndreMiras:
@goffi-contrib https://github.com/kivy/python-for-android/pull/1437 just got merged. Do you mind rebasing to see if Travis is now happy about it?

goffi-contrib:
@AndreMiras done :)

AndreMiras:
Damn it, they all rely on `libffi` which requires autoconf.
So it looks like that one can be added to just like in https://github.com/kivy/python-for-android/pull/1441#issuecomment-437568145
I'm sorry this is embarrassing :confused: 
The CI still needs a lot of improvements, but it will come.
It could be easy to just add autoconf install to the Dockerfile, but I'm thinking of a more powerful solution.

AndreMiras:
We actually decided adding the system dependencies to the container https://github.com/kivy/python-for-android/pull/1457/files#diff-3254677a7917c6c01f55212f86c57fbfR51
So I'm going to give your PR a try manually now and if it seems OK, I'll see if I can fix the conflict and merge without disturbing you one more time :smile:

Issue #1445, opened 2018-11-05
lxml recipe doesn't work with Python 3


JonasT:
### Versions

* Python: python3crystax
* OS: ubuntu in docker
* Kivy: not used
* Cython: not relevant

### Description

I saw an lxml recipe was recently merged and I'm very interested in that, but sadly it seems to be limited to python2 for some reason. It would be neat if that could be changed!

I already tried changing the dependencies to `("python2", "python3crystax")` instead, but something seems to hardcode `-lpython2.7` into the linker flags, so it breaks nevertheless.

JonasT:
I made a pull request that takes the hacks out: https://github.com/kivy/python-for-android/pull/1446 This makes it work flawlessly with python3crystax, as far as I can tell. **It might break Python 2,** if any of those hacks were actually workarounds to core problems. In that case, that core problems should probably be fixed...

PR #1446, opened 2018-11-05
Remove python 2-specific hardcoded hacks from lxml recipe

JonasT:
This pull request removes the python 2-specific hacks from the `lxml`-recipe that break Python 3 support. **I expect that this possibly will break the Python 2 lxml recipe build.** However, to fix that, the recipe shouldn't get any Python 2 specific flags hardcoded. Instead, either the core should be fixed if possible, or if that really isn't doable, then at least the Python 2 changes should be added as conditionals.

I therefore suggest that someone with Python 2 interest (I don't have any legacy Python 2 code) checks the state of this, and that these Python 2-specific hardcoded things get removed as in this pull request.

**What was tested:** I applied this patch and successfully tested both lxml install, `python-docx` install (which uses lxml), and generating DOCX documents from scratch directly on Android. Afterwards, I opened the document with the android libreoffice viewer to verify that it looks as expected. *All of this was tested with python3crystax.*

### DIFF
--- a/pythonforandroid/recipes/lxml/__init__.py
+++ b/pythonforandroid/recipes/lxml/__init__.py
@@ -1,12 +1,13 @@
 from pythonforandroid.toolchain import Recipe, shutil
 from pythonforandroid.recipe import CompiledComponentsPythonRecipe
-from os.path import exists, join, dirname
+from os.path import exists, join
+from os import listdir
 
 
 class LXMLRecipe(CompiledComponentsPythonRecipe):
     version = "3.6.0"
     url = "https://pypi.python.org/packages/source/l/lxml/lxml-{version}.tar.gz"
-    depends = ["python2", "libxml2", "libxslt"]
+    depends = [("python2", "python3crystax"), "libxml2", "libxslt"]
     name = "lxml"
 
     call_hostpython_via_targetpython = False  # Due to setuptools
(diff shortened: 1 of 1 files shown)


AndreMiras:
Thanks CI looks happy.
However something weird happened with the CI (https://api.travis-ci.org/v3/job/450779313/log.txt):
```
[INFO]:    Found multiple valid dependency orders:
[INFO]:        ['hostpython2', 'libxml2', 'libxslt', 'python2', u'lxml']
[INFO]:        ['hostpython3crystax', 'libxml2', 'libxslt', 'python3crystax', u'lxml']
[INFO]:    Using the first of these: ['hostpython2', 'libxml2', 'libxslt', 'python2', u'lxml']
[INFO]:    Trying to find a bootstrap that matches the given recipes.
...
[INFO]:    Found 4 acceptable bootstraps: ['webview', 'service_only', 'sdl2_gradle', 'pygame']
[INFO]:    Found a single valid recipe set: ['hostpython2', 'libxml2', 'libxslt', 'python2', u'lxml']
[INFO]:    -> directory context testapps/
('recipes modified:', set([u'lxml']))
('recipes to build:', set([u'lxml']))
falling back to python2
('recipes to build (no broken):', set([u'lxml']))
('requirements:', u'lxml,python2')
```
So well the good news is that it doesn't seem to break the build on Python2 (ping @jerome-poisson).
But the funny thing is the CI seems to get mislead in thinking it's still not python3 compatible. I'm not too sure why.

Pro-tip for next PR, you can add a `fixes #1445` so it gets closed automatically on merge.

PR #1452, opened 2018-11-09
C++ recipe didn't set all instances of `ctx.ndk_dir` in format string

dbrnz:
Closing parentheses were in the wrong place...


AndreMiras:
Good catch, it's looking good to me. Actually we could also skip the `+` inside the brackets.
These kinda bugs you just fixed could have been avoided if we had recipe unit testing. This will come very soon.
I'll let Travis finish building and merge.

AndreMiras:
Merged, thanks!

PR #1453, opened 2018-11-09
Show actual values in info messages

dbrnz:


AndreMiras:
I agree this is a good improvement, but I worry this could conflict with the soon to be merged https://github.com/kivy/python-for-android/pull/1412
So I prefer to put on hold before merging if you don't mind.

AndreMiras:
Hi @dbrnz #1412 just got merged so you can rebase your changes on it.

Also just wondering if we could simplify the "wording" further by replacing:
```python
info('Found Android API target in $ANDROIDAPI: it is {}'.format(android_api))
```
By directly:
```python
info('Found Android API target in $ANDROIDAPI: {}'.format(android_api))
```
So skipping the "`it is`" part in each case. And we can also probably skip the line break (if tox doesn't mind).
It's just a suggestion, but what do you think?

dbrnz:
Hmm, my attempt at rebasing has pulled in all of #1412's patches... I'm closing this PR and will create a new one that has just my stuff. And yes to removing the "it is" (another similar message had "it is" so copied the wording...).

PR #1456, opened 2018-11-10
Added python3 testapp build to Travis matrix

inclement:
I omitted this from the original PR because the Travis build failed at pyjnius for some reason. Trying again here to debug it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,7 +44,7 @@ RUN apt update -qq && apt install -qq --yes --no-install-recommends \
 RUN dpkg --add-architecture i386 && apt update -qq && apt install -qq --yes --no-install-recommends \
     build-essential ccache git libncurses5:i386 libstdc++6:i386 libgtk2.0-0:i386 \
     libpangox-1.0-0:i386 libpangoxft-1.0-0:i386 libidn11:i386 python2.7 python2.7-dev \
-    openjdk-8-jdk unzip zlib1g-dev zlib1g:i386
+    openjdk-8-jdk zip unzip zlib1g-dev zlib1g:i386
 
 # specific recipes dependencies (e.g. libffi requires autoreconf binary)
 RUN apt install -qq --yes --no-install-recommends \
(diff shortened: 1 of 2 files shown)


inclement:
Okay, so the python3 recipe has errors like:

          working: jnius/jnius.c:46896:11: error: 'PyThreadState' has no member named 'exc_type'  
           working:      tstate->exc_type = local_type;                                            
           working:            ^                                                                   
           working: jnius/jnius.c:46897:11: error: 'PyThreadState' has no member named 'exc_value' 
           working:      tstate->exc_value = local_value;                                          

([full log](https://api.travis-ci.org/v3/job/453319180/log.txt))

Interestingly, the Python 2 PyThreadState does have these members, which makes me suspect a cython version problem. @tito I remember you had problems using cython2 with the python3 recipe, did you get errors like this?

inclement:
https://github.com/cython/cython/issues/1978 <- confirmed the issue is cython version related. I'm not clear if just upgrading the cython version will work, but we'll see.

AndreMiras:
Great catch. I'm actually updating cython to 0.28.6 in an upcoming pull request https://github.com/kivy/python-for-android/pull/1457

inclement:
Testing 0.29 here because that's what I have working locally. If 0.28.6 works too then that would obviously also be fine, but I found some indications that at least some versions of 0.28 still had this problem.

AndreMiras:
By the way keep in mind you can always run your tests directly on the Docker environment to make debugging faster than via Travis.
So basically to play with this you can:

1. build the image: `docker build --tag=p4a .`
2. run it in an interactive shell `docker run -it --rm p4a`
3. prepare your env from this shell e.g. `pip install cython==0.29`
4. and play with p4a just like we have it in Travis e.g. `python setup_testapp_python3.py apk...`

inclement:
Thanks, I was going to ask how to do that, although I was also hoping that upgrading cython would just work and leave me to look at more interesting things.

inclement:
The python3 build worked, so I squashed the experimental commits and will let the tests pass again (as they all should now). When that happens, this is good to merge.

Dockerfile
```diff
@@ -93,7 +93,7 @@ RUN useradd --create-home --shell /bin/bash ${USER}
 # with sudo access and no password
 RUN usermod -append --groups sudo ${USER}
 RUN echo "%sudo ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
-RUN pip install --quiet --upgrade Cython==0.28.6
+RUN pip install --quiet --upgrade cython==0.28.6
```
AndreMiras: Minor, why changing the case, it just create "noise" in the history?

inclement: I didn't, you did, and I rebased on your changes :p

AndreMiras: https://github.com/kivy/python-for-android/commit/19d2de09f3bd42e910ef631c6d111f58507dd4fa#diff-3254677a7917c6c01f55212f86c57fbfR96
In my changes it's with a capital plus the version change. In your changes you're only updating the case.
It's minor anyway, but I think it could be avoided

Issue #1455, opened 2018-11-10
Python 3 recipe follow up issues


inclement:
The new Python 3 recipe (#1412) is in a fully working state, but there are a few important follow up issues to work through now:

- [x] Make python3 recipe use p4a's arch system as much as possible, including making p4a understand clang toolchains (and ideally at the same time try using clang for other recipes)
- [x] openssl recipe compatibilty
- [x] sqlite recipe compatibility
- [x] Fully optimize the zipped distribution size by removing unnecessary things (and making this customisable?)
- [x] Make the python3 build include ctypes, not sure why it doesn't build right now as it did in bpo-30386 #1465
- [x] Get a python3 CI test working #1456

pmp-p:
>     * [ ]  Make the python3 build include ctypes, not sure why it doesn't build right now as it did in bpo-30386

setup.py won't always detect libffi of target when cross compiling, but now python3.7 - unlike the bpo - *need* a system ffi to build ctypes.

cf: https://bugs.python.org/issue31710

pmp-p:
maybe a bug, i found that when using a python host 3.7.1+ ( some distributions have started shipping it ) 
the jinja2 from ./pythonforandroid/bootstraps/pygame/build/buildlib/jinja2.egg is seriously updated ( 3 errors due to py3 )

inclement:
@pmp-p Thanks for the note, the pygame bootstrap is deprecated and has many issues like this, we may look to just remove it soon.

inclement:
Follow up issues essentially complete thanks to various contributors :)

PR #1460, opened 2018-11-12
[WIP] Refactor python recipes and update python2's to version 2.7.15

opacam:
This pr is intended to complement the amazing work of @inclement in his python3's recipe #1412. It refactors the python3's recipe into the recipe module and updates the python2 version to 2.7.15 using the same build process than the refactored python3's recipe. So, python2 and python3 will share the same build system and allow us to remove some specific source code for python2. Right now the build process completed successfully for both python versions and the python3 recipe also works as expected when installed on android devices...so...

Things to do:

  - [x] Fix python2's initialization on android device
  - [x] Make neatest the python3's refactoring
  - [x] Update openssl libraries to grant support for python2 and python3 (trying to not lose support for python2legacy)
  - [x] Enable libraries support (for both python2 and python3)
  - [x] Review the patches
  - [x] Test that python3crystax is not broken
  - [x] Fix some hardcoded python2 entries
  - [x] Fix webview's bootstrap (for both versions of python)
  - [x] Fix service_only's bootstrap (for both versions of python)
  - [x] Fix python2/python3 compatibility for most of the recipes that uses the new openssl lib (ffmpeg and LibSecp256k1 not faced here and are broken...)
  - [x] Enhance library loading mechanism on android side (no more harcoded libs in PythonUtil.java)
  - [x] Add tests apps to ci for webview and service_only bootstraps
  - [x] Add tests apps to ci for most of the encryption recipes (openssl dependant)
  - [x] Reduce travis log output, to not face the travis 4mb limit
  - [ ] Check/fix if image related recipes are working (i dont think so): libjpeg, libpng and pillow
  - [ ] Fix pygame using the old python recipe or remove it (it could be done in another pr)
  - [ ] Properly document the changes applied in this pr


pythonforandroid/recipes/sdl2/__init__.py
```diff
@@ -24,12 +24,12 @@ def get_recipe_env(self, arch=None):
         env['PYTHON_INCLUDE_ROOT'] = self.ctx.python_recipe.include_root(arch.arch)
         env['PYTHON_LINK_ROOT'] = self.ctx.python_recipe.link_root(arch.arch)
 
-        if 'python2' in self.ctx.recipe_build_order:
-            env['EXTRA_LDLIBS'] = ' -lpython2.7'
-
-        if 'python3' in self.ctx.recipe_build_order:
-            env['EXTRA_LDLIBS'] = ' -lpython{}m'.format(
+        if 'python2' in self.ctx.recipe_build_order or \
+                'python3' in self.ctx.recipe_build_order:
```
AndreMiras: I know people would probably not like it, but I have to give it a try :joy: 
```python
if set('python2', 'python3') & set(self.ctx.recipe_build_order):
```

pythonforandroid/recipe.py
```diff
@@ -965,20 +946,13 @@ def build_cython_components(self, arch):
...
-                info('Stripping object files')
-                shprint(sh.find, '.', '-iname', '*.so', '-exec',
-                        '/usr/bin/echo', '{}', ';', _env=env)
-                shprint(sh.find, '.', '-iname', '*.so', '-exec',
-                        env['STRIP'].split(' ')[0], '--strip-unneeded',
-                        # '/usr/bin/strip', '--strip-unneeded',
-                        '{}', ';', _env=env)
+            info('Stripping object files')
+            shprint(sh.find, '.', '-iname', '*.so', '-exec',
+                    '/usr/bin/echo', '{}', ';', _env=env)
```
AndreMiras: Looks like it was already like that, but now I'm getting curious why this is working, since:
```
ls /usr/bin/echo
ls: cannot access '/usr/bin/echo': No such file or directory
```

pythonforandroid/recipe.py
```diff
@@ -1037,17 +1011,147 @@ def get_recipe_env(self, arch, with_flags_in_cc=True):
...
+        'test',
+        'tests',
+        'lib2to3',
+        'ensurepip',
+        'idlelib',
+        'tkinter',
+    }
+    '''The directories that we want to omit for our python bundle'''
+
+    stdlib_filen_blacklist = [
```
AndreMiras: Why not also having it as a set? I actually liked the idea

pythonforandroid/recipe.py
```diff
@@ -1056,12 +1160,14 @@ def prebuild_arch(self, arch):
             exit(1)
         self.ctx.python_recipe = self
 
-    def include_root(self, arch):
+    def include_root(self, arch_name):
         '''The root directory from which to include headers.'''
-        raise NotImplementedError('Not implemented in TargetPythonRecipe')
+        return join(self.get_build_dir(arch_name),
+                    'Include')
```
AndreMiras: Minor nitpick: here and other place, it looks like some very agressive line length, isn't it?

pythonforandroid/recipe.py
```diff
@@ -1090,6 +1244,115 @@ def reduce_object_file_names(self, dirn):
...
+        recipe_build_dir = self.get_build_dir(arch.arch)
+
+        # Create a subdirectory to actually perform the build
+        build_dir = join(recipe_build_dir, self.build_subdir)
+        ensure_dir(build_dir)
+
+        if not exists(join(build_dir, 'python')):
+            with current_directory(recipe_build_dir):
+                # Configure the build
+                with current_directory(build_dir):
```
AndreMiras: I'm curious to know if we can we skip the `with current_directory(recipe_build_dir)` above and just use this one?
Unless or maybe we're dealing with relative directories.
If it's the case and both are still needed, maybe you want to give a try to:
```python
with current_directory(recipe_build_dir), current_directory(build_dir):
```
So you don't have to indent two levels.

AndreMiras:
I appreciate all the effort, specially since I would not have the energy to maintain python2 related things :smile: 
@inclement is the one that knows the core components the most now, so I let him deal with it :running_man:

pythonforandroid/bootstraps/sdl2/build/build.py
```diff
@@ -126,7 +127,8 @@ def make_python_zip():
 
     if not exists('private'):
         print('No compiled python is present to zip, skipping.')
-        print('this should only be the case if you are using the CrystaX python or python3')
+        print('this should only be the case if you are using'
+              'the CrystaX python, python2 or python3')
```
inclement: If we were to merge this, we could just ditch this message.

pythonforandroid/recipes/sdl2/__init__.py
inclement: If you want to write it neatly, `if self.ctx.python_recipe.name in ('python2', 'python3')` seems nicer. It seems a minor point though.

opacam:
Ohh guys both of you are right with all the comments and reviews (I will apply it in some specific commit)  but @inclement touch the point, the python initialization should be solved otherwise all of this is useless....so...I've been working on this point and I fixed the issue.

About the "python legacy" recipe, already thought about that, in matter fact my first test was in that direction, but after a coding a while I realised that we will complicate things a lot and we will end up with a duplicated python2 and moreover, if the pygame bootstrap can be run in an updated python recipe (I'm not sure about that but I suspect that this can be achieved with some minor changes and building with an api minor than 21), what is the purpose of maintain the python legacy recipe? We will end up with a lot more of code and  harder to maintain...anyway...I will adapt to whatever @inclement decide, no problem.

pythonforandroid/recipe.py
tshirtman: I'm not sure how it would work, but i think the -ls option of find is what one would want to use.

opacam: I've been testing  this from terminal, and it fails on the echo stage, because, As @AndreMiras pointed,  the echo command is not located at /usr/bin/echo, actually should be /bin/echo or echo...but even changing this  the result is exactly the same...so I suspect that the combination of this two commands cause the issue...so the @tshirtman's proposal maybe a  solution...can you give me any clue about of how to apply the find command with -ls, filtering  by  extension ".so" and printing line by line?

inclement: I suppose this line just wasn't doing anything on systems where the path to echo is wrong. `sh` doesn't actually raise an error if this happens, it would just do nothing.

inclement: Also, you can just take out the -exec stuff, you can still extract a list of files from the result (good enough for printing). Or just remove the print entirely, it isn't really important.

opacam:
Update:

I already have some of the points mentioned in the first post done in my local machine:

 - Reintroduce old python recipe as a python2legacy recipe (following @inclement's guidelines)
 - Fix pygame's bootstrap 

I did not pushed them because I prefer to review it  and create the apropiate commits, I will try to do it as soon as possible (I plan to do it  today night or tomorrow).

The openssl update should not be a problem, the job has been already done by @tito while helping with @inclement's python3 recipe (the only thing to do here should be some kind of mechanism that allow us to build the old openssl libs for python2legacy). 

It must be mentioned that the new python recipe has the same problem than the python3 recipe, ctypes is not working (as the python3 recipe, see pr #1455), but I thing that I got the solution for that...I suspect that we must add our libffi's recipe into the optional dependencies in order to build the ctypes module...¿what do you think about this ctypes solution @inclement?

pythonforandroid/recipe.py
opacam: In the next commit, I was planning to restore this lines at his previous state (to introduce python2legacy recipe) and remove entirely this print statement as you suggested, but now, I'm thinking to go a little further...perhaps we can create a new method  like **strip_cython_files** into TargetPythonRecipe (as you do with **create_python_bundle**)?

pythonforandroid/recipe.py
opacam: Because I suspect that @inclement did it on purpose, this way the data types for the "file extensions blacklist" match, and notice that one of this lists is empty and, as far as I know we cannot initialize an empty set with {}, we should use set(), I think that is more elegant the way it is  :wink:

inclement:
>  but I thing that I got the solution for that...I suspect that we must add our libffi's recipe into the optional dependencies in order to build the ctypes module

I'm aware of this, but hadn't got around to checking what the python2 and crystax python recipes do instead. It looks like both of these include mechanisms to build libffi as part of the Python build. In the case of python2, it's partially related to the patching we do, perhaps some small change would enable that in your current build.

Overall though, making the libffi recipe a dependency would be fine, it looks like this would be an appropriate way for p4a to manage things.

Beyond this, my main concern with this PR is that it *will* break things for people, probably in unexpected ways such as finding that other recipes are no longer compatible, or that people depended on biglinking in some weird way (this may also effectively raise the lowest supported API number, although we can probably live with just no longer 'supporting' the very old ones). It's fine for the python3 recipe because it's new, but the existing python2 build has been unchanged for years.

I don't think this should ultimately prevent this being merged or the python2 recipe being replaced, but I think it will require some planning to smooth the transition (e.g. documenting clearly that there is a legacypython recipe).

inclement:
Of course we could also take the opposite tack and just officially drop the old python2 method much sooner than expected. Maybe we should! But it requires some thought.

AndreMiras:
Thanks @inclement for considering all these uses cases that I may have overlooked.
I'm always for improving the code base and maintainability and I don't think we should keep holding ourselves because of compatibility to something legacy. We have tags anyway so people who don't want to move on or can't move on can still run on `python-for-android==0.6.0`.
I really think trying to keep track with legacy things for retro compatibility is slowly us down too much and sucks our energy a lot.
Amazing contribution @opacam by the way :muscle:

opacam:
I see that there is some doubts (I also have some of the concerns that @inclement mention) I am delaying the push of the python2 legacy recipe...yesterday I pushed some branch with the minimal required changes to introduce the original python recipe by @tito as a python2legacy to help us to decide if it's worth it to reintroduce it. This is only to get a global idea of what we have to touch and how it will looks like, and also if anyone want to test it .

Some notes about his possible python2legacy recipe:

  - I must say that the pygame bootstrap will partially work, I meant, it compiles and runs fine on android device but crash on pyjnius vibration test (pygame uses "renpy" instead of "kivy").
   - It will require an old android ndk (tested on r13b), and it must be before the "unified headers" feature (r15b I think)...and this should be properly documented
  - Also it will require to modify all python2 compatible recipes one by one to add compatibility to python2legacy
  - The python2legacy recipe is almost the same than the current p4a's master branch python2 recipe, except that was cleaned up a little (split long lines, fixed imports and added a new method **create_python_install** which behaves similar than create_python_bundle with the difference than this method will only be for python2legacy and triggered whenever necessary)

Regarding deprecating all python2 versions as suggested by @inclement, if it was my call, i will not do it, at least until the python team oficially deprecates it (2020), but guys, it's your decision, and whatever is your decision I'm sure that will be the right one and I will adapt to it, no problem, even if you want that I close this pr. I think than the updated python2recipe make easier to maintain both python versions, thanks to the great work of @inclement, but probably, things will stop work for python2 over the time (then I will start deprecate things for python2).

If the team decide that this pr it's better without the python2legacy, then we probably must start deprecating (or directly removing) the pygame related stuff alongside with the old sdl libs.

Ok, saying all that, I will continue to fix all stuff without the python2legacy (for now), the code is made and we could apply whenever we want...or not...just tell me...but if we introduce that, I'm afraid that then, what mentioned @AndreMiras about slowing down the development it will happen (In matter fact it does now  :joy: :joy:)

My plan is to work with this pr during this weekend and I will  try to be connected to discord from afternon on.

opacam:
> Overall though, making the libffi recipe a dependency would be fine, it looks like this would be an appropriate way for p4a to manage things. Thanks for looking into it. If you're thinking to try implementing this, how about even making it a separate PR for the python3...

Sorry @inclement  :pray:,  I forgot about this, I'll try to create a new pr for this, as you propose, today.

hackalog:
My humble opinion, but I thought I'd wade in: @inclement re: dropping the python2 method, I would definitely hate that. We've been holding off moving our kivy code to python3 until kivy-ios supported it, as we need our apps to support both android and ios targets, but migrating our code isn't going to be a trivial process. A modern 2.7 (such as 2.7.15) would be an excellent place to stabilize things before making the move to python 3.

That said, I believe python 2 has to be deprecated by 2020. It just shouldn't do so until python3 is a first-class citizen across the supported kivy targets.

opacam:
Those last commits grant libraries support for the python recipes (sqlite3, openssl and libffi). So far works well, at least with the test apps, the modules which depends on the optional libraries are getting build for both version of python, so I think that the core is good, and we are ready to move to the other remaining points mentioned at the first post: recipes/bootstraps compatibility and python2 harcoded entries.

Would like to mention that two of this last commits, the clang feature (c7edf85) and the update of the openssl libs (71e855a) are not entirely my work, those are based on some patch created by @tito while he was helping with the @inclement's python3 recipe...so...all the credit goes to @tito to solve both problems :satisfied:

PD: If @tito wants to introduce those features himself, no problem, I will solve the merge conflicts when merged to the master branch :wink:

opacam:
Ei guys, I Rewroted the history to be more readable and to fix some nasty conflicts appeared due to last changes in master branch and I will push it today. I squashed some commits of the start of the history (includes the fixes/enhancements courtesy of the reviewers). Also I split some of the last commits pushed (the ones related with non sdl2 bootstraps) because those were the source of the conflict, difficult to review...too many changes at once...now I think that it is more readable and easy to solve possible future conflicts with the master branch.

Things to be  mentioned:

 - I remove one step from the previous history. The refactor of the python3 recipe was done in two steps: first I moved to recipe module and then was refactored again into python module. This make difficult to track the code, so the new history avoid the first step.
 - The previous history directly updates the python2 recipe to a new version. Now I move the pithon2 recipe into python2legacy and then I create the new one so we can keep the git history more clear when we decide if this python2legacy recipe it's worth it to maintain or remove it. Keep in mind that in the current source code state the python2legacy recipe is unusable.
- If you want to know what commits changed respect the last history, you should see the commit date. All the commit older than 4th of November  were "cherry picked" from a backup branch  created just before the history's rewrote.
- All this history rework were done building and testing the corresponding apps
- Please, stay tuned to all my C/Java code, I don't usually write in those languages, I understand it and I can do simple things but are not my area

JonasT:
I see you touched e.g. some of the renpy files, like `AssetExtract.java`. Would it make sense to move these into the `common` bootstrap folder maybe? Or are there differences with these as well between the bootstraps?

(Please note you can make the whole `src/main/java/org/renpy/...` folder structure in both e.g. `webview` and `common`, and have some files in `common` and some in `webview` - they'll get merged automatically on build. So there's no need to move *everything* into the common area if it doesn't make sense for some files as they are right now)

JonasT:
But just to add: if in doubt, this could also be done later. This looks like we want to merge it *really fast*, because it changes so many things. E.g. my SDL2 pull request looks like it might already cause new conflicts (which is fine, I'll resolve them, but it makes sense to get this in quickly before other stuff like my large amount of requests breaks it again)

pythonforandroid/bootstraps/webview/build/src/main/java/org/kivy/android/PythonActivity.java
```diff
@@ -87,13 +95,20 @@ protected void onCreate(Bundle savedInstanceState) {
         resourceManager = new ResourceManager(this);
 
         Log.v(TAG, "Ready to unpack");
-        unpackData("private", getFilesDir());
+        File app_root_file = new File(getAppRoot());
+        unpackData("private", app_root_file);
 
-        this.mActivity = this;
+        Log.v(TAG, "About to do super onCreate");
```
JonasT: this looks familiar, isn't this change on master already? https://github.com/kivy/python-for-android/blob/master/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonActivity.java#L66 
is this pull request rebased, or is master merged into it somehow?

it would be quite good if we could see what is actually new, so if the history could be cleaned up that would be very useful

JonasT:
One additional thing I noticed: you added many more tests to `.travis.yml` - but it would make more sense to put them into some sort of `.travis.yml.disabled` and then see which ones we want to prioritize, because we're already at the limit of concurrent tests right now, so this change will *double* the build to take half an hour for any change tested by travis... that would be a bit unfortunate

opacam:
> I see you touched e.g. some of the renpy files, like AssetExtract.java. Would it make sense to move these into the common bootstrap folder maybe?

@JonasT , yes, it should.
The thing was that both of us were working on the same files at the same time (but I did not refactor all the stuff as you did, your solution is far better), so I moved on, to work in other stuff and I only fixed the merge conflicts when the changes that you made were merged into master branch...so...it need more work...and probably you can do better than me (I don't usually code in c or java).

opacam:
> Could you please checkout into the `master` branch, pull the changes from `kivy/python-for-android` repo to your local `master` and then...

@KeyWeeUsr, all the last commits I pushed are the whole history of this pr rebased on the top of the latest changes on p4a master branch, applying the commands that you mentioned...so this is already done...no need to do it again...but...keep in mind that at some point I rewrote the original history (I mentioned that in some comment before)

JonasT:
@opacam you should be able to `git diff` what you have right now against master I think (pull it on a separate branch, and `git diff <thatbranch>` should give the differences). I think if you output that diff into a patch file and then apply it on master and make a new commit/pull request from that, we might get a pull request that cleanly shows what you actually changed. Would that be possible for you to do?

(I assume rebasing given the convoluted merges might no longer easily be possible, but if it is, that would of course be even better)

JonasT:
@opacam `...no need to do it again...` the reason we ask for it because this is incorporating merge commits, and they mess up what is actually changed in the pull request - as you can see by the section I quoted above showing a change that isn't actually part of the pull request but already on master. This is highly confusing and makes it really difficult for us to see what you actually did.

What you could also try is squash all your commits after latest master to one, and then rebase that on latest master and see what that gives us. (make sure to make a safety copy of your repo before you do that!!!) but whatever we have right now, it's not a clean rebase

opacam:
Ok guys you are right, I squashed all the changes made in this pr until now into the new pr. No changes about the tests as @JonasT suggested, for now...I was planning to reduce the number of tests but I wanted that all of you see that all the builds success...we have to make sure that this works before merge or it will break the whole project...

If you are in hurry to merge this, do it, still things to do, but most of the work it's done, plus the remaining points could be done apart and maybe it would be even better, so anyone can contribute. Moreover, this touches a lot of files and each time is harder to mantain, to review..so maybe it's time to merge...this is your decision guys  :wink:

opacam:
Moving this to a new pr #1534...because of the mess in the history.

PR #1462, opened 2018-11-16
Fixes P4A_FULL_DEBUG that is broken

tito:


pythonforandroid/logger.py
```diff
@@ -148,8 +148,10 @@ def shprint(command, *args, **kwargs):
     kwargs["_bg"] = True
     is_critical = kwargs.pop('_critical', False)
     tail_n = kwargs.pop('_tail', None)
+    full_debug = False
     if "P4A_FULL_DEBUG" in os.environ:
```
AndreMiras: Minor suggestion: we could also write a one-liner that is still very readable:
```python
full_debug = "P4A_FULL_DEBUG" in os.environ
```

tito: Look below, it also have an impact on tail_n

Issue #1454, opened 2018-11-09
Conditional builds skip conflicting dependencies


AndreMiras:
In the current version of [rebuild_updated_recipes.py](https://github.com/kivy/python-for-android/blob/3e561b0/ci/rebuild_updated_recipes.py), if two conflicting recipes got updated, the build would fail.
One way to work around it is to check it upfront and completely skip the compilation on conflicts.
Another way would be to compile deps isolated (one by one) on conflicts.
While the first way is pretty straight forward to implement, it means we won't have any coverage for conflicting recipes.

PR #1463, opened 2018-11-17
Add support for ctypes to python3's recipe

opacam:
Should be mentioned that the current test app for python3 has been modified by adding libffi to the requirements because the ui for the app has some button to test the ctypes module.

References: This commit partially fixes some of the points mentioned in issue number #1455

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonUtil.java
+++ b/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonUtil.java
@@ -37,6 +37,7 @@ public class PythonUtil {
         ArrayList<String> libsList = new ArrayList<String>();
         addLibraryIfExists(libsList, "crystax", libsDir);
         addLibraryIfExists(libsList, "sqlite3", libsDir);
+        addLibraryIfExists(libsList, "ffi", libsDir);
         libsList.add("SDL2");
         libsList.add("SDL2_image");
         libsList.add("SDL2_mixer");
--- a/pythonforandroid/recipes/python3/__init__.py
+++ b/pythonforandroid/recipes/python3/__init__.py
@@ -1,4 +1,4 @@
-from pythonforandroid.recipe import TargetPythonRecipe
+from pythonforandroid.recipe import TargetPythonRecipe, Recipe
 from pythonforandroid.toolchain import shprint, current_directory
 from pythonforandroid.logger import logger, info
 from pythonforandroid.util import ensure_dir, walk_valid_filens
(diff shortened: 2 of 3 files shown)


opacam:
> I presume you've tested the final APK on your device? :)

Yes  :wink:

KeyWeeUsr:
Ah, so it does brag after all :smile: https://travis-ci.org/kivy/python-for-android/jobs/456426763#L526
Please remove that single space, so that the build is green :+1:

opacam:
One of the tests is failing, I suspect that affected by issue #1454

AndreMiras:
Thanks!
Yep indeed for the failing test, you're right we can ignore that one. It's not exactly what's described #1454 but yes I've created https://github.com/kivy/python-for-android/pull/1464 for it.

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -33,13 +33,53 @@
...
+        '--without-ensurepip',
+        'ac_cv_little_endian_double=yes',
+        '--prefix={prefix}',
+        '--exec-prefix={exec_prefix}')
+
+    def set_libs_flags(self, env, arch):
+        '''Takes care to properly link libraries with python depending on our
+        requirements and the attribute :attr:`opt_depends`.
+        '''
+        def add_flags(include_flags, link_flags):
```
inclement: Since you only use this once it doesn't really need to be a function.

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -33,13 +33,53 @@
...
+
+        if 'libffi' in self.ctx.recipe_build_order:
+            info('Activating flags for libffi')
+            recipe = Recipe.get_recipe('libffi', self.ctx)
+            include = ' -I' + ' -I'.join(recipe.get_include_dirs(arch))
+            ldflag = ' -L' + join(recipe.get_build_dir(arch.arch),
+                                  recipe.get_host(arch), '.libs') + ' -lffi'
+            add_flags(include, ldflag)
+            # libffi needs some extra configurations
+            env['LIBFFI_CFLAGS'] = env.get('CFLAGS', '') + include
```
inclement: I'm not clear where these environment variables get used?

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -33,13 +33,53 @@
...
+            env['LIBFFI_LIBS'] = ldflag
+            self.configure_args += ('--with-system-ffi',)
```
inclement: I think `--with-system-ffi` should be unnecessary on linux, and on osx it seems cpython adds `/usr/include/ffi` to the include path automatically. Have you tried on osx, and with omitting this argument, I'm not sure if I've interpreted it correctly?

It doesn't really matter too much, just wanting to understand what's going on.

pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonUtil.java
```diff
@@ -37,6 +37,7 @@ protected static void addLibraryIfExists(ArrayList<String> libsList, String patt
         ArrayList<String> libsList = new ArrayList<String>();
         addLibraryIfExists(libsList, "crystax", libsDir);
         addLibraryIfExists(libsList, "sqlite3", libsDir);
+        addLibraryIfExists(libsList, "ffi", libsDir);
```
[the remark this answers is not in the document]
opacam: I tested without these line and dynamic access seems to work, thanks! :)

pythonforandroid/recipes/python3/__init__.py
opacam: You are right, I will fix it

pythonforandroid/recipes/python3/__init__.py
opacam: I tested without it and they are not needed, I've been using those variables for a while in python2 (when cross compiling)...you see...I still am attached to the old python recipe habits...thanks again :smile:

pythonforandroid/recipes/python3/__init__.py
opacam: Again, You are right, so I will revert all the changes affecting the attribute configure_args, there is no need to touch it. So many thanks for your clinical eyes @inclement with all the changes this becomes a lot more cleaner  :blush:

PR #1467, opened 2018-11-18
Add instructions for using Docker image for building APKs

KeyWeeUsr:
This way (including the `gradlew` fix) I built a runnable app on Windows 8.1 (in the old Docker for Windows with VirtualBox backend).

### DIFF
--- a/doc/source/docker.rst
+++ b/doc/source/docker.rst
@@ -1,0 +1,69 @@
+.. _docker:
+
+Docker
+======
+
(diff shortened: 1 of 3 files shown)


AndreMiras:
We can ignore the conditional build failing. Seems like a glitch:
```
('recipes modified:', set([u'python3']))
('recipes to build:', set([u'python3']))
falling back to python2
('recipes to build (no broken):', set([u'python3']))
('requirements:', u'python3,python2')
```

PR #1468, opened 2018-11-18
Replaced many `exit(1)`s with exception raising

inclement:
This both makes the code neater and will make it easier to test things.

### DIFF
--- a/pythonforandroid/distribution.py
+++ b/pythonforandroid/distribution.py
@@ -2,9 +2,8 @@ from os.path import exists, join
 import glob
 import json
 
-from pythonforandroid.logger import (info, info_notify, warning,
-                                     Err_Style, Err_Fore, error)
-from pythonforandroid.util import current_directory
+from pythonforandroid.logger import (info, info_notify, Err_Style, Err_Fore)
+from pythonforandroid.util import current_directory, BuildInterruptingException
 from shutil import rmtree
 
 
@@ -131,6 +130,7 @@ class Distribution(object):
         # If there was a name match but we didn't already choose it,
         # then the existing dist is incompatible with the requested
         # configuration and the build cannot continue
+<<<<<<< HEAD
         if name_match_dist is not None and not allow_replace_dist:
             error('Asked for dist with name {name} with recipes ({req_recipes}) and '
                   'NDK API {req_ndk_api}, but a dist '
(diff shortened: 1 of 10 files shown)


AndreMiras:
I was wondering if `test_invalid_recipe_order_and_bootstrap` would pass, but indeed it didn't :confused:

pythonforandroid/toolchain.py
```diff
@@ -58,8 +59,7 @@ def check_python_dependencies():
                 ok = False
 
     if not ok:
-        print('python-for-android is exiting due to the errors above.')
-        exit(1)
+        print('python-for-android is exiting due to the errors logged above')
```
JonasT: Should there be an error raised here? Or is this change intended?

inclement: I originally did, but actually this code is run before most of the build process (because it has to run before the imports) so I think it needs a different treatment.

inclement:
@AndreMiras Just rebased this to fix merge conflicts, any objections to merging it (if the tests pass)?

JonasT:
Just to barge in, I have been eyeing this for a while and I think it's a really good code smell fix. Haven't commented so far since I haven't had the chance to test if it actually doesn't break stuff, but it looks like a thing really worth improving

pythonforandroid/toolchain.py
JonasT: well but currently, this seems to be a big change in behavior since the code flow just seems to resume. Is that intended? Maybe the `exit` should be left in if a runtime error doesn't make sense here? Or will this problem of `not ok` be evaluated again & properly dealt with later?

inclement: I also added the exit(1) back in with the previous comment, I missed that I'd accidentally removed it before so thanks for pointing it out!

JonasT: Ah right, the removing line is gone now. Didn't see that :laughing:

AndreMiras:
Hi @inclement can you give me write access to your `improve_fault_handling` branch so I can help to finish with this one?
We're almost there, I would be a pity to wait for that PR to conflict with other changes

pythonforandroid/distribution.py
```diff
@@ -129,6 +128,7 @@ def get_distribution(cls, ctx, name=None, recipes=[],
         # If there was a name match but we didn't already choose it,
         # then the existing dist is incompatible with the requested
         # configuration and the build cannot continue
+<<<<<<< HEAD
```
AndreMiras: Looks like something went wrong during the rebasing :confused: 
Apparently I have write access to the branch so I'll pick this up

AndreMiras: Fixed

PR #1469, opened 2018-11-18
Allow dist overwrite by default

inclement:
My python3 change fixed a bug where dists could be overwritten if a previous dist with the same name but different recipes was present. This wasn't the original intention of the code. However, it looks like the old buggy behaviour was more desirable in practice, so this PR allows dist overwriting by default but adds an option to configure it.

### DIFF
--- a/pythonforandroid/distribution.py
+++ b/pythonforandroid/distribution.py
@@ -5,6 +5,7 @@ import json
 from pythonforandroid.logger import (info, info_notify, warning,
                                      Err_Style, Err_Fore, error)
 from pythonforandroid.util import current_directory
+from shutil import rmtree
 
 
 class Distribution(object):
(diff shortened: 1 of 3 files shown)


inclement:
@tito This should fix your dist replacement issue

AndreMiras:
looking good to me, if you could just fix the linter :grimacing: 
```
pythonforandroid/distribution.py:8:1: F811 redefinition of unused 'exists' from line 1
```
Use `tox` locally to verify if you don't want to wait for Travis each time.

pythonforandroid/distribution.py
```diff
@@ -5,6 +5,8 @@
 from pythonforandroid.logger import (info, info_notify, warning,
                                      Err_Style, Err_Fore, error)
 from pythonforandroid.util import current_directory
+from os.path import exists
```
AndreMiras: linter for the win :grimacing: 
`tox`

PR #1471, opened 2018-11-18
Remove crystax docker and optimize Dockerfile

KeyWeeUsr:
Prioritize SDK and NDK downloads before everything because once downloaded we don't care about redownloading unless the version is changed and layer cache works this way more efficiently. Redirect Licenses to `/dev/null`, move more important layers up. Move `&&` operators to the front - Docker sometimes, (personally I've encountered it only with `apt` and old `coreutils`), is very critical of the amount of spaces *after* the operator but does not care what's in front of it that much.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,22 +16,16 @@ ENV WORK_DIR="${HOME_DIR}" \
     PATH="${HOME_DIR}/.local/bin:${PATH}"
 # get the latest version from https://developer.android.com/ndk/downloads/index.html
 ENV ANDROID_NDK_VERSION="16b"
-# get the latest version from https://www.crystax.net/en/download
-ENV CRYSTAX_NDK_VERSION="10.3.2"
 # get the latest version from https://developer.android.com/studio/index.html
 ENV ANDROID_SDK_TOOLS_VERSION="3859397"
 
 ENV ANDROID_HOME="/opt/android"
 ENV ANDROID_NDK_HOME="${ANDROID_HOME}/android-ndk" \
-    CRYSTAX_NDK_HOME="${ANDROID_HOME}/crystax-ndk" \
     ANDROID_SDK_HOME="${ANDROID_HOME}/android-sdk"
 ENV ANDROID_NDK_HOME_V="${ANDROID_NDK_HOME}-r${ANDROID_NDK_VERSION}" \
-    CRYSTAX_NDK_HOME_V="${CRYSTAX_NDK_HOME}-${CRYSTAX_NDK_VERSION}"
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,103 +1,137 @@
...
-ENV USER="user"
-ENV HOME_DIR="/home/${USER}"
-ENV WORK_DIR="${HOME_DIR}" \
-    PATH="${HOME_DIR}/.local/bin:${PATH}"
-# get the latest version from https://developer.android.com/ndk/downloads/index.html
+RUN apt-get -y update -qq \
+    && apt-get -y install -qq --no-install-recommends curl unzip \
+    && apt-get -y autoremove \
+    && apt-get -y clean \
+    && rm -rf /var/lib/apt/lists/*
```
AndreMiras: Minor can we replace the old fashion `apt-get` by simply `apt`?

Dockerfile
```diff
@@ -1,103 +1,137 @@
...
-    --exclude=crystax-ndk-${CRYSTAX_NDK_VERSION}/toolchains/x86_64-* \
-    --exclude=crystax-ndk-${CRYSTAX_NDK_VERSION}/toolchains/llvm-* \
-    --exclude=crystax-ndk-${CRYSTAX_NDK_VERSION}/toolchains/aarch64-* \
-    --exclude=crystax-ndk-${CRYSTAX_NDK_VERSION}/toolchains/mips64el-* && \
-    ln -sfn "${CRYSTAX_NDK_HOME_V}" "${CRYSTAX_NDK_HOME}" && \
-    rm -rf "${CRYSTAX_NDK_ARCHIVE}"
+# get the latest version from https://developer.android.com/studio/index.html
+ENV ANDROID_SDK_TOOLS_VERSION="3859397"
+ENV ANDROID_SDK_TOOLS_ARCHIVE="sdk-tools-linux-${ANDROID_SDK_TOOLS_VERSION}.zip"
+ENV ANDROID_SDK_TOOLS_DL_URL="https://dl.google.com/android/repository/${ANDROID_SDK_TOOLS_ARCHIVE}"
```
AndreMiras: Why don't we keep all the env configuration top most? Is it for layer caching reasons? For me the idea is if it's on top we can easily see what can be "configured"

Dockerfile
AndreMiras: You can still use `autoremove` and stuff through `apt` just like we were doing back then with `apt-get`.
No specific reasons, it just feels too 2010 to use `apt-get` :joy: I like to stay up to date with the new "upstreamness" :wink:

PR #1472, opened 2018-11-19
Moved ndk_api parsing earlier in program initialisation

inclement:
This is actually quite important because we need to know a reasonable number for the NDK API even in places where the build environment has not been prepared.

I've removed the ability to set this parameter via environment variable because, on reflection, I don't think it's actually really important or useful. The environment variables we do accept are there partly for legacy reasons, and are mainly useful for the ndk/sdk locations rather than the api number to target.

The control flow around argument parsing could do with revisiting, but I think this is the neatest solution for now.

### DIFF
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -350,20 +350,7 @@ class Context(object):
                     'set it with `--ndk-version=...`.')
         self.ndk_ver = ndk_ver
 
-        ndk_api = None
-        if user_ndk_api:
-            ndk_api = user_ndk_api
-            info('Getting NDK API version (i.e. minimum supported API) from user argument')
-        elif 'NDKAPI' in environ:
-            ndk_api = environ.get('NDKAPI', None)
-            info('Found Android API target in $NDKAPI')
-        else:
-            ndk_api = min(self.android_api, DEFAULT_NDK_API)
-            warning('NDK API target was not set manually, using '
-                    'the default of {} = min(android-api={}, default ndk-api={})'.format(
-                        ndk_api, self.android_api, DEFAULT_NDK_API))
-        ndk_api = int(ndk_api)
-        self.ndk_api = ndk_api
+        self.ndk_api = user_ndk_api
 
         if self.ndk_api > self.android_api:
             error('Target NDK API is {}, higher than the target Android API {}.'.format(
--- a/pythonforandroid/toolchain.py
+++ b/pythonforandroid/toolchain.py
@@ -197,6 +197,15 @@ def split_argument_list(l):
     return re.split(r'[ ,]+', l)
 
 
+def select_ndk_api(user_ndk_api, user_android_api):
+    ndk_api = min(DEFAULT_NDK_API, user_android_api)
+    if user_ndk_api == 0:
+        warning(('No valid --ndk-api received, using the default of {} = '
+                 'min(android-api={}, default ndk-api={})').format(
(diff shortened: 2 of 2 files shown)


Zen-CODE:
Testing with this, it seems the API level does not pull through. It picks up API 27 earlier, but then it's 0 when looking in the ndk...
```
[WARNING]: ndk_platform doesn't exist: /home/richard/Android/android-ndk-r16b/platforms/android-0/arch-arm
```
https://gist.github.com/Zen-CODE/6d3f1e7daa1d7c305ed3380fbfd18e16

inclement:
Thanks @Zen-CODE , I've pushed a correction for that.

pythonforandroid/toolchain.py
```diff
@@ -528,7 +537,9 @@ def add_parser(subparsers, *args, **kwargs):
         self.ndk_dir = args.ndk_dir
         self.android_api = args.android_api
         self.ndk_version = args.ndk_version
-        self.ndk_api = args.ndk_api
+        args.ndk_api = select_ndk_api(args.ndk_api, args.android_api)
+        self.ndk_api = args.ndk_Api
```
Zen-CODE: Typo here ("ndk_api")

Zen-CODE:
Pulling these changes,I still get the same problem. It happens with API 21 and 27....

https://gist.github.com/Zen-CODE/9c1b2ea9022086565f5a7900bc7ea272

pythonforandroid/toolchain.py
AndreMiras: Travis didn't like that either :smile: 
```
AttributeError: 'Namespace' object has no attribute 'ndk_Api'
```

Zen-CODE:
Still get the same error...

pythonforandroid/toolchain.py
bunmalik: I have also tried, but got this error:

    args.ndk_api = select_ndk_api(args.ndk_api, args.android_api)
    NameError: name 'select_ndk_api' is not defined

inclement:
@Zen-CODE @AndreMiras Is it working now? Can this be merged?

pythonforandroid/build.py
```diff
@@ -350,20 +350,7 @@ def prepare_build_environment(self,
...
-            ndk_api = environ.get('NDKAPI', None)
-            info('Found Android API target in $NDKAPI')
-        else:
-            ndk_api = min(self.android_api, DEFAULT_NDK_API)
-            warning('NDK API target was not set manually, using '
-                    'the default of {} = min(android-api={}, default ndk-api={})'.format(
-                        ndk_api, self.android_api, DEFAULT_NDK_API))
-        ndk_api = int(ndk_api)
-        self.ndk_api = ndk_api
+        self.ndk_api = user_ndk_api
```
JonasT: Can we possibly keep NDKAPI env var support in? I think this is really important for all options that are more dependent on the installed build env (NDK/SDK/... downloaded) rather than the project itself - because otherwise, these can't be preconfigured to proper defaults in e.g. a docker build env

Zen-CODE:
From my side, I still get
```
[WARNING]: ndk_platform doesn't exist: /home/fruitbat/Android/android-ndk-r16b/platforms/android-0/arch-arm
```
as posted earlier....

inclement:
Closing as obseleted by more recent changes.

PR #1475, opened 2018-11-20
Fix android module JAVA_NAMESPACE and JNI_NAMESPACE with unicode

KeyWeeUsr:
If not switched to unicode, anything with formating such as `+` will trigger an error for py3 when accessing `JAVA_NAMESPACE` and likes:

    b'1' + u'2'
    u'1' + b'1'

and py2 just doesn't care.

### DIFF
--- a/pythonforandroid/recipes/android/__init__.py
+++ b/pythonforandroid/recipes/android/__init__.py
@@ -30,7 +30,7 @@ class AndroidRecipe(IncludedFilesBehaviour, CythonRecipe):
         th = '#define {} {}\n'
         tpy = '{} = {}\n'
 
-        bootstrap = bootstrap_name = self.ctx.bootstrap.name
+        bootstrap = bootstrap_name = self.ctx.bootstrap.name.decode('utf-8')
         is_sdl2 = bootstrap_name in ('sdl2', 'sdl2python3', 'sdl2_gradle')
         is_pygame = bootstrap_name in ('pygame',)
         is_webview = bootstrap_name in ('webview',)
@@ -38,8 +38,8 @@ class AndroidRecipe(IncludedFilesBehaviour, CythonRecipe):
         if is_sdl2 or is_webview:
             if is_sdl2:
                 bootstrap = 'sdl2'
-            java_ns = 'org.kivy.android'
-            jni_ns = 'org/kivy/android'
+            java_ns = u'org.kivy.android'
+            jni_ns = u'org/kivy/android'
         elif is_pygame:
             java_ns = 'org.renpy.android'
             jni_ns = 'org/renpy/android'
--- a/pythonforandroid/recipes/android/src/android/_android.pyx
+++ b/pythonforandroid/recipes/android/src/android/_android.pyx
@@ -175,13 +175,13 @@ api_version = autoclass('android.os.Build$VERSION').SDK_INT
 version_codes = autoclass('android.os.Build$VERSION_CODES')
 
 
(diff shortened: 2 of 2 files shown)


pythonforandroid/recipes/android/__init__.py
```diff
@@ -30,16 +30,16 @@ def prebuild_arch(self, arch):
         th = '#define {} {}\n'
         tpy = '{} = {}\n'
 
-        bootstrap = bootstrap_name = self.ctx.bootstrap.name
+        bootstrap = bootstrap_name = self.ctx.bootstrap.name.decode('utf-8')
```
AndreMiras:
```
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/python-for-android-new-toolchain/pythonforandroid/toolchain.py", line 1032, in <module>
    main()
  File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/python-for-android-new-toolchain/pythonforandroid/toolchain.py", line 1028, in main
    ToolchainCL()
  File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/python-for-android-new-toolchain/pythonforandroid/toolchain.py", line 550, in __init__
  File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/python-for-android-new-toolchain/pythonforandroid/toolchain.py", line 149, in wrapper_func
  File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/python-for-android-new-toolchain/pythonforandroid/toolchain.py", line 188, in build_dist_from_args
  File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/python-for-android-new-toolchain/pythonforandroid/build.py", line 635, in build_recipes
    recipe.prebuild_arch(arch)
  File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/python-for-android-new-toolchain/pythonforandroid/recipes/android/__init__.py", line 33, in prebuild_arch
    bootstrap = bootstrap_name = self.ctx.bootstrap.name.decode('utf-8')
AttributeError: 'str' object has no attribute 'decode'
```
:cry:

KeyWeeUsr: I guess this is due to the `Dockerfile` using Python 2, therefore it worked for me *on the device* with py3 but it fails when you want to compile with py3 on the host. I think we should fully switch to unicode literals otherwise in case something is generated like `variable = 'value'` it'll always split to bytes on py2 and unicode on py3 and we either start checking each damn occurrence of this schrodinger's string or do this:

```python
>>> b'x'.encode('utf-8').decode('utf-8')
u'x'
>>> u'x'.encode('utf-8').decode('utf-8')
u'x'
```

and get kicked by anything that is not `ASCII` encodable.

Perhaps a fix for this would be to remove the `decode()` and enforce `self.ctx.bootstrap.name` to be always unicode with an `assert`.

AndreMiras: I'm in to fully switch since it's one step further the migration to python3 and we can still make it compatible with python2 with e.g. `unicode_literals` from future

AndreMiras: Fixed by https://github.com/kivy/python-for-android/pull/1512

PR #1478, opened 2018-11-22
Update OpenSSL to 1.1.1

tito:
This PR updates OpenSSL to 1.1.1, and uses clang for compilation.
There is now a distinction between the version we use for downloading, and the version used for creating the library: OpenSSL 1.1 now have prefix with "1.1.so". This explains why others recipes that depends on it require the "library" version.

Let me know if i must remove and make a separate PR for others recipes that are broken on python3 because there was a missing deps (and forget my commit, we can squash and merge for that)

### DIFF
--- a/pythonforandroid/archs.py
+++ b/pythonforandroid/archs.py
@@ -37,7 +37,6 @@ class Arch(object):
             '-DANDROID',
             '-fomit-frame-pointer',
             '-D__ANDROID_API__={}'.format(self.ctx.ndk_api)]
-            # '-D__ANDROID_API__={}'.format(self.ctx._android_api)]
         if not clang:
             cflags += ['-mandroid']
         else:
--- a/pythonforandroid/archs.py
+++ b/pythonforandroid/archs.py
@@ -30,13 +30,25 @@ class Arch(object):
                 d.format(arch=self))
             for d in self.ctx.include_dirs]
 
-    def get_env(self, with_flags_in_cc=True):
+    def get_env(self, with_flags_in_cc=True, clang=False):
         env = {}
 
-        env['CFLAGS'] = ' '.join([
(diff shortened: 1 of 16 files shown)


tito:
Ok, this is breaking Python 2, as Python 2 is not compatible with OpenSSL 1.1. :[

tito:
So maybe get https://github.com/kivy/python-for-android/pull/1460 first, then i'll adjust this PR?

opacam:
Or maybe I could rewrite the history of pr #1460 and integrate your "openssl update and the clang feature" with the git commit option --author with your info.

You could send me a patch or two with the changes you made in here and the appropriate commit messages and I will solve the conflicts and merge it into my branch but with your user info.

I still will not close this pr because, maybe #1460 will not be merged to master branch, if merged, then you can close this (and your authory will be preserved) and if not, then this pr can be merged as usual.

I think that this should be done after someone review this pr and approve it (doesn't matter that the tests fails here, we will see if it works in pr #1460, in matter fact we know that it does with the appropriate environment: proper ndk version and python recipes)

What do you think about that?

SomberNight:
What version of Android NDK are you using?

opacam:
Referenced this pr, because I rewritted the history for #1460 on a separate branch, If all goes well with the CI tests, I will force push to the main branch. This history rewrite includes some of the @tito commits squashed into two and reflecting @tito's authory. 

Should be mentioned that all the changes from this pr affecting openssl dependant recipes are excluded because I changed the way of dealing with openssl version in another commit after the ones from @tito in order to avoid changing other recipes. Also I did not include the addition of compatibility to python3 to some essential recipes (setuptools, requests...) because I did that in another commits.

I think that, with this changes, all the git history will be more clear , easy to merge branches and also it will be easier for @tito to add/change anything if he considers necessary  :wink:

tito:
Then let's close this one :)

PR #1480, opened 2018-11-22
Fix libxslt not getting proper libxml2 path, fixes #1479

### LINKED ISSUE #1479: libxml2 build broken on latest p4a master with python 3
With latest p4a master & python3crystax & host python 3, the libxml2 recipe gets built inside `/root/.local/share/python-for-android/build/other_builds/libxml2/armeabi-v7a__ndk_target_19/libxml2/ `.

This is a folder name change, previously it used to be in `/root/.local/share/python-for-android/build/other_builds/libxml2/armeabi-v7a/libxml2/` - and as a result, the path detection inside the libxml2 fails for some reason.

### Versions

* Python: host python 3.x, python3crystax target
* OS: ubuntu in docker
* Kivy: not used
* Cython: not sure, but I don't think it's relevant

Edit: corrected lxml -> libxml2

JonasT:
Some recent change to how the build dir is named has broken the libxslt recipe (and with it, lxml2). This pull request makes it build again!

### DIFF
--- a/pythonforandroid/recipes/libxslt/__init__.py
+++ b/pythonforandroid/recipes/libxslt/__init__.py
@@ -1,5 +1,5 @@
 from pythonforandroid.toolchain import Recipe, shprint, shutil, current_directory
-from os.path import exists, join, dirname
+from os.path import exists, join
 import sh
 
 
@@ -22,10 +22,8 @@ class LibxsltRecipe(Recipe):
             # If the build is done with /bin/sh things blow up,
             # try really hard to use bash
             env["CC"] += " -I%s" % self.get_build_dir(arch.arch)
-            libxml = (
-                dirname(dirname(self.get_build_container_dir(arch.arch)))
-                + "/libxml2/%s/libxml2" % arch.arch
-            )
+            libxml = Recipe.get_recipe(
+                'libxml2', self.ctx).get_build_dir(arch.arch)
             shprint(
                 sh.Command("./configure"),
                 "--build=i686-pc-linux-gnu",


pythonforandroid/recipes/libxslt/__init__.py
```diff
@@ -24,7 +24,8 @@ def build_arch(self, arch):
             env["CC"] += " -I%s" % self.get_build_dir(arch.arch)
             libxml = (
                 dirname(dirname(self.get_build_container_dir(arch.arch)))
-                + "/libxml2/%s/libxml2" % arch.arch
+                + "/libxml2/%s/libxml2" % basename(
+                    self.get_build_container_dir(arch.arch))
```
inclement: I think lines 25 to 28 can just be `libxml = self.get_build_dir(arch.arch)`. Am I missing something here, since the rest of the recipe uses `get_build_dir` normally I don't see why this part was originally written differently?

JonasT:
@inclement this is in the `libxslt` recipe, so this would return the `libxslt` build dir. However, the recipe tries to access the `libxml2` recipe build dir, not its own build dir.

Also, I missed an import and forgot to push that fix :woman_facepalming: just corrected that

pythonforandroid/recipes/libxslt/__init__.py
inclement: Oh, I see. In that case you can do `Recipe.get_recipe('libxml2', self.ctx).get_build_dir(arch.arch)`. This has the advantage of being fully future proof.

JonasT: I added in your suggested change! <3

inclement:
Thanks!

JonasT:
Thanks for the merge! :heart:

Issue #1473, opened 2018-11-20
.pxd files of dependency targeted recipe'd library not found


JonasT:
### Description

It seems that when having two Cython libraries in the dependencies (both with a CythonRecipe with no special options) and one `cimport`s things from the other (therefore requiring its `.pxd` files), something incorrect in python-for-android's Cython setup will break locating any of the `.pxd` files. I'm just getting errors like `'wobblui/cache.pxd' not found`, `'wobblui/font/manager.pxd' not found`, ... for all such "external" pxd files.

On desktop, the build works perfectly fine, even when installing the dependency target with the `.pxd` files remotely using `pip3 install https://github.com/JonasT/wobblui/archive/master.zip` - which should be how python-for-android fetches and installs it as well. I confirmed with `updatedb && locate` that the `.pxd` files are indeed installed, including in the python-for-android build environment. Just for some reason, Cython inside the python-for-android build doesn't "see" them.

TL;DR: `.pxd`s of Cython dependency target libs aren't found. Breaks build, not good :cry:

### Versions

* Python: python3crystax, with python 3 build host
* OS: ubuntu in docker
* Kivy: not used
* Cython: 0.29

### buildozer.spec
not used

JonasT:
Are recipe dependencies properly added to the site-packages paths *before* the respective next recipe is built/its setup.py is run? Are they available in the follow up recipe's virtualenv at all?

Because they would really need to be, otherwise it's no surprise this breaks. If this is the reason that'd be bad, because that makes any more performant `cimport`s between Cython libraries impossible... (in addition to the fact that my libraries are now written to **require** this, so I'd have to rip it out again even for the desktop version, or do branching code from hell to somehow work around this)

JonasT:
Just to record the insights from chat, I checked the python-for-android code.
It has the following design errors:

- It just runs Cython manually, assuming all `.pyx` files in the project tree need to build. (Not a bad guess, but simply not a necessarily correct assumption)
- It runs Cython with host python outside of any virtualenv. This is simply incorrect, because `.pxd` files from previous dependencies might be required for build. (This is why I got the error for which I made this ticket)

In overall the main problem seems to be that it manually runs Cython at all, when really the `setup.py` should get the correct env setup and be trusted to Cythonize. This ties in with the borked Cython defaults without a CythonRecipe to start with, because if all was set up correctly, none of this completely unnecessary manual Cythonize handling that makes all these wrong assumptions would be required in the first place.

[...]

PR #1483, opened 2018-11-23
Make Cython work without recipe

JonasT:
**What this does:** This merge request makes Cython work for the non-recipe regular build. This means CythonRecipe is now unnecessary if the project doesn't need Android-specific extra changes. That's good because a recipe with no actual content is redundant, and because CythonRecipe is problematic:

**Why CythonRecipe is problematic:** I noticed CythonRecipe results in the packages own `setup.py` not run for Cythonization, and instead, p4a engages some manual Cython compilation which is broken per design for a couple of reasons:

1. it assumes all `.pyx` files anywhere in the project dir must compile. This breaks projects that have any `.pyx` files that are *not meant* to be built for the main build and break when Cythonized out of the blue
2. it uses the wrong `site-packages` because it uses host python. This breaks `.pxd` cross-includes from any dependencies, since this requires proper site-packages during Cythonize (this affected me, see #1473 )
3. it assumes the `setup.py` won't produce new `.pyx` files on its own. This breaks projects where that happens

(Please note this pull request doesn't fix CythonRecipe in any way, it just makes it avoidable for Android-aware Cython projects)

**Needs quite some testing, obviously!**

### DIFF
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -779,12 +779,14 @@ class PythonRecipe(Recipe):
             hppath.append(join(dirname(self.hostpython_location), 'Lib'))
             hppath.append(join(hppath[0], 'site-packages'))
             builddir = join(dirname(self.hostpython_location), 'build')
-            hppath += [join(builddir, d) for d in listdir(builddir)
-                       if isdir(join(builddir, d))]
-            if 'PYTHONPATH' in env:
-                env['PYTHONPATH'] = ':'.join(hppath + [env['PYTHONPATH']])
-            else:
-                env['PYTHONPATH'] = ':'.join(hppath)
+            if exists(builddir):
+                hppath += [join(builddir, d) for d in listdir(builddir)
+                           if isdir(join(builddir, d))]
+            if len(hppath) > 0:
+                if 'PYTHONPATH' in env:
+                    env['PYTHONPATH'] = ':'.join(hppath + [env['PYTHONPATH']])
+                else:
+                    env['PYTHONPATH'] = ':'.join(hppath)
         return env
 
     def should_build(self, arch):
@@ -955,16 +957,6 @@ class CythonRecipe(PythonRecipe):
 
         env = self.get_recipe_env(arch)
 
-        if self.ctx.python_recipe.from_crystax:
-            command = sh.Command('python{}'.format(self.ctx.python_recipe.version))
-            site_packages_dirs = command(
-                '-c', 'import site; print("\\n".join(site.getsitepackages()))')
-            site_packages_dirs = site_packages_dirs.stdout.decode('utf-8').split('\n')
-            if 'PYTHONPATH' in env:
-                env['PYTHONPATH'] = env['PYTHONPATH'] + ':{}'.format(':'.join(site_packages_dirs))
-            else:
-                env['PYTHONPATH'] = ':'.join(site_packages_dirs)
-
         with current_directory(self.get_build_dir(arch.arch)):
             hostpython = sh.Command(self.ctx.hostpython)
             shprint(hostpython, '-c', 'import sys; print(sys.path)', _env=env)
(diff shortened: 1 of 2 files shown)


JonasT:
Side note: I'm unhappy about how this pull request makes things *based* on CythonRecipe for the env vars of the regular build, since I think the CythonRecipe should be a special case of special needs Cython projects, not the other way around. However, I didn't take care of this yet because I wanted to first see if this works for other projects at all, and then gradually move stuff out of CythonRecipe and into the core later, stripping down CythonRecipe to its special uses in that process.

(Also, it would move quite a lot of stuff around, so just for cleanliness I think it'd be nicer to have it in a separate pull request later)

tito:
About your 1. just to let you know Kivy source code have pyx that doesn't work and required on Android (like window_x11.pyx, etc). So maybe you should have an option to give a list of files to ignore?

JonasT:
> About your 1. just to let you know Kivy source code have pyx that doesn't work and required on Android (like window_x11.pyx, etc). So maybe you should have an option to give a list of files to ignore?

That's like a hack of another hack. The real problem is that the `setup.py` already does this for any sane project, and p4a's CythonRecipe basically assumes the `setup.py` never works for any project for Cythonizing which IMHO is just a bad assumption. (It would be ok as a manual per-recipe enabled workaround, but as a default behavior for Cython projects I don't think it makes a lot of sense)

Also, I don't have any non-compiling `.pyx` files either, I was just explaining why CythonRecipe kinda sucks.

JonasT:
It appears to be working mostly fine, but I can't fully test since I'm currently stuck on #1487

JonasT:
Okay, so I finally made it through an app build to test this, and while the build succeeds and the app is packaged with the Cython parts successfully in the site-packages, this happens when it is launched:

> ImportError: dlopen failed: cannot locate symbol "PyExc_DeprecationWarning" referenced by "/mnt/expand/50b713c2-bef0-4226-aea1-000678d55d07/user/0/com.example.myapp/files/app/_python_bundle/site-packages/wobblui/`__core__.so`"...

I wonder, since there is no ELF loading error, whether the architecture/toolchain worked out correctly, but the linked python version or headers are possibly wrong? Anyone having an idea what the problem could be?

JonasT:
(this is still WIP though, I want to clean up some stuff - but I'd **love** to hear test results!! Even if your app doesn't use Cython, just to make sure other things aren't broken by this)

JonasT:
Okay, I cleaned up the parts I still wanted to clean up.

**To test this and give me useful feedback, it is enough to just build your project with this merge request.** While it needs some adjustments to use the new functionality, the much more important question is *whether it breaks existing stuff.* It changes the regular non-recipe python module install env, so it has the potential to cause some interference, so some wider testing wouldn't be a bad idea.

So anyone who's interested in testing, please do and report back! It works fine for me is all I can say

JonasT:
Just a tiny note: the latest force pushes were purely rebasing to latest master since I rely on this pull request for my own builds quite deeply, not actual changes. As I wrote above, **it is done and ready for testing.**

JonasT:
@AndreMiras by the way, any clue on when we're going to merge this? It's pretty much been sitting here for two weeks, I use it for my active project and it's done as far as I'm concerned... so IMHO it would make sense to review & merge this *if* you guys want this feature. I personally think moving away from recipes where they aren't strictly necessary is always an improvement, but that's just my view on things

inclement:
Thanks, sorry for the delay on this one.

pythonforandroid/recipe.py
```diff
@@ -989,8 +981,15 @@ def build_cython_components(self, arch):
...
 
+            self.strip_object_files(arch, env)
+
+    def strip_object_files(self, arch, env, build_dir=None):
+        if build_dir is None:
+            build_dir = self.get_build_dir(arch.arch)
+        with current_directory(build_dir):
             info('Stripping object files')
             if self.ctx.python_recipe.name == 'python2legacy':
+                info('Stripping object files')
```
AndreMiras: Minor: isn't this redundant with the one two lines above?

JonasT:
oops yeah, I had  a conflict there and must have messed that line during merge/rebase

AndreMiras:
Well that's not a big deal anyway, very good work as usual and sorry for letting the PR rot for so long. Great that inclement took over and merged

PR #1484, opened 2018-11-23
Fix broken --ndk-api default value

JonasT:
I ran into this by accident while making the other pull requests. `--ndk-api` defaults to `0` in the parser, but code in other places (e.g. `distribution.py`) expects it to be `None` if not set. This breaks locating the proper final distribution in some cases when it's not specified, therefore I suggest this should be fixed.

### DIFF
--- a/pythonforandroid/toolchain.py
+++ b/pythonforandroid/toolchain.py
@@ -261,7 +261,7 @@ class ToolchainCL(object):
             help=('The version of the Android NDK. This is optional: '
                   'we try to work it out automatically from the ndk_dir.'))
         generic_parser.add_argument(
-            '--ndk-api', type=int, default=0,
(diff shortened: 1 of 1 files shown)


JonasT:
There was no change with that force push, I just rebased onto latest master so I could locally pile up all the pull requests I currently depend on :cry:

inclement:
https://github.com/kivy/python-for-android/pull/1472 provides a different fix that addresses the root cause better, but if this improves things then it seems fine to merge for now. Thanks.

DanShai:
When it was None I got tons of errors and couldnt build the apk, but @tshirtman solution worked like charm ! 0 is better and buit the apk on the fly. if you woried bout that line just replace:
 if ndk_api is not None   --to-->  if ndk_api 
thats it  and leave it at 0

PR #1486, opened 2018-11-23
add .tox to git ignore files

tito:


JonasT:
Ah, ran into this as well yesterday but forgot about it. Useful addition! :+1:

Issue #1488, opened 2018-11-23
Project's setup.py is not run, it should be (at least as an option)


JonasT:
The main project's setup.py is never run. This breaks projects that need to be installed to work. I think this is wrong and should be changed, at least with an optional `p4a` option - the recipes can be installed beforehand so running the `setup.py` shouldn't trigger any misguided non-reciped dependency installs, so I don't see why this couldn't be done.

(My current workaround to this is to copy the entire project into another folder and specify it again as a dependency - but that's kind of hacky and I also ran into #1487 )

### Versions

* Python: all versions affected
* OS: ubuntu docker
* Kivy: not used
* Cython: not relevant

JonasT:
> The current behaviour is essentially a deliberate choice to keep things simple, but it's definitely open to a rethink.

Yes, please! Because this is why p4a is a huge headache for me, and I pile up so many pull requests. **I think defaulting to such simplifications is wrong** - and I mean almost always.

Don't get me wrong, we need those hacks. Projects need them, recipes need them. The problem is, these hacks break things. And there is no way to turn them off without recipe, which a non-broken project fundamentally just shouldn't need. (IMHO)

*Example of the headaches of this `--no-deps` issue:*

E.g. to work around this issue, I need to add the project itself as a dep so it is build. However, just adding the project folder will make p4a think it has a recipe, since it's the project itself, which messes up Cython and more - so I need to copy it to a duplicate folder, and put that as a dependency. Can you imagine how long it took me just to figure that out...?

I just don't think this is good. Let python tools do their work: 1. lowers entrance barrier (less need to learn p4a quirks), 2. doesn't make some builds impossible without fundamental hacks, 3. recipes are needed for "picky"/broken projects anyway

But that's just my opinion, of course!

JonasT:
Also let me add it's completely not obvious to any python expert even, that the `setup.py` will simply not run. `p4a` spews out so much output, how is anyone new to it supposed to see that?

For that reason alone, I think at the very least, this should be a `p4a` command line option, and default to running the `setup.py` with maybe a fat warning that it could break which refers to the option, similar to how regular python modules are treated.

JonasT:
This will be fixed one #1625 is merged

PR #1489, opened 2018-11-23
Fix handling of file paths in module dependencies

JonasT:
This fixes the handling of file paths in module dependencies in `--requirements`, which previously led to them simply being ignored (and hence breaking any build that would try to use this).

### DIFF
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -8,6 +8,7 @@ import glob
 import sys
 import re
 import sh
+import subprocess
 
 from pythonforandroid.util import (ensure_dir, current_directory)
 from pythonforandroid.logger import (info, warning, error, info_notify,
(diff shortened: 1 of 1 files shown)


JonasT:
This is by the way rather relevant to work around #1488 (because you can copy your project folder and specify that as a dep to make sure it's `setup.py` installed) so it's not quite as theoretical as it may sound like

inclement:
Seems fine, although I'll note that this it not so much fixing the handling of file paths as adding support for modules supplied as file paths. I don't think this has ever been intended to work. It's interesting that it takes so little work to get it going.

JonasT:
@inclement I can't rule out other broken corner cases, but it worked fine in my tests. Please note this doesn't fix recipe's not handling them though, it just fixes them not working properly for `--requirements`

JonasT:
Updated the code a little such that it leaves URLs alone. Other than that, it works for me, and I encourage others to give it a try. I am depending on an increasing pile of unmerged changes to get my project to build, so it'd be neat to see particularly this file path support go in some time soon

PR #1491, opened 2018-11-26
Fix reportlab GPL issue by removing problematic component

JonasT:
This fixes the reportlab GPL issue by removing problematic component. (By default, reportlab has a GPL'ed font that is included into the `.apk` which is quite bad for licensing reasons.)

I tested that:

- the DarkGarden font is gone from the `.apk` (by manually unzipping it and opening the `private.mp3` tarball)
- that the build with `python3crystax` still works
- that a `python3crystax` app can still successfully export PDF with text

So as far as I'm concerned, this should be ready for merging


JonasT:
Sorry for the indentation / flake 8 error, forgot to check that. Should be fine now

PR #1494, opened 2018-11-26
when listing distributions, if one has no ndk_api, consider it to be 0

tshirtman:
this avoids crashes, but also ensure that it doesn't consider it a
match, is that something we want?

### DIFF
--- a/pythonforandroid/distribution.py
+++ b/pythonforandroid/distribution.py
@@ -190,7 +190,7 @@ class Distribution(object):
                 dist.recipes = dist_info['recipes']
                 if 'archs' in dist_info:
                     dist.archs = dist_info['archs']
-                dist.ndk_api = dist_info['ndk_api']
+                dist.ndk_api = dist_info.get('ndk_api', 0)
                 dists.append(dist)
         return dists
 


tshirtman:
@inclement @AndreMiras opinions welcome!

AndreMiras:
Thanks for looking into this!
I thought we had a default value setup somewhere. But seeing [Distribution.ndk_api](https://github.com/kivy/python-for-android/blob/1f2291e/pythonforandroid/distribution.py#L25) is `None` that's probably not what I was looking for :smile: 
What about just totally skipping it when the key is not available, just like we did for `'archs' in dist_info` above?
I haven't played with it so I'm not sure if that would make sense. @inclement is your man since he made a large refactoring on this recently.

tshirtman:
Yeah i was wondering about just setting to None, but i now think it's not going to make much of a difference, since this is only checked if user asked for a specific api, this way may trigger a few unnecessary distribution rebuild, and some unused space, but it's probably a lesser evil than trying to guess what api was used in an older distribution if we fail.

DanShai:
Thanx @tshirtman  your solution solved my problem !

inclement:
The intended solution is to delete and rebuild the dist as ultimately using a dist with an unknown API target can't be recommended, but this wasn't a major point. Setting a default value is acceptable, it may as well be zero. It would also be good to print a warning in this case.

tshirtman:
Ok, i'll add a warning, would something like:

"Distribution {dist} has been built with an unknown api target, ignored, you might want to delete it".

represent the intention well?

inclement:
Sounds good to me!

pythonforandroid/distribution.py
```diff
@@ -190,7 +190,18 @@ def get_distributions(cls, ctx, extra_dist_dirs=[]):
...
+                if 'ndk_api' in dist_info:
+                    dist.ndk_api = dist_info['ndk_api']
+                else:
+                    dist.ndk_api = 0
+                    warning(
+                        "Distribution {distname}: (distdir) has been "
+                        "built with an unknown api target, ignoring it, "
+                        "you might want to delete it".format(
+                            distname=dist.name,
+                            distdir=dist.dist_dir
```
AndreMiras: `distdir` seems unused here

tshirtman: I though it would make it easier for user to remove it… but i can remove it if it's not relevant.

JonasT:
Please note an `ndk_api` of `0` breaks this line:

https://github.com/kivy/python-for-android/blob/master/pythonforandroid/distribution.py#L98

This is why I made this pull request: https://github.com/kivy/python-for-android/pull/1484

So I highly recommend against merging this unless you either address this, or you're certain that this won't be an issue (I might be wrong, obviously).

DanShai:
When it was None I got tons of errors and couldnt build the apk, but @tshirtman solution worked like charm ! 0 is better and buit the apk on the fly. if you woried bout that line just replace:
if ndk_api is not None --to--> if ndk_api
thats it and leave it at 0

JonasT:
Right, that might be one way to address it, but then also the argument parser default value would need to be set to `0`. My main intention was to point out that in the current form this pull request might clash with some things, not that `None` is the way to go or anything like that

JonasT:
Another line that may need changing in relation to this pull request: https://github.com/kivy/python-for-android/blob/master/pythonforandroid/distribution.py#L215

Might be worth searching for `ndk_api is None` all over the code before going ahead with this change

JonasT:
Hm. Thinking about this some more, why can't we fix the code that breaks handling `None`? It is obviously a special case, using `0` might risk running through code without warnings/errors that *should* actually care about the `ndk_api`... so I'm not sure this is the right direction. But I wrote too much already, I'll let you guys discuss this and figure something out

tshirtman:
It's good feedback, i had looked at the line you mention and assumed that it wouldn't be an issue, but didn't think that `ndk_api` could be 0, indeed, that wouldn't be good, maybe setting to `None` would be better, that way we are sure this condition will evaluate to True and the distribution is not matched.

Anything against that?

pythonforandroid/distribution.py
AndreMiras: No sorry what I mean is the format string seems to add a param that's not used, right?

JonasT:
As elaborated above, I like `None`, but I can't claim to know all the potential consequences of this change :grimacing:

pythonforandroid/distribution.py
JonasT: Yea kind of looks like `(distdir)` should be `{distdir}`

tshirtman: uh, i misread that >_<, thanks for your patience :laughing:

AndreMiras:
Thank you for clarifying @inclement then do we want to address it this way in this case?
I mean if at the end the apk is not going to work, why don't we fail fast with a meaningful error during the build? Rather than failing late at runtime?

JonasT:
@AndreMiras I might be wrong, but I think the change here means that rather than the presence of an old, non-ndk build crashing p4a *always*, it will no longer disrupt build of a **proper** distribution with ndk version when any old distribution is *still present* - with this warning advising the user to remove that old outdated distribution, since it won't be built or be recognized anyway.

I think that's useful, because the warning is much more useful than a random backtrace, and I also find it useful to not block building a more recent project just because an older, incompatible distribution happens to be lying around. But of course one could still abort the entire build always but with a nice error. However, I prefer the solution here (given I understood things correctly)

JonasT:
@tshirtman what do you think about changing this line:

https://github.com/kivy/python-for-android/blob/master/pythonforandroid/distribution.py#L98

to:

`if (ndk_api is not None and dist.ndk_api != ndk_api) or dist.ndk_api is None:`
?

(The idea being that a distribution with `.ndk_api` is not set, and this would ensure it's never picked up even when `--ndk-api` is unspecified by the user)

PR #1496, opened 2018-11-29
Rework common bootstrap area based on kollivier's work

JonasT:
This is a rework of a common shared bootstrap area based on @kollivier 's work started here: https://github.com/kivy/python-for-android/pull/991

**Ignore comments below about errors, got it to work!** :smile:


JonasT:
Okay so pygame has all jni stuff in `jni/application/src/` and others in `jni/src/`, which in my humble opinion is extremely ugly. I am now trying to unify this, but this will most likely break a lot of things, so I could use some help getting this to work again. However, without this, common files like `start.c` cannot be reasonably shared without major pain, so I don't see any way not to do this (apart from this difference in structure for no good reason simply being ugly).

JonasT:
**What do I need help with**: checkout this pull request, then build with the "sdl2" and "webview" targets, and report back here how all the `Android.mk` files and similar could be modified so things work again without an error, so I can work those things in and repair everything.

JonasT:
Ok it seems that with the folder change from `jni/src` to `jni/application/src`, currently, the `main.so` for the SDL2 entry point will either not be built, or not be included after the build. (Can't find it in the .apk and the SDLActivity errors in the `public void loadLibraries()` function, most likely failing to load `main.so`.)

Can someone point me to where this path is stored/accessed for building & packaging? Is it somewhere in the gradle files? Some help by someone familiar with how the SDL2 bootstrap works would really be greatly appreciated

pythonforandroid/bootstrap.py
```diff
@@ -11,6 +13,26 @@
 from pythonforandroid.recipe import Recipe
 
 
+def copy_files(src_root, dest_root, override=True):
```
tshirtman: any reason not to use `shutil.copytree` ?

JonasT: `shutil.copytree` can't override I think, or at least I am not aware of it. Also, according to Stack Overflow, it can't skip existing files either which is really needed here, and which is what the function does with `override=False`. So I don't want to rule out that this could be done with `shutil` functions somehow, but this seemed to be a good way to do it. I didn't write this code by the way, this is `99%` @kollivier 's part - but I would have done something similar

tshirtman: ah right, it should work in python3 with a copy_function that ignore existing files, but that option is not present in python2, there is the `ignore` argument, probably with set intersection to ignore, but not sure how much simpler that would be. Nothing wrong with having a specific function anyway.

JonasT:
Sorry, still had a small mistake: the wrong `start.c` was in the common area, I fixed that now. Somehow managed to only fix that on my custom testing branch (which merges a couple of other unmerged pull requests to get my app to build). Travis build should hopefully be happy with that change

pythonforandroid/toolchain.py
```diff
@@ -141,6 +141,9 @@ def wrapper_func(self, args):
                                       user_ndk_ver=self.ndk_version,
                                       user_ndk_api=self.ndk_api)
         dist = self._dist
+        bs = Bootstrap.get_bootstrap(args.bootstrap, ctx)
+        # recipes rarely change, but during dev, bootstraps can change from
+        # build to build, so run prepare_bootstrap even if needs_build is false
```
AndreMiras: The comment says we run `prepare_bootstrap()` but I don't see that we do :confused: 
Also why are we calling `Bootstrap.get_bootstrap()` for?

JonasT: Interesting! Might be an oversight by @kollivier and maybe it was supposed to have a `ctx.prepare_bootstrap(bs)` right after? That would explain why `bs` is unused and there's the confusing comment there. I don't reuse any of the build layout for subsequent builds, so I assume that's why I didn't run into a situation in my tests where this was relevant

pythonforandroid/bootstrap.py
```diff
@@ -11,6 +13,26 @@
...
+    for root, dirnames, filenames in walk(src_root):
+        for filename in filenames:
+            subdir = root.replace(src_root, "")
+            if subdir.startswith(sep):
+                subdir = subdir[1:]
```
AndreMiras: Minor suggestion/question:
I wonder if it's not cleaner to use something like `os.path.normpath()`?
```
>>> dest_root = '/foo/'
>>> subdir = 'bar/'
>>> os.path.join(os.path.normpath(dest_root), os.path.normpath(subdir))
'/foo/bar'
```
Which also covers he case were `subdir` by chance ends with two slashes.
So basically we could replace the three lines above by one:
```python
subdir = os.path.normpath(root.replace(src_root, ""))
```

JonasT: normpath won't remove a leading(!) slash, only a trailing one. It could still be used in addition, but I'm pretty certain the point of this code is to make sure `subdir` becomes a relative path, and `normpath` alone wouldn't achieve that

JonasT:
I suggest that the missing `ctx.prepare_bootstrap(bs)` should still be added though. I'm currently testing a build with that change

JonasT:
(I suppose it could also still be added in a minor merge request afterwards)

pythonforandroid/toolchain.py
kollivier: The code wasn't used in the original patch, either, so whatever I was intending to do, I never actually made it happen. I think it can just be removed. 

From the comment, I suspect that during some tests I found a case where switching the bootstrap didn't trigger a rebuild. Even if that was still true, it's likely not a common case (I was triggering it no doubt because I was testing with different bootstraps!) and can be fixed with a clean rebuild.

kollivier: Also, thanks for cleaning this up and getting it running against the latest code! :)

JonasT:
Follow-up pull request for outstanding possible issues & fixes with the common area handling code itself is here: #1499

pythonforandroid/toolchain.py
JonasT: @kollivier you're welcome, and this would have been infinitely more work without your foundation, so thanks again for starting this!

PR #1499, opened 2018-11-30
Minor fixes to basic common bootstrap handling code

JonasT:
Minor fixes to basic common bootstrap handling code as suggested by @AndreMiras here: #1496 (see the code review comments)

Please note I added `normpath` *without* replacing the two code lines after it as suggested in the code review, since that one does something different (not a path normalization, but ensures it is a relative path) and I'm pretty sure it needs to stay in.

**Tested, ready for merge**

### DIFF
--- a/pythonforandroid/bootstrap.py
+++ b/pythonforandroid/bootstrap.py
@@ -1,4 +1,4 @@
-from os.path import (join, dirname, isdir, splitext, basename)
+from os.path import (join, dirname, isdir, normpath, splitext, basename)
 from os import listdir, walk, sep
 import sh
 import glob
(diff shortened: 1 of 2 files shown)


JonasT:
The `prepare_bootstrap` line does actually blow up, so I'll have to investigate deeper why exactly the comment suggests it should be there and see if I can fix it. Will report back once I have some more insight

JonasT:
Okay, I have modified this as per @kollivier 's suggestion with the accepted caveat that if a `p4a` user changes the bootstrap without forcing a clean build, things could break.

Then I tested that the resulting merge request as it is now works fine and builds my app with no problems, and it does! Travis is also happy. Therefore, as far as I'm concerned, this is ready for merge.

PR #1500, opened 2018-12-02
Unify start.c of all bootstraps to one file

JonasT:
I have started reviewing the `start.c` files of all bootstraps for changes, and preparing a unification process to remove the different `start.c` files. This pull request is the result of this.

**WIP because I still need to test this myself.**


JonasT:
Okay, so I tested the `sdl2` bootstrap with `python3crystax` and a real world app, and it builds & still works! Now the others still need to be tested. Anyone has some actual apps lying around to test with that?

inclement:
At a quick glance, looks very solid so far.

I've been meaning for a while to make the start.c code less awful, do you have any intention to try that? It might be an opportunity to avoid some of the ifdef mess we currently have. However, it isn't essential, just merging the files is a great improvement.

JonasT:
> I've been meaning for a while to make the start.c code less awful, do you have any intention to try that?

Eventually, but so far I'm still mostly busy with fixing architectural issues to get my app back to building from p4a master. This involves rescheduling my `ctypes.util.find_library()` merge request (which changes `start.c`, hence is best to go in after this one) to finally get this working for Python 3 based on this.

I am interested in also cleaning up the code itself some more, I just haven't really gotten around yet. But this pull request should be a splendid foundation for anyone to do that in the near future

JonasT:
Right, since the only thing remaining is more testing, I'll remove the `WIP` now. I tested `sdl2`, I think @AndreMiras wanted to look into `service_only`, might be worth looking into `webview` still working (if it wasn't broken before) before merging

JonasT:
*(latest force push is rebase onto master with gradle issue fixed, no change in this pull request)*

**Testing progress:**

- *python3crystax / sdl2 bootstrap*: tested build & deploy & launch of actual real-world app with p4aspaces `p4a-py3crystax-api19-ndkbundle` environment
- *python3 / sdl2 bootstrap*: tested build of `testapp_keyboard`, to reproduce: `sudo p4aspaces cmd p4a-py3-api28ndk21 testbuild --p4a https://github.com/JonasT/python-for-android/archive/startcunify.zip`
- *python2 / sdl2 bootstrap*: in progress

pythonforandroid/bootstraps/common/build/jni/application/src/start.c
```diff
@@ -14,9 +14,15 @@
 #include <sys/types.h>
 #include <errno.h>
 
+#include "bootstrap_name.h"
+#ifndef HAS_NO_SDL_BOOTSTRAP
```
inclement: Since this includes every bootstrap but SDL2, would it be simpler to just check for BOOTSTRAP_NAME_SDL2, or are do you have something in mind for this?

JonasT: I just named this badly, it's supposed to mean "this bootstrap uses SDL headers". I think at least one other (might have been pygame) uses SDL headers as well. Should rename it for clarity, good catch

inclement: I thought that was the intention, but it looks like your pygame file includes "#define HAS_NO_SDL_BOOTSTRAP", is that a mistake?

My instinct would be to just hardcode `if defined(BOOTSTRAP_NAME_PYGAME) || defined(BOOTSTRAP_NAME_SDL2)`, and revisit it if in fact we add another SDL bootstrap. Especially because I'd consider pygame/sdl1 and sdl2 to be so different as to be different libraries that happen to have a header file with the same name.

I don't feel that strongly about it though, it's simple enough to change if we do something in the future that would encourage a different scheme.

JonasT: Nice, this was actually a legitimate mistake. This was defined for `pygame` when it shouldn't have been, I checked back on the pre-merge source. Fixed now, and renamed to `BOOTSTRAP_USES_NO_SDL_HEADERS`. :+1:

JonasT:
I did another Python 2 sdl 2 bootstrap testbuild with the force-pushed version that fixes the `HAS_NO_SDL_BOOTSTRAP`/`BOOTSTRAP_USES_NO_SDL_HEADERS` thing, and it worked fine. Would have been surprised about anything else but wanted to be sure

AndreMiras:
Also looks good to me, I'll give it a run-time try with `service_only` and crystax later today.
Thanks guys!

PR #1503, opened 2018-12-03
Make SDL2 & services_only bootstrap properly error with missing --private

JonasT:
The SDL2 & services_only bootstraps don't support building without `--private`, but the package assembling `build.py` will give an obscure backtrace if the option is omitted. This pull request adds in a proper error.


JonasT:
No it's really just these two, I'm pretty sure. I didn't test them all, but scanned over the entire build/packaging code to see if it gracefully handles args.private being `None` or not.

I think the historical reason is that `webview` can launch a webserver even without custom app code, and `pygame` maybe launches some sort of pygame default window? While `sdl2`/`service_only` have no "default" graphical UIs that could show or do anything, so they need a custom entry point to work.

Issue #1506, opened 2018-12-08
p4a latest master / Python 2.7 / API target 28 / NDK 21 / openjdk8 crashes during gradle step


JonasT:
When combining p4a latest master, Python 2.7, API target 28, NDK api 21, and openjdk8 to build the **official testapp_keyboard** out of the repo, then the build will crash during the gradle step:

```
# cd ~/testapp && p4a apk --arch=armeabi-v7a --name test --package com.example.test --version 1 --requirements=kivy,python2 --private .
...
[INFO]:    Selecting java build tool:
[INFO]:    Detected highest available build tools version to be 28.0.3
[INFO]:        Building with gradle, as gradle executable is present
[INFO]:    -> running gradlew assembleDebug
           working:                                                                                 Exception in thread background thread for pid 23040:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/usr/local/lib/python2.7/dist-packages/sh.py", line 1540, in wrap
    fn(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/sh.py", line 2459, in background_thread
    handle_exit_code(exit_code)
  File "/usr/local/lib/python2.7/dist-packages/sh.py", line 2157, in fn
    return self.command.handle_command_exit_code(exit_code)
  File "/usr/local/lib/python2.7/dist-packages/sh.py", line 815, in handle_command_exit_code
    raise exc
ErrorReturnCode_1: 

  RAN: /root/.local/share/python-for-android/dists/unnamed_dist_1/gradlew assembleDebug

  STDOUT:
Downloading https://services.gradle.org/distributions/gradle-3.3-all.zip
[...]

PR #1507, opened 2018-12-09
Update gradle version

opacam:
To solve the ci errors when trying to download the corresponding build tools which seems to not exist anymore...

Resolves: #1506

### DIFF
--- a/pythonforandroid/toolchain.py
+++ b/pythonforandroid/toolchain.py
@@ -847,7 +847,9 @@ class ToolchainCL(object):
 
                 # gradle output apks somewhere else
                 # and don't have version in file
-                apk_dir = join(dist.dist_dir, "build", "outputs", "apk")
+                apk_dir = join(dist.dist_dir,
+                               "build", "outputs", "apk",
+                               args.build_mode)
                 apk_glob = "*-{}.apk"
                 apk_add_version = True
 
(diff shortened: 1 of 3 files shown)


JonasT:
I did an additional build of the `testapp_keyboard` with Python 3 non-crystax / API 28 + NDK 21 / openjdk8 (p4aspaces' `p4a-py3-api28ndk21` env) because you know, I don't have much better things to do... and that one also worked. Haven't tested the `.apk` though, but the build runs through fine including the gradle step.

opacam:
I did test an apk at run time and it seems to work as expected (only python2 tested but it should work for python 3 because I tested with the changes made in pr #1460, so both versions of python shares the same build system). The nature of the failure makes me think that all the ci builds made after the removed build tools will fail, unless restored the build tools...I don't know if we can rely on this...

The only visible difference (in functionality) I find is that the resulting apk's path has been changed into a sub folder (debug or release).

I must say that with this pr the most of the gradle dependencies will be downloaded from google...this will make the build system more reliable? hope that yes...

inclement:
Thanks!

pythonforandroid/toolchain.py
```diff
@@ -847,7 +847,9 @@ def apk(self, args):
 
                 # gradle output apks somewhere else
                 # and don't have version in file
-                apk_dir = join(dist.dist_dir, "build", "outputs", "apk")
+                apk_dir = join(dist.dist_dir,
+                               "build", "outputs", "apk",
+                               args.build_mode)
```
tito: Just find out that this line breaks buildozer APK searchs. :[

tito: Buildozer was searching in the apk directory, not a subdirectory.

opacam: Ei @tito, it seems that @AndreMiras already fixed that in buildozer's master branch, see:

kivy/buildozer#766

PR #1511, opened 2018-12-10
No need to decode into unicode when running in python 3

jtoledo1974:
python-for-android fails to build when running under a python 3 interpreter with the android recipe. There is no decode method in python3 strings.

### DIFF
--- a/pythonforandroid/recipes/android/__init__.py
+++ b/pythonforandroid/recipes/android/__init__.py
@@ -4,6 +4,7 @@ from pythonforandroid.patching import will_build
 from pythonforandroid import logger
 
 from os.path import join
+from sys import version_info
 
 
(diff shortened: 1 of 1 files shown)


AndreMiras:
I was wondering why this was closed, but probably a twin of #1512 by misclicking or another GitHub glitch :smile:

jtoledo1974:
Missed a reassignment to bootstrap_name. In all honesty I don't see the need of two different variables, bootstrap and bootstrap_name, but it's not my code so I'm just providing the most minimal change that runs the code :-)

AndreMiras:
I don't see the need either, but I haven't looked deeper.
Anyway then it's good for you to know you don't have to create a new pull request every time since it's possible to update existing ones. You can simply commit and push again.
Then you have more advanced things to do to keep the history clean, like rebasing and squashing. Take a quick look that might help for next time :wink:

PR #1512, opened 2018-12-10
No need to decode into unicode when running in python 3

jtoledo1974:
python-for-android fails to build when running under a python 3 interpreter with the android recipe. There is no decode method in python3 strings.


AndreMiras:
Thanks for the pull request @jtoledo1974 
It's not the most satisfying approach, but probably the safest one. Also it looks better than what we have now which is broken for host Python3.
Let's see if Travis is happy about it before merging.

AndreMiras:
Merged, thanks for the fix

KeyWeeUsr:
Ref #1516.

PR #1517, opened 2018-12-10
[WIP] Fix ctypes.util.find_library() crash with live patch on startup

JonasT:
Fix ctypes.util.find_library() crash with live patch on startup. This is a new variant based on the new common bootstrap area of the previous attempt in #1433


JonasT:
Sorry I would provide more testing insight but I'm stuck on #1520 right now - as soon as I got that one figured out I can post some more in-depth test results

JonasT:
Tested it with my actual real world app, and it works! Since this is vital for using Python 3 + ctypes without monkey-patching around, it would be very useful to have this merged for everyone

JonasT:
I confirm the semantics fix to be working with my real-world app, including with the PySDL2 handling removed. As far as I'm concerned, this is ready for merging

pythonforandroid/bootstraps/common/build/jni/application/src/start.c
```diff
@@ -243,9 +275,69 @@ int main(int argc, char *argv[]) {
...
+    "        for lib_search in [app_root, app_root_arm, sys_dir]:\n"
+    "            if not os.path.exists(lib_search):\n"
+    "                continue\n"
+    "            for filename in os.listdir(lib_search):\n"
+    "                if check_name(filename, name):\n"
+    "                    return os.path.join(lib_search, filename)\n"
+    "        return orig_func(*args)\n"
+    "    ctypes.util.find_library = android_find_library_hack\n"
+    "except ImportError:\n"
+    "    pass\n");
```
inclement: Maybe could log something here, like "Could not import ctypes, build your app with libffi if you want ctypes to be present". I think that could help people.

JonasT: I put that in now!

I'm just curious though, is `libffi` shipped per default and would need to be opted out of with some option, or does it actually need to be put into `--requirements` for ctypes to be present at all? Because an opt-out intuitively would seem to make more sense to me, in which case people would already be aware and the warning wouldn't be that important, would it?

pythonforandroid/bootstraps/common/build/jni/application/src/start.c
```diff
@@ -243,9 +275,71 @@ int main(int argc, char *argv[]) {
...
   LOGP("AND: Ran string");
 
+#if PY_MAJOR_VERSION >= 3
+  /* Workaround crash in ctypes.find_library() which is caused by its
+   * internal use of /bin/sh, and looking in the wrong folders on Android.
+   *
+   * The code needs 'env_argument' as baked path, so we need to insert this:
+   */
+  LOGP("ctypes workaround insert");
+  char code_with_placeholder[] = (""
```
AndreMiras: Would it make sense and be easy enough to have this large string loaded from a file instead? That way it would make it easier to edit in case it's needed?

JonasT: The `clean` way of how to do that would I think be to install it into the site-packages, but that might be a little complex to look at. (Like, harder to figure out for people who try to understand how the boot process works) The other alternative I can think of is adding the file to the asset folder, and then directly `open()`ing it in that inline python and using `exec()` or something.

I'm a bit undecided on how much 'better' these two solutions are. They would certainly be easier to edit, but I'm not sure if it wouldn't also be more confusing about what is going on at boot.

JonasT: Basically I can't think of the most obvious mechanism and it would introduce additional complexity, so I'm a bit torn

AndreMiras: OK thanks :+1:

pythonforandroid/bootstraps/common/build/jni/application/src/start.c
```diff
@@ -243,9 +275,71 @@ int main(int argc, char *argv[]) {
...
+    "                    return os.path.join(lib_search, filename)\n"
+    "        try:\n"
+    "            return orig_func(*args)\n"
+    "        except OSError:\n"  // catch bogus error about missing "sh"
+    "            return None\n"
+    "    ctypes.util.find_library = android_find_library_hack\n"
+    "except ImportError:\n"
+    "    print('Could not import ctypes, build your app with '\n"
+    "          'libffi if you want ctypes to be present')\n"
+    "    pass\n");
```
AndreMiras: Super minor: you can now remove that `pass` right?

JonasT: oh right I forgot about that when putting the `print` in (without it, it was still necessary)

pythonforandroid/bootstraps/common/build/jni/application/src/start.c
```diff
@@ -243,9 +275,70 @@ int main(int argc, char *argv[]) {
...
 
+#if PY_MAJOR_VERSION >= 3
+  /* Workaround crash in ctypes.find_library() which is caused by its
+   * internal use of /bin/sh, and looking in the wrong folders on Android.
+   *
+   * The code needs 'env_argument' as baked path, so we need to insert this:
+   */
+  LOGP("ctypes workaround insert");
+  char code_with_placeholder[] = (""
+    "try:\n"
```
AndreMiras: This is a rather super large try/except block :open_mouth: 
What would you think about some more like:
```c
char code_with_placeholder[] = (""
    "orig_func = None\n"
    "try:\n"
    "    import ctypes.util\n"
    "    orig_func = ctypes.util.find_library\n"
    "except ImportError:\n"
    "    print('Could not import ctypes, build your app with '\n"
    "          'libffi if you want ctypes to be present')\n"
    "def android_find_library_hack(*args):\n"
    "    import os\n"
    "    name = args[0]\n"
    "    \n"
    "    # Truncate ending for easier comparison:\n"
    "    if name.find('.so.') >= 0:\n"
    "        name = name.partition('.so.')[0]\n"
    "    if name.endswith('.so'):\n"
    "        name = name[:-len('.so')]\n"
    "    if not name.endswith('.'):\n"
    "        name += '.'\n"
    "    \n"
    "    # Helper function to check lib name:\n"
    "    def check_name(lib_name, search_name):\n"
    "        if filename.endswith('.so') and (\n"
    "                filename.startswith('lib' + search_name) or\n"
    "                filename.startswith(search_name)):\n"
    "            return True\n"
    "        return False\n"
    "    \n"
    "    # Check the user app lib dir and system dir:\n"
    "    app_root = os.path.normpath(os.path.abspath(os.path.join(\n"
    "        '%s', '..', '..', 'lib')))\n"
    "    app_root_arm = os.path.join(app_root,\n"
    "                                'arm')  # fixme: other archs?\n"
    "    sys_dir = '/system/lib'\n"
    "    for lib_search in [app_root, app_root_arm, sys_dir]:\n"
    "        if not os.path.exists(lib_search):\n"
    "            continue\n"
    "        for filename in os.listdir(lib_search):\n"
    "            if check_name(filename, name):\n"
    "                return os.path.join(lib_search, filename)\n"
    "    try:\n"
    "        return orig_func(*args)\n"
    "    except OSError:\n"  // catch bogus error about missing "sh"
    "        return None\n"
    "if orig_func is not None:\n" // only patch if ctypes module is present
    "    ctypes.util.find_library = android_find_library_hack\n");
```

pythonforandroid/bootstraps/common/build/jni/application/src/start.c
```diff
@@ -72,6 +72,38 @@ int file_exists(const char *filename) {
   return 0;
 }
 
+/* Helper function to escape all occurrences of ' as \', and all
+ * occurances of \n and \r with \\n and \\r for inline string literals.
```
AndreMiras: Minor: `s/occurances/occurrences`

pythonforandroid/bootstraps/common/build/jni/application/src/start.c
AndreMiras: Was very minor anyway :+1:

inclement:
> I think ctypes is simply shipped by default, unless explicitly blacklisted via p4a blacklist.txt files. What do you guys think about it?

The python configure/setup.py simply don't build ctypes unless they can find libffi. I think the other python builds bypassed this by implicitly building libffi in the python2 case, and by explicitly building it during the python bundling in the crystax case. With this in mind, the libffi requirement actually looks like the correct p4a way to handle things, although I didn't realise this until looking into it to work out why ctypes wasn't already working.

AndreMiras:
Good to know, thanks @inclement

JonasT:
@inclement ok but then it might make sense to throw it into the default base recipe set anyway, wouldn't it? (similar to pyjnius) in that case the usual app would always ship with it included, although of course having a sane error handling message still doesn't hurt

inclement:
@JonasT Yep, I agree. Sorry, I know you suggested that before, I forgot to reply to that one.

JonasT:
I did? I'm very forgetful :joy:

JonasT:
Tested the new change in real world app, works!

JonasT:
@inclement ok wait, isn't the new pull request by @opacam fixing openssl for python 3?

If yes, isn't this pull request here kinda stupid now? Given the new python3 recipe always builds from scratch so a patch would be a cleaner solution, and crystax is kinda unnecessary, outdated and overly complicated anyway?

Seems to me like we should not merge this and I need to do a proper patch as it is done against python 2 instead

AndreMiras:
Agree to see if #1537 gets merged with its patch. And also agree for not putting CrystaX specific efforts :+1:

PR #1521, opened 2018-12-12
Add google repository at allprojects

wo01:
Add google repository for lint-gradle

https://stackoverflow.com/questions/46972122/could-not-find-com-android-tools-lintlint-gradle-android-studio-3


JonasT:
@AndreMiras @inclement I just came to the same conclusion that this is apparently required now - the google repos should be the first point of download, jcenter only if it's not on there, and due to their recent mass deletion hiccup some libs may not come back to jcenter (what a google employee said in the issue tracker - they didn't really seem to fully know themselves yet). So this change makes a lot of sense. Can we get this tested & merged soon? I'm test-building it for my project right now

JonasT:
Upstream issue is here: https://issuetracker.google.com/issues/120759347

The detailed build error message is here: #1520

PR #1522, opened 2018-12-12
Unify configChanges manifest entry and add missing values

JonasT:
The current manifests don't specify some values in configChanges, e.g. the ones relevant for the side-by-side feature of new Android versions. If these values are missing, the SDLActivity will be restarted - however, SDL2 as a result will simply quit even when the application is in foreground/active without restarting properly, so the application crashes/instant-quits at any of these events.

In this spirit, the pygame bootstrap also seemed to have a lot more of these values added already, avoiding such config-triggered crashes. But for some reason these weren't in the other bootstraps. I unified the values for all bootstraps, and added the missing ones of newer API levels like `density`.

**Note:** to fully get side by side working, we also need #1528 - but with these combined, it works perfectly!


JonasT:
Worked fine for my app! More testing & merging welcome!

PR #1524, opened 2018-12-12
Unify build.py contents

JonasT:
This unifies the different build.py files to get the different options in sync, and make it easier to expand them without all the duplicate code.

Tested it, and it works fine with my app / `sdl2` bootstrap!


pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -212,66 +237,84 @@ def compile_dir(dfn):
     '''
     Compile *.py in directory `dfn` to *.pyo
     '''
+
+    if get_bootstrap_name() != "sdl2":
+        # HISTORICALLY DISABLED for other than sdl2. NEEDS REVIEW! -JonasT
```
AndreMiras: what's up with this?

JonasT: I retained the exact behavior pre-merge, which is that it's disabled for webview and service_only. As the comment says it might be a good idea to look at this and maybe enable it for all, or figure out why - I simply considered this out of the scope of the merge which is mostly supposed to just keep the current behavior for now, but get rid of the duplicate files

JonasT:
That definitely would make sense. However, similarly to the common bootstrap area merge request which could have incorporated this but didn't, I personally prefer to go about this in multiple separate requests. While I agree it looks a bit messy with the conditionals, it is *IMHO* still a lot easier to do changes now, given it's at least in one central obvious place of what the bootstrap differences are

pythonforandroid/bootstraps/common/build/build.py
AndreMiras: OK it makes sense with that explanation. Perhaps you can update the comment to just what you said here

JonasT: I thought with the `historically disabled` it'd be obvious it needs to get looked at & refactored some time soon :grimacing: but I can add some more explanation if you think that'd be helpful

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -296,9 +339,9 @@ def make_package(args):
         with open(args.intent_filters) as fd:
             args.intent_filters = fd.read()
 
-    args.add_activity = args.add_activity or []
-
-    args.activity_launch_mode = args.activity_launch_mode or ''
+    if get_bootstrap_name() == "sdl2":
+        args.add_activity = args.add_activity or []
+        args.activity_launch_mode = args.activity_launch_mode or ''
```
AndreMiras: kinda out of scope, but:
people tend to misused Python ternary operator. It's fine here because you just move the lines, but I find it pretty misleading and can lead to unsuspected bugs

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -355,50 +418,78 @@ def make_package(args):
...
-
-    render(
-        'strings.tmpl.xml',
-        'src/main/res/values/strings.xml',
-        args=args,
-        url_scheme=url_scheme,
-        private_version=str(time.time()))
+    if get_bootstrap_name() == "sdl2":
+        if exists('AndroidManifest.xml'):
+            remove('AndroidManifest.xml')
```
AndreMiras: Minor/question: won't the line below already override the existing file?

pythonforandroid/bootstraps/common/build/build.py
JonasT: Yeah I touched as little as possible to not make things even more complex :grimacing: but keep those comments coming, so I can look into them once I refactor `build.py` in a follow-up request!

pythonforandroid/bootstraps/common/build/build.py
JonasT: Not for sdl2, because there manifest_path points to `src/main/AndroidManifest.xml`. This is also why only for sdl2, this copy line is meaningful. These subtle folder differences are also something that would make a lot of sense to get rid of, once I looked deeper at this - it's just a historical thing again that I just kept as-is, for now

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -355,50 +418,78 @@ def make_package(args):
...
+        # Historically, service_only doesn't support ant anymore.
+        # Maybe we should also drop this for the others? -JonasT
+        render(
+            'build.tmpl.xml',
+            'build.xml',
+            args=args,
+            versioned_name=versioned_name)
+
+    # String resources:
+    if get_bootstrap_name() != "service_only":
```
AndreMiras: I'm confused why this block isn't shared with the previous one since we have two adjacent `if get_bootstrap_name() != "service_only":`

JonasT: Really just a visual thing to keep the blocks separate. It's kind of random/an accident that both happen to concern only the other bootstraps - since these differences are again historical for not any obviously good reason, and should probably be removed anyway. I could move things around some more to try to get everything concerning one bootstrap closer together, but honestly, I think the whole thing should be revamped fully into some inheritance-style variant anyway. So I can move it around if you want, but I'd rather revamp the whole thing later

JonasT: (The order of things in general is also quite historical, I just tried to minimize changes in overall since the diff/changeset is already massive enough)

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -534,8 +645,23 @@ def parse_args(args=None):
...
+            return
+        print("Reading .p4a configuration")
+        with open(".p4a") as fd:
+            lines = fd.readlines()
+        lines = [shlex.split(line)
+                 for line in lines if not line.startswith("#")]
+        for line in lines:
+            for arg in line:
+                args.append(arg)
+    _read_configuration()
```
AndreMiras: What is this function for?
It seems to update `args`, but then it's being overridden again next lines :thinking:

JonasT: `args` before that overriding line is a string array, and afterwards is an ArgumentParser args object with the resulting values from parsing it. This approach is again historical - I suppose it could be renamed for clarity, I'll try to keep that in mind for the refactor later.

That parsing function itself is also historical from the webview and allows specifying additional argument strings that get injected into the parser, which I found quite neat - so I kept it for all bootstraps, since I saw no good reason to limit it. I could also remove it of course, but it does seem like something useful

Issue #1183, opened 2017-11-28
p4a doesn't handle runtime permissions


inclement:
Probably we need some code for this, and maybe an api to make it simple to use.

brentpicasso:
Google play now increased the minimum API version, which requires the app to request runtime permissions for all new google play installations.  It would seem this feature's priority should be raised as a result, as any new app install would need to do this for it's needed permissions.

inclement:
@zworkb Nice work!

JonasT:
Unless the support library is useful for lots of other things, maybe it would be a better idea to not use it? (since it will just further add on top of the complicated combinations of things to install)

With reflection, apparently the permission dialog can be triggered without using it: https://codemammoth.blogspot.com/2016/06/how-to-invoke-checkselfpermission.html so that might be an interesting approach when adding it to the `PythonActivity.java` that could be preferable to depending on the support library

brentpicasso:
Nice job @zworkb. 

Good point @JonasT .  How would this look in PythonActivity - would we scan all of the permissions set in the manifest and then trigger a permission request automatically?

JonasT:
No IMHO that's a bad idea. The user expects the permission request to happen exactly when an action actually runs that needs it - otherwise, it seems like a malicious app that wants unrestricted permissions of everything right at launch to mess your phone up. (Exactly why Google introduced this system)

p4a would need to add API calls for each permission type for the app to consciously request these at runtime, and return True/False depending on whether the user gave permission. For a phone that has an older API without checkSelfPermission() being present, those calls should ideally just return True.

Also it would be good to make sure the functions in PythonActivity are thread-safe and don't access global state, because they are blocking and some apps may want to call them from a side thread to keep some sort of action (e.g. network connections) running. Please note this also means the GIL should be released while PythonActivity is in this function

JonasT:
This is now a blocker for the SDL2 update, see #1528 . I could take a look at this, but **I am really looking for some input** - like, where should the final end user API be? plyer seems like a good place, but then it would be *a requirement for all kivy apps*. Is that what we want?

JonasT:
This is now part of the SDL2 update, see #1528 :smile: :smile:

brentpicasso:
Nice job, you have saved us all!

Is this a generalized approach that deals with any 'high risk' permission, or just addresses SD card access?

JonasT:
@brentpicasso it can be trivially expanded for other permissions, for now I put in external storage read & write.

[...]

Issue #1081, opened 2017-07-22
Remove WRITE_EXTERNAL_STORAGE default permission


inclement:
This may need a small change in Kivy, but is otherwise non-essential.

KeyWeeUsr:
It's forced [here](https://github.com/kivy/python-for-android/blob/0d491b916efd3fdc1671deb8028a89a7936d0c41/pythonforandroid/bootstraps/sdl2/build/templates/AndroidManifest.tmpl.xml#L28), it's needed for https://github.com/kivy/kivy/issues/4777 and fixed in the linked PR.

inclement:
Oh, great, I hadn't seen that PR. I was going to propose also checking the old location first, but otherwise defaulting to the new (better) one. I think that would be worthwhile to fix this annoying problem on Android, even if it's slightly more complex.

We could also use the app's official external storage directory, found by querying the Android API. That would be more consistent with the spirit of the current behaviour.

KeyWeeUsr:
I was thinking about it, but then it'd mean including at least `pyjnius` to each android application, which isn't the happiest solution. We could maybe use the old-ish `android` package/recipe, but that might still need some editing and also requiring from user to include it (or add it to `kivy` as a requirement).

If we can't do any of these cases, we end up with an environment variable. As that thing is set in(/for) [`start.c`](https://github.com/kivy/python-for-android/blob/c45bd81ea15b345815450dc2fd14ff884fb12fbb/pythonforandroid/bootstraps/sdl2/build/src/org/kivy/android/PythonActivity.java#L351), we can use `sys.executable`

inclement:
`pyjnius` is already a Kivy dependency on Android.

KeyWeeUsr:
That kind of skipped my mind completely 😄

fyookball:
I am having a problem and I don't know whether it is related to this issue, but I just built an apk for my project, which gives a parsing error (can't install)...but it works fine if i manually copy the apk file to the phone with a usb cable.  Is there something I can do to force my android phone to properly download the apk?

inclement:
@fyookball There is no reason to think that this issue is related to your problem. Please post instead on the [kivy-users mailing list](https://groups.google.com/forum/#!forum/kivy-users).

JonasT:
Is there a switch to remove this permission for non-kivy projects that use python-for-android?

inclement:
[...]

PR #1528, opened 2018-12-13
Bump SDL2 to 2.0.9 & Add API >=23 runtime permissions API

JonasT:
Bump to SDL2 to 2.0.9.

State: **ready for merge**

- [x] Expand `build.py` to apply `SDLActivity.java` patch (required so this pull request will work for others)
- [x] ~~Add option to enable/disable immersive mode to `build.py` and `AndroidManifest.tmpl.xml`, and make it disabled per default to keep old behavior - might conflict with #1522~~ **upstream SDL2 bug:** https://bugzilla.libsdl.org/show_bug.cgi?id=4424 (makes navigation annoyingly slide away in windowed when it shouldn't, workaround is to set window fullscreen & back to windowed at runtime)
- [x] Find a new approach to trigger `keepActive` for `PythonActivity.java` to fix loading screen not disappearing on its own (can be worked around by user by tabbing in & out of the app until it works)
- [x] Get in some way of triggering sd card runtime permission dialog, because SDL2 requires target API >= 26 where sd card access isn't otherwise possible, and document it - also see #1183
- [ ] Fix kivy not dealing properly with runtime sd card permission - also see #1081 and https://github.com/kivy/kivy/issues/6081
- [x] Fix kivy not properly obeying fullscreen/windowed mode on launch by making it use env vars
- [x] ~~Document new API requirement >= 26, and possibly add an instant abort for lower API levels when using the `sdl2` bootstrap~~ **added clear note in docs that other target API levels than suggested 27 are less likely to work** (such that I didn't need to specify an exact minimum that we'll forget to update later)
- [x] Back button might be broken. Investigate... --> works fine when new `SDL_ANDROID_TRAP_BACK_BUTTON` hint is set to `1`

Example use for new runtime permissions API added by this pull request:

```
from android.permissions import Permission, request_permission,\
    check_permission
request_permission(Permission.WRITE_EXTERNAL_STORAGE)
```

Gains:

- CTRL key **finally** works properly on physical keyboards
- Side-by-side works properly (with #1522 merged), including contents resizing properly to half the screen
- All other bugfixes that happened between 2.0.4(?) and 2.0.9
- API >=23 runtime permission API added to p4a's `android` module

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -14,6 +14,7 @@ import shutil
 import subprocess
 import sys
 import tarfile
+import tempfile
 import time
 from zipfile import ZipFile
 
@@ -273,8 +274,17 @@ main.py that loads it.''')
     # construct a python27.zip
     make_python_zip()
 
+    # Add extra environment variable file into tar-able directory:
+    env_vars_tarpath = tempfile.mkdtemp(prefix="p4a-extra-env-")
+    with open(os.path.join(env_vars_tarpath, "p4a_env_vars.txt"), "w") as f:
+        f.write("P4A_IS_WINDOWED=" + str(args.window) + "\n")
+        if hasattr(args, "orientation"):
+            f.write("P4A_ORIENTATION=" + str(args.orientation) + "\n")
+        f.write("P4A_NUMERIC_VERSION=" + str(args.numeric_version) + "\n")
+        f.write("P4A_MINSDK=" + str(args.min_sdk_version) + "\n")
+
     # Package up the private data (public not supported).
-    tar_dirs = []
+    tar_dirs = [env_vars_tarpath]
     if args.private:
         tar_dirs.append(args.private)
     for python_bundle_dir in ('private', 'crystax_python', '_python_bundle'):
@@ -286,6 +296,9 @@ main.py that loads it.''')
         make_tar(
             join(assets_dir, 'private.mp3'), tar_dirs, args.ignore_path,
             optimize_python=args.optimize_python)
+
+    # Remove extra env vars tar-able directory:
+    shutil.rmtree(env_vars_tarpath)
 
     # Prepare some variables for templating process
     res_dir = "src/main/res"
@@ -481,6 +494,31 @@ main.py that loads it.''')
     else:
         if exists('build.properties'):
             os.remove('build.properties')
+
+    # Apply java source patches if any are present:
+    if exists(join('src', 'patches')):
+        print("Applying Java source code patches...")
+        for patch_name in os.listdir(join('src', 'patches')):
+            patch_path = join('src', 'patches', patch_name)
+            print("Applying patch: " + str(patch_path))
+            try:
+                subprocess.check_output([
+                    # -N: insist this is FORWARd patch, don't reverse apply
+                    # -p1: strip first path component
+                    # -t: batch mode, don't ask questions
+                    "patch", "-N", "-p1", "-t", "-i", patch_path
+                ])
+            except subprocess.CalledProcessError as e:
+                if e.returncode == 1:
+                    # Return code 1 means it didn't apply, this will
+                    # usually mean it is already applied.
+                    print("Warning: failed to apply patch (" +
+                          "exit code 1), " +
+                          "assuming it is already applied: " +
+                          str(patch_path)
+                         )
+                else:
+                    raise e
 
 
 def parse_args(args=None):
--- a/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonActivity.java
+++ b/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonActivity.java
@@ -7,33 +7,40 @@ import java.io.FileOutputStream;
 import java.io.FileWriter;
 import java.io.File;
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.UnsatisfiedLinkError;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
-import java.util.ArrayList;
+import java.util.Timer;
+import java.util.TimerTask;
 
-import android.view.ViewGroup;
-import android.view.SurfaceView;
 import android.app.Activity;
-import android.content.Intent;
-import android.util.Log;
-import android.widget.Toast;
-import android.os.AsyncTask;
-import android.os.Bundle;
-import android.os.PowerManager;
-import android.graphics.PixelFormat;
-import android.view.SurfaceHolder;
 import android.content.Context;
+import android.content.Intent;
 import android.content.pm.ActivityInfo;
 import android.content.pm.PackageManager;
 import android.content.pm.ApplicationInfo;
-import android.content.Intent;
-import android.widget.ImageView;
-import java.io.InputStream;
 import android.graphics.Bitmap;
 import android.graphics.BitmapFactory;
 import android.graphics.Color;
+import android.graphics.PixelFormat;
+import android.Manifest;
+import android.os.AsyncTask;
+import android.os.Bundle;
+import android.os.PowerManager;
+import android.util.Log;
+import android.view.SurfaceHolder;
+import android.view.SurfaceView;
+import android.view.View;
+import android.view.ViewGroup;
+import android.view.Window;
+import android.view.WindowManager;
+import android.widget.ImageView;
+import android.widget.Toast;
 
+import org.libsdl.app.SDL;
 import org.libsdl.app.SDLActivity;
 
 import org.kivy.android.PythonUtil;
@@ -51,16 +58,16 @@ public class PythonActivity extends SDLActivity {
     private ResourceManager resourceManager = null;
     private Bundle mMetaData = null;
     private PowerManager.WakeLock mWakeLock = null;
+    private static boolean appliedWindowedModeHack = false;
 
     public String getAppRoot() {
         String app_root =  getFilesDir().getAbsolutePath() + "/app";
         return app_root;
     }
 
-
     @Override
     protected void onCreate(Bundle savedInstanceState) {
-        Log.v(TAG, "My oncreate running");
+        Log.v(TAG, "PythonActivity onCreate running");
         resourceManager = new ResourceManager(this);
 
         Log.v(TAG, "About to do super onCreate");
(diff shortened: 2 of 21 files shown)


JonasT:
After looking at this some more I'm actually seeing a good chance I can get it to work, but I'm currently blocked on #1524 (because it looks like I'll need to do some minor changes to `build.py` which would conflict).

JonasT:
Small update: SDL2 requires minimum target API/SDK 26, and Crystax only supports <= 21. So this won't work with the CrystaX NDK.

Since I don't see a good way to have two SDL2 versions in parallel, this might be a good point to check where we're at with abandoning CrystaX and somehow coordinate it with the SDL2 version upgrade

JonasT:
Works with a few caveats:

- target API must be 26 or higher. this means sd card access is no longer possible without the runtime permission API that p4a doesn't support yet (I'll look into that next, but as a separate pull request)

- hint `SDL_ANDROID_TRAP_BACK_BUTTON` must be set to `1` to get key presses for back button. This is new behavior of SDL2, and kivy might need to be changed to do this

- had to rewrite/change the loading screen removal because that poll method no longer exists. As a result, it will disappear *earlier*, right at the launch of the main method - there's usually a short black screen visible as a result (duration of app's `import`s). However, it's more reliable unlike the old method, which often removed it *way too late* for me. But the black screen is not ideal, so I'm open for improvement ideas

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -500,6 +500,16 @@ def make_package(args):
         if exists('build.properties'):
             os.remove('build.properties')
 
+    # Apply java source patches if any are present:
+    if exists('src/patches'):
+        print("Applying Java source code patches...")
+        for patch_name in os.listdir('src/patches'):
+            patch_path = 'src/patches/' + patch_name
```
AndreMiras: Minor: we usually make path not OS specific, e.g. using `os.path.join` or `os.path.sep`

JonasT:
So yea, everything works other than 2.0.9 apparently breaking windowed mode ( https://discourse.libsdl.org/t/search-for-workaround-for-broken-windowed-mode-on-android/25467 ) and it seems to be a native C code part bug. Sadly, none about the SDL2 devs has responded about it yet. So in theory this merge request is fully complete, but kinda blocked on this 2.0.9 regression

OptimusGREEN:
Nice!

I already have my permissions up and running, the only thing I couldn't get around was the onRequestPermissionsResult issue as this can only be done from the activity.

Is this something that will be implemented into PythonActivity with a callback accessible in python? The reason this is required is if someone has ticked the 'don't ask again box or clicked the deny either by accident or deliberately, then it allows the dev to show a dialog or something to explain that the permission is denied.

I'd do it myself if I understood java enough but to be honest it hurts my head hence why Kivy is the man! :)

JonasT:
@OptimusGREEN I just haven't added it yet because I wanted to get the basics in first :smile: but yes it absolutely makes sense to add it later, and it shouldn't be particularly hard to do. (and you didn't miss anything, right now that information / callback is not accessible)

OptimusGREEN:
@JonasT Superb, looking forward to it.

Zen-CODE:
@JonasT. Thanks for this :+1: I think this is a going to be a great step for Android support. Once it's merged, I plan to help document the changes as put together a simple example showing how to use the runtime permission API. I think we need to have a run-able example to illustrate exactly this stuff works :-)

misl6:
I've tested it on my real-world app with a huge set of recipes.

I've encountered 2 bugs.

Bug 1)
Every **even** compilation fails during java build. I suppose that you're patching the SDLActivity.java even if it is already.

Bug 2)
During the startup process, app automatically rotates to landscape.

This is my current buildozer setup for orientation:
`# (str) Supported orientation (one of landscape, portrait or all)`
`orientation = portrait`

If, during runtime, I'm changing the rotation programmatically, the app set-up to requested orientation without any problem.

JonasT:
Regarding 1) should be fixed (although if you tried the previous version before I fixed this just now, you might need to delete your distribution folder)

Regarding 2): I tested windowed mode portrait in different configurations (windowed manifest + SDL_Window created windowed, fullscreen manifest + SDL_Window created windowed) and it worked fine, so this needs to be something connected to fullscreen mode. I'll test that next

inclement:
It turns out the issue with the app automatically rotating is because SDL-2.0.9 ignores the app's normal orientation setting and always tries to set a new value based on the contents of the "SDL_IOS_ORIENTATIONS" environment variable (despite the name, this is documented as applying to both Android an iOS). By default it uses `"LandscapeLeft LandscapeRight`, although I'm not sure where it gets that from.

If the environment variable doesn't have a valid value, SDL2 still sets the orientation to landscape or portrait based on the aspect ratio of the display, so there seems to be no simple way to skip this.

Also, when I did set the env var to have value `""` or `"Portrait"`, the window is fullscreen despite using `--window` (maybe a different and similar SDL hint issue?), and the drawing surface is in the wrong place until I tap the screen.

JonasT:
I think what should be the correct way is to set the env var, then find out/get upstream to fix the related fullscreen/placement issue. From seeing the code I don't think SDL can reasonably know the orientation otherwise, so this is the only way that makes sense to me this can work unless the app/kivy is expected to set orientation on its own

Still on the train right now, but I'll look into it and see what can be done

inclement:
Well, SDL2 wasn't doing this before, right? I guessed this is more just a new hint they introduced is a simpler cross-platform api, its behaviour is okay except that it's mildly annoying to have no 'do nothing' option that will let the app use the AndroidManifest.xml orientation.

Do you know why the app isn't windowed any more when `--window` is used? Is this the issue you explained above?

JonasT:
Yes, I have a theory: SDL2's windowed mode was broken because they added code to get rid of status and action bar dynamically to ensure constant fullscreen mode. This code can't distinguish windowed mode and reappearing bars in fullscreen, and therefore needs to check SDL_Window's creation flags to know whether the app is meant to be in fullscreen right now (which it failed to do properly, that was the bug that broke windowed mode always). I suspect this code, which is now properly being triggered only in fullscreen mode, also resets the orientation inadvertendly - hence this only happening for fullscreen apps with this pull request right now, and never happening in early SDL2 versions (which didn't have this function at all, and therefore had unreliable fullscreenmode on new android versions - something we don't want either)

Now I suspect this function has no clue about orientation either APART from the hints - which is why that seems like the obvious approach to me. I suspect the hints triggering fullscreen mode themselves is a separate bug and once that one is fixed, I imagine this could work nicely. Without the hints I'm not sure the function can easily know the orientation the app is supposed to have and the hints already exist after all, so it makes sense to me to use them

inclement:
For reference (as discussed on discord), the problem is that SDL-2.0.9 uses certain SDL2 settings to control the Android window, whereas the older SDL2 ignored them on Android. Kivy passes some of these settings, which causes problems.

The following code should make windowed Kivy apps work, place it before any other imports.
```
import os
os.environ['SDL_IOS_ORIENTATIONS'] = ""
from kivy.config import Config
Config.set('graphics', 'fullscreen', '0')
```
Alternatively setting fullscreen to `'1'` should work too, if you don't pass the --window argument.

We'll need to either retire the `--window` and `--orientation` options of the SDL2 bootstrap, or make sure these are stored and used when the code runs on the device. In principle it might be best to remove the options and leave it to applications to do what they want when they create the SDL2 window, but that would be a relatively big api change (since these are pretty core options) and also would need documenting.

Alternatively we could add some more code to pass these options into the device.

JonasT:
just also for reference, I just agreed on discord I will make sure the options are passed as env vars as part of this pull request. then kivy just needs to be changed to pick them up and things should work as before

Fak3:
Could someone please tell if it will be possible to build app that supports Android 4.4 api level 19 after this is merged?

JonasT:
@Fak3 I think current minimum API level is actually 21, so I don't think so. This is however unrelated to this pull request, that only changes the target api level's minimum, which doesn't change compatibility with older devices

JonasT:
I added environment options now, P4A_IS_WINDOWED (which can be "True" or "False") and another one being P4A_ORIENTATION (which can be e.g. "user", "portrait", ... just whatever that setting is set to).

They're set in `build.py` and it shouldn't be too hard to expand to more in the future if we ever should want to add any

inclement:
I opened https://github.com/kivy/kivy/pull/6107 to fix Kivy to work with these env vars. It works fine except for a back button issue mentioned there, which might have a simple fix.

JonasT:
For what it's worth, we discussed the issue @guysoft had in chat, and from what I can read from the docs this is somewhat expected. (The permission doesn't appear to allow writing *anywhere*, just to a few common folders like `Documents`, ... To pick other locations, the graphical dialog API probably needs to be used, but this is really beyond the scope of this pull request) So as far as I'm concerned this is still good & ready for merging!

guysoft:
I can confirm that here.
Path that does work is ``/storage/external_SD/Android/data/com.your.app.name/files`` on my lollipop device.
Where ``external_SD`` changes in different vendors. 
Also it seem that latter Nugat threre was an API added for this but I don't have a device to test.

JonasT:
Tested this with the fixed keyboard app, and I can back out with the return button just fine. @inclement is there more that needs to be tested to verify it? Otherwise, this is ready for merge as far as I'm concerned

inclement:
Works super, thanks again!

pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonActivity.java
```diff
@@ -364,118 +385,219 @@ public static void stop_service() {
...
+        } catch (IllegalAccessException | NoSuchMethodException |
+                 InvocationTargetException e) {
+        }
+        return false;
+    }
+
+    /**
+     * Used by android.permissions p4a module to request a permission
+     **/
+    public void requestNewPermission(String permission) {
```
AndreMiras: I've started playing with this recently and it seems to work, thanks for this @JonasT 
Any reason why we didn't stick to official API signature? I mean why isn't this called `requestPermissions` and why doesn't it take a list of permission string?

OptimusGREEN:
any word on a getPermissionsRequestResult callback for python?q that way we could get the result instantly if the "don't ask again" box has previously been ticked?

pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonActivity.java
JonasT: @AndreMiras oh right that would have been smarter I guess :grimacing:

AndreMiras: Thanks @JonasT :pray: 
I just wanted your feedback on it. I've created a ticket to address it
https://github.com/kivy/python-for-android/issues/1704

PR #1532, opened 2018-12-15
Use API 27 as new default for travis & docs

JonasT:
Use API 27 as default for travis & examples, including `armeabi-v7a` as arch default.
26 is new absolute minimum of the Play Store, and otherwise the new SDL2 2.0.9 pull request won't build.


JonasT:
Good point, I properly bumped the actual test files, and also the introduction docs! I think it's actually better to have the environment setup in the test files rather than the `.travis.yml`, I only did that because I hadn't seen them :laughing:

PR #1534, opened 2018-12-17
Refactor python recipes and update python2's to version 2.7.15 (fixed history)

opacam:
**Originally started in pr #1460** (moved here to solve the history conflicts)

This pr is intended to complement the amazing work of @inclement in his python3's recipe #1412. It refactors the python3's recipe into the recipe module and updates the python2 version to 2.7.15 using the same build process than the refactored python3's recipe. So, python2 and python3 will share the same build system and allow us to remove some specific source code for python2. Right now the build process completed successfully for both python versions and the python3 recipe also works as expected when installed on android devices...so...

Things to do:

  - [x] Fix python2's initialization on android device
  - [x] Make neatest the python3's refactoring
  - [x] Update openssl libraries to grant support for python2 and python3 (trying to not lose support for python2legacy)
  - [x] Enable libraries support (for both python2 and python3)
  - [x] Review the patches
  - [x] Test that python3crystax is not broken
  - [x] Fix some hardcoded python2 entries
  - [x] Fix webview's bootstrap (for both versions of python)
  - [x] Fix service_only's bootstrap (for both versions of python)
  - [x] Fix python2/python3 compatibility for most of the recipes that uses the new openssl lib (ffmpeg and LibSecp256k1 not faced here and are broken...)
  - [x] Enhance library loading mechanism on android side (no more harcoded libs in PythonUtil.java)
  - [x] Add tests apps to ci for webview and service_only bootstraps
  - [x] Add tests apps to ci for most of the encryption recipes (openssl dependant)
  - [x] Reduce travis log output, to not face the travis 4mb limit
  - [x] Properly document the changes applied in this pr

Things to do, probably better to do it apart:

  - Check/fix if image related recipes are working: libjpeg, libpng and pillow
  - Fix pygame using the old python recipe or remove it


pythonforandroid/bootstrap.py
```diff
@@ -257,6 +266,81 @@ def _unpack_aar(self, aar, arch):
...
+        # special shared libs, like stl libs used at compile time or libs that
+        # the bootstrap may be dependant).
+        # Todo: All the recipes using gnustl_shared should be migrated to use
+        # c++_shared because in ndk r18 it will be the only stl available:
+        # https://developer.android.com/ndk/guides/cpp-support
+        stl_gnu_shared_recipes = (
+            'icu', 'leveldb', 'libzmq', 'protobuf_cpp',
+            # CppCompiledPythonRecipes depends on gnustl_shared so...
+            'atom', 'kiwisolver',)
+        stl_cxx_shared_recipes = ('boost',)
```
JonasT: Should this be here? That seems like something that should be set on each of these recipes directly, then scanned here from the global recipe list to compile this on-demand.

But that's just a minor nitpick and maybe this can be improved upon later, in overall, I like the changes I see!

JonasT:
Ok, I have two nitpicks:

- the list that need C++ or C shared libs linked in are hardcoded. I think that should be set in the recipe instead (see comment)

- some `.java` files seem to be copied into each separate bootstrap that kinda look like they actually should be in `common` instead (please note you *can* have a file in `common` and still override it by also adding it to a single bootstrap - so there's no need to copy it separately into *all* bootstraps if just one of them needs a different version)

But in overall, at least from the central changes I checked (I didn't look much into the changed/new recipes), this all looks like considerable improvements. I didn't test run it, but I like what I'm seeing.

I'm therefore in favor of merging this relatively quickly and improving things later if necessary, because it'd be sad to see all this work buried in another pile of conflicts :+1:

pythonforandroid/bootstrap.py
opacam: Yes, you are right, this shoudn't go here,  I will deal with this in the next commit many thanks!! :smile:

opacam:
@JonasT, all done...probably more refactoring could be done for the java files, but at least the new files that I introduced are refactored. Also reduced some tests, maybe should be reduced even more...now we will have 5 jobs (I think that is the limit). For you to know, the most stressant recipe is libtorrent+boost...by removing libtorrent we would improve even more the build timings...

.travis.yml
```diff
@@ -18,11 +18,14 @@ env:
     - ANDROID_SDK_HOME=/opt/android/android-sdk
     - ANDROID_NDK_HOME=/opt/android/android-ndk
   matrix:
-    - COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_python3.py apk --sdk-dir $ANDROID_SDK_HOME --ndk-dir $ANDROID_NDK_HOME --requirements sdl2,pyjnius,kivy,python3'
+    # test encryption recipes
+    - COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_python_encryption.py apk --dist-name bdisttest_python3_encryption --sdk-dir $ANDROID_SDK_HOME --ndk-dir $ANDROID_NDK_HOME --requirements python3,libffi,openssl,sdl2,pyjnius,kivy,cryptography,pycrypto,scrypt,m2crypto,pysha3,pycryptodome,libtorrent'
```
AndreMiras: I would not build all as we're trying to make trade off between build time and "recipe criticality".
Some of these recipes are not that critical to always have to be built, e.g. `libtorrent` and few other being listed here.
We still have some kinda of coverage however thanks to the `./ci/rebuild_updated_recipes.py` script.

.travis.yml
```diff
@@ -31,5 +34,14 @@ before_script:
...
+  - trap "echo \"Error...last docker's log build lines:\" & tail -500 build.log & echo \"Error last tests log lines:\" &tail -1000 tests.log" ERR
+  # The Docker image build
+  - while sleep 1m; do echo "=====[ $SECONDS seconds, docker image still building... ]====="; done & PING_LOOP1_PID=$!
+  - docker build --tag=p4a --file Dockerfile.py3 . > build.log 2>&1
+  - kill ${PING_LOOP1_PID}
+  # The Tests build
+  - while sleep 1m; do echo "=====[ $SECONDS seconds, tests still building... ]====="; done & PING_LOOP2_PID=$!
+  - docker run p4a /bin/sh -c "$COMMAND" > tests.log 2>&1
+  - kill ${PING_LOOP2_PID}
+  - echo "All operations performed...last build log lines are:" & tail -500 tests.log
```
AndreMiras: I would rather try to avoid all these "ping-hacks" when it's possible.
If we need them sometimes it might that we're misusing the CI.
Even though I agree Travis is pretty limited sometimes

.travis.yml.unused
```diff
@@ -0,0 +1,11 @@
...
+
+# Test encryption recipes
+- COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_python_encryption.py apk --dist-name bdisttest_python2_encryption --sdk-dir $ANDROID_SDK_HOME --ndk-dir $ANDROID_NDK_HOME --requirements python2,libffi,openssl,sdl2,pyjnius,kivy,cryptography,pycrypto,scrypt,m2crypto,pysha3,pycryptodome,libtorrent'
+
+# test webview bootstrap (plus numpy)
+- COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_flask.py apk --dist-name bdisttest_python2_flask_webview --sdk-dir $ANDROID_SDK_HOME --ndk-dir $ANDROID_NDK_HOME --requirements python2,flask,pyjnius,numpy'
+
+# test service only bootstrap (plus numpy)
+- COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_service.py apk --dist-name bdisttest_python3_service --sdk-dir $ANDROID_SDK_HOME --ndk-dir $ANDROID_NDK_HOME --requirements python3,genericndkbuild,numpy'
+- COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_service.py apk --dist-name bdisttest_python2_service --sdk-dir $ANDROID_SDK_HOME --ndk-dir $ANDROID_NDK_HOME --requirements python2,genericndkbuild,numpy'
```
AndreMiras: then I would drop this file if it's not used

Dockerfile.py2
```diff
@@ -26,7 +26,7 @@ RUN apt -y update -qq \
 
 
 ENV ANDROID_NDK_HOME="${ANDROID_HOME}/android-ndk"
-ENV ANDROID_NDK_VERSION="16b"
+ENV ANDROID_NDK_VERSION="17c"
```
AndreMiras: doesn't this kind of change already deserve it's own PR for instance?
That why it's easier to justify why it's needed and eventually make it early (or not) to the tree

Dockerfile.py3
```diff
@@ -15,7 +15,7 @@
 #     Use 'docker run' without '--rm' flag for keeping the container and use
 #     'docker commit <container hash> <new image>' to extend the original image
 
-FROM ubuntu:18.04
+FROM python:3.7-stretch
```
AndreMiras: I've always been puzzled about base image not clearly coming from a well known distrib and stuff.
What I mean here, is even though the Ubuntu base image is larger than some other ones, it makes it explicit the project compiles/run fine on Ubuntu.
Of course if changing the base image make a 50% time gain, then it can be discussed/justified. If it's a 10% then I'm not sure personally.
So this kinda of change also could be or not hard to make it to master sometimes even as single PR.
It deserves to be discussed, but I would rather have it on a dedicated PR since it could start some long discussions.

Dockerfile.py3
```diff
@@ -104,7 +104,7 @@ RUN dpkg --add-architecture i386 \
 # specific recipes dependencies (e.g. libffi requires autoreconf binary)
 RUN apt -y update -qq \
     && apt -y install -qq --no-install-recommends \
-        autoconf automake cmake gettext libltdl-dev libtool pkg-config \
+        libffi-dev autoconf automake cmake gettext libltdl-dev libtool pkg-config \
```
AndreMiras: I'm still not sure if we want by design to pickup the recipes or the system lib. Since the recipe could also be installed on host. I don't know what's the orinal plan for p4a and libffi.

Dockerfile.py3
JonasT: good catch. I'm also in favor of sticking to the ubuntu base image, simply because it has a more widely known configuration

pythonforandroid/archs.py
```diff
@@ -30,13 +30,24 @@ def include_dirs(self):
                 d.format(arch=self))
             for d in self.ctx.include_dirs]
 
-    def get_env(self, with_flags_in_cc=True):
+    def get_env(self, with_flags_in_cc=True, clang=False):
```
AndreMiras: So basically that file change mainly makes it possible to compile with `clang`.
That's very interesting and probably deserves a PR on its own to make it faster to the tree

pythonforandroid/bootstrap.py
```diff
@@ -257,6 +266,83 @@ def _unpack_aar(self, aar, arch):
...
+            # `pythonforandroid.build.copylibs_function`?
+            for l in listdir(self.ctx.get_libs_dir(arch.arch)):
+                if l.startswith('libboost_'):
+                    libs_to_load.append(l[3:-3])
+        if 'libtorrent' in self.ctx.recipe_build_order:
+            libs_to_load.append('torrent_rasterbar')
+        if 'libiconv' in self.ctx.recipe_build_order:
+            libs_to_load.append('iconv')
+        if 'libzbar' in self.ctx.recipe_build_order:
+            libs_to_load.append('zbar')
```
AndreMiras: It feels kinda weird to have these specific use cases in a core module rather than in a dedicated recipe. What do you think?

pythonforandroid/bootstrap.py
```diff
@@ -269,14 +355,11 @@ def strip_libraries(self, arch):
...
-                             join(self.dist_dir, 'libs'),
-                             '-iname', '*.so', _env=env).stdout.decode('utf-8')
-        else:
-            filens = shprint(sh.find, join(self.dist_dir, '_python_bundle', '_python_bundle', 'modules'),
-                             join(self.dist_dir, 'libs'),
-                             '-iname', '*.so', _env=env).stdout.decode('utf-8')
+        filens = shprint(sh.find, join(self.dist_dir, '_python_bundle',
+                                       '_python_bundle', 'modules'),
+                         join(self.dist_dir, 'libs'),
+                         '-iname', '*.so', _env=env).stdout.decode('utf-8')
```
AndreMiras: here the logic is unified for both `python2` and `else` and that works? Then that's actually is a good thing and might be something we want to isolate/demonstrate in a dedicated PR

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -342,9 +338,9 @@ def make_package(args):
         with open(args.intent_filters) as fd:
             args.intent_filters = fd.read()
 
-    if get_bootstrap_name() == "sdl2":
-        args.add_activity = args.add_activity or []
-        args.activity_launch_mode = args.activity_launch_mode or ''
+    # if get_bootstrap_name() == "sdl2":
```
AndreMiras: minor: why is this commented rather than deleted?

pythonforandroid/bootstraps/common/build/gradlew.bat
```diff
@@ -1,90 +1,90 @@
...
+:fail
+rem Set variable GRADLE_EXIT_CONSOLE if you need the _script_ return code instead of
+rem the _cmd.exe /c_ return code!
+if  not "" == "%GRADLE_EXIT_CONSOLE%" exit 1
+exit /b 1
+
+:mainEnd
+if "%OS%"=="Windows_NT" endlocal
+
+:omega
```
AndreMiras: I cannot grasp why this file was entirely changed? I saw the `git mv` from `sdl2` to `common`, but I don't get why the whole file is "diff-ed". Is it because of line ending?

pythonforandroid/bootstraps/service_only/build/src/main/java/org/kamranzafar/jtar/Octal.java
```diff
@@ -1,141 +1,141 @@
...
+     * @return The long value of the octal bytes.
+     */
+    public static int getLongOctalBytes(long value, byte[] buf, int offset, int length) {
+        byte[] temp = new byte[length + 1];
+        getOctalBytes( value, temp, 0, length + 1 );
+        System.arraycopy( temp, 0, buf, offset, length );
+        return offset + length;
+    }
+
+}
```
AndreMiras: Also curious why this file and other java files are diffing so much. Is it a github glitch or something else happened here? Or it's something I'm missing?

pythonforandroid/bootstraps/webview/build/src/main/java/org/kivy/android/PythonService.java
```diff
@@ -12,7 +12,9 @@
 import android.app.Notification;
 import android.app.PendingIntent;
 import android.os.Process;
+import java.io.File;
```
AndreMiras: Minor: why is this import for?

pythonforandroid/bootstraps/webview/build/templates/strings.tmpl.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
     <string name="app_name">{{ args.name }}</string>
-    <string name="private_version">0.1</string>
+    <string name="private_version">{{ private_version }}</string>
```
AndreMiras: Is this something we were missing?
It may deserve it's own PR then

pythonforandroid/logger.py
```diff
@@ -187,14 +197,21 @@ def shprint(command, *args, **kwargs):
...
+                        Err_Style.RESET_ALL, msg_hdr,
+                        'wait...', width=msg_width))
+                    stdout.flush()
+                    need_closing_newline = True
             else:
                 logger.debug(''.join(['\t', line.rstrip()]))
+            n_line += 1
         if need_closing_newline:
             stdout.write('{}\r{:>{width}}\r'.format(
                 Err_Style.RESET_ALL, ' ', width=(columns - 1)))
```
AndreMiras: Definitely deserves to be discussed in its own PR

AndreMiras:
Kudos for that amazing PR, the amount of work is just insane! Thank you for all your follow-up efforts.
Unfortunately the PR is too big for my human brain to process as I'm more used to break down problems into smaller pieces.
I left some random comment here and there. But there's so much code involved that I don't know where to start reviewing, what to comment.... I personally think breaking down PR usually make them faster to the tree.
This is 200+ files of diff and while walking through it I felt like it could be "easily" be broken down into ~10 PR.
For instance I would totally take some recipes out of it even it it means leaving them broken. Random example, the `idna` recipe update that has a version bump and `python3` added to the `depends` doesn't necessarily need to be in that PR. That just a random trivial example, but I would pretty much generalise it for other recipes as well.
I have a lot more things to share or ask about that PR, but it would probably not be so productive to write a massive comment about it and not convenient for you to tackle my comments/questions in another huge reply.
I hope you don't get me wrong. I love to see contributions especially the ones that require such level of knowledge to improve the core deeply like you did. But we really need to find a more workable way to understand/test and integrate changes and I believe it's by doing smaller increments.

opacam:
Ok guys, don't worries, all the heavy job is done so split into small pieces should not be a problem, basically because when I did all this stuff I tried to follow a temporal line and I tried to group the parts, so it could be easily followed, all of that were lost at the moment of squashing...so reviewing all this changes here is not a good idea, let's do it the right way.

 I also didn't realize that this will ended up with soo many changes...anyway...all the work until now has allowed us to see that this works and it was necessary to properly test some parts. Saying that, I already pushed a new pr with only the core of the python refactoring, still huge (about the half of all the stuff...more or less), but hope that more manageable. The next part should be the bootstraps fixes, this also will be big but not as big as the core part. All the other stuff shouldn't be a problem. By the way, all the changes related with supressing the output of ci tests were necessary to pass the bootstrap tests and the encryption test, cause this tests will not be enabled in travis, we can omit all of that.

Issue #1501, opened 2018-12-02
Need libpython3.7m.so.1.0 in android phone


homdx:
Hello.
I build with this instruction:

https://groups.google.com/forum/#!topic/kivy-users/Lf4zlYmLVPo

Apk compiled and builded. After setup and launch in my Android phone, I have a error:
```
SDL Error

An error occurred while trying to start the applicaion. Please try again and/or reinstall.
Error: dl open failed: could not load library: "libpython3.7m.so.1.0" needed by "libmain.so"; caused by library "libpython3.7m.so.1.0" not foud

log from logcat:

V/pythonutil(11885): Checking pattern libcrystax\.so against libSDL2_mixer.so
V/pythonutil(11885): Checking pattern libcrystax\.so against libSDL2_ttf.so
V/pythonutil(11885): Checking pattern libcrystax\.so against libmain.so
V/pythonutil(11885): Checking pattern libcrystax\.so against libSDL2.so
V/pythonutil(11885): Checking pattern libcrystax\.so against libSDL2_image.so
V/pythonutil(11885): Checking pattern libcrystax\.so against libpython3.7m.so
V/pythonutil(11885): Checking pattern libsqlite3\.so against libSDL2_mixer.so
V/pythonutil(11885): Checking pattern libsqlite3\.so against libSDL2_ttf.so
V/pythonutil(11885): Checking pattern libsqlite3\.so against libmain.so
V/pythonutil(11885): Checking pattern libsqlite3\.so against libSDL2.so
V/pythonutil(11885): Checking pattern libsqlite3\.so against libSDL2_image.so
V/pythonutil(11885): Checking pattern libsqlite3\.so against libpython3.7m.so
V/pythonutil(11885): Checking pattern libssl.*\.so against libSDL2_mixer.so
V/pythonutil(11885): Checking pattern libssl.*\.so against libSDL2_ttf.so
V/pythonutil(11885): Checking pattern libssl.*\.so against libmain.so
V/pythonutil(11885): Checking pattern libssl.*\.so against libSDL2.so
V/pythonutil(11885): Checking pattern libssl.*\.so against libSDL2_image.so
V/pythonutil(11885): Checking pattern libssl.*\.so against libpython3.7m.so
V/pythonutil(11885): Checking pattern libcrypto.*\.so against libSDL2_mixer.so
V/pythonutil(11885): Checking pattern libcrypto.*\.so against libSDL2_ttf.so
[...]

Issue #1525, opened 2018-12-12
Android NDK 18b issues


tito:
Libffi latest stable version is incompatible with Android NDK 18b:
```
../src/arm/sysv.S:363:2: error: invalid instruction, did you mean: fldmiax?
 fldmiadgt ip, {d0-d7}
 ^
../src/arm/sysv.S:399:2: error: invalid instruction
 stmeqia r2, {r0, r1}
 ^
../src/arm/sysv.S:412:2: error: invalid instruction, did you mean: fstmiax?
 fstmiadeq r2, {d0-d3}
 ^
/tmp/sysv-5d4296.s:263:2: error: invalid instruction
 fstmfdd sp!, {d0-d7}
 ^
/tmp/sysv-5d4296.s:313:2: error: invalid instruction, did you mean: fldmiax?
 fldmiad sp, {d0-d1}
 ^
/tmp/sysv-5d4296.s:316:2: error: invalid instruction, did you mean: fldmiax?
 fldmiad sp, {d0-d3}
```

It appears to be because of clang 7.0, which doesn't accept deprecated functions. 
The Android libffi version have fixes: https://android.googlesource.com/platform/external/libffi/+/ca22c3cb49a8cca299828c5ffad6fcfa76fdfa77

Didn't tested if libffi master have the fixes, but stable don't.

inclement:
I found this still happened with NDK 19 in #1612, so we'll need to deal with this it seems.

dbrnz:
FYI, libffi can be made to compile with NDK 19 by using setting  `version = 3.3-rc0` and `patches = []` in the libffi recipe.

inclement:
Thanks @dbrnz, that's great to know.

misl6:
Fixed via #1722

PR #1537, opened 2018-12-18
[CORE UPDATE - PART I] Refactor python recipes + openssl + sqlite3

opacam:
This pr unifies the build system for the python recipes using the `native build method` used by @inclement in his python3 recipe. This is a big change because using the new build method allows us to bump the version number for the python2 recipe to version 2.7.15 and the old python2 recipe has been moved to a new recipe `python2legacy`, in order that we have a failsafe python recipe, in case that we must target a minimum api lower than 21 (because the new python recipe has the same limitation than the python3 recipe, `minimum api >= 21`).

Actions taken care here:
  - [x] Refactor python recipes into a new core module `~pythonforandroid.python`
  - [x] Add ability to compile recipes with clang **This is @tito's works, ¡¡¡thanks for that @tito!!!**
  - [x] Update openssl version to 1.1 **This is @tito's works, ¡¡¡thanks for that @tito!!!**
  - [x] Add libs support for both versions of python: sqlite3 and openssl
  - [x] Make sure that we can build with all architectures (Fix some of the hardcoded entries we used in our archs module)
  - [x] Make it work the old python2 recipe as a new python recipe `python2legacy` (this one is the one originally created by @tito and reworked by @inclement)

The build has been tested for all archs, but I only could test in a device with armv7a...so it will be great if someone could test this with another archs.

*Note for python2legacy*:
  - **This python2legacy recipe may be removed soon** and, if anyone needs to use it, it may require to change the attribute `depends` of some recipe...because only has been granted a minimal compatibility, to create a simple application using sdl2 bootstrap with kivy, openssl and sqlite. Also should be mentioned that if you make use of the python2legacy recipe, the openssl libs installed will be the old ones (version 1.0) because the new ones are incompatible with the old python. This add a complexity in our build system that we may want to remove very soon...but for now...it may allow us to make a softer transition between python2legacy to python3, first migrating all the recipes to be compatible with the new python2 build system so we can safely remove python2legacy, and  once we get in there...then we probably will slowly face out python2 support towards 2020 (the last planned release for python2)

*Note for non sdl2 bootstraps*:
  - **pygame bootstrap**: Contains a lot of hardcoded entries for the old python2 build, so it will not work until properly fixed in another pr. My intention here was to safe the pygame bootstrap using the python2legacy recipe...but we will see...because the python2legacy recipe may be removed soon and I'm not sure if we must maintain this pygame bootstrap because without the python2legacy recipe it may be difficult to make it work. 
  - **service_only bootstrap**: Again, hardcoded entries for python 2, but this is fixed in pr #1541 plus in that pr we will make it work for python3 too
  - **webview bootstrap**: Again, hardcoded entries for python 2, but this is fixed in pr #1541 plus in that pr we will make it work for python3 too

*Note for openssl affected recipes:*

Well, the openssl update and the new python2 build system breaks most of the recipes that relies on openssl because most of them contains hardcoded entries of the old python2. All those affected recipes has been modified to work with the new python build system, plus we granted python3's compatibility to all of them in the corresponding prs:

  - [CORE UPDATE - PART IV] Fix cryptography+cffi (#1545)
  - [CORE UPDATE - PART V] Fix pycrypto (#1546)
  - [CORE UPDATE - PART VI] Fix scrypt (#1547)
  - [CORE UPDATE - PART VII] Apsw for both versions of python (#1548)
  - [CORE UPDATE - PART VIII] Netifaces for both versions of python (updates the netifaces version) (#1549)
  - [CORE UPDATE - PART IX] Pymunk for both versions of python and enhance flags (#1550)
  - [CORE UPDATE - PART X] Protobuf_cpp fixes and updated version (#1551)
  - [CORE UPDATE - PART XI] M2crypto for both versions of python and updated version (#1552)
  - [CORE UPDATE - PART XII] Pycryptodome for both versions of python (#1553)
  - [CORE UPDATE - PART XIII] Pysha3 for both versions of python (#1554)
  - [CORE UPDATE - PART XIV] Libtorrent+boost for both versions of python and updated versions (#1555)
  - *[CORE UPDATE - PART XV] Add encryption test app (#1556)*

*Note for graphical affected recipes*:
  - **Pillow and pil needs to be updated** to work with new python build system. I already have done that (I mean fix the hardcoded entries and grant compatibility for both versions of python) but still not created the proper prs...because I touch some libraries needed by this recipes (libjpeg, png, freetype, harfbuzz) and still needs to review it and properly test. 

*Other recipes that may need touches*:
  - I think that lxml was failing with all this changes (now I'm not sure about that)...anyway...also I had granted compatibility for both versions of python for lxml (Also I touch  libxml2 and lxslt). For this still not created prs... I need to review before push those changes...

**Important note**

This pr needs one last fix, it may be some users that face the issues described in #1501, if that is the case, you can use a patch to solve this. Look at the **reverted** commit 99ee28b, extract a patch and apply it to your fork. We plan to fix it in #1568.

**This has been tested with ndk r17c and also I made some builds using ndkr18b, but this last one needs more testing**

*Development note: This pr was started in #1460 and also discussed in  #1534. Has been split in several prs to make easier to discuss and merge or discard all the initial work made in those prs*


opacam:
> Which openssl-dependant things exactly are broken by this?

All the recipes which uses openssl, mostly the ones I fixed in original pr (plus some others like ffmpeg or libsec..)...except for the python recipes.

opacam:
> And in what way are the other bootstraps broken, is it a paths related issue?

Yes, and the way we pack and unpack the new python's build system is different.

JonasT:
The libraries/stl unpacking? I think these changes are splendid, so I hope since they're now split off that you'll make a pull request for that right after :+1: I really liked those.

Nevertheless, I think we should get this one in ASAP (after a review of course) and not waste more time about what exactly should be in or not, so my vote is on keeping this pull request here as it is (minus minor change requested by the others) and just getting it merged.

opacam:
> Maybe it would be good to at least add in the bootstrap parts as well?

I'm creating a new branch for that right now, with the original content of the pr with the changes we discussed earlier..the .java files unified...

But if you want we could put all in here, but will increase considerably the diff ...because we move all the files to be gradle compatible...so I thought it will be better to put in a new pr as a part 2.

JonasT:
Sure make it as a part 2, because I think the others prefer it as smaller bits. I think I'm the one here who minds large pull requests the least but I don't have a direct vote in merges so I'm kind of irrelevant in that regard :smile: :woman_shrugging:

opacam:
> The libraries/stl unpacking? I think these changes are splendid, so...

Thanks man :smile:, I though that should be the part3, but for that we depend on part 2...and that is only the initial work...all of the libs should be adapted because I suspect that if we build something against any shared lib, we will must load  that lib when we initialize the app, otherwise we will get errors about missing symbols/functions...(at least this is true for all the libraries which python depends on, plus libtorrent, boost, stl...)

pythonforandroid/bootstraps/common/build/jni/application/src/start.c
```diff
@@ -161,6 +157,9 @@ int main(int argc, char *argv[]) {
   Py_Initialize();
 
 #if PY_MAJOR_VERSION < 3
+  // Can't Py_SetPath in python2 but we can set PySys_SetPath, which must
```
JonasT: I'm wondering if we should just omit this change, since SSL works with python 3 now and crystax becomes increasingly unnecessary. I think the better plan is to dump Crystax very soon. OpenSSL is as far as I know the only part remaining that doesn't work with @inclement 's new python 3 recipe, and you're fixing this right here in this pull request.

Or is this also useful/necessary in some way for non-crystax builds?

pythonforandroid/recipes/hostpython2/__init__.py
```diff
@@ -1,60 +1,18 @@
+from pythonforandroid.python import HostPythonRecipe
 
-from pythonforandroid.toolchain import Recipe, shprint, current_directory, info, warning
-from os.path import join, exists
-import os
-import sh
 
+class Hostpython2Recipe(HostPythonRecipe):
```
JonasT: Love how much simpler it looks now! :heart: (assuming it actually still works)

pythonforandroid/recipes/python2/patches/fix-api-minor-than-21.patch
```diff
@@ -0,0 +1,145 @@
...
+ {
+diff -Naurp Python-2.7.15.orig/Modules/signalmodule.c Python-2.7.15/Modules/signalmodule.c
+--- Python-2.7.15.orig/Modules/signalmodule.c	2018-04-30 00:47:33.000000000 +0200
++++ Python-2.7.15/Modules/signalmodule.c	2018-07-05 16:40:46.601022385 +0200
+@@ -32,6 +32,13 @@
+ #include <process.h>
+ #endif
+ 
++//PMPP API<21
++#if __ANDROID_API__ < 21
```
JonasT: shouldn't this be `  +#if defined(__ANDROID_API__) && __ANDROID_API__ < 21` instead? to make it more mergeable for upstream

pythonforandroid/recipes/python2/patches/fix-init-site.patch
```diff
@@ -0,0 +1,17 @@
+--- Python-2.7.2.orig/Python/pythonrun.c	2010-10-29 05:45:34.000000000 +0200
++++ Python-2.7.2/Python/pythonrun.c	2011-04-20 17:52:12.000000000 +0200
+@@ -254,9 +254,13 @@
+     _PyGILState_Init(interp, tstate);
+ #endif /* WITH_THREAD */
+ 
++    /* For PGS4A, we don't want to call initsite, as we won't have the
```
JonasT: Can we add a proper `#define` flag here as well? Again to make it more suitable for a potential upstream merge. Something like `#ifdef NOINITSITE` or similar.

pythonforandroid/recipes/python2legacy/patches/custom-loader.patch
```diff
@@ -0,0 +1,67 @@
...
+         struct stat statb;
+--- Python-2.7.2.orig/Python/pythonrun.c	2010-10-29 05:45:34.000000000 +0200
++++ Python-2.7.2/Python/pythonrun.c	2011-04-20 17:52:12.000000000 +0200
+@@ -254,9 +254,13 @@
+     _PyGILState_Init(interp, tstate);
+ #endif /* WITH_THREAD */
+ 
++    /* For PGS4A, we don't want to call initsite, as we won't have the
++       library path set up until start.pyx finishes running. */
++#if 0
```
JonasT: again a proper `#define` flag like `#ifdef NOINITSITE` would be nice

pythonforandroid/recipes/python2/patches/fix-api-minor-than-21.patch
JonasT: Also I'm thinking, maybe the entire lower-than-21 patch should just be dropped. If you look here: https://developer.android.com/about/dashboards/ pre-5.0 phones are now at less than 25% and this problem will kinda resolve itself soonish, so maybe we just shouldn't support building with lower than NDK API 21 anymore.

pythonforandroid/python.py
```diff
@@ -0,0 +1,385 @@
...
+            recipe = Recipe.get_recipe('openssl', self.ctx)
+            add_flags(recipe.include_flags(arch), recipe.link_flags(arch))
+        return env
+
+    def prebuild_arch(self, arch):
+        super(TargetPythonRecipe, self).prebuild_arch(arch)
+        if self.from_crystax and self.ctx.ndk != 'crystax':
+            error('The {} recipe can only be built when '
+                  'using the CrystaX NDK. Exiting.'.format(self.name))
+            exit(1)
```
AndreMiras: Minor: how about using `BuildInterruptingException`, e.g.
```python
raise BuildInterruptingException('The {} recipe can only be built when '
                  'using the CrystaX NDK. Exiting.'.format(self.name))
```

pythonforandroid/recipes/python2/patches/fix-api-minor-than-21.patch
opacam: > shouldn't this be ` +#if defined(__ANDROID_API__) && __ANDROID_API__ < 21` instead?

Yes
Note: this patch has not been tested (is disabled and I put in there in the case that we will need it)

inclement:
> I suspect that if we build something against any shared lib, we will must load that lib when we initialize the app, otherwise we will get errors about missing symbols/functions

I've never been sure why this explicit loading appears necessary sometimes but not others. I think I've even seen differences between devices, although this was some time ago with older android versions.

I'm not sure if the real problem is ultimately that the libraries could be dynamically loaded, but aren't found in that directory. We know that dynamic loading in general does work, because of e.g. all the compiled components of the python build.

doc/source/quickstart.rst
```diff
@@ -93,11 +93,34 @@ named ``tools``, and you will need to run extra commands to install
...
-the latest useable NDK version is r10e, which can be downloaded here:
+operating system. The minimal, and recommended, NDK version to use is r17c:
+
+ - `Linux (64-bits) <https://dl.google.com/android/repository/android-ndk-r17c-linux-x86_64.zip>`_
+ - `Macos X (64-bits) <https://dl.google.com/android/repository/android-ndk-r17c-darwin-x86_64.zip>`_
+ - Windows users should create a virtual machine with an GNU Linux os
+   installed, and then you can follow the described instructions from within
+   your virtual machine.
+
+.. note::
```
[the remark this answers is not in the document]
JonasT: I agree, the problem with such a note is that nobody will think about updating it, so it'll get people to download a super old NDK version for no good reason some time in the future (that may not even work properly)

pythonforandroid/recipes/python2legacy/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+
+        # # if exists(join(self.get_build_dir(arch.arch), 'libpython2.7.so')):
+        # if exists(join(self.ctx.libs_dir, 'libpython2.7.so')):
+        #     info('libpython2.7.so already exists, skipping python build.')
+        #     if not exists(join(self.ctx.get_python_install_dir(), 'libpython2.7.so')):
+        #         info('Copying python-install to dist-dependent location')
+        #         shprint(sh.cp, '-a', 'python-install', self.ctx.get_python_install_dir())
+        #     self.ctx.hostpython = join(self.ctx.get_python_install_dir(), 'bin', 'python.host')
+
+        #     return
```
AndreMiras: Minor: we can drop this commented code above

pythonforandroid/recipes/python2legacy/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+                        'HOSTPGEN={}'.format(hostpgen),
+                        'CROSS_COMPILE_TARGET=yes',
+                        'INSTSONAME=libpython2.7.so',
+                        _env=env)
+            except sh.ErrorReturnCode_2:
+                print('First python2 make failed. This is expected, trying again.')
+
+            print('Second install (expected to work)')
+            shprint(sh.touch, 'python.exe', 'python')
+            shprint(make, '-j5', 'install', 'HOSTPYTHON={}'.format(hostpython),
```
AndreMiras: minor-out-of-scope-suggestion: we may want to use `cpu_count` from `multiprocessing` e.g. `'-j{}'.format(cpu_count())`

pythonforandroid/recipes/python2legacy/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+            # info('Copying python-install to dist-dependent location')
+            # shprint(sh.cp, '-a', 'python-install', self.ctx.get_python_install_dir())
+
+            # print('Copying hostpython binary to targetpython folder')
+            # shprint(sh.cp, self.ctx.hostpython,
+            #         join(self.ctx.get_python_install_dir(), 'bin', 'python.host'))
+            # self.ctx.hostpython = join(self.ctx.get_python_install_dir(), 'bin', 'python.host')
+
+        # print('python2legacy build done, exiting for debug')
+        # exit(1)
```
AndreMiras: Minor: we can drop these comments

pythonforandroid/recipes/python2legacy/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+        libdir = join(dirn, 'lib', 'python2.7')
+        site_packages_dir = join(libdir, 'site-packages')
+        with current_directory(libdir):
+            removes = []
+            for dirname, root, filenames in walk("."):
+                for filename in filenames:
+                    for suffix in EXCLUDE_EXTS:
+                        if filename.endswith(suffix):
+                            removes.append(filename)
+            shprint(sh.rm, '-f', *removes)
```
AndreMiras: Minor: We could simplify and kill at least one for since `endswith()` can take a list, e.g.:
```python
for dirname, root, filenames in walk("."):
    for filename in filenames:
        if filename.endswith(EXCLUDE_EXTS):
            removes.append(filename)
```

doc/source/quickstart.rst
```diff
@@ -93,11 +93,34 @@ named ``tools``, and you will need to run extra commands to install
 the SDK packages needed. 
 
 For Android NDK, note that modern releases will only work on a 64-bit
-operating system. If you are using a 32-bit distribution (or hardware),
-the latest useable NDK version is r10e, which can be downloaded here:
+operating system. The minimal, and recommended, NDK version to use is r17c:
```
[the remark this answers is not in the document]
opacam: Well, I have done all the tests using 17c, when I first test it with travis, it failed, I don't remember why, I thing that it was related with the openssl libs (if you are interested I could search some travis build with the error). The thing is that upgrading to ndk17c solve the issue. Also, if I remember well, the unified headers feature was introduced in version 15, so all the previous versions will not work because we use that feature in our builds...maybe that version will work,  but if ndk16 fails I assume that ndk 15 will also fail. Moreover, we have to consider the stl libs, c++ is the new android default, starting from ndk17 and it is the only one in version 18, so the ndkr17 has the two lib, is prepared to successfully build with c++ (I suspect that the previous version were still immature) and is the most modern ndk  that we can use which still have the gnustl lib..so I think that is the best version to use to make the transition to the new standard...and moreover,  this version it is proved to work, so...better that we stick to the version which hast not troubles right?

pythonforandroid/recipes/python2/__init__.py
```diff
@@ -1,225 +1,68 @@
...
     conflicts = ['python3crystax', 'python3']
-    opt_depends = ['openssl', 'sqlite3']
 
-    patches = ['patches/Python-{version}-xcompile.patch',
-               'patches/Python-{version}-ctypes-disable-wchar.patch',
-               'patches/disable-modules.patch',
+    patches = [
+               # new 2.7.15 patches
+               # ('patches/fix-api-minor-than-21.patch',
+               #  is_api_lt(21)), # Todo: this should be tested
```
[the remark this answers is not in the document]
opacam: Yes

pythonforandroid/recipes/python2legacy/patches/ctypes-find-library.patch
```diff
@@ -0,0 +1,28 @@
...
+ 
+-if os.name == "posix" and sys.platform == "darwin":
++# this test is for android specifically shoudl match here and ignore any
++# of the other platform tests below
++if os.name == "posix":
++    def find_library(name):
++        """ hack to find librarys for kivy and android 
++        split the path and get the first parts which will give us 
++        the app path something like /data/data/org.app.foo/"""
++        app_root = os.path.abspath('./').split(os.path.sep)[0:4]
```
AndreMiras: won't work when deployed on external storage, see 58f7618

pythonforandroid/recipes/python2legacy/__init__.py
```diff
@@ -0,0 +1,232 @@
+from pythonforandroid.recipe import TargetPythonRecipe, Recipe
```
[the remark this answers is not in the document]
opacam: You are right, is broken, we must decide if we maintain it with the pygame bootstrap or we remove it.

pythonforandroid/recipes/requests/__init__.py
```diff
@@ -4,7 +4,7 @@
 class RequestsRecipe(PythonRecipe):
     version = '2.13.0'
     url = 'https://github.com/kennethreitz/requests/archive/v{version}.tar.gz'
-    depends = [('hostpython2', 'hostpython3crystax'), 'setuptools']
+    depends = [('python2', 'python3', 'python3crystax'), 'setuptools']
```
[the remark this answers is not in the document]
opacam: Great!! :smile:

testapps/setup_pygame.py
```diff
@@ -4,11 +4,13 @@
 
 options = {'apk': {'debug': None,
                    'requirements': 'pygame,pyjnius,kivy,python2,android',
-                   'android-api': 19,
+                   'android-api': 27,
+                   'ndk-api': 19,
```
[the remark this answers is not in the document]
opacam: Nope, the pygame bootstrap is broken,

testapps/setup_testapp_python3_sqlite_openssl.py
```diff
@@ -0,0 +1,30 @@
...
+                   'arch': 'armeabi-v7a',
+                   'permissions': ['INTERNET', 'VIBRATE'],
+                   }}
+
+package_data = {'': ['*.py',
+                     '*.png']
+                }
+
+packages = find_packages()
+print('packages are', packages)
```
AndreMiras: out-of-scope-personal-taste: I would drop these two lines above. But yes it mainly comes from copy pasting so I understand it could be a bit out of scope

opacam:
> About the legacy python recipes, am I right to think that these will not work in the PR as it stands? Do your other changes fix this?

The changes introduced in the big pr, not contains any fix to python2legacy but I fixed that in some local branch (The fix was oriented to run the pygame bootstrap with the old python2 recipe). If the team decide to remove the pygame bootstrap then maybe we should remove the python2legacy recipe too...or move it to some historic folder in honor to @tito...that is your decision...If the team decide that it's worth it to save it I could fix it...for now the files are in there...in hold...I think that wehatever the decision is it would be better to do it in a separate pr...we already have a lot of code in this pr don't you think?

JonasT:
I'm all for removing the pygame bootstrap, the less the better and `service_only`/`webview` have much more distinct purposes

opacam:
Ei guys, all the comments about the python2legacy recipe are great, but I only moved those files in preparation to fix the pygame bootstrap or remove it and added some note about that the recipe is useless in the current state...so...all the that code has been there for a while I didn't touch it in this pr or the big one. I have some fixes in other branches but not in here or the big pr...

pythonforandroid/recipes/hostpython2/__init__.py
opacam: It does work ;)

opacam:
I need your opinion:

I was testing the part 2 (bootstraps fixes) and I realise that if we introduce the library mechanism before, we could unify the PythonUtil.java into common/bootstraps...so we could do it in one shot, because those files must be modified anyway to fix the bootstraps and we could avoid job for all of us...so i was thinking that maybe it will be better to put all of that in one pr, what do you think?

JonasT:
@opacam as for this pull request, I think we should leave it as-it-is (unless more changes are requested, but it looks good to me) and @inclement and @AndreMiras should merge it. For the next one, it makes sense to put all your bootstrap changes including everything they need into one thing, IMHO

pythonforandroid/archs.py
```diff
@@ -30,13 +30,24 @@ def include_dirs(self):
...
+            '-DANDROID',
+            '-fomit-frame-pointer',
+            '-D__ANDROID_API__={}'.format(self.ctx.ndk_api)]
+        if not clang:
+            cflags += ['-mandroid']
+        else:
+            cflags += ['-target armv7-none-linux-androideabi']
+            android_host = 'arm-linux-androideabi'
+            platform_dir = join(self.ctx.ndk_dir, 'platforms', 'android-{}'.format(self.ctx.ndk_api), 'arch-arm')
+            toolchain = '{android_host}-4.9'.format(android_host=android_host)
```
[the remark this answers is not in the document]
KeyWeeUsr: I'd at least move it to some variable if not to some global config for `Recipe` for extensions. This is quite a hidden place and I think it'll bite us later.

AndreMiras:
FYI I gave builds a try under `python3crystax` and `libffi` fails with:
https://gist.github.com/AndreMiras/ddb751cc8e04bb4792824b844da5c134
```
libtool: compile:  /usr/bin/ccache arm-linux-androideabi-gcc -DANDROID -fomit-frame-pointer -D__ANDROID_API__=19 -mandroid -I/home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -isysroot /home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -I/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/include/python3.6 -DHAVE_CONFIG_H -I. -I.. -I. -I../include -Iinclude -I../src -I. -I../include -Iinclude -I../src -DANDROID -fomit-frame-pointer -D__ANDROID_API__=19 -mandroid -I/home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -isysroot /home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -I/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/include/python3.6 -I/home/andre/.buildozer/crystax-ndk-10.3.2/sources/python/3.6/include/python/ -march=armv7-a -mfloat-abi=softfp -mfpu=vfp -mthumb -MT src/arm/sysv.lo -MD -MP -MF src/arm/.deps/sysv.Tpo -c ../src/arm/sysv.S -o src/arm/sysv.o >/dev/null 2>&1
libtool: compile:  /usr/bin/ccache arm-linux-androideabi-gcc -DANDROID -fomit-frame-pointer -D__ANDROID_API__=19 -mandroid -I/home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -isysroot /home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -I/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/include/python3.6 -DHAVE_CONFIG_H -I. -I.. -I. -I../include -Iinclude -I../src -DANDROID -fomit-frame-pointer -D__ANDROID_API__=19 -mandroid -I/home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -isysroot /home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -I/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/include/python3.6 -I/home/andre/.buildozer/crystax-ndk-10.3.2/sources/python/3.6/include/python/ -march=armv7-a -mfloat-abi=softfp -mfpu=vfp -mthumb -Wall -fexceptions -MT src/arm/ffi.lo -MD -MP -MF src/arm/.deps/ffi.Tpo -c ../src/arm/ffi.c  -fPIC -DPIC -o src/arm/.libs/ffi.o
libtool: compile:  /usr/bin/ccache arm-linux-androideabi-gcc -DANDROID -fomit-frame-pointer -D__ANDROID_API__=19 -mandroid -I/home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -isysroot /home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -I/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/include/python3.6 -DHAVE_CONFIG_H -I. -I.. -I. -I../include -Iinclude -I../src -DANDROID -fomit-frame-pointer -D__ANDROID_API__=19 -mandroid -I/home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -isysroot /home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -I/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/include/python3.6 -I/home/andre/.buildozer/crystax-ndk-10.3.2/sources/python/3.6/include/python/ -march=armv7-a -mfloat-abi=softfp -mfpu=vfp -mthumb -Wall -fexceptions -MT src/arm/ffi.lo -MD -MP -MF src/arm/.deps/ffi.Tpo -c ../src/arm/ffi.c -o src/arm/ffi.o >/dev/null 2>&1
/bin/bash ./libtool  --tag=CC   --mode=link /usr/bin/ccache arm-linux-androideabi-gcc -DANDROID -fomit-frame-pointer -D__ANDROID_API__=19 -mandroid -I/home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -isysroot /home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -I/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/include/python3.6  -DANDROID -fomit-frame-pointer -D__ANDROID_API__=19 -mandroid -I/home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -isysroot /home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -I/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/include/python3.6 -I/home/andre/.buildozer/crystax-ndk-10.3.2/sources/python/3.6/include/python/ -march=armv7-a -mfloat-abi=softfp -mfpu=vfp -mthumb  -Wall -fexceptions -no-undefined -avoid-version  -Xcompiler '--sysroot' '/home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm' '-lm' '-L/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/libs_collections/etheroll/armeabi-v7a' '-L/home/andre/.buildozer/crystax-ndk-10.3.2/sources/crystax/libs/armeabi-v7a' '-lcrystax'  --sysroot /home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -lm -L/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/libs_collections/etheroll/armeabi-v7a -L/home/andre/.buildozer/crystax-ndk-10.3.2/sources/crystax/libs/armeabi-v7a -lcrystax -o libffi.la -rpath /home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/lib src/prep_cif.lo src/types.lo src/raw_api.lo src/java_raw_api.lo src/closures.lo                          src/arm/sysv.lo src/arm/ffi.lo
libtool: link: /usr/bin/ccache arm-linux-androideabi-gcc -DANDROID -fomit-frame-pointer -D__ANDROID_API__=19 -mandroid -I/home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -isysroot /home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -I/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/include/python3.6 -shared  -fPIC -DPIC  src/.libs/prep_cif.o src/.libs/types.o src/.libs/raw_api.o src/.libs/java_raw_api.o src/.libs/closures.o src/arm/.libs/sysv.o src/arm/.libs/ffi.o
  -L/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/libs_collections/etheroll/armeabi-v7a -L/home/andre/.buildozer/crystax-ndk-10.3.2/sources/crystax/libs/armeabi-v7a -lm -lcrystax  -mandroid -isysroot /home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -mandroid -isysroot /home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm -march=armv7-a -mfloat-abi=softfp -mfpu=vfp -mthumb --sysroot --sysroot /home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm   -Wl,-soname -Wl,libffi.so -o .libs/libffi.so
/home/andre/.buildozer/crystax-ndk-10.3.2/toolchains/arm-linux-androideabi-5/prebuilt/linux-x86_64/bin/../lib/gcc/arm-linux-androideabi/5.3/../../../../arm-linux-androideabi/bin/ld: error: cannot open crtbegin_so.o: No such file or directory
/home/andre/.buildozer/crystax-ndk-10.3.2/toolchains/arm-linux-androideabi-5/prebuilt/linux-x86_64/bin/../lib/gcc/arm-linux-androideabi/5.3/../../../../arm-linux-androideabi/bin/ld: fatal error: /home/andre/.buildozer/crystax-ndk-10.3.2/platforms/android-19/arch-arm: pread failed: Is a directory
collect2: error: ld returned 1 exit status
Makefile:1176: recipe for target 'libffi.la' failed
make[1]: *** [libffi.la] Error 1
make[1]: Leaving directory '/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/other_builds/libffi/armeabi-v7a__ndk_target_19/libffi/arm-unknown-linux-androideabi'
Makefile:2368: recipe for target 'libffi.la' failed
make: *** [libffi.la] Error 2
```
I just wanted you to know. But I'm not sure if it makes sense investigating into `python3crystax` since it was already dropped from the CI and `python3` should replace it. So I won't be blocking specially if the pure `python3` build pass (haven't tested yet).

AndreMiras:
> I was testing the part 2 (bootstraps fixes) and I realise that if we introduce the library mechanism before, we could unify the PythonUtil.java into bootstraps/common...so we could do it in one shot, because those files must be modified anyway to fix the bootstraps and we could avoid job for all of us...so i was thinking that maybe it will be better to put all of that in one pr, what do you think?

I also think it's fine as it is and prefer to tackle more refactoring in another PR because otherwise my brain will explode.
I'd like @inclement or @tito to take a(nother) look and merge if they're happy with it. Of course it's not perfect, but I don't expect that PR to be perfect before merging

testapps/setup_testapp_python3_sqlite_openssl.py
```diff
@@ -0,0 +1,30 @@
+
+from distutils.core import setup
+from setuptools import find_packages
+
+options = {'apk': {'requirements': 'libffi,openssl,sqlite3,requests,peewee,sdl2,pyjnius,kivy,python3',
```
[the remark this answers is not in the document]
JonasT: I'm not sure I'd call less than 8% super significant: https://developer.android.com/about/dashboards/ But maybe that's just me. But I think we should move on at some point, especially if this makes the patching more complex in this particular case

pythonforandroid/bootstraps/common/build/jni/application/src/start.c
```diff
@@ -182,18 +181,6 @@ int main(int argc, char *argv[]) {
    * replace sys.path with our path
    */
   PyRun_SimpleString("import sys, posix\n");
-  if (dir_exists("lib")) {
```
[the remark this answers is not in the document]
opacam: Because those paths were the location for the old python, now we use the same packaging system as python3 (the paths are changed  and we zip some files...), so I think that those lines are useless but maybe I missed something ...Do you have some reason to maintain those lines?

pythonforandroid/python.py
```diff
@@ -0,0 +1,385 @@
...
+    opt_depends = ['sqlite3', 'libffi', 'openssl']
+    '''The optional libraries which we would like to get our python linked'''
+
+    def __init__(self, *args, **kwargs):
+        self._ctx = None
+        super(GuestPythonRecipe, self).__init__(*args, **kwargs)
+
+    def get_recipe_env(self, arch=None, with_flags_in_cc=True):
+        if self.from_crystax:
+            return \
```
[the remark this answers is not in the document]
opacam: I know that I can split the lines in several ways, depending the circumstance I use the methods you mentioned, but I usually use tha backslash in the return calls when  long lines are involved, I dind't thought about being more pythonic, only thought in making the line shorter...thanks @KeyWeeUsr !!!

JonasT:
@opacam any ideas on the `time.sleep`, `peewee` and hardcoded gcc version issues? I could also offer to take a look at these if I get write access to the pull request - I really want to move this ahead because I have other changes pending that are potentially conflicting, and if we're not merging one or the other at some point, we're going to get stuck with this

opacam:
Ei @JonasT, until tonight I cannot work with this, I started to fix the hardcoded toolchain version and arch for python.py and arch.py but I didn't finished...so if you want to work with other stuff, be me guess, no problem, I think that you should be able to access the pr, I always check the give administrator access check box...is this will be enough to access the pr?
If yes , suit yourself, if not I'm afraid that until night I cannot do nothing...

testapps/setup_testapp_python3_sqlite_openssl.py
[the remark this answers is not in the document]
JonasT: Ah right, the existing test app already uses that. That makes sense :+1:

testapps/setup_testapp_python3_sqlite_openssl.py
```diff
@@ -0,0 +1,30 @@
...
+                }
+
+packages = find_packages()
+print('packages are', packages)
+
+setup(
+    name='testapp_python3_sqlite_openssl_googlendk',
+    version='1.1',
+    description='p4a setup.py test',
+    author='Alexander Taylor',
```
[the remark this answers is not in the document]
JonasT: @KeyWeeUsr the sqlite test app isn't new though, just the setup file is (that is also why the `peewee` use isn't new, I missed it also on first read)

opacam:
Mmmm...the time.sleep(3) issue for openssl...I don't know how to face this, I committed the changes, but @tito did it and I applied some more changes later..so...I don't know why we need this call to time.sleep... maybe to wait that some files has been created and sync in the hard drive?

testapps/setup_testapp_python3_sqlite_openssl.py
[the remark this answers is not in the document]
opacam: > I think you should use own name/contact info ...

@KeyWeeUsr, I did it that way because it's almost a copy of the testap by @inclement, I have no problem to put my info in there...if @inclement wants that I will change it I will do it, no problem.

opacam:
Ok, it seems that the old version of openssl lib was failing to build some symbols and we introduced a function to check if that symbols were build or not, if not, we forced to rebuild again with a pause of 3 seconds between builds, so I think that the call to time.sleep of 3 seconds is only there to have time to read the message that the build will start again.

But I went a little further...I removed the lines that check and rebuild the libs, and I builded the test app with some different archs (x86_64, arm64-v8a and armeabi-v7a)...I checked the libs with the readelf command and it seems that the functions are there, so probably we can remove all the related lines to this "check/rebuild", because whatever was happening with the old libs, isn't happening anymore. For now I leave the lines in there in the case I miss something...

Do you think that we should remove those lines?

JonasT:
> Do you think that we should remove those lines?

yes definitely, we have too many lines around where nobody knows for sure why they are in, and whether they are even needed. it sounds like a better idea to me to investigate this deeper if it breaks again for someone, and then document & understand properly why this is necessary (and otherwise leave it out if it works for now)

opacam:
Update: The last changes I made allow us to build for almost all archs, I have tested the changes made with different archs (with the test app sqlite3 and openssl) here are the results:

  - python2:
    - x86: **OK**
    - x86_64: **OK**
    - armeabi-v7a: **OK** (tested in device with numpy)
    - arm64-v8a: **OK**

  - python3 (all tested with numpy):
    - x86: **OK**
    - x86_64: *OK* without numpy (fails on linking... -lm file not found error)
    - armeabi-v7a: **OK** (tested in device)
    - arm64-v8a: **OK**

Note: The apps that I tested in a device ( armeabi-v7a) for both versions of python, were tested removing the openssl lines mentioned in some previous comment and without removing the lines...and always worked as expected.

doc/source/quickstart.rst
```diff
@@ -93,11 +93,34 @@ named ``tools``, and you will need to run extra commands to install
...
+
+ - `Linux (64-bits) <https://dl.google.com/android/repository/android-ndk-r17c-linux-x86_64.zip>`_
```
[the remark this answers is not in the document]
opacam: @inclement, I put the direct link because the main page for downloads does not contain the ndk version r17c, plus the ndk r18 seems to have issues documented by @tito in #1525 which right now maybe solved with the latests changes I applied en here...

pythonforandroid/recipes/python2legacy/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+            # python$EXE ../../Tools/scripts/h2py.py -i '(u_long)' /usr/include/netinet/in.h  # noqa
+            # /home/tito/code/python-for-android/build/python/Python-2.7.2/python: 1: Syntax error: word unexpected (expecting ")")  # noqa
+            # because at this time, python is arm, not x86. even that,
+            # why /usr/include/netinet/in.h is used ?
+            # check if we can avoid this part.
+
+            make = sh.Command(env['MAKE'].split(' ')[0])
+            print('First install (expected to fail...')
+            try:
+                shprint(make, '-j5', 'install', 'HOSTPYTHON={}'.format(hostpython),
```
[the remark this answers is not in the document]
opacam: This has been hardcoded for a while, I didn't touch that, plus this pr doesn't cover to fix the python2legacy recipe, which will not work (I have fixed the python2legacy recipe...but not in here or in the big pr)

JonasT:
@KeyWeeUsr are all the things you care about resolved? I can't see any still to be taken care of that I remarked (or saw regarding your interesting remarks). I'm asking since it's still in the *"KeyWeeUsr requested changes"* state

pythonforandroid/recipes/hostpython2legacy/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+                self.ctx.hostpython = join(self.get_build_dir(), 'hostpython')
+                self.ctx.hostpgen = join(self.get_build_dir(), 'hostpgen')
+                return
+
+            if 'LIBS' in os.environ:
+                os.environ.pop('LIBS')
+            configure = sh.Command('./configure')
+
+            shprint(configure)
+            shprint(sh.make, '-j5')
```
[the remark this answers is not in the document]
opacam: No, but this is has been there for a while, I only move the old python recipe to python2legacy, in preparation to repair it or remove it in a future pr...

opacam:
Ok guys, I created the remaining prs which completes the most of the work of the original pr. All those pr should be easily reviewed  and they will not cause the troubles that causes the part I and part II.

I labeled all the new prs with a WIP label because most of them needs the new openssl lib or some recipe which must be compatible with python3 and in the master branch those recipes are only python2 compatible.

Once we move on with the core-update we could remove the WIP label and properly test all those recipes which has been proved to work in the original pr and should work with the new core.

Note: I see that they are some other pr that touches the same recipes that I do but probably did not take into account the python core update, so I created them anyway, I will close the conflicting ones if any one is detected, no problem.

testapps/setup_testapp_python3_sqlite_openssl.py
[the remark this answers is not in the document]
KeyWeeUsr: Ah, ok, I get it. :) Also, pep8 this line.

pythonforandroid/recipes/python2legacy/patches/ctypes-find-library.patch
JonasT: for what it's worth as I understand it the python2legacy recipe will be broken anyway, and I personally suggest it should be removed. I don't think having the <10% of devices with <21 API support is worth it, giving that problem will only get "better" on its own as time passes. But if we intend to fix the python2legacy recipe this may be worth addressing

pythonforandroid/recipes/python2legacy/__init__.py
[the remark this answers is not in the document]
KeyWeeUsr: Moving the request to a separate issue → https://github.com/kivy/python-for-android/issues/1558.

pythonforandroid/recipes/python2legacy/__init__.py
KeyWeeUsr: Moving the request to a separate issue → https://github.com/kivy/python-for-android/issues/1558.

pythonforandroid/recipes/hostpython2legacy/__init__.py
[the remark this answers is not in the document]
KeyWeeUsr: Moving the request to a separate issue → https://github.com/kivy/python-for-android/issues/1558.

opacam:
Oh sorry, I hided the comments to see the big picture...mmm...plus seems that whenever I refresh the page firefox hides them without my consent...and seems that still things to do, thanks for to put them in the comment above I will look into it.

pythonforandroid/bootstraps/pygame/build/jni/application/Android.mk
```diff
@@ -18,9 +18,7 @@ LOCAL_CFLAGS := $(foreach D, $(APP_SUBDIRS), -I$(LOCAL_PATH)/$(D)) \
 				-I$(LOCAL_PATH)/../jpeg \
 				-I$(LOCAL_PATH)/../intl \
 				-I$(LOCAL_PATH)/.. \
-                -I$(LOCAL_PATH)/../../../../other_builds/$(PYTHON2_NAME)/$(ARCH)/python2/python-install/include/python2.7
-				# -I$(LOCAL_PATH)/../../../../python-install/include/python2.7
-				# -I$(LOCAL_PATH)/../../../build/python-install/include/python2.7
+                -I$(LOCAL_PATH)/../../../../other_builds/$(MK_PYTHON_INCLUDE_ROOT)
```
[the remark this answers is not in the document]
opacam: I don't think that we can replace them because the variables you propose are absolute paths, and we need them relative paths.

pythonforandroid/recipe.py
```diff
@@ -613,6 +618,30 @@ def get_build_dir(self, arch):
...
+        env['EXTRA_LDLIBS'] = ' -lpython{}'.format(
+            self.ctx.python_recipe.major_minor_version_string)
+        if 'python3' in self.ctx.python_recipe.name:
+            env['EXTRA_LDLIBS'] += 'm'
+
+        # set some env variables that may be needed to build some bootstrap ndk
+        # recipes that needs linking with our python via mk files, like
+        # recipes: sdl2, genericndkbuild or sdl
+        other_builds = join(self.ctx.build_dir, 'other_builds') + '/'
+        env['MK_PYTHON_INCLUDE_ROOT'] = \
```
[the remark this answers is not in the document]
opacam: Because those ones are relative paths...and the others are absolute paths

pythonforandroid/recipe.py
```diff
@@ -799,19 +821,6 @@ def install_python_package(self, arch, name=None, env=None, is_dir=True):
             elif self.call_hostpython_via_targetpython:
```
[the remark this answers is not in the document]
opacam: Wow @inclement, you are right, I think that we should discuss about if we should fix the python2legacy recipe, because if we fix it, we will need those lines, otherwise I think that we could remove it and probably we could remove all the references to call_hostpython..., but this will touch a lot of files...
About python2legacy recipe, I already said in discord that I have it done in another branch, so it's only matter to decide if we safe it or remove it...so...

what do you think @AndreMiras, @inclement , @tito and @KeyWeeUsr,

**Should we keep the python2legacy recipe?**

opacam:
I pushed some changes, but still have to check more comments, but I'm afraid that i cannot do more today, probably tomorrow night can work with it...so...

**¡¡¡Happy Christmas!!!**

pythonforandroid/bootstraps/common/build/jni/application/src/start.c
opacam: I think we cannot remove this line because we need that for the new version of python2 which also use the `python_bundle_dir` like python3.

pythonforandroid/recipes/python2/patches/fix-api-minor-than-21.patch
opacam: Update: I have tested the mentioned patch, building with minimum api 19 and, even that the patch successfully applies, the python2 build will fail at linking stage complaining about missing  localeconv...As far as I know localeconv is not fully implemented in android until sdk 21 (https://stackoverflow.com/questions/44736135/ndk-clang-error-undefined-reference-to-localeconv) so I'm afraid that we will not be able to run with a minimum api lower than 21, like the python3 or numpy (#1450) recipes...unless someone find some solution, which I searched but not found :<...for now I leave the patch but maybe we should remove it as proposed by @JonasT...

pythonforandroid/recipes/python2legacy/__init__.py
AndreMiras: I agree this is minor and out of context since this minor debt was already existing, but just moved.
I've created a ticket to track it https://github.com/kivy/python-for-android/issues/1560

pythonforandroid/archs.py
```diff
@@ -82,8 +101,19 @@ def get_env(self, with_flags_in_cc=True):
             env['NDK_CCACHE'] = self.ctx.ccache
             env.update({k: v for k, v in environ.items() if k.startswith('CCACHE_')})
 
-        cc = find_executable('{command_prefix}-gcc'.format(
-            command_prefix=command_prefix), path=environ['PATH'])
+        if clang:
+            clang_path = join(
+                self.ctx.ndk_dir, 'toolchains', 'llvm', 'prebuilt',
```
AndreMiras: FYI I had to symlink `llvm-3.7` to `llvm`:
```
ln -sfn ~/.buildozer/crystax-ndk-10.3.2/toolchains/llvm-3.7/ ~/.buildozer/crystax-ndk-10.3.2/toolchains/llvm
```
We may not want to tackle it in this PR, it's just for us to know

pythonforandroid/archs.py
```diff
@@ -93,20 +123,20 @@ def get_env(self, with_flags_in_cc=True):
                 'installed. Exiting.')
```
AndreMiras: The `{}` string argument is not rendered.
This is minor since it's out of context, but FWIW: https://github.com/kivy/python-for-android/pull/1561

pythonforandroid/archs.py
opacam: mmm...you are right, I did not consider crystax and I see that the paths are different, do you want that I fix it?

Something like:
```
llvm_dirname = 'llvm'
if self.ctx.python_recipe and self.ctx.python_recipe.from_crystax:
    llvm_dirname = split(glob(join(self.ctx.ndk_dir, 'toolchains', 'llvm-*'))[-1])[-1]
    clang_path = join(self.ctx.ndk_dir, 'toolchains', llvm_dirname, 'prebuilt', 'linux-x86_64', 'bin')
```

with the proper imports, will solve the problem.

AndreMiras: Thanks for the reply.
I don't consider it critical since we're dropping crystax support slowly.
But yes still if having it working during the transition is not that painful, then yes.
I like the approach you suggest, however we increase the cyclomatic complexity slightly with that if.
What about going directly for `'llvm*'` (note that I removed the dash `-`) and kill the if? Do you think that could introduce issues to non-crystax NDK or you it should be safe?
So basically the code would be:
```python
llvm_dirname = split(glob(join(self.ctx.ndk_dir, 'toolchains', 'llvm*'))[-1])[-1]
clang_path = join(self.ctx.ndk_dir, 'toolchains', llvm_dirname, 'prebuilt', 'linux-x86_64', 'bin')
```

opacam: It works, I tested via python console and hardcoded paths. Your solution is far better :smile: thanks!!!

opacam:
Ei @AndreMiras, I think that I already solved the issue #1560 with the commit 1a4830b

pythonforandroid/recipes/python2legacy/__init__.py
AndreMiras: Fixed in https://github.com/kivy/python-for-android/commit/1a4830b7ddbb6d2f2d0fec0782f76545219b49b9 of this PR

opacam:
Ok , I think that now this is in a pretty good shape, you should take a look when you have some time, if any of you feel that I miss something just tell me.

With the latest changes introduced we will be able to use the python2legacy recipe with sqlite3 and openssl almost as before. To achieve this I was forced to reintroduce the old openssl recipe but the new openssl will also work  (7a7d9f6). But I must warn you that some applications may require to add python2legacy compatibility to some recipes in order to work, because I only granted python2legacy compatability to some minimal set of recipes (66d6394).

I must say that with this changes we will be able to target sdk minimun api 19 via the python2legacy recipe which is a concern for @KeyWeeUsr, and also solves one of the concerns of @inclement about merging this pr with a broken python2legacy recipe.

The commits needed to make it work the python2legacy recipe has been named with the prefix `Python2legacy:`, to make it easier to revert the changes applied in case that the team decide that python2legacy must be removed.

Also it would be great that @AndreMiras could try again to build his etheroll app with crystax so we would know if openssl pass the build stage and works at runtime.

JonasT:
> The commits needed to make it work the python2legacy recipe has been named with the prefix Python2legacy:, to make it easier to revert the changes applied in case that the team decide that python2legacy must be removed.

Not sure how useful that is, most likely we'll need to merge it squashed anyway.

My suggestion would still be to just drop the old recipe and all changes that would be required to get it to work. (which would give us minimum API of 21, right? That's not ideal, but given the device stats - this seems to affect ~9%ish devices right now - I personally find it somewhat acceptable given this problem will only solve itself)

JonasT:
Happy new year everyone :tada:  :tada: 

For what it's worth, there is already some user interest in this pull request here #1559 and I'm also interested in moving my app from CrystaX to the regular python 3 after this is merged and help fix the `ctypes` issue (with a better approach than I attempted in the ill-fated #1517 ).

Is there anything I can do to help move this pull request along? Any input/opinions on whether `python2legacy` should be just dumped (which is my preferred solution)?

opacam:
Happy new year!!

I also think that the python2legacy should be removed, but the administrator team must decide, I put in here because I thought that with it we will have more possibilities....I mean...we can slowly fase out python2:

  - First we could remove python2legacy, when the statics for devices that we will leave out  are in our favour (lower than 5%) and when most of the recipes works with the new python2 build system.
  - Then we could begin to migrate all the recipes to be python3 compatible, and when we have most of them working in python3, we could remove python2.

Also, by acting this way, we will have some kind of fail safe python recipe, in case someone must target a minimum api lower than 21, or needs some recipe that have troubles to build with the new python formula...then ...the python2legacy recipe could be useful (at least for a while...until we begin to touch recipes...then ...I think that will be more problematic than useful to have this python2legacy recipe).

Anyway, I have no problem in removing the python2legacy recipe and the latest commits which make it work, if the team decide that it must be done, I will do it  :wink:

pythonforandroid/python.py
```diff
@@ -291,9 +296,8 @@ def create_python_bundle(self, dirn, arch):
         python_lib_name = 'libpython' + self.major_minor_version_string
         if self.major_minor_version_string[0] == '3':
             python_lib_name += 'm'
-        for lib in [python_lib_name + '.so', python_lib_name + '.so.1.0']:
-            shprint(sh.cp, join(python_build_dir, lib),
-                    'libs/{}'.format(arch.arch))
+        shprint(sh.cp, join(python_build_dir, python_lib_name + '.so'),
```
JonasT: I don't understand, why would you remove the loop here? The only thing that seems to achieve is that in case it is ever named `.so.1.0` again in the future (which I assume can happen depending on NDK or compiler version) it will break. Or does this actually fix something?

opacam: If we don't remove the versioned library we will get an error when running the copy command, because now the versioned library doesn't exist anymore

JonasT: @opacam I'm just thinking, is this make option guaranteed to exist in the future? Is it some standard thing? Maybe it would be better to make the `for lib` loop based on `os.listdir` instead?

JonasT:
@opacam just as a general note regarding pull requests, you're making it also increasingly unlikely that we get this merged if you pile on fixes that weren't caused by this pull request. I recommend not changing anything anymore unless it is a requested change or something broken as part of this request, so we can finally settle on some final version

I think the main question is still whether python2legacy should go since it introduce some complexity, @AndreMiras @inclement and @KeyWeeUsr should probably weigh in on that

pythonforandroid/python.py
JonasT: Also is this fixing something broken by this pull request? If not, better make a separate one. This one is huge enough and the more things are added on top, the longer it will take us to look through it and get it merged :grimacing:

opacam: > I'm just thinking, is this make option guaranteed to exist in the future?

Nothing is guaranteed to exist in the future  :laughing: :laughing:...but this option exists at least since the python version 2.7.2...we have it in our python2legacy recipe...so...probably will continue there for a while...

opacam: > Also is this fixing something broken by this pull request?...

No, but it will cause troubles, and we know it...so I though that it will be better get fixed as soon as possible because it requires only a few lines, plus the affected files are already touched for this pr, it affects directly the work done in here and moreover...there still some changes that must be reviewed...**but you are also right...better to  not be touched anymore unless requested.**

JonasT: That is a non-issue, we have code comments for that :wink: (you could think that with the old loop as well)

I don't think it would be much longer than the previous code, and avoid that Makefile option. I guess using the Makefile option is kind of neater, but I'm not sure if we should trust it to stay around...? Might be worth checking in the Makefile or python install docs if this is something official or rather some internal variable we hacked into that isn't meant to be used from the outside

**Anyway**, in conclusion I would suggest you remove this commit, and make a separate pull request (since this is only bringing up new discussion and isn't strictly part of this, we should stop delaying things)

JonasT: > but you are also right...better to not be touched anymore unless requested.

true, and I would actually suggest you undo that latest commit and make a separate pull request. it's no good to stuff more things into this, even if they are necessary

> will be better get fixed as soon as possible

Even more of a reason to make a separate pull request! piling up more things into a single one we won't get it in any sooner due to the complexity of all the changes

opacam: **INSTNOME It's official**, and it has been there for a long, long, long time...checking the cpython code I see that was introduced at year 2002 (python/cpython@1142de3)...seems to me that has been pretty well tested, even for us because @tito make use of it in his original recipe (bdefea1) and @inclement did the same with his rework of the recipe (when the recipe.sh was removed, 4c8b5bc) so...we are currently having this discussion thanks to this variable jajajaja...at the point that we remove it in python3 recipe... we begin to have this linkage problems...so I have no doubt that we can rely on this INSTNOME variable...don't you think?

About removing the commit...mmmm...I already shot the bullet...even if I remove the commit...the damage has been done...lets the administrator team decide ok  :wink:?

**Note:** If we create the pr you mention and gets merged before this, we may have troubles resolving the merge conflicts in here and we could delay this even more...that is the reason to not  create the pr in first place...and the reason to create a proper patch and post it into the proper issue and to notify it in discord.

JonasT: > About removing the commit...mmmm...I already shot the bullet...even if I remove the commit...the damage has been done...lets the administrator team decide ok wink?

not really, the commits will all be squashed anyway, so if you do a revert it will be truly removed with no issues. there is really no problem if you just revert it.

you could already squash the commit history as a preparation if you want, it would need to be done on the final merge with github's squash button anyway

KeyWeeUsr:
@opacam Thanks again for your hard work and patience :+1:

PR #1541, opened 2018-12-19
[CORE UPDATE - PART II] Fix bootstraps for webview and service_only

opacam:
This is the second part for the pr #1537 (discussed previously in #1460 and #1534)

I created a parallel branch with the part I in the current state and the part II, with two more commits that allow us to view if the build is success. The first commit adds travis tests. The second one applies the docker image change (from ubuntu to debian). This last one is mandatory to pass the webview test...we need python 3.7 and ubuntu ships 3.6. I put this on here for all us to know and because we already started a discussion about that in the previous pr. We don't need those changes in master branch because the webview tests will not be enabled but I think that this information can safe time to anyone who wants to build those tests in ci.

Here is the travis build with the tests enabled: https://travis-ci.org/opacam/python-for-android/builds/469953483

For now I leave out the libraries loading mechanism, it should be the part 3, but for that we need to merge this pr because we touch files modified in here so maybe better do it after merged this don't you think?

If you think that we must implement in here let me know and I will apply the changes on the top.

Additional info for future reference:
  - The commit that enables webview and service_only tests (for both version of python): https://github.com/opacam/python-for-android/commit/157f0af82ff23fc9bc04e5a8d630ef224b739d02 
  - The commit that change the docker image: https://github.com/opacam/python-for-android/commit/6c35c93b50afa20a7ef9534c33dcbeaa04fe8978


pythonforandroid/bootstraps/webview/build/src/main/java/org/kivy/android/PythonActivity.java
```diff
@@ -181,19 +200,22 @@ public boolean shouldOverrideUrlLoading(WebView view, String url) {
...
+
         // make sure all child threads (python_thread) are stopped
         android.os.Process.killProcess(android.os.Process.myPid());
     }
 
     public void loadLibraries() {
-        PythonUtil.loadLibraries(getFilesDir());
+        String app_root = new String(getAppRoot());
+        File app_root_file = new File(app_root);
+        PythonUtil.loadLibraries(app_root_file);
```
JonasT: This, together with the entire pull request https://github.com/kivy/python-for-android/pull/1543/files , is for me definitely the most critical part.

I like the loading mechanism as a general construct since it seems well enough done, however, I don't understand why it is necessary. Shouldn't this all just work through dynamic linking? Because this and the entire code of https://github.com/kivy/python-for-android/pull/1543/files do add quite a lot of additional complexity in the loading process, which to me at a first glance shouldn't be needed, so I'd like to understand what is going on and if we really need & want this

JonasT:
@opacam could you possibly take out the library loading related things of this pull request? It is very useful to have the bootstraps repaired, but since I think the library loading really needs some deeper investigation first, I would suggest that should be merged in later (if it turns out to be necessary). So if you split it up from this one, I think we can move ahead all the awesome bootstrap refactoring which definitely looks like stuff we want to have in as soon as possible

opacam:
> could you possibly take out the library loading related things of this pull request?

Yes, seems that I commited those changes by error, thanks!!

JonasT:
> don't you agree?

No, not unless we decided we actually make use of it with the third bootstrap (and I'm kind of opposed to that as of now, since it seems to me like something that is unnecessary). I understand that given how you separated the files it was easiest to keep this in, but they are really *the* key uncertainty of this pull request for me. Maybe we could wait for the feedback of the others, but I would personally be love to have everything merged *except* for these lines

JonasT:
Sorry, it appears this is my mistake! It really looks like you just moved the existing somewhat hardcoded library list to another place, but it's essentially the same thing. Looks good from my side then :+1:

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -340,9 +336,9 @@ def make_package(args):
         with open(args.intent_filters) as fd:
             args.intent_filters = fd.read()
 
-    if get_bootstrap_name() == "sdl2":
-        args.add_activity = args.add_activity or []
-        args.activity_launch_mode = args.activity_launch_mode or ''
+    # if get_bootstrap_name() == "sdl2":
```
AndreMiras: Minor: I would simply delete that comment

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -340,9 +336,9 @@ def make_package(args):
...
+    args.add_activity = args.add_activity or []
+    args.activity_launch_mode = args.activity_launch_mode or ''
```
AndreMiras: Minor: I know it was already that way before, but people misuse this as ternary operator, this is not the python ternary operator

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -708,7 +687,7 @@ def _read_configuration():
     if args.meta_data is None:
         args.meta_data = []
 
-    if args.services is None:
+    if (not hasattr(args, 'services')) or args.services is None:
```
AndreMiras: minor: `if getattr(args, 'service', None) is None`

pythonforandroid/bootstraps/common/build/build.py
AndreMiras: So with Python-style ternary that would be:
```
args.add_activity = args.add_activity if args.add_activity else []
args.activity_launch_mode = args.activity_launch_mode if args.activity_launch_mode or ''
```

PR #1543, opened 2018-12-19
Remove the libraries loading mechanism

opacam:
~~This is the third part for the pr #1537 (discussed previously in #1460 and #1534)~~

Here we remove the library loading mechanism because we don't need that, we only need to load a single library `libmain.so`. By removing this mechanism we also can remove the PythonUtil.java file because in there, we only have the loading mechanism.


pythonforandroid/bootstrap.py
```diff
@@ -257,6 +266,83 @@ def _unpack_aar(self, aar, arch):
...
+        if 'boost' in self.ctx.recipe_build_order:
+            # Todo: check boost loading dependencies and create a proper
+            # loading order. For now it works for libtorrent's recipe, the only
+            # recipe that depends on boost...so...for now it's fine, but should
+            # be enhanced, maybe using similar method than
+            # `pythonforandroid.build.copylibs_function`?
+            for l in listdir(self.ctx.get_libs_dir(arch.arch)):
+                if l.startswith('libboost_'):
+                    libs_to_load.append(l[3:-3])
+        if 'libtorrent' in self.ctx.recipe_build_order:
```
JonasT: I don't think it makes much sense to have libtorrent/libiconv/libzbar code hardcoded here. `boost`, *maybe*, but even that is a relatively unimportant lib for the p4a ecosystem as a whole. Would it be possible to somehow also put this into the respective recipes and have it accessed here?

But in general, I also don't quite understand why preloading libs that should be linked dynamically (and hence automatically loaded by the OS loader of Android) is necessary in the first place. Do you know why this is required? I'm very curious

opacam: > Do you know why this is required?

Ok, this is what I think is happening and is based on my experience, maybe android is capable to load the system libs but I suspect that android os has no way to know the library order loading, for instance, take the libpython3.X.so as example and imagine that we build with our libffi.so. Given that situation if  android loads first the libpython3X.so it will crash because we build it as a shared library and at the time that the library gets loaded will not have all the functions/symbols needed declared in libffi.so, because still is not loaded, so the application will crash. I also use the libtorrent since introduced in p4a and that also applies to it...boost must be loaded after python (because depends on python) and libtorrent after boost libraries (because depends on boost)...if the order is not right...buuuum...crash complaining about missing symbols/functions. Those are the cases I remember..there is more. So...I suspect that the android os has no way to know the right loading order for external libraries.

JonasT: That sounds to me like the dynamic link is just wrong. E.g. you built libtorrent wrong without properly depending on libboost - like the main program depends on it properly, but your `libtorrent.so` doesn't. I highly doubt the OS loader in Android cannot resolve this properly, since it's just the Linux loader, so IMHO this should be investigated a lot deeper before we merge this

JonasT: Also a common problem: maybe it depends on libboost, but the wrong version or with the wrong library file name referenced? It could be any of those things, and we should be very sure that's not the culprit before we add all this extra complexity

opacam: @JonasT, I don't think that I have linkage problems with the boost/libtorrent recipe, I already tested in one of my apps (which integrates a torrent client, running with libtorrent rasterbar) and I downloaded some file without issues...well...only one problem...the version I introduce is a release candidate version and some changes has been made in the source code that may require to adapt some calls inside the app code...but this has nothing to do with this talk. Also I should mention that the libtorrrent recipe is kind of special:

  - It is build with the standalone toolchain (which isolates an android environment with the proper files...so it's easier to set the proper flags because all the environment files are in one place)
  - also contains two libraries: One is the library which contains all the functions/symbols like any standard library and **It's not python dependant**. The other one has the python bindings and that one is installed into site-packages like a regular python module (and that one does not require any special treatment). The first one should be treated as a regular library, that is why we need to load it like any standard library.

Saying that, if you wanna take a look into the libtorrent recipe I created a pr #1555, You will see the relevant code in there and despite it works, I may did something wrong with the flags or with something else...so you tell me please...:smile:...I will fix it if that is the case.

JonasT: Well again, if it doesn't load at runtime then your dynamic link is most likely wrong.

The recipe isn't that interesting, what would be interesting if you used `nm` and `objdump` to actually check on your produced `.so` files to see what they link, and if these links are correct. (Most notably, do they link the libraries you hacked into your `libraries_to_load` mechanism, with exactly the correct name? I doubt they do, otherwise it shouldn't be necessary)

JonasT: Can you give me an affected `.apk` maybe? One that doesn't use `libraries_to_load` and therefore doesn't work? You could e-mail it to `jonas@thiem.email` and I could try `nm` and `objdump` to see if i can find out what exactly is wrong with the linking, or why it doesn't launch properly

opacam: @JonasT, no need that pass you any application, I just made some test it in a new branch (using the branch python-core-update **on his current status** (pr #1537 as a base),  adding the update of the libtorrent recipe and adapting one of the test apps (the one with openssl and sqlite3 for python3) to libtorrent instead of using the numpy module and...whatever was happening before...it's not happening now :smile: ...so...good news for all of us!!!

I suspect that my last commit in the python-core-update branch (bdd2b7c) maybe related with this, because now it works the ctypes module without importing libffi...and I did the pr for the libffi's python3 compatibility (#1463) and at that time it was failing...not now using the new core...

In my last test I comment all the libs, only with the libpythonX.X.so and libmain.so seems that the test app works...

If we remove the libpythonX.X.so (with only the libmain.so in PythonUtil.java) then the app not starts and I see some error in logcat similar to:

`dlopen failed: could not load library "libpython3.7m.so.1.0"...`

*Note: We already seen something similar in the issue #1501...mmm*

Anyway, here are the branch I used to make the test:

  - https://github.com/opacam/python-for-android/commits/python-core-libs-loading-test

  - and the commit that I thing that fixed the libraries loading problems is bdd2b7c.

Still not sure about all this, let me do some more test with a real app but I suspect that we can remove most of the libs of the PythonUtil.java and probably drop this pr and maybe you can test some of your app without the ctypes fix (#1517) using the python-core-update branch (maybe we doesn't need that fix now). Also it would be great that you confirm my words in this post, so we can remove this pr.

JonasT: `dlopen failed: could not load library "libpython3.7m.so.1.0"` yes the problem here seems to be that this file exists in the distribution (the `.so`) but gradle somehow throws it out (gone from the `.apk`). So it should most likely be only `libmain.so`, and whatever goes wrong with gradle should be fixed / #1501 should be fixed

opacam: @JonasT, seems to me that we do not need this pr
¿Do you want to create a new pr, something like `simplify loading mechanism` to remove the unneeded stuff? If you don't wanna to do it I could reconvert this pr to do so or create a new one...I ask you because maybe you already have some work done on this, you have been one step ahead all the way with this matter...so...

JonasT: @opacam no I have been just waiting... haven't touched this part. So feel free to update it, I really hope the others will wake up soon and actually do some merging

opacam:
Ei @JonasT, I changed the title of this pr and applied the necessary changes to remove the loading mechanism...but I suspect that this may conflict with your sdl2 update (#1528)...that is why I leave the `[WIP]` status. Take a look on this when you have time, **we are not hurry in this**. But your **sdl2 update** is important for us (**we need that as soon as we can**). So it's your call to decide if we remove the [WIP] status in here because I think that will be easier to fix conflicts in here than in there :wink:

JonasT:
@opacam the SDL update is ready for merging, it's just that kivy in its current state will break with it. So how fast we can get it in depends on how fast @inclement and @tshirtman will address the remaining kivy issues (these being the windowed/fullscreen startup behavior that needs to obey the env vars, and the config file location which mustn't default to `/sdcard/`), so it's really kind of out of my hands at this point :woman_shrugging: 

But I agree, it might be easier to resolve conflicts here. But I really don't know how fast the SDL2 will get in

opacam:
Ok, no problem at all, then we leave this in `WIP`, this will not affect at all at the current functionality of p4a, it's only a clean up. Thanks for the info about your sdl2 update status by the way :smile: I'm sure that @inclement and @tshirtman will fix the kivy issues you mention as soon as they can.

opacam:
Ui, I forgot to remove the `wip` status on this...sorry...It should had done some days ago... :grimacing:

@JonasT, can you take a look on this when you have some time?

Note: please look at it carefully, because my `C` programing level is far to be as good as yours so I can easily make mistakes in here  :wink:

opacam:
My guess is that the library loading problem was solved at some point after the gradle update was introduced (for sdl2), I think that, right now, the following line solves the problem for bootstraps `sdl2, webview and service_only`:

https://github.com/kivy/python-for-android/blob/master/pythonforandroid/bootstraps/common/build/templates/build.tmpl.gradle#L62

See also: https://developer.android.com/studio/projects/gradle-external-native-builds

As a side note: I can say that I've been testing with the same device for a quite while, more than a year without any system update and I can say that I had the library loading problem before, and not now (with the same device and the same android version)

JonasT:
@opacam looks promising, I'm all for removing unnecessary code! I'll build my app with this and see what happens, but given this doesn't break anything then this is a really good change :+1:

inclement:
I'm concerned that if we don't know whether it's actually fixed, merging this could break things in a really annoying way for some users. Do you know why the gradle change would have fixed things, does it actually signal Android to load the libraries in some different way?

JonasT:
https://github.com/realm/realm-java/pull/1771 yeah it sounds a little like it is very device dependent.

For what it's worth, does p4a use relinker? Because SDL references it as well: https://github.com/kivy/python-for-android/blob/dec1badc3bd134a9a1c69275339423a95d63413e/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/libsdl/app/SDL.java#L45

Maybe if we used that library, we could let go of our custom loading hacks?

opacam:
> For what it's worth, does p4a use relinker? 

As far as I know we don't use relinker

> Maybe if we used that library, we could let go of our custom loading hacks?

Oh yes, it may simplify our code..maybe we should test relinker. And if we keep going with this pr maybe we should do that in here...@inclement, what do you thing about that?

inclement:
I have nothing against using relinker, it looks like there is no real cost to doing so.

It does say in its README "Note that this library fixes intermittent link errors; if you get an error every time you use your app, you may have a configuration issue". I think the errors we were seeing were more the latter, although there may also be intermittent problems I'm not directly aware of that this would fix. Most users would have no idea what happened if the app randomly failed, so it could definitely be happening.

Most of all though, I'm reluctant to change the current code without at least a test case that it fixes (i.e. an actual device with problems before but not after), or some specific documented change in the Android build tools that explains why the issue occurred and when it was fixed.

inclement:
My understanding was that jniLibs.srcDir doesn't affect where System.loadLibrary loads things from on the device, it just tells gradle what folders to pull .so files from to package into the APK. Am I wrong about that? I also didn't check yet, did we add `jniLibs.srcDir` to the gradle project only some time after the original gradle support PR, or some time afterwards?

The change in the current code that made things start working was specifically to explicitly add System.loadLibrary for each different library in turn, because calling System.loadLibrary only for the main one wasn't successfully automatically finding the others even though they were present in the expected place. That's consistent with the System.loadLibrary failures in the realm issue you linked, except that it was happening every time for some users rather than being due to System.loadLibrary flakiness as mentioned in the relinker docs.

inclement:
https://github.com/kivy/python-for-android/pull/1071/files#diff-e6d6c9254ef94f6f1ea119b95c029668R55

^^ looks like jniLibs.srcDir was set even in the original gradle PR, so if @opacam is correct that the loading issue was solved only some time after the original gradle introduction, we probably can't associate it with that.

JonasT:
`because calling System.loadLibrary only for the main one wasn't successfully automatically finding the others even though they were present in the expected place`

Did it lead to libmain.so loading failures with errors that the dependencies cannot be found? That sounds like ReLinker might help

If they just weren't loaded at all and that lead to missing symbols, then maybe it's rather a toolchain failure/bulid problem with the dynamic linking...?

I don't think we'll get anywhere unless someone can recall a device where this originally broke, or puts in the time to find one

inclement:
@JonasT I think what we actually saw was libpython.so failing to load the openssl and sqlite libs, but I don't remember for sure.

opacam:
Well...after the conversation...it seems risky to remove the loading mechanism plus is harmless if you don't suffer of the loading problem...so...I was thinking:

**What if we do a removal in steps?**

  1. remove a library of the loading mechanism (sqlite3 I was thinking)
  2. wait some days/weeks/months, staying alert of  any complain about sqlite3
  3. if we start to get complains about sqlite3 loading problems it will mean that we need the loading mechanism and then we quickly add sqlite3 again or we test relinker to see if we can use that...and if not ... :scissors:

Maybe it's a crazy idea...

JonasT:
That could mean it's broken for some end users for months. I don't think that's a good idea if we don't have some basic clue about what's going on

opacam:
> That could mean it's broken for some end users for months. 

 :scream:, no please, If we merge this and we really need the libraries loading mechanism...then that will be the situation for us and then will be hard to fix it, as @inclement pointed, plus we could have issues with multiple libraries (not only sqlite3) and it may be related with this or not...

> I don't think that's a good idea if we don't have some basic clue about what's going on

That is the key, **we don't have an idea what is going on** (at least myself), that is why i propose this, a controlled  :bomb: (we only have two issues regarding sqlite3...by the way...so many thanks @inclement :heart:...you closed a lot of the issues and you too @JonasT :heart:)...so...it would be easy to find any issue about that. Plus I think that we only affect the master branch (well...that can affect a lot of people but I'm confident that we would have some notice...of course I could be wrong) and we can provide a solution for this end user via discord or via issues section. Anyway, I agree with you @JonasT, we can cause some problems to some end users...but it will cause less troubles than merging the pr...

Maybe we should put this in `wip` again for starters?

inclement:
>  Plus I think that we only affect the master branch

We're actually experimenting with having the master branch be the default for everyone, since that's consistently been a better experience recently.

Overall, I really support just not messing with that right now. There's no need to experiment on the users because this code isn't actually really a problem, there is very little cost to maintaining PythonUtils.java because it doesn't ever change.

opacam:
> Overall, I really support just not messing with that right now. There's no need to experiment on the users because this code isn't actually really a problem, there is very little cost to maintaining PythonUtils.java because it doesn't ever change.

@inclement, I totally agree with you

I close this pr because I don't see that we can safely remove the library loading mechanism

!!!Soo many thanks for your contributions in this pr @JonasT and @inclement!!!

inclement:
@opacam If you'd like to leave this open as something for the future, that would be fine. There's nothing wrong with the idea of the change, and it would log that there's something to be done here, but also log the reason it hasn't been merged yet. I feel bad for the work to be lost, even if I think it's the safest way for now :)

opacam:
Again,  you are right, I will leave this open...so we have a record that we have something to look into and it's not solved...hopefully the time will give us the answer :wink:

¡¡¡Thanks @inclement!!!

PR #1544, opened 2018-12-20
fix wrong conditional for build custom_rules.tmpl.xml

bit4bit:
without this, **ant debug** stop with message:

~~~
/root/.buildozer/android/platform/android-sdk-20/tools/ant/build.xml:655: /root/fleet_dispatch_client/.buildozer/android/platform/build/dists/myapp/tmp-src does not exist.
~~~

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -483,7 +483,7 @@ main.py that loads it.''')
             join(res_dir, 'values/strings.xml'),
             **render_args)
 
-    if exists("custom_rules.tmpl.xml"):
+    if exists(join("templates", "custom_rules.tmpl.xml")):
         render(
             'custom_rules.tmpl.xml',
             'custom_rules.xml',


pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -483,7 +483,7 @@ def make_package(args):
             join(res_dir, 'values/strings.xml'),
             **render_args)
 
-    if exists("custom_rules.tmpl.xml"):
+    if exists(join("templates", "custom_rules.tmpl.xml")):
```
AndreMiras: I'm not sure how this works if you fixed the first path in the `exists` condition, but not the `render` one which I would have expected to be the same :thinking:

bit4bit: hi, because **os.path.exists** check on filesystem and not use **jinja2.Environment**, **render** it's a local function.

JonasT: @AndreMiras it works because of this line: https://github.com/kivy/python-for-android/blob/41b740eac3d4271eade67438cafa3ccaea9b98e9/pythonforandroid/bootstraps/common/build/build.py#L79 (which sets up the environment to be located inside `templates`, so jinja2 naturally looks in there)

AndreMiras: Thank you guys, clear!

AndreMiras:
Looking good thank you. I'm just restarting the build because it flaky failed and will merge pretty soon

PR #1546, opened 2018-12-22
[CORE UPDATE - PART V] Fix pycrypto

opacam:

Here I fix pycrypto to make it work with the core-update pull request, which also contains a new version of the openssl libs. I Also update pycrypto to latest version and grants compatibility for both versions of python.

Note: this has to be merged only when the core-update has been merged because it depends on the new python2 recipe and the openssl libs (1.1)

### DIFF
--- a/pythonforandroid/recipes/pycrypto/__init__.py
+++ b/pythonforandroid/recipes/pycrypto/__init__.py
@@ -4,26 +4,26 @@ from pythonforandroid.toolchain import (
     info,
     shprint,
 )
-from os.path import join
 import sh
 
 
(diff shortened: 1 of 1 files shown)


opacam:
No @AndreMiras, look at the LDFLAGS...I pass libraries in there instead of doing that in LIBS (for openssl)...let me fix that please!...all the other cryptography recipes of my prs  (all those with status WIP) may have the same issue, be careful with that!!

opacam:
Now should be fine :smile:

AndreMiras:
Merged :beer:

PR #1549, opened 2018-12-22
[CORE UPDATE - PART VIII] Netifaces for both versions of python (updates the netifaces version)

opacam:

Grants compatibility for both versions of python and cleans up from unneeded source code as well as update the netifaces version.

Note: this has to be merged only when the core-update has been merged because it has been tested with the new python core and depends on some recipes which aren't python3  compatible in the current master branch (setuptools)

### DIFF
--- a/pythonforandroid/recipes/netifaces/__init__.py
+++ b/pythonforandroid/recipes/netifaces/__init__.py
@@ -3,25 +3,17 @@ from pythonforandroid.recipe import CompiledComponentsPythonRecipe
 
 class NetifacesRecipe(CompiledComponentsPythonRecipe):
 
-    version = '0.10.4'
+    version = '0.10.7'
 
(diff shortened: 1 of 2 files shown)


pythonforandroid/recipes/netifaces/__init__.py
```diff
@@ -3,25 +3,17 @@
 
 class NetifacesRecipe(CompiledComponentsPythonRecipe):
 
-    version = '0.10.4'
+    version = '0.10.7'
 
-    url = 'https://pypi.python.org/packages/18/fa/dd13d4910aea339c0bb87d2b3838d8fd923c11869b1f6e741dbd0ff3bc00/netifaces-{version}.tar.gz'
+    url = 'https://files.pythonhosted.org/packages/81/39/4e9a026265ba944ddf1fea176dbb29e0fe50c43717ba4fcf3646d099fe38/netifaces-{version}.tar.gz'
```
AndreMiras: Try this instead:
```python
url = 'https://files.pythonhosted.org/packages/source/n/netifaces/netifaces-{version}.tar.gz'
```

opacam:
Nop, already removed :)

AndreMiras:
Merged thank you!

PR #1554, opened 2018-12-22
[CORE UPDATE - PART XIII] Pysha3 for both versions of python

opacam:

Grants compatibility for both versions of python, remove unneeded flags and fix the hardcoded ones.

Note: this has to be merged only when the core-update has been merged because it has been tested with the new python core and depends on some recipes which aren't python3  compatible in the current master branch (setuptools)


AndreMiras:
Thanks!

PR #1555, opened 2018-12-22
[WIP][CORE UPDATE - PART XIV] Libtorrent+boost for both versions of python and updated versions

opacam:

Grants compatibility for both versions of python, maintaining the same build method (android's ndk standalone toolchain and b2 tool) but now we will use clang as the compiler, plus starts the migration of the stl lib from gnustl_shared to c++_shared (the only one stl lib starting from ndk18)

Note: this has to be merged only when the core-update has been merged because it has been tested with the new python core and depends on the new openssl lib version 1.1 introduced in  #1537

### DIFF
--- a/pythonforandroid/recipes/boost/__init__.py
+++ b/pythonforandroid/recipes/boost/__init__.py
@@ -1,5 +1,6 @@
 from pythonforandroid.toolchain import Recipe, shprint, shutil, current_directory
 from os.path import join, exists
+from os import environ
 import sh
 
 """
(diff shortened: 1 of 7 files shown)


inclement:
Looks good to me. @opacam would you mind rebasing one last time to resolve the conflicts? Otherwise I can do it and get someone else to merge it.

The failing test is only because the log gets too long for travis.

opacam:
Done @inclement

inclement:
Merged even though the tests aren't working, as they seem to fail repeatedly due to issues with downloading the SDK or its components. The tests passed before, and the PR doesn't touch any recipes that are actually in the tests, so it should be fine.

Thanks!

PR #1566, opened 2019-01-04
[OMEMO] updated omemo recipe

goffi-contrib:
upstream omemo module has been splitted, with the wireformat put in a backend module.
This patch update omemo package, and create the new recipe for the
backend, "omemo_backend_signal".
It has been tested working with python2.


goffi-contrib:
Hi @AndreMiras I've updated the recipe you have merged the last time, but I'm not sure how I can fix travis here. Any hint? Thanks!

PR #1572, opened 2019-01-09
Fix APP_PLATFORM not properly passed in NDKRecipe

JonasT:
Fix APP_PLATFORM not properly passed in NDKRecipe. This is needed so certain NDKAPI/--minimum-api values won't lead to build errors with NDK based recipes, e.g. when using NDKAPI=21 with the current `jpeg` recipe


JonasT:
The additional pushes you see are just rebases to latest master. I used this during a couple of builds now without any issues :+1: (all python 3 though, no python 2 builds)

inclement:
Thanks

PR #1573, opened 2019-01-09
Rework for Pillow/pil recipes & update jpeg and png

opacam:
Several actions taken here:

  - Update jpeg and moved to mainline
  - Update png
  - Force to build freetype/harfbuzz **only** as static libraries (to make sure that Pillow/pil gets linked with the static libraries, plus  we cannot use those libraries in his shared form because we will have conflicts with the ones distributed by android os...so no need to build those libraries unless properly versioned as we do with openssl libs)
  - Make it work Pillow with python3/python2/python2legacy
  - Make it work pil with python2 and python2legacy
  - Add freetype/harfbuzz support for Pillow/pil
  - Formats the touched recipes and fixes imports

**Note:** This should work, tested in python3. I must review the code and create the proper commits plus I may change something that we may not need. I push it in this state in case that anyone has troubles building Pillow/pil with the new core.

### DIFF
--- a/pythonforandroid/recipes/jpeg/__init__.py
+++ b/pythonforandroid/recipes/jpeg/__init__.py
@@ -1,35 +1,78 @@
-from pythonforandroid.recipe import NDKRecipe
+from pythonforandroid.recipe import Recipe
 from pythonforandroid.logger import shprint
 from pythonforandroid.util import current_directory
 from os.path import join, exists
+from os import environ, uname
+from glob import glob
 import sh
 
 
-class JpegRecipe(NDKRecipe):
+class JpegRecipe(Recipe):
+    '''
+    .. versionchanged:: 0.6.0
+        rewrote recipe to be build with clang and updated libraries to latest
+        version of the official git repo.
+    '''
     name = 'jpeg'
-    version = 'linaro-android'
-    url = 'git+https://git.linaro.org/people/tomgall/libjpeg-turbo/libjpeg-turbo'
+    version = '2.0.1'
+    url = 'https://github.com/libjpeg-turbo/libjpeg-turbo/archive/{version}.tar.gz'  # noqa
+    # we will require this below patch to build the shared library
+    # patches = ['remove-version.patch']
 
-    patches = ['build-static.patch']
-
-    generated_libraries = ['libjpeg.a']
-
-    def prebuild_arch(self, arch):
-        super(JpegRecipe, self).prebuild_arch(arch)
-
-        build_dir = self.get_build_dir(arch.arch)
-        app_mk = join(build_dir, 'Application.mk')
-        if not exists(app_mk):
-            shprint(sh.cp, join(self.get_recipe_dir(), 'Application.mk'), app_mk)
-        jni_ln = join(build_dir, 'jni')
-        if not exists(jni_ln):
-            shprint(sh.ln, '-s', build_dir, jni_ln)
+    def should_build(self, arch):
+        return not exists(join(self.get_build_dir(arch.arch),
+                               'libturbojpeg.a'))
 
     def build_arch(self, arch):
         super(JpegRecipe, self).build_arch(arch)
-        with current_directory(self.get_lib_dir(arch)):
-            shprint(sh.mv, 'libjpeg.a', 'libjpeg-orig.a')
-            shprint(sh.ar, '-rcT', 'libjpeg.a', 'libjpeg-orig.a', 'libsimd.a')
+        build_dir = self.get_build_dir(arch.arch)
+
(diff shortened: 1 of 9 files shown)


JonasT:
I am just testing this, and I ran into some detail: maybe https://github.com/kivy/python-for-android/blob/master/doc/source/quickstart.rst#installing-dependencies should be expanded to add `- cmake (required for some native code recipes)`? (Before this pull request, I never needed to have it installed)

JonasT:
:+1: I tested the build with this and python 3 (new non-crystax), and even though I couldn't test runtime yet (app wouldn't run due to unrelated other issue) without this Pillow won't even build, so in any case this is definitely an upgrade :laughing:

JonasT:
Tested runtime now, works :ok:  so given without this it won't even build, this is really an essential upgrade :+1:

JonasT:
@opacam could you add an edit to this pull request which changes [this file](https://github.com/kivy/python-for-android/blob/master/doc/source/quickstart.rst#installing-dependencies) and adds `- cmake (required for some native code recipes)` to the dependencies? Just so we don't forget

pythonforandroid/recipes/png/__init__.py
```diff
@@ -3,8 +3,8 @@
 
 class PngRecipe(NDKRecipe):
     name = 'png'
-    version = '1.6.15'
-    url = 'https://github.com/julienr/libpng-android/archive/{version}.zip'
+    version = '1.6.29'
+    url = 'https://github.com/julienr/libpng-android/archive/master.zip'
```
AndreMiras: Oops, I guess you hardcoded the branch/tag by mistake right?

JonasT: oh wow didn't notice that one :open_mouth: now I feel blind. yeah, would be nice to know if that is an accident or not

opacam: Unfortunately, there is no error...well...I forgot to notify it (**my apologies for that...I completely forgot about it...**)...

Well...the thing is that the author of the github's repo never released/tagged this version, despite He performed the necessary changes in master branch and notify about it in the README file...I thing that we should try to move it from that repo to the mainline repo, plus considering that the last commit is more than a year old...mmm...I have bad feelings about that...

Anyway...maybe we should target a commit instead of a version, this way we wouldn't hardcode the url...I still think that we should try to move it to mainline but maybe would be better to do it in another pr (this way we could solve the current Pillow/pil issues, and move the png recipe to the mainline repository when we can).

@JonasT and @AndreMiras, what do you think?

JonasT: @opacam yea hardcoding a commit seems like the best idea for now :+1:

pythonforandroid/recipes/jpeg/__init__.py
```diff
@@ -1,35 +1,78 @@
...
+                                  'build/cmake/android.toolchain.cmake')
 
-        build_dir = self.get_build_dir(arch.arch)
-        app_mk = join(build_dir, 'Application.mk')
-        if not exists(app_mk):
-            shprint(sh.cp, join(self.get_recipe_dir(), 'Application.mk'), app_mk)
-        jni_ln = join(build_dir, 'jni')
-        if not exists(jni_ln):
-            shprint(sh.ln, '-s', build_dir, jni_ln)
+            shprint(sh.rm, '-f', 'CMakeCache.txt', 'CMakeFiles/')
```
AndreMiras: Would `rm -f` work on a directory without the `-r` flag?

pythonforandroid/recipes/jpeg/__init__.py
```diff
@@ -1,35 +1,78 @@
...
+            shprint(sh.cmake, '-G', 'Unix Makefiles',
+                    '-DCMAKE_SYSTEM_NAME=Android',
+                    '-DCMAKE_SYSTEM_PROCESSOR={cpu}'.format(cpu='arm'),
+                    '-DCMAKE_POSITION_INDEPENDENT_CODE=1',
+                    '-DCMAKE_ANDROID_ARCH_ABI={arch}'.format(arch=arch.arch),
+                    '-DCMAKE_ANDROID_NDK=' + self.ctx.ndk_dir,
```
AndreMiras: Minor, why don't you use the same format string  syntax as before/after for consistency?

pythonforandroid/recipes/jpeg/__init__.py
AndreMiras: Got the bug report on Discord 2 years later :exploding_head: 
https://discord.com/channels/423249981340778496/678695971164520523/876867985355186228

Issue #1294, opened 2018-06-06
ffmpeg breaks buildozer android debug  compilation process.


littlegreenbean33:
Hi,

I just downloaded everything from scratch  into my ubuntu 16.04 after facing some difficulties.

My application is a minimal app just to test I can display video on android.

Dependencies kivy/ffmpg
The build process breaks in 

```
 STDOUT:
GEN     libavdevice/libavdevice.ver
CC      libavformat/network.o
CC      libavformat/options.o
CC      libavformat/os_support.o
CC      libavformat/protocols.o
In file included from libavformat/network.h:29:0,
                 from libavformat/network.c:22:
libavformat/os_support.h:67:32: error: expected declaration specifiers or '...' before '(' token
  #define lseek(f,p,w) lseek64((f), (p), (w))
                                ^
libavformat/os_support.h:67:37: error: expected declaration specifiers or '...' before '(' token
  #define lseek(f,p,w) lseek64((f), (p), (w))
                                     ^
libavformat/os_support.h:67:42: error: expected declaration specifiers or '...' before '(' token
   # define lseek(f,p,w) lseek64((f), (p), (w))
                                          ^
common.mak:60: recipe for target 'libavformat/network.o' failed
make: *** [libavformat/network.o] Error 1
make: *** Waiting for unfinished jobs....
```

I understand this is not a buildozer error. Just to reflect the inability to produce apps that support video playing while this is not resolved.

I wonder if there are alternatives to this libs for playing video

regards.

JC

inclement:
Thanks for the report, I guess the ffmpeg recipe is broken. Interesting that the error seems to be an actual syntax error, I'm not really sure what's gone wrong.

AndreMiras:
It compiled OK in the Docker env.
I ran the following command:
```
python setup_testapp_python2_sqlite_openssl.py apk --sdk-dir /opt/android/android-sdk --ndk-dir /opt/android/android-ndk --requirements sdl2,kivy,ffmpeg
```
Relevant part of the log is:
```
[INFO]:    Building sdl2 for armeabi-v7a
[INFO]:    -> directory context /root/.local/share/python-for-android/build/bootstrap_builds/sdl2_gradle-python2/jni
[INFO]:    -> running ndk-build V=1
[INFO]:    <- directory context /app/testapps
[INFO]:    Building six for armeabi-v7a
[INFO]:    six apparently isn't already in site-packages
[INFO]:    Installing six into site-packages
[...]

PR #1579, opened 2019-01-11
Fixes ffmpeg compilation w/ openssl 1.1.1

misl6:
This PR fixes ffmpeg compilation when openssl is in requirements list.
I've also added the `scale` filter.
Now, the `ffpyplayer` recipe is compatible with the non-crystax python3.


misl6:
I think that fixes #1294

PR #1580, opened 2019-01-11
[WIP] Rework libxml2, libxslt and lxml (update versions)

opacam:
The lxml recipe depends on those two libraries, if we update lxml we must update the two libraries (libxml2 and libxslt). This rework takes care of that and also:

  - replace double quotes to single quotes because I think that it's more readable
  - fix imports
  - shortened lines

**Note:** Successfully build for python3 and python2 (tested some days ago in a real python3's app but it would be great that someone could test it again)

### DIFF
--- a/pythonforandroid/recipes/libxml2/__init__.py
+++ b/pythonforandroid/recipes/libxml2/__init__.py
@@ -1,53 +1,58 @@
-from pythonforandroid.toolchain import Recipe, shprint, shutil, current_directory
+from pythonforandroid.recipe import Recipe
+from pythonforandroid.toolchain import shprint, shutil, current_directory
 from os.path import exists, join
 import sh
 
 
 class Libxml2Recipe(Recipe):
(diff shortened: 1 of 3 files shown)


AndreMiras:
Looking overall good to me. Let's see if Travis is happy about it.
I disagree with this (minor) point however `replace double quotes to single quotes because I think that it's more readable` I think it's a matter of taste, unless there's a PEP8 rule for it. So in general I don't touch something like that

opacam:
Yes, you are right @AndreMiras, it's a personal taste the quotes  :grimacing: (I knew before hand that this will be problematic  :smirk:...if you want I will change to double quotes no problem). Anyway, as far as I know, it is recommended to be consistent along all the project, and now we have mixed quotes and before the change I checked some of the core modules and it seemed to me that single quotes are predominant...but I cannot be sure about that...so...the said...just tell me and I will change it .

PR #1581, opened 2019-01-11
 fix missing gethostbyname_r on Android 5.1

opacam:
@tito faced the `gethostbyname_r` issue and send me a fix for that.

This also should fix the second issue described in https://github.com/kivy/python-for-android/pull/1217#issuecomment-453608150, which is the same that @tito found:

See Also: https://bugs.python.org/issue26857

¡¡¡Thanks @tito!!!


homdx:
Thanks to @opacam . Are good pull request. Tested under Python2 (https://github.com/kivy/python-for-android/pull/1217#issuecomment-453679133)

AndreMiras:
Thanks

PR #1583, opened 2019-01-12
[WIP] Rework zbar (add python3 compatibility + add recipes: pyzbar and zbarlight)

opacam:
- Add ability to use libzbar for python2 and python3
- Add two more recipes (pyzbar and zbarlight) because those two are compatible with python3, the old zbar recipe seems unmaintained and working only for python2.

**Note: this should be properly tested with some application using this recipes, I only tested the build for python 2 and python 3...so...any feedback about this will be great...**

### DIFF
--- a/pythonforandroid/recipes/libzbar/__init__.py
+++ b/pythonforandroid/recipes/libzbar/__init__.py
@@ -11,7 +11,7 @@ class LibZBarRecipe(Recipe):
 
     url = 'https://github.com/ZBar/ZBar/archive/{version}.zip'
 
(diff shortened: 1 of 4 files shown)


opacam:
Because I only tested the build of those recipes. I usually modify some app and try to import and make some calls...but in this case I hadn't got time for that...it seems to build fine, the packages are installed where they should be...it should work...but not properly tested...

fszymanski:
[mchehab](https://github.com/mchehab) continues to work on [zbar](https://github.com/mchehab/zbar), it is `Python 3` compatible. You should use this version.

PR #1584, opened 2019-01-12
Move libffi to mainline repo

opacam:


inclement:
Thanks (and to hackalog).

PR #1585, opened 2019-01-12
[CORE FIX/ENHANCEMENT] Speedup copy that can be very very long (up to 2 minutes)

opacam:
This fixes the slow copy issue we have when copying files to our distribution folder and affects to python2 and python3.

I tested this in python 3 and @tito made and tested this commit in his own builds for python 2. Travis will say the final words but I suspect that  we can  reduce travis's build timings...will see...

@inclement please...take a look on this when you have time...this affects the core!

Again :smile:, ¡¡¡Thanks @tito!!!

### DIFF
--- a/pythonforandroid/python.py
+++ b/pythonforandroid/python.py
@@ -4,6 +4,7 @@ build our python3 and python2 recipes and his corresponding hostpython recipes.
 '''
 
 from os.path import dirname, exists, join
+from shutil import copy2
 from os import environ
 import glob
 import sh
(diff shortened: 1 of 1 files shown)


pythonforandroid/python.py
```diff
@@ -266,16 +267,19 @@ def create_python_bundle(self, dirn, arch):
                 '2' if self.version[0] == '2' else '',
                 self.major_minor_version_string
             ))
-        module_filens = (glob.glob(join(modules_build_dir, '*.so')) +
-                         glob.glob(join(modules_build_dir, '*.py')))
+        module_filens = list(glob.glob(join(modules_build_dir, '*.so')) +
```
KeyWeeUsr: Is there any reason for calling `list()` when `glob.glob()` already returns a `list` and doing `[] + []` is implemented for lists?

pythonforandroid/python.py
```diff
@@ -266,16 +267,19 @@ def create_python_bundle(self, dirn, arch):
...
-            shprint(sh.cp, filen, modules_dir)
+            info(" - copy {}".format(filen))
+            copy2(filen, modules_dir)
 
         # zip up the standard library
         stdlib_zip = join(dirn, 'stdlib.zip')
         with current_directory(join(self.get_build_dir(arch.arch), 'Lib')):
-            stdlib_filens = walk_valid_filens(
-                '.', self.stdlib_dir_blacklist, self.stdlib_filen_blacklist)
+            stdlib_filens = list(walk_valid_filens(
```
KeyWeeUsr: This is perhaps the same thing as with `glob.glob()`.

pythonforandroid/python.py
opacam: Oh yes, you are right, we don't need that...this commit was coming from an outdated branch of the python core...I don't know how I didn't see that  :sweat_smile: ... so many thanks @KeyWeeUsr  :blush:

pythonforandroid/python.py
opacam: This one is not the same, we need that because `walk_valid_filens` it's a generator and we add a `info` message later where we print the number of files that we will zip using `len`...here we need a list...or remove the message that @tito added:

`info("Zip {} files into the bundle".format(len(stdlib_filens)))`

I think that this may be useful...what do you think?

KeyWeeUsr: Yeah, for this use-case it makes sense.

PR #1586, opened 2019-01-13
Minor corrections for `Fix bootstraps for webview and service_only` (recently merged)

opacam:
We merged by mistake the mentioned pr #1541 with some minor issues that should had been fixed in there. We solve that in here.

¡¡¡Thanks @AndreMiras!!!

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -336,9 +336,11 @@ main.py that loads it.''')
         with open(args.intent_filters) as fd:
             args.intent_filters = fd.read()
 
-    # if get_bootstrap_name() == "sdl2":
-    args.add_activity = args.add_activity or []
-    args.activity_launch_mode = args.activity_launch_mode or ''
+    if not args.add_activity:
+        args.add_activity = []
+
+    if not args.activity_launch_mode:
+        args.activity_launch_mode = ''
 
     if args.extra_source_dirs:
         esd = []
@@ -687,7 +689,7 @@ tools directory of the Android SDK.
     if args.meta_data is None:
         args.meta_data = []
 
-    if (not hasattr(args, 'services')) or args.services is None:
+    if getattr(args, 'services', None) is None:
         args.services = []
 
     if args.try_system_python_compile:


pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -336,9 +336,11 @@ def make_package(args):
         with open(args.intent_filters) as fd:
             args.intent_filters = fd.read()
 
-    # if get_bootstrap_name() == "sdl2":
-    args.add_activity = args.add_activity or []
-    args.activity_launch_mode = args.activity_launch_mode or ''
+    if not args.add_activity:
```
AndreMiras: Ternary is still possible if you like:
```
args.add_activity = args.add_activity if args.add_activity else []
```

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -687,7 +689,7 @@ def _read_configuration():
     if args.meta_data is None:
         args.meta_data = []
 
-    if (not hasattr(args, 'services')) or args.services is None:
+    if getattr(args, 'services', None) is None:
         args.services = []
```
AndreMiras: Now I'm thinking we may go even more concise with:
```python
args.services = getattr(args, 'services', [])
```
What do you think?

pythonforandroid/bootstraps/common/build/build.py
JonasT: @AndreMiras doesn't that work in the same way though? won't `or` return the value to the right if the left-hand side would also evaluate in a conditional to false? (which would then give the same result as your ternary operator) or is there an actual semantic difference in what it produces?

While I agree it's not a ternary operator, I don't see why that is a problem. Is it specifically discouraged or something? I've seen `X or Y` quite often so it does seem to be a common pattern

pythonforandroid/bootstraps/common/build/build.py
JonasT: yes that looks exactly how the default value of `getattr` is supposed to be used like :+1:

pythonforandroid/bootstraps/common/build/build.py
AndreMiras: Yes there is totally a semantic difference. I also see it very often, but most of the time people use it as they think it's ternary (because other languages follow this pattern).
But yes it's semantically different, but I suspect we wanted a ternary and not this

opacam: I know, I prefer shorten lines, also I thing that is more consistent with the module code like I did, but if you want I will change it, do you want that? 
;)

AndreMiras: I don't really mind @opacam :+1:

JonasT: My vote is to keep it as `args.add_activity = args.add_activity or []`. I don't see any benefit in turning this into a ternary operator

opacam: Aaa..we can do the same in here:

`args.add_activity = getattr(args, 'add_activity', [])`

and

`args.activity_launch_mode = getattr(args, 'activity_launch_mode', '')`

JonasT: That doesn't do the same thing. `getattr` returns the value if the attribute *isn't present*, while `X or Y` will return `Y` if `X` is None and error if the attribute isn't present.

AndreMiras: Yes my bad, I got mislead by other things I've seen too often, but it involves the `and` keyword like here https://github.com/kivy/python-for-android/blob/dab2a8a/pythonforandroid/bootstraps/pygame/build/buildlib/jinja2.egg/jinja2/filters.py#L300
Thanks for the head up!

pythonforandroid/bootstraps/common/build/build.py
AndreMiras: True, the semantic would not be the same for the `args.services is None` use case. But is that even an use case with `add_argument('--service', dest='services', action='append')`? I'm really wondering

JonasT: hmm only if the default is wrong, I suppose. if we checked that the default has a proper value and never `None` then it would be fine

opacam:
Well, thanks to @KeyWeeUsr that noticed that the `--service` argument was disabled.

This is related with the problematic line discussed above, which I think that  all the discussion started because of the `--service` argument being disabled (which leaded me to change the discussed line to avoid compilation errors)...so...I restored the problematic line as it was originally and enabled the `--service` argument again....I think that this should close the above discussion...
 
I also modified the testapp_service in order that the test made in there matches one of the options explained in the python-for-android's docs to create a service (thanks again to @KeyWeeUsr :smile:...who  pointed me to the right direction via discord)

**This changes has been tested for python2 and python3 with the  service's test app (the one I modify in here)**

Note: While testing the new app, I realised that the file `Service.tmpl.java` needed a slight modifications to make it work with the new python core (a wrong path and a couple of missing imports) so I also fixed it.

Issue #1582, opened 2019-01-11
python2legacy - various warnings then no valid dependency graphs


hottwaj:
I have just tried to build my app using the new (post API 26) toolchain.  Unfortunately my app still needs to use pygame/sdl on android, so I saw in the docs that I need to use the python2legacy recipe

When I run P4A with the python2recipe selected, I get various warnings like this:

```
/home/jonathan/prog/src/p4a-2019/pythonforandroid/recipes/pyjnius/__init__.py:3: RuntimeWarning: Parent module 'pythonforandroid.recipes' not found while handling absolute import
  from pythonforandroid.patching import will_build
```

This happens for all recipes I have included in my build.

Finally, I get:

```
[ERROR]:   Build failed: Didn't find any valid dependency graphs. This means that some of your requirements pull in conflicting dependencies.
```

I guess that this is because some recipes depend on python2 (and not python2legacy)?

Thanks!

homdx:
Hello.
You not show in issue .p4a file or command line parameters or/and requirements.txt

AndreMiras:
Since #1582 was recently merged you could give it a try on master see if the problem is fixed.
e.g. via `buildozer android update`

hottwaj:
Thanks, I will give that a try - and sorry for the lack of p4a cmd line parameters etc, my concern (which I wasn't clear on) was that the python2legacy route was not currently usable because of the recent overhaul

I will close for now and come back with anything specific that I have trouble with.

Cheers!

PR #1587, opened 2019-01-14
Minimal fixes to make pygame bootstrap work with python2legacy

opacam:
The pygame bootstrap is not working with the latest changes in master branch:

  -  because the old python2 recipe has been moved to python2legacy and the pygame bootstrap was made for python2legacy recipe (the android paths for the new python2 are different...so...it's unlikely that it works without touching anything)
  - The java build method used for pygame bootstrap is ant (which has been deprecated in modern android ndks and the default java version is not anymore in linux distribution packages (1.5)...now we need version 1.7)
- The distributed jpeg library in sdl does not build fine with modern ndks (assembler issues)

So...saying that, this pr fixes all the mentioned issues and makes that the pygame bootstrap uses the refactored distribution methods (already in python recipe, no need to touch that). 

This has been tested with ndk-r17c, ndk-api=19, android-api=27 and arch `armeabi-v7a`.

This is a quick fix because **still will be issues to solve...pyjnius** is not working at runtime, despite it compiles/installs fine (pygame bootstrap uses the old `renpy` entries instead of the `kivy` ones...).

**Build Note**: because we still use ant build for pygame bootstrap, the path of the final apk is not the same as the ones used by other bootstraps (the others uses gradle)...so...buildozer may have troubles with that (I dont know that for sure...it's only my guess)

**Note about pyjnius**: I suspect that we can make pyjnius work by moving from ant to gradle (and move the relevant code from renpy folder to the kivy one)...but this is quite a work...I'm not sure if we should do that...

I know that has been some talks **about removing pygame...maybe it's the time...but even if we do that...this code does not affect the core of the project...** so I think that could be safely merged after the proper reviews/discussions because all the changes introduced in here only affects files needed by pygame/sdl...so...easy to remove later.

### DIFF
--- a/pythonforandroid/recipes/pygame_bootstrap_components/__init__.py
+++ b/pythonforandroid/recipes/pygame_bootstrap_components/__init__.py
@@ -25,5 +25,11 @@ class PygameJNIComponentsRecipe(BootstrapNDKRecipe):
         info('Unpacking was successful, deleting original container dir')
         shprint(sh.rm, '-rf', self.get_build_dir(arch))
 
+        info('Applying jpeg assembler patch')
+        ndk_15_plus_patch = join(self.get_recipe_dir(), 'jpeg-ndk15-plus.patch')
+        shprint(sh.patch, '-t', '-d',
+                join(self.get_build_container_dir(arch), 'jpeg'), '-p1',
+                '-i', ndk_15_plus_patch, _tail=10)
+
 
 recipe = PygameJNIComponentsRecipe()
--- a/pythonforandroid/bootstraps/common/build/jni/application/src/start.c
+++ b/pythonforandroid/bootstraps/common/build/jni/application/src/start.c
@@ -17,7 +17,9 @@
 #include "bootstrap_name.h"
 #ifndef BOOTSTRAP_USES_NO_SDL_HEADERS
 #include "SDL.h"
+#ifndef BOOTSTRAP_NAME_PYGAME
 #include "SDL_opengles2.h"
+#endif
 #endif
 #ifdef BOOTSTRAP_NAME_PYGAME
 #include "jniwrapperstuff.h"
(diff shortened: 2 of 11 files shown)


opacam:
**Update:** this pr may fix the issue described in #1582

pythonforandroid/recipes/pygame_bootstrap_components/__init__.py
```diff
@@ -25,5 +25,11 @@ def prebuild_arch(self, arch):
         info('Unpacking was successful, deleting original container dir')
         shprint(sh.rm, '-rf', self.get_build_dir(arch))
 
+        info('Applying jpeg assembler patch')
+        ndk_15_plus_patch = join(self.get_recipe_dir(), 'jpeg-ndk15-plus.patch')
+        shprint(sh.patch, '-t', '-d',
+                join(self.get_build_container_dir(arch), 'jpeg'), '-p1',
+                '-i', ndk_15_plus_patch, _tail=10)
```
AndreMiras: Just wondering why don't we apply the patch using the `patches` attribute?

AndreMiras:
Looking good to me. I'm just curious about why would you use `pygame` over `sdl2`?
The Travis conditional build is of course failing because it doesn't support other bootstraps yet, see https://github.com/kivy/python-for-android/issues/1588

Error was: `Didn't find any valid dependency graphs. This means that some of your requirements pull in conflicting dependencies.`

pythonforandroid/recipes/pygame_bootstrap_components/__init__.py
opacam: As far as I remember...it's for the jpeg path, If we use the default patching method the file will not be found...but I'm thinking now that maybe we can add the needed levels to the patch, so we could refactor this into the default patching mechanism...I suppose that this should work...I will try to make some tests about that ;)

AndreMiras: Yes that would be awesome, that would simplify the code a little bit :+1:

Issue #1570, opened 2019-01-09
p4a apk crash "IndexError: list index out of range"


homdx:
I probe compile hello world with env by @JonasT in docker:
https://github.com/JonasT/p4a-build-spaces/

./p4aspaces shell p4a-py3-api28ndk21
#p4a apk
```
log:

[INFO]:    -> running mv /root/.local/share/python-for-android/dists/unnamed_dist_1/_python_bundle/_python_bundle/site-packages/kivy/_clock.cpython-37m-x86_64-linux-gnu.so /root/.local/share/python-for-android/...(and 79 more)
[INFO]:    -> running mv /root/.local/share/python-for-android/dists/unnamed_dist_1/_python_bundle/_python_bundle/site-packages/kivy/weakproxy.cpython-37m-x86_64-linux-gnu.so /root/.local/share/python-for-andro...(and 85 more)
[INFO]:    -> running mv /root/.local/share/python-for-android/dists/unnamed_dist_1/_python_bundle/_python_bundle/site-packages/kivy/properties.cpython-37m-x86_64-linux-gnu.so /root/.local/share/python-for-andr...(and 87 more)
[INFO]:    <- directory context /root/test
[INFO]:    Stripping libraries
[INFO]:    -> running find /root/.local/share/python-for-android/dists/unnamed_dist_1/_python_bundle/_python_bundle/modules /root/.local/share/python-for-android/dists/unnamed_dist_1/libs -iname *.so
[INFO]:    Stripping libraries in private dir                                                                                                                                                                                                
[INFO]:    Frying eggs in /root/.local/share/python-for-android/dists/unnamed_dist_1/_python_bundle/_python_bundle/site-packages
[INFO]:    -> directory context /root/.local/share/python-for-android/dists/unnamed_dist_1
[INFO]:    Saving distribution info
[INFO]:    <- directory context /root/test
[INFO]:    # Your distribution was created successfully, exiting.
[INFO]:    Dist can be found at (for now) /root/.local/share/python-for-android/dists/unnamed_dist_1
[INFO]:    Of the existing distributions, the following meet the given requirements:
[INFO]:         unnamed_dist_1: min API 21, includes recipes (hostpython3, sdl2_image, sdl2_mixer, sdl2_ttf, python3, sdl2, six, pyjnius, kivy, requests3), built for archs (armeabi-v7a)
[INFO]:    unnamed_dist_1 has compatible recipes, using this one
Traceback (most recent call last):
[...]

PR #1591, opened 2019-01-15
Fix --private and others showing weird error when used without argument

JonasT:
When using some of the options like `--private` without an argument, which shouldn't be done but users make mistakes sometimes, this currently causes a weird backtrace error that isn't obvious to the average user. This pull request addresses that problem such that better error handling can happen

### DIFF
--- a/pythonforandroid/toolchain.py
+++ b/pythonforandroid/toolchain.py
@@ -772,7 +772,7 @@ class ToolchainCL(object):
                 if len(argx) > 1:
                     unknown_args[i] = '='.join(
                         (argx[0], realpath(expanduser(argx[1]))))
-                else:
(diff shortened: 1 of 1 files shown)


homdx:
Very good!!! @JonasT , thank you !

checked in added link
https://github.com/kivy/python-for-android/issues/1570#issuecomment-454531634

JonasT:
@inclement argparse supports `--` though after which `-` in front of a path is allowed. So this might break legitimate path arguments if that's not taken into account... could make things complicated

inclement:
@JonasT Sure, it's not that we need to reinvent the argparse logic so much as that it would probably improve more errors than it makes worse.

Issue #1588, opened 2019-01-14
Travis conditional boostrap build support


AndreMiras:
Currently the Travis conditional build script only support the `sdl2` bootstrap. This is because [sdl2 the bootstrap is hardcoded](https://github.com/kivy/python-for-android/blob/12ac837/ci/rebuild_updated_recipes.py#L67).
It would be great if it could at least try to rebuild another bootstrap if it gets updated. That could be done using the same way we detect for a recipe update.

Julian-O:
Travis no longer supported. Closing as Not Planned.

PR #1594, opened 2019-01-15
Make bootstrap dynamic for travis conditional builds

opacam:
This should solve: #1588

### DIFF
--- a/ci/constants.py
+++ b/ci/constants.py
@@ -63,6 +63,7 @@ BROKEN_RECIPES_PYTHON2 = set([
     'zeroconf',
     'zope',
 ])
+# to be created via https://github.com/kivy/python-for-android/issues/1514
 BROKEN_RECIPES_PYTHON3 = set([
     'brokenrecipe',
     # enum34 is not compatible with Python 3.6 standard library
--- a/ci/constants.py
+++ b/ci/constants.py
@@ -3,8 +3,7 @@ from enum import Enum
 
 class TargetPython(Enum):
     python2 = 0
-    python3crystax = 1
-    python3 = 2
+    python3 = 1
 
 
 # recipes that currently break the build
--- a/ci/rebuild_updated_recipes.py
+++ b/ci/rebuild_updated_recipes.py
@@ -48,7 +48,7 @@ def modified_recipes(branch='origin/master'):
     return recipes
 
 
-def build(target_python, requirements):
+def build(target_python, target_bootstrap, requirements):
     """
     Builds an APK given a target Python and a set of requirements.
     """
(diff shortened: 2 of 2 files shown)


AndreMiras:
Thanks for it @opacam I like the approach and we can probably merge after some testing.
The approach I was thinking of was to play with `git diff --name-only` and compile on bootstrap changes. So for instance if a PR touches `service_only` we recompile using this bootstrap.
With your approach it feels more like if the recipe is compatible with `webview` bootstrap only then that's the one that will be picked up. Am I right?

opacam:
Oh yes, uses the system that we use to resolve the dependencies and pick up one bootstrap (just the same as we do when building without specifying a bootstrap)

ci/constants.py
```diff
@@ -96,9 +97,15 @@ class TargetPython(Enum):
     'sympy',
     'vlc',
 ])
+BROKEN_RECIPES_PYTHON2 = set([
+])
+# to be created via https://github.com/kivy/python-for-android/issues/1514
+BROKEN_RECIPES_PYTHON3 = set([
```
AndreMiras: This list is already defined here https://github.com/kivy/python-for-android/blob/0563ef00ab5a64184f1290941ff2d8ee564cacbc/ci/constants.py#L70 why would we override?
I would keep that comment about #1514 however, good idea

ci/rebuild_updated_recipes.py
```diff
@@ -65,34 +65,71 @@ def build(target_python, requirements):
...
+    bootstrap = None
+    target_python = None
+    for n, target_python in enumerate(target_python_priorities):
+        logger.info(
+            'trying to get a bootstrap forcing target python: {}'.format(
+                target_python.name))
+        bootstrap = get_bootstrap(recipes | {target_python.name})
+        if bootstrap:
+            break
+        else:
```
AndreMiras: I understand you wanted to keep some meaningful logs to see the fallback in action, but we could greatly simplify the code if we drop that feature actually. Having the `trying to get a bootstrap forcing target python` above would be more than enough to guess when falling back. If we do so you can drop all the `else` block and the `enumerate` requirement

ci/rebuild_updated_recipes.py
```diff
@@ -65,34 +65,71 @@ def build(target_python, requirements):
...
+            # fallback to next TargetPython if current target is not compatible
+            logger.info('\t- incompatible with {}'.format(target_python.name))
+            if len(target_python_priorities) < n + 1:
+                logger.info('\t- falling back to {}'.format(
+                    target_python_priorities[n + 1].name))
+                continue
+    if not bootstrap:
+        logger.warning('we didn\'t find any valid combination of bootstrap and'
```
AndreMiras: Minor: this is a good example where I would use double quotes in my string, just to avoid escaping inner single quotes

ci/rebuild_updated_recipes.py
```diff
@@ -65,34 +65,71 @@ def build(target_python, requirements):
...
+                       ' target python...rebuild updated recipes cancelled.'
+                       'The recipes we couldn\'t rebuild are:'
+                       '\n\t-{}'.format('\n\t-'.join(list(recipes))))
```
AndreMiras: Minor: why would you need the `list()` here, can't you join directly?

opacam:
Thanks for the comments, tomorrow I will look into it..:sleeping:
:new_moon:

ci/constants.py
```diff
@@ -96,9 +98,13 @@ class TargetPython(Enum):
     'sympy',
     'vlc',
 ])
+# to be created via https://github.com/kivy/python-for-android/issues/XXXX
```
opacam: I think that we should create an issue page for python2.
What do you think @AndreMiras ?

ci/constants.py
```diff
@@ -5,12 +5,13 @@ class TargetPython(Enum):
     python2 = 0
     python3crystax = 1
```
opacam: Maybe we should remove `python3crystax`?

ci/rebuild_updated_recipes.py
```diff
@@ -65,34 +65,65 @@ def build(target_python, requirements):
...
-        logger.info('falling back to {}'.format(target_python.name))
+        pass
+    return bootstrap
+
+
+def main():
+    target_python_priorities = [
+        TargetPython.python3,
+        TargetPython.python2,
+        TargetPython.python2legacy
```
opacam: `python2legacy` is unlikely to be used unless we modify some `pygame` dependant recipe, which will be removed soon, so...I'm not sure if we have to introduce this :thinking:

opacam:
Again @AndreMiras, so many thanks for your comments of last night, you were right with all of that, and I must say, a lot better :smile:. I don't plan touch this pull request unless you request me, so feel free to modify wathever you consider necessary, no problem :wink: (ahhh...and I don't have any more `un-pushed` work related with this pr...so we shouln't have problems If you want to add some commit, I could `pull` your changes...so feel free to modify whatever you want)

ci/constants.py
AndreMiras: I don't know if it's worth the effort, but feel free if want

ci/constants.py
AndreMiras: Yes feel free :+1:

ci/rebuild_updated_recipes.py
AndreMiras: Yes by default I would be lazy and drop

opacam: Yes, but we already have the work done in here, so maybe we should maintain this, at least until @inclement can finish his `pygame removal work`, then maybe we could start to remove all the `python2legacy` stuff. It's easy to remove this, I'm more concerned with other `python2legacy` code...but we will add some more code to remove...mmm...@inclement, what is your opinion on this?

opacam: Ok, after thinking about it, I took the lazy road :grin:, removed `python2legacy`

Issue #1575, opened 2019-01-10
reportlab broken, possibly using wrong python header or compiler flags with python 3?


JonasT:
With latest master & `python3` (non-crystax), it seems there is some issue with reportlab including a Python header that expects `crypt.h` on the **host** (which I'm pretty sure shouldn't be included from there):
```
[INFO]:    Building reportlab for armeabi-v7a
[INFO]:    reportlab apparently isn't already in site-packages
[INFO]:    Building compiled components in reportlab
[INFO]:    -> directory context /root/.local/share/python-for-android/build/other_builds/reportlab-python3/armeabi-v7a__ndk_target_21/reportlab
[DEBUG]:   -> running python setup.py build_ext -v
================================================
Attempting build of _rl_accel & pyHnj
extensions from '/root/.local/share/python-for-android/build/other_builds/reportlab-python3/armeabi-v7a__ndk_target_21/reportlab/src/rl_addons/rl_accel'
================================================
================================================
Attempting build of _renderPM
extensions from '/root/.local/share/python-for-android/build/other_builds/reportlab-python3/armeabi-v7a__ndk_target_21/reportlab/src/rl_addons/renderPM'
================================================
will use package libart 2.3.12
# installing with freetype version 21
################################################
not downloading T1 font curve files
/root/.local/share/python-for-android/build/other_builds/hostpython3/desktop/hostpython3/Lib/distutils/dist.py:274: UserWarning: Unknown distribution option: 'install_requires'
  warnings.warn(msg)
running build_ext
building 'reportlab.lib._rl_accel' extension
creating build
creating build/temp.linux-x86_64-3.7
creating build/temp.linux-x86_64-3.7/root
creating build/temp.linux-x86_64-3.7/root/.local
creating build/temp.linux-x86_64-3.7/root/.local/share
creating build/temp.linux-x86_64-3.7/root/.local/share/python-for-android
creating build/temp.linux-x86_64-3.7/root/.local/share/python-for-android/build
creating build/temp.linux-x86_64-3.7/root/.local/share/python-for-android/build/other_builds
creating build/temp.linux-x86_64-3.7/root/.local/share/python-for-android/build/other_builds/reportlab-python3
creating build/temp.linux-x86_64-3.7/root/.local/share/python-for-android/build/other_builds/reportlab-python3/armeabi-v7a__ndk_target_21
[...]

PR #1595, opened 2019-01-17
Fix reportlab's recipe `crypt.h` error

opacam:
Caused because we try to compile the package with hostpython and we need to use target python or we will not have the right python headers.

Resolves: #1575


DanantST:
Hello! I'm new to building apk. How am I supposed to apply this fix?

PR #1597, opened 2019-01-17
Fix hardcoded entries (build platform) for core modules: archs and python

opacam:
This commit fixes the hardcoded entries for `build platform` that we still have in some of our core modules (arch and python). This should make possible to successfully build an apk from a macos...but I cannot confirm this.

Note: we still have some hardcoded entries regarding the `build platform`  in some recipes...


OptimusGREEN:
Yup, got me past that error, thanks

PR #1598, opened 2019-01-17
Add functions for obtaining the default storage paths

JonasT:
We just discussed this on discord and the situation with the storage available on Android is a little bit confusing, so I read through the docs a little and wrote an additional storage module to add after #1528 is in. Still untested / to be polished

### DIFF
--- a/doc/source/apis.rst
+++ b/doc/source/apis.rst
@@ -5,6 +5,60 @@ Working on Android
 This page gives details on accessing Android APIs and managing other
 interactions on Android.
 
+Storage paths
+-------------
+
+If you want to store and retrieve data, you shouldn't just save to
+the current directory, and not hardcode `/sdcard/` or some other
+path either - it might differ per device.
+
+Instead, the `android` module which you can add to your `--requirements`
+allows you to query the most commonly required paths::
+
+      from android.storage import app_storage_path
+      settings_path = app_storage_path()
+
+      from android.storage import primary_external_storage_path
+      primary_ext_storage = primary_external_storage_path()
+
+      from android.storage import secondary_external_storage_path
+      secondary_ext_storage = secondary_external_storage_path()
+
+`app_storage_path()` gives you Android's so-called "internal storage"
+which is specific to your app and cannot seen by others or the user.
+It compares best to the AppData directory on Windows.
+
+`primary_external_storage_path()` returns Android's so-called
+"primary external storage", often found at `/sdcard/` and potentially
+accessible to any other app.
+It compares best to the Documents directory on Windows.
+Requires `Permission.WRITE_EXTERNAL_STORAGE` to read and write to.
+
+`secondary_external_storage_path()` returns Android's so-called
+"secondary external storage", often found at `/storage/External_SD/`.
+It compares best to an external disk plugged to a Desktop PC, and can
+after a device restart become inaccessible if removed.
+Requires `Permission.WRITE_EXTERNAL_STORAGE` to read and write to.
+
+.. warning::
+   Even if `secondary_external_storage_path` returns a path
+   the external sd card may still not be present.
+   Only non-empty contents or a successful write indicate that it is.
+
+Read more on all the different storage types and what to use them for
+in the Android documentation:
+
+https://developer.android.com/training/data-storage/files
+
+A note on permissions
+~~~~~~~~~~~~~~~~~~~~~
+
+Only the internal storage is always accessible with no additional
+permissions. For both primary and secondary external storage, you need
+to obtain `Permission.WRITE_EXTERNAL_STORAGE` **and the user may deny it.**
+Also, if you get it, both forms of external storage may only allow
+your app to write to the common pre-existing folders like "Music",
+"Documents", and so on. (see the Android Docs linked above for details)
 
 Runtime permissions
 -------------------
--- a/pythonforandroid/recipes/android/src/android/storage.py
+++ b/pythonforandroid/recipes/android/src/android/storage.py
@@ -1,0 +1,103 @@
+
+from jnius import autoclass, cast
+import os
+
+
+Environment = autoclass('android.os.Environment')
+
+
+def _android_has_is_removable_func():
+    VERSION = autoclass('android.os.Build$VERSION')
+    return (VERSION.SDK_INT >= 24)
+
+
+def _get_sdcard_path():
+    """ Internal function to return getExternalStorageDirectory()
+        path. This is internal because it may either return the internal,
+        or an external sd card, depending on the device.
+        Use primary_external_storage_path()
+        or secondary_external_storage_path() instead which try to
+        distinguish this properly.
+    """
+    return (
+        Environment.getExternalStorageDirectory().getAbsolutePath()
+    )
+
+
+def app_storage_path():
+    """ Locate the built-in device storage used for this app only.
+
+        This storage is APP-SPECIFIC, and not visible to other apps.
+        It will be wiped when your app is uninstalled.
+
+        Returns directory path to storage.
+    """
+    PythonActivity = autoclass('org.kivy.android.PythonActivity')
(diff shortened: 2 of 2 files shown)


JonasT:
By the way I'm aware this might be covered by `plyer`, but this is so basic (especially `app_storage_path()`) that it really should be available easily without additional larger dependencies, and without this apps like even kivy actually (or mine, admittedly) may hardcode `/sdcard` which *may not actually exist* on some oddball devices! So I think this should be added & documented in the p4a core android module (I will add documentation once the SDL2 pull request is in, since that one touches the section I'd also need to modify here so that will most likely leave me in conflict territory if I do that now)

pythonforandroid/recipes/android/src/android/storage.py
```diff
@@ -0,0 +1,83 @@
...
+        Environment = autoclass('android.os.Environment')
+        try:
+            sdpath = (Environment.getExternalStorageDirectory().
+                                  getAbsolutePath())
+            # Apparently this can both return primary (built-in) or
+            # secondary (removable) external storage depending on the device,
+            # therefore check that we got what we wanted:
+            if not Environment.isExternalStorageRemovable(sdpath):
+                return sdpath
+        except Exception:
```
inclement: Can we be more explicit about what Exception could occur?

pythonforandroid/recipes/android/src/android/storage.py
```diff
@@ -0,0 +1,83 @@
...
+        Environment = autoclass('android.os.Environment')
+        try:
+            sdpath = (Environment.getExternalStorageDirectory().
+                                  getAbsolutePath())
+            # Apparently this can both return primary (built-in) or
+            # secondary (removable) external storage depending on the device,
+            # therefore check that we got what we wanted:
+            if Environment.isExternalStorageRemovable(sdpath):
+                return sdpath
+        except Exception:
```
inclement: Same here about the Exception.

pythonforandroid/recipes/android/src/android/storage.py
```diff
@@ -0,0 +1,83 @@
...
+            pass
+    if "EXTERNAL_STORAGE" in os.environ:
+        return os.environ["EXTERNAL_STORAGE"]
+    raise RuntimeError(
+        "unexpectedly failed to determine " +
+        "primary external storage path"
+    )
+
+def secondary_external_storage_path():
```
inclement: Am I right to read that this function is actually identical to the previous one?

pythonforandroid/recipes/android/src/android/storage.py
JonasT: the `isExternalStorageRemovable` condition is different, which marks the difference between internal and external storage

pythonforandroid/recipes/android/src/android/storage.py
JonasT: the isExternalStorageRemovable condition is different, which marks the difference between internal and external storage

pythonforandroid/recipes/android/src/android/storage.py
AndreMiras: Good catch, also seems addressed, closing

doc/source/apis.rst
```diff
@@ -5,19 +5,72 @@ Working on Android
...
+      settings_path = app_storage_path()
+
+      from android.storage import primary_external_storage_path
+      primary_ext_storage = primary_external_storage_path()
+
+      from android.storage import most_portable_storage_path
+      portable_storage = most_portable_storage_path()
+
+`app_storage_path()` gives you Android's so-called "internal storage".
+It compares best to the AppData directory on Windows.
```
AndreMiras: We may say that unlike the two external storage, this one gets wiped out when the application is removed.

pythonforandroid/recipes/android/src/android/storage.py
```diff
@@ -0,0 +1,77 @@
...
+def is_external_storage_available();
+    """ Returns whether the secondary external storage is available &
+        usable (True) or not (False).
+    """
+    try:
+        p = secondary_external_storage_path()
+        if p is not None:
+            if os.path.exists(os.path.join(p, "Documents")):
+                # Looks like it's properly mounted!
+                return True
```
AndreMiras: Minor: simply go with:
```python
return os.path.exists(os.path.join(p, "Documents"))
```

AndreMiras:
This is looking promising @JonasT what about fixing the little conflict and make it to the tree?

JonasT:
Sounds good, although it should still be tested (I haven't so far :joy: ) - I'll give it a try and report back, and anyone else is also open to do so

pythonforandroid/recipes/android/src/android/storage.py
```diff
@@ -0,0 +1,90 @@
...
+
+        This is storage is SHARED, and visible to other apps and the user.
+        It will remain untouched when your app is uninstalled.
+
+        Returns directory path to storage.
+
+        WARNING: You need storage permissions to access this storage.
+    """
+    Build = autoclass('android.os.Build')
+    if Build.VERSION.SDK_INT >= 24:
```
AndreMiras: Fails at runtime:
```
07-18 23:09:07.928 24730 25122 I python  :    File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/android/storage.py", line 36, in primary_external_storage_path
07-18 23:09:07.929 24730 25122 I python  :  AttributeError: type object 'android.os.Build' has no attribute 'VERSION'
```
So what you want actually is:
```python
VERSION = autoclass('android.os.Build$VERSION')
if VERSION.SDK_INT >= 24:
```

pythonforandroid/recipes/android/src/android/storage.py
```diff
@@ -0,0 +1,90 @@
...
+        It will remain untouched when your app is uninstalled.
+
+        Returns None if not found, otherwise path to storage.
+
+        WARNING: You need storage permissions to access this storage.
+                 If it is not writable and presents as empty even with
+                 permissions, then the external sd card may not be present.
+    """
+    Build = autoclass('android.os.Build')
+    if Build.VERSION.SDK_INT >= 24:
```
AndreMiras: Same here:
```python
VERSION = autoclass('android.os.Build$VERSION')
if VERSION.SDK_INT >= 24:
```
And since we use it twice maybe we want to make it global, same for `Environment`.
What do you think?

AndreMiras:
Thanks for updating, it looks good overall, but fails runtime with:
```
AttributeError: type object 'android.os.Build' has no attribute 'VERSION'
```
The fix is fairly easy and I'd be happy to test once you have updated the PR

AndreMiras:
Hi @JonasT I'm about to give that one another try. Could you also rebase upstream/develop?

AndreMiras:
I gave it a try and it's crashing. I'm working on a fix and I'll add commits to the PR.
One of the crash could have been easily caught by unit testing. I'll work on that too later

JonasT:
Let me know if you need branch access on the repo fork itself, I really wouldn't mind giving you! I want to know how these "adding a commit" things work :smile: and thanks for taking over, that's nice :slightly_smiling_face: (I can also work on some tests though if you want me to)

AndreMiras:
I think access to your fork is only if I want to push force, for adding commits it should be just fine now.
I'll let you know.
I can work on the unit test I don't mind and I know what broke hence what to test.

AndreMiras:
I edited through the GitHub UI, I hope I didn't do a mess :grin: 
I'm going to re-test it from scratch now see how it goes. If things run OK and build is green I'll merge

JonasT:
Neat! Actually looking forward to use this myself in my apps :heart: it's an overdue addition for sure

Sahil-pixel:
path of secondary storage is none ..please fix this 
: secondary_external_storage_path: None

PR #1601, opened 2019-01-19
[WIP] Restores the ability to compile the python files into .pyo/.pyc (for both versions of python)

opacam:
We already had before this feature for python2 (python2legacy now)  and we loosed it with the recently merged core update. Also we will be able to compile our python install files for python 3 (new feature for python3). But be aware that to achieve this we must modify a little our source code because as of python 3.5 the `.pyo extension` for compiled python files has been removed in favour of  `.pyc extension`, so it requires to modify `start.c` and `PythonActivity.java` files, which only contemplated the .pyo extension.

Notes:
  - the default behaviour will be to compile all python files to the corresponding compiled extension, so **this way the apk will run faster**
  - also we compile the code with the option `-00` which will strip the docs, so we **will generate lighter apks**
  - this change **affects** bootstraps: **sdl2, webview and service_only**

See also:
  - `PEP 488 -- Elimination of PYO files` (https://www.python.org/dev/peps/pep-0488/)
  - `PEP 3147 -- PYC Repository Directories` as to why a separate directory is used (https://www.python.org/dev/peps/pep-3147/)

Status: **I leave the `WIP` status because this may conflict with the sdl2 update by** @JonasT (#1528) and, probably, the merge conflicts will be aesier to solve here than in there, so...let @JonasT decide if we must remove the `WIP` status (be aware that this will also be an enhancement)

**Thanks to @tito which make me notice that we had loose that feature**

### DIFF
--- a/pythonforandroid/python.py
+++ b/pythonforandroid/python.py
@@ -6,6 +6,7 @@ build our python3 and python2 recipes and his corresponding hostpython recipes.
 from os.path import dirname, exists, join
 from shutil import copy2
 from os import environ
+import subprocess
 import glob
 import sh
 
@@ -71,7 +72,7 @@ class GuestPythonRecipe(TargetPythonRecipe):
     '''The directories that we want to omit for our python bundle'''
 
     stdlib_filen_blacklist = [
-        '*.pyc',
+        '*.py',
         '*.exe',
         '*.whl',
     ]
@@ -84,12 +85,22 @@ class GuestPythonRecipe(TargetPythonRecipe):
     '''The directories from site packages dir that we don't want to be included
     in our python bundle.'''
 
-    site_packages_filen_blacklist = []
+    site_packages_filen_blacklist = [
+        '*.py'
+    ]
     '''The file extensions from site packages dir that we don't want to be
     included in our python bundle.'''
 
     opt_depends = ['sqlite3', 'libffi', 'openssl']
     '''The optional libraries which we would like to get our python linked'''
+
+    compiled_extension = '.pyc'
+    '''the default extension for compiled python files.
+
+    .. note:: the default extension for compiled python files has been .pyo for
+        python 2.x-3.4 but as of Python 3.5, the .pyo filename extension is no
+        longer used and has been removed in favour of extension .pyc
+    '''
 
     def __init__(self, *args, **kwargs):
         self._ctx = None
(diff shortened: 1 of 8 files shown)


hackalog:
Just a note (learned in https://github.com/kivy/kivy-ios/pull/340) If you are going to ship only `.pyc` files, it is *mandatory* to run python with -OO, or to set PYTHONOPTIMIZE=2.

opacam:
Ahhh, I didn't know about that, I leaved a `todo` in there to make it optional this option. I will remove that and make sure that `-00` is always aplied...and I will update the post.

¡¡¡Thanks @hackalog!!!

:smile:

pythonforandroid/python.py
```diff
@@ -249,26 +260,47 @@ def include_root(self, arch_name):
...
+        .. note:: python2 compiles the files into extension .pyo, but in
+            python3, and as of Python 3.5, the .pyo filename extension is no
+            longer used...uses .pyc (https://www.python.org/dev/peps/pep-0488)
+        '''
+        args = [self.ctx.hostpython]
+        if self.ctx.python_recipe.name == 'python3':
+            args += ['-OO', '-m', 'compileall', '-b', '-f', dir]
+        else:
+            args += ['-OO', '-m', 'compileall', '-f', dir]
+        subprocess.call(args)
```
AndreMiras: Why not using the `sh` module?

opacam: Because it fails:

  - in python 3 fails when the compilation is about to finish (after compile zipfile module) without a clear message of why it failed...still seems that the files are compiled well but to bypass the error we would had to put a `try/except` (mmm...I don't like that without knowing why it fails...)
  - in python 2  the compilation not even starts and complains about elf32 ...

Then I realized that we have another call to compile all files in meth:`~pythonforandroid.bootstraps.common.build.build.compile_dir` and in there we used the `subprocess.call`...so I tried with that...and it works fine for both versions of python :smile:

:thinking:...my guess is that some of our environment variables is causing some kind of conflict between hostpython and python when we try to use the compileall command with sh...but I'm not sure about that...maybe @inclement, @tito, @hackalog or @JonasT  could give us some more light in here...

opacam:
Thanks @AndreMiras :blush:

**Update:** I have done a `rebase` for the branch of this pr on top of the current master branch, to make sure that the tests are properly passed so we test this pr and the recently merged fix for sdk license error...here we go...

pythonforandroid/python.py
JonasT: Just to add as extra explanation: why you have to look up how `sh` works because it's unclear how it handles 1. the environment, 2. stdout/stderr buffering, 3. stdout/stderr redirect, and similar things. (for some of which p4a is using a really horrible wrapper in the logging file) So sure, the call itself is simple enough that it's not much of an issue, but if you care for details the `sh` module IMHO is really kind of a problem just because it isn't obvious for someone only familiar with `submodule` what it'll do or which parameters make it do certain things

inclement: I'm not opposed to using subprocess, but I do think the sh module is pretty nice. I don't really see anything you mention as an issue, it isn't obvious how subprocess works either until you read the documentation, although it's fair to say that more people will have done so in advance. Most people will never have to read the details of either.

That said, sh is overused in p4a partly due to me being lazy when originally duplicating the functionality of the old-p4a shell scripts.

I would most like to use some nice pythonic file management module for a lot of things that in a shell script would be calls to cp, rm, mkdir etc.. It annoys me that `shutil` and `os` are both quite poorly structured. https://github.com/inclement/nosh is the kind of thing I had in mind.

JonasT: @inclement not sure I agree with `Most people will never have to read the details of either`, at a point where that starts to matter people will have, and that's exactly the point where `sh` then ends up problematic. Therefore I personally prefer to stick to core modules for such basic functionality even if they might arguably be not nearly as pythonic as they should be, but of course that's just my opinion

AndreMiras: @opacam OK failing is a fair enough point, I was just wondering anyway :stuck_out_tongue: 
@JonasT I didn't know the `sh` module before knowing p4a, I can't tell if it's that great, but I can tell it's not that bad. I'm also more a fan of going with core modules as much as possible. I was asking @opacam why he didn't use, because this is what's being used by the project so far, so not using it could have a reason I was curious about

inclement:
@opacam I'm looking to merge this but it has conflicts due to some recent merges, would you be able to rebase it again? I would normally do it myself, but now that we have mandatory PR approval I wouldn't then be able to merge it.

opacam:
@inclement, no problem, let me do one more test with the conflicts solved and I will push the changes....testing right now

inclement:
@opacam Thanks, I'm also testing it, I just don't think I can push my merged version without someone else approving it.

opacam:
python3 is also ok (tested with sqlite openssl app)

@inclement, now it should be ready

inclement:
I've also tested it, seems great. Thanks!

PR #1604, opened 2019-01-20
Conditional build script fixes

AndreMiras:
- explicitly sets default target python3 to build order
- uses logger rather than print
- updates libzbar recipe to demo python2 fallback (E126 lint fix)


ci/rebuild_updated_recipes.py
```diff
@@ -74,21 +77,24 @@ def build(target_python, requirements):
...
-    print('recipes to build:', recipes)
+    logger.info('recipes to build: {}'.format(recipes))
     context = Context()
-    build_order, python_modules, bs = get_recipe_order_and_bootstrap(
-        context, recipes, None)
-    # fallback to python2 if default target is not compatible
-    if target_python.name not in build_order:
-        print('incompatible with {}'.format(target_python.name))
+    # forces the default target
+    recipes_and_target = recipes | set([target_python.name])
```
AndreMiras: required because of https://github.com/kivy/python-for-android/blob/4ed9dff/pythonforandroid/graph.py#L131

pythonforandroid/recipes/libzbar/__init__.py
```diff
@@ -17,7 +17,7 @@ class LibZBarRecipe(Recipe):
 
     def should_build(self, arch):
         return not os.path.exists(
-                os.path.join(self.ctx.get_libs_dir(arch.arch), 'libzbar.so'))
+            os.path.join(self.ctx.get_libs_dir(arch.arch), 'libzbar.so'))
```
AndreMiras: Minor E126 lint fix to demonstrate conditional recipe rebuild.
The `libzbar` recipe is currently `python2` only so the conditional build script should first try with current default target `python3` and then fallback to `python2`

PR #1605, opened 2019-01-21
Fixes crash on ctypes with Python 3

tito:
Without the patch, it is possible that the Python 3 `configure` catch libffi from the system, and use it as the flags are inserted before ours. So basically it compiles with the headers of the system libffi, but runs with the target libffi. And lead to a crash on the device:

```
01-20 20:38:17.272 13814 13814 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
01-20 20:38:17.272 13814 13814 F DEBUG   : Build fingerprint: 'HUAWEI/BAH/HWBAH-Q:7.0/HUAWEIBAH-W09/C100B253:user/release-keys'
01-20 20:38:17.272 13814 13814 F DEBUG   : Revision: '0'
01-20 20:38:17.272 13814 13814 F DEBUG   : ABI: 'arm'
01-20 20:38:17.272 13814 13814 F DEBUG   : pid: 13790, tid: 13810, name: SDLThread  >>> fr.connectage.connectage <<<
01-20 20:38:17.272 13814 13814 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0xffffffff
01-20 20:38:17.272 13814 13814 F DEBUG   :     r0 ffffffff  r1 cfe7e650  r2 00000019  r3 cfe7e728
01-20 20:38:17.272 13814 13814 F DEBUG   :     r4 cf9eb7c0  r5 e66dc000  r6 cf9eb56c  r7 00000001
01-20 20:38:17.272 13814 13814 F DEBUG   :     r8 cf9eb548  r9 00000000  sl 00000000  fp d16fc310
01-20 20:38:17.272 13814 13814 F DEBUG   :     ip d1c06218  sp d16fc308  lr deb3a798  pc deb3a7cc  cpsr a00e0010
01-20 20:38:17.273 13814 13814 F DEBUG   : 
01-20 20:38:17.273 13814 13814 F DEBUG   : backtrace:
01-20 20:38:17.273 13814 13814 F DEBUG   :     #00 pc 0000d7cc  /data/data/fr.connectage.connectage/files/app/_python_bundle/modules/_ctypes.cpython-37m.so
```

Starting from 3.6, `--with-system-ffi`/`--without-system-ffi` got deprecated, and ignored in 3.7.

2 possibles ways to fix:
- patching `configure`, like this PR
- change the way p4a works, aka install somewhere the recipes we do (like kivy-ios does with a root directory installation) and then point `PKG_CONFIG_PATH` to it, to pick our `libffi.pc` instead of the system.

### DIFF
--- a/pythonforandroid/recipes/python3/__init__.py
+++ b/pythonforandroid/recipes/python3/__init__.py
@@ -24,6 +24,10 @@ class Python3Recipe(GuestPythonRecipe):
     depends = ['hostpython3']
     conflicts = ['python3crystax', 'python2', 'python2legacy']
 
+    patches = [
+        'patches/fix-ctypes-with-system-ffi.patch'
+    ]
+
     configure_args = (
         '--host={android_host}',
         '--build={android_build}',
--- a/pythonforandroid/recipes/python3/patches/fix-ctypes-with-system-ffi.patch
+++ b/pythonforandroid/recipes/python3/patches/fix-ctypes-with-system-ffi.patch
@@ -1,0 +1,10 @@
+--- a/configure.orig	2019-01-21 16:30:12.359453312 +0100
++++ b/configure	2019-01-21 16:31:07.481526333 +0100
+@@ -10246,6 +10246,7 @@
+ else
+     LIBFFI_INCLUDEDIR=""
+ fi
++LIBFFI_INCLUDEDIR=""
+ 
+ 
+ # Check for use of the system libmpdec library


inclement:
I merged @opacam's #1609 for this instead as it seemed nicer to avoid introducing a patch for python3. Hope that's okay!

PR #1607, opened 2019-01-25
Fix strip in bootstrap

j-devel:
Due to hardcoding, the arch-dependent `strip` command is not invoked when building for `arch` other than `--arch=armeabi-v7a` (e.g. `strip` not detected when `--arch=x86`).

This PR fixes

- the problem above, and
- the case of applying `strip` erroneously when the file name is `''` (an empty string).

### DIFF
--- a/pythonforandroid/bootstrap.py
+++ b/pythonforandroid/bootstrap.py
@@ -278,6 +278,8 @@ class Bootstrap(object):
 
         logger.info('Stripping libraries in private dir')
         for filen in filens.split('\n'):
+            if not filen:
+                continue  # skip the last ''
             try:
                 strip(filen, _env=env)
             except sh.ErrorReturnCode_1:
--- a/pythonforandroid/bootstrap.py
+++ b/pythonforandroid/bootstrap.py
@@ -263,11 +263,11 @@ class Bootstrap(object):
             info('Python was loaded from CrystaX, skipping strip')
             return
         env = arch.get_env()
-        strip = which('arm-linux-androideabi-strip', env['PATH'])
+        strip = env['STRIP'].split(' ')[0]
         if strip is None:
             warning('Can\'t find strip in PATH...')
             return
-        strip = sh.Command(strip)
+        strip = sh.Command(strip).bake('--strip-unneeded')
 
         libs_dir = join(self.dist_dir, '_python_bundle',
                         '_python_bundle', 'modules')
--- a/pythonforandroid/bootstrap.py
+++ b/pythonforandroid/bootstrap.py
@@ -10,7 +10,7 @@ import shutil
 from pythonforandroid.logger import (warning, shprint, info, logger,
                                      debug)
 from pythonforandroid.util import (current_directory, ensure_dir,
-                                   temp_directory, which)
+                                   temp_directory)
 from pythonforandroid.recipe import Recipe
 
 
--- a/pythonforandroid/bootstrap.py
+++ b/pythonforandroid/bootstrap.py
@@ -265,7 +265,9 @@ class Bootstrap(object):
             return
         env = arch.get_env()
         tokens = shlex.split(env['STRIP'])
-        strip = sh.Command(tokens[0]).bake(tokens[1:])
+        strip = sh.Command(tokens[0])
+        if len(tokens) > 1:
+            strip = strip.bake(tokens[1:])
 
         libs_dir = join(self.dist_dir, '_python_bundle',
                         '_python_bundle', 'modules')
--- a/pythonforandroid/bootstrap.py
+++ b/pythonforandroid/bootstrap.py
@@ -1,6 +1,7 @@
 from os.path import (join, dirname, isdir, normpath, splitext, basename)
 from os import listdir, walk, sep
 import sh
+import shlex
 import glob
 import importlib
 import os
@@ -263,8 +264,8 @@ class Bootstrap(object):
             info('Python was loaded from CrystaX, skipping strip')
             return
         env = arch.get_env()
-        tokens = env['STRIP'].split(' ')
-        strip = reduce(lambda cmd, t: cmd.bake(t), tokens, sh.Command(tokens.pop(0)))
+        tokens = shlex.split(env['STRIP'])
+        strip = sh.Command(tokens[0]).bake(tokens[1:])
 
         libs_dir = join(self.dist_dir, '_python_bundle',
                         '_python_bundle', 'modules')
--- a/pythonforandroid/bootstrap.py
+++ b/pythonforandroid/bootstrap.py
@@ -263,11 +263,8 @@ class Bootstrap(object):
             info('Python was loaded from CrystaX, skipping strip')
             return
         env = arch.get_env()
-        strip = env['STRIP'].split(' ')[0]
-        if strip is None:
-            warning('Can\'t find strip in PATH...')
-            return
-        strip = sh.Command(strip).bake('--strip-unneeded')
+        tokens = env['STRIP'].split(' ')
+        strip = reduce(lambda cmd, t: cmd.bake(t), tokens, sh.Command(tokens.pop(0)))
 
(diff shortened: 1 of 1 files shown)


j-devel:
Thanks for reviewing code. I further considered and tested the following point:

> it might result in the option being specified twice. 

As a result, I think the possibility of doubly applying `STRIP` is none.  Here's my findings.

The `STRIP` of our concern is part of `strip_libraries()` and this function takes care of only python-core and android-jni related shared objects, which are specifically,
```
DIST/_python_bundle/_python_bundle/modules/*.so
DIST/libs/ARCH/*.so
```
And these shared objects are NOT stripped devoid of this function.

On the other hand, shared objects belonging to regular recipes are independent of
`strip_libraries()` because they reside under `DIST/_python_bundle/site-packages/` (e.g.  `DIST/_python_bundle/site-packages/jnius/jnius.so`).

I performed a test build with `--arch=x86`, got stripped shared objects (that are stripped only once by `strip_libraries()`), and verified Python running without errors on Android.

JonasT:
@j-devel regarding `split(' ')[0]`: I oppose that idea, that seems very problematic to me since it's not unthinkable the strip command contains other options that are needed. (e.g. I'm pretty sure it's not completely unthinkable a more unusual compiler toolchain might require something like `compiler-name --strip` to invoke a strip tool) So that seems like a bad idea to me.

I was more thinking of considering where `--strip-unneeded` should actually be added. After all, the problem appears to be that this isn't completely obvious, hence done twice. But this shouldn't hold back this pull request, since it's more of a nitpick. But please don't add `split(' ')[0]`, that seems like far too much into the other direction than is good for anything

pythonforandroid/bootstrap.py
```diff
@@ -263,11 +263,11 @@ def strip_libraries(self, arch):
             info('Python was loaded from CrystaX, skipping strip')
             return
         env = arch.get_env()
-        strip = which('arm-linux-androideabi-strip', env['PATH'])
+        strip = env['STRIP'].split(' ')[0]
```
JonasT: I don't think this `split` is a good idea. Also see my comment in the conversation

j-devel: > it's not unthinkable the strip command contains other options that are needed. (e.g. I'm pretty sure it's not completely unthinkable a more unusual compiler toolchain might require something like `compiler-name --strip` to invoke a strip tool) So that seems like a bad idea to me.

This is a very important point that I wasn't well aware of.  In 73d13ad, I have updated code so that the `strip` command object reflects the original `STRIP` command specified in `env['STRIP']` including option strings.  This way, we don't need to add `--strip-unneeded` in this part again (already done in archs.py).

pythonforandroid/bootstrap.py
```diff
@@ -263,11 +263,8 @@ def strip_libraries(self, arch):
             info('Python was loaded from CrystaX, skipping strip')
             return
         env = arch.get_env()
-        strip = which('arm-linux-androideabi-strip', env['PATH'])
-        if strip is None:
-            warning('Can\'t find strip in PATH...')
-            return
-        strip = sh.Command(strip)
+        tokens = env['STRIP'].split(' ')
+        strip = reduce(lambda cmd, t: cmd.bake(t), tokens, sh.Command(tokens.pop(0)))
```
JonasT: Ok I played around with this, and it appears this can be written as:

```
tokens = env['STRIP'].split(' ')
strip = sh.Command(tokens[0]).bake(tokens[1:])
```
I would suggest that would be a little more readable

(This also works when `len(token) == 1` and consequently `tokens[:1] == []`, I tested that special case)

Sorry I really didn't want to go on about this forever from my initial duplicate option nitpick, I hope I'm not wearing you out too much :grimacing: but the reduce/lambda variant is a little hard to read

pythonforandroid/bootstrap.py
JonasT: I like the functionality you have now, but I would suggest writing it a little simpler:

```
tokens = env['STRIP'].split(' ')
strip = sh.Command(tokens[0]).bake(tokens[1:])
```
(also works when `len(tokens) == 1`/no options are used)

JonasT: Wait, I'm just noticing that will break with space character arguments. I think this should be done instead:

```
tokens = shlex.split(env['STRIP'])
strip = sh.Command(tokens[0]).bake(tokens[1:])
```

pythonforandroid/bootstrap.py
JonasT: Additional note: as written above I just noticed the space char splitting could break things.
```
tokens = shlex.split(env['STRIP'])
strip = sh.Command(tokens[0]).bake(tokens[1:])
```
This should work correctly (requires `import shlex` at the beginning of the file, which is a core/stdlib module as far as I'm aware)

pythonforandroid/bootstrap.py
```diff
@@ -263,11 +264,8 @@ def strip_libraries(self, arch):
             info('Python was loaded from CrystaX, skipping strip')
             return
         env = arch.get_env()
-        strip = which('arm-linux-androideabi-strip', env['PATH'])
-        if strip is None:
-            warning('Can\'t find strip in PATH...')
-            return
-        strip = sh.Command(strip)
+        tokens = shlex.split(env['STRIP'])
+        strip = sh.Command(tokens[0]).bake(tokens[1:])
```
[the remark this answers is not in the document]
j-devel: Added the check in 7ad0207.

> Looks like there are some pep8 issues breaking travis

I would like to look into this issue soon and come back.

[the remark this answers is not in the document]
JonasT: @j-devel @inclement .bake works fine with an empty list. the if condition is not necessary, the above code works fine even for a single argument

pythonforandroid/bootstrap.py
```diff
@@ -263,11 +264,10 @@ def strip_libraries(self, arch):
...
             return
         env = arch.get_env()
-        strip = which('arm-linux-androideabi-strip', env['PATH'])
-        if strip is None:
-            warning('Can\'t find strip in PATH...')
-            return
-        strip = sh.Command(strip)
+        tokens = shlex.split(env['STRIP'])
+        strip = sh.Command(tokens[0])
+        if len(tokens) > 1:
```
JonasT: have you tested this without the `if` and a single argument? I tested bake with an empty list and it worked fine, I'm pretty sure this conditional is not required

inclement: Oh, you're quite right, I misremembered what list slice syntax would do outside the bounds of the list. It's fine to revert this.

j-devel: @JonasT Yes, I tested the `len(tokens) == 1` case, and it basically works.  But, I found that `.bake()` raises a complaint when it's argument is `[]`  (see the following REPL test).
```
>>> import sh
>>> sh.Command('strip').bake([])
__main__:1: UserWarning: Empty list passed as an argument to '/opt/local/bin/strip'. If you're using glob.glob(), please use sh.glob() instead.
```
If we want to eliminate this warning, I think we still need the `if` check.  Thoughts? @JonasT @inclement

JonasT: @j-devel oh weird, have never seen that (maybe I got a different version?). In that case definitely keep the check in

pythonforandroid/bootstrap.py
```diff
@@ -263,11 +264,10 @@ def strip_libraries(self, arch):
             info('Python was loaded from CrystaX, skipping strip')
             return
         env = arch.get_env()
-        strip = which('arm-linux-androideabi-strip', env['PATH'])
```
AndreMiras: Since `which` is not used anymore, would you mind removing the import to please the CI :pray: 
```
pythonforandroid/bootstrap.py:12:1: F401 'pythonforandroid.util.which' imported but unused
```

inclement:
Great, thanks!

PR #1608, opened 2019-01-25
Fix getting empty "modules" directory when arch is not armeabi-v7a

j-devel:
In `create_python_bundle()`, copy source path of built modules (*.so) is hardcoded to `arm`. This results in an empty `modules` directory when build `arch` is other than `armeabi-v7a`.

This PR fixes the above issue; accordingly, the `modules` directory is generated as expected (when arch is `x86`, `x86_64`,`arm64-v8a`, etc).

### DIFF
--- a/pythonforandroid/python.py
+++ b/pythonforandroid/python.py
@@ -263,8 +263,9 @@ class GuestPythonRecipe(TargetPythonRecipe):
             self.get_build_dir(arch.arch),
             'android-build',
             'build',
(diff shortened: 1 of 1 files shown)


j-devel:
Thanks!

> playing with the `command_prefix` string feels a bit hacky, I didn't see what else we could use.

Indeed.  In future, if this `arch.command_prefix.split('-')[0]` stuff starts appearing multiple times in other part of code, I think it's time to consider refactoring it as a member/method of the `Arch` class.

PR #1609, opened 2019-01-25
Fix libffi/ctypes - wrong libffi headers when building python

opacam:
It has been reported of errors with `ctypes`, even when the `ctypes` module seems to be built fine.
@tito is one of the affected by this and he found that, in some system configuration, python could be:

```compiled with the headers of the system libffi, but runs with the target libffi```

That will end in an app crash...

So...this pr makes that the python build make use of our libffi headers.

This would be a third solution to add to the ones proposed by @tito in #1605

**This solution avoids to patch python3 and will also work for python2**

!!!So many thanks @tito to found where the problem started :smile:!!!

### DIFF
--- a/pythonforandroid/python.py
+++ b/pythonforandroid/python.py
@@ -178,6 +178,10 @@ class GuestPythonRecipe(TargetPythonRecipe):
         if 'libffi' in self.ctx.recipe_build_order:
             info('Activating flags for libffi')
             recipe = Recipe.get_recipe('libffi', self.ctx)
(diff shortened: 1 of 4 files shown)


JonasT:
:+1: this looks like the more comprehensive solution that avoids unnecessary patches to the project itself. While it appears to add a little complexity, it is limited to the libffi-recipe and I personally think this approach is the better one (not to say that @tito 's solution is bad or anything)

PR #1610, opened 2019-01-25
Fix incorrect site-packages path breaking keyboard test app at runtime

JonasT:
The keyboard app crashes at launch right now because it has outdated assumptions about the site packages folder and it tries to list things from there. This changes the path accordingly, and avoids a crash in the future if the path shouldn't be present/moved again.


JonasT:
Ah right, I added a duplicate `sys` import. I'll remove it & retest, hang on

JonasT:
tested, can run keyboard app fine :+1:

PR #1615, opened 2019-01-27
Make python flags relative for Android.mk (bootstraps/common...)

opacam:
For bootstraps/common/build/jni/application/src/Android.mk

To fix possible linkage problems in some OS and because it's the recommended way.

See also: https://stackoverflow.com/questions/19720421/android-mk-relative-or-absolute-paths


inclement:
I think I misunderstood the diff when you mentioned this before - what is this actually fixing? It seems counterintuitive to me that switching to relative paths should be better, because they are difficult to understand by eye and more vulnerable to falling out of date.

I couldn't find where this is recommended in the NDK docs, as claimed in the linked stackoverflow post.

inclement:
Ah, found it at https://developer.android.com/ndk/guides/android_mk

> We recommend avoiding absolute file paths; relative paths make your Android.mk file more portable.

This isn't true for python-for-android because we are autogenerating everything anyway, so we have different concerns. I think the file paths should remain absolute.

The only portability gain of using the relative paths is really that the whole build folder could be moved elsewhere, but that seems only a very minor benefit considering there's no workflow that requires it.

opacam:
@inclement, you are right

but I was thinking that we never finished the conversation that we started in the core-update pr and I applied some changes in there that we never finished to discuss and right now those are already in master branch...
so maybe we should revert the changes I made in there which involves those `MK_PYTHON...` variables..so I just created pr for that in here:

https://github.com/kivy/python-for-android/pull/1619

I close this ;)

PR #1600, opened 2019-01-19
Migrates gevent to new python3 recipe

AndreMiras:
Also adds unit tests for checking flags are being cleaned up properly.

PR #1618, opened 2019-01-28
Create a `dumb` librt recipe and refactor the affected recipes to make use of this

opacam:
Because some recipes inserts flags without our control (`-lrt`) and this is a problem because librt does not exist in android, but has been integrated into libc, so, this recipe create a temporary symbolic link from libc to librt so the compilation can find the necessary `symbols/functions` to finish the compilation

Note: Once we end the build we will remove the created symbolic link

¡¡¡Thanks @AndreMiras, for this idea of "dumb" librt recipe!!!

### DIFF
--- a/pythonforandroid/recipes/gevent/__init__.py
+++ b/pythonforandroid/recipes/gevent/__init__.py
@@ -1,30 +1,13 @@
 import re
-import os
-import sh
-from pythonforandroid.logger import info, shprint
+from pythonforandroid.logger import info
 from pythonforandroid.recipe import CythonRecipe
 
 
 class GeventRecipe(CythonRecipe):
     version = '1.3.7'
     url = 'https://pypi.python.org/packages/source/g/gevent/gevent-{version}.tar.gz'
-    depends = ['greenlet']
+    depends = ['librt', 'greenlet']
     patches = ["cross_compiling.patch"]
-
-    def build_cython_components(self, arch):
-        """
-        Hack to make it link properly to librt, inserted automatically by the
-        installer (Note: the librt doesn't exist in android but it is
-        integrated into libc, so we create a symbolic link which we will
-        remove when our build finishes)
-        """
-        link_c = os.path.join(self.ctx.ndk_platform, 'usr', 'lib', 'libc')
-        link_rt = os.path.join(self.ctx.ndk_platform, 'usr', 'lib', 'librt')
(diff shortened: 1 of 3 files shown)


opacam:
Aaaa...I see:

https://github.com/kivy/python-for-android/pull/1600#issuecomment-457958760

It's the same error

JonasT:
@opacam @AndreMiras this appears to break build for me: #1621

JonasT:
Oops, seems like this was broken before. I just never tried `lxml` in that environment since it was broken for a while and previously I used another way to build things :woman_shrugging: but still, something that probably should be fixed :grimacing:

opacam:
@JonasT, I have a solution for docker, still not created a pr, but I will, don't worry ;)

opacam:
@JonasT, If you are in hurry with this I will do it now...do you need that?

JonasT:
@opacam take your time it's not urgent. `lxml` is an optional component for my application so I cna just disable it for now. but yeah, would be pretty nice to see a solution - I was already pondering some LD_LIBRARY_PATH hacks to get this to work just out of curiosity, but I don't mind if you are going to care of it instead :smile: :+1:

opacam:
Here it is the commit that will solve the docker issue with `lxml` and `greenlet`:

https://github.com/opacam/python-for-android/commit/ad9c35aaa55f776924f7dc2993e86511e8f80c9d

I will create the pr as soon as I have the explanations (I only move code in there, from one place to another, except for the variable `ANDROID_HOME`)

JonasT:
@opacam did you just make the NDK folder writable? Because I would not consider that a fix to this, I think the problem is that it is assumed to be writable in the first place, so the fault is IMHO definitely with the `librt` recipe and not the environment

opacam:
No, I change the download directory into a user directory

JonasT:
Which has the same outcome really. You are changing it from a system-wide to a user install to make it writable - this shouldn't be required, and isn't required for any other recipe. In my eyes that's a workaround hack, not a fix :man_shrugging: not that I don't think that's not possibly useful for other reasons, but I don't think this should be required to be able to build `librt`

AndreMiras:
yes I do would prefer that we make it work just like other recipes. So adding build artifact to a dedicated recipe build folder and then link it via `-Ldir`, see https://github.com/kivy/python-for-android/issues/1621#issuecomment-458480369

PR #1619, opened 2019-01-28
Make python flags to be absolute paths for Android.mk files

opacam:
Because it make the reading easy, more maintainable and also will allow us to remove the recently introduced variables: MK_PYTHON_LINK_ROOT and PYTHON_INCLUDE_ROOT
Which  only adds unneeded complexity...so...we also take care of that in here.

¡¡¡Thanks @inclement!!!


inclement:
Thanks!

PR #1623, opened 2019-01-28
Fix librt recipe requires that NDK folder is writable

### LINKED ISSUE #1621: Build crashes if NDK is installed system-wide without write permissions
### Versions

shouldn't be relevant

### Description

There is an issue (I thought it was a regression introduced by #1618 at first, but I think this was broken before) where in an environment where the NDK is installed system-wide, the build will crash because then files cannot simply be changed:

```
[INFO]:    Building lxml for armeabi-v7a
[INFO]:    lxml apparently isn't already in site-packages
[DEBUG]:   -> running ln -sf /ndk/platforms/android-21/arch-arm/usr/lib/libc.so /ndk/platforms/android-21/arch-arm/usr/lib/librt.so
[DEBUG]:   	/bin/ln: failed to create symbolic link '/ndk/platforms/android-21/arch-arm/usr/lib/librt.so': Permission denied
Exception in thread background thread for pid 18688:
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/home/userhome/.local/lib/python3.6/site-packages/sh.py", line 1540, in wrap
    fn(*args, **kwargs)
  File "/home/userhome/.local/lib/python3.6/site-packages/sh.py", line 2459, in background_thread
    handle_exit_code(exit_code)
  File "/home/userhome/.local/lib/python3.6/site-packages/sh.py", line 2157, in fn
  File "/home/userhome/.local/lib/python3.6/site-packages/sh.py", line 815, in handle_command_exit_code
    raise exc
sh.ErrorReturnCode_1: 

  RAN: /bin/ln -sf /ndk/platforms/android-21/arch-arm/usr/lib/libc.so /ndk/platforms/android-21/arch-arm/usr/lib/librt.so

  STDOUT:
/bin/ln: failed to create symbolic link '/ndk/platforms/android-21/arch-arm/usr/lib/librt.so': Permission denied


  STDERR:


  File "/home/userhome/.local/bin/p4a", line 11, in <module>
    sys.exit(main())
  File "/home/userhome/.local/lib/python3.6/site-packages/pythonforandroid/toolchain.py", line 1033, in main
    ToolchainCL()
  File "/home/userhome/.local/lib/python3.6/site-packages/pythonforandroid/toolchain.py", line 553, in __init__
  File "/home/userhome/.local/lib/python3.6/site-packages/pythonforandroid/toolchain.py", line 150, in wrapper_func
  File "/home/userhome/.local/lib/python3.6/site-packages/pythonforandroid/toolchain.py", line 191, in build_dist_from_args
  File "/home/userhome/.local/lib/python3.6/site-packages/pythonforandroid/build.py", line 634, in build_recipes
    recipe.build_arch(arch)
  File "/home/userhome/.local/lib/python3.6/site-packages/pythonforandroid/recipe.py", line 902, in build_arch
    self.build_compiled_components(arch)
  File "/home/userhome/.local/lib/python3.6/site-packages/pythonforandroid/recipes/lxml/__init__.py", line 35, in build_compiled_components
    shprint(sh.ln, '-sf', link_c + '.so', link_rt + '.so')
  File "/home/userhome/.local/lib/python3.6/site-packages/pythonforandroid/logger.py", line 178, in shprint
    for line in output:
  File "/home/userhome/.local/lib/python3.6/site-packages/sh.py", line 863, in next
    self.wait()
  File "/home/userhome/.local/lib/python3.6/site-packages/sh.py", line 792, in wait
    self.handle_command_exit_code(exit_code)
  File "/home/userhome/.local/lib/python3.6/site-packages/sh.py", line 815, in handle_command_exit_code
    raise exc
sh.ErrorReturnCode_1: 

  RAN: /bin/ln -sf /ndk/platforms/android-21/arch-arm/usr/lib/libc.so /ndk/platforms/android-21/arch-arm/usr/lib/librt.so

  STDOUT:
/bin/ln: failed to create symbolic link '/ndk/platforms/android-21/arch-arm/usr/lib/librt.so': Permission denied
```

Now that I am pondering this, this also seems somewhat ugly to me (not the symlink in itself, but actually writing around inside the NDK folder - I don't think this is to be expected from the user). Therefore I think a better solution should be found that doesn't require write permissions in the NDK folder

JonasT:
This fixes that the librt recipe requires the NDK folder to be writable. The NDK folder can be installed system-wide and may not be writable, so without this fix the librt build crashes in such an environment. This fixes #1621 

Fixes the build for me! Feel free to test & merge! (or to tell me what to change)

### DIFF
--- a/pythonforandroid/archs.py
+++ b/pythonforandroid/archs.py
@@ -1,8 +1,9 @@
-from os.path import (exists, join, dirname, split)
-from os import environ, uname
-from glob import glob
-import sys
 from distutils.spawn import find_executable
+from os import environ, uname
+from os.path import (exists, join, dirname, split)
+from glob import glob
+import shlex
+import sys
 
 from pythonforandroid.recipe import Recipe
 from pythonforandroid.util import BuildInterruptingException, build_platform
@@ -19,6 +20,12 @@ class Arch(object):
     def __init__(self, ctx):
         super(Arch, self).__init__()
         self.ctx = ctx
+
+        # Allows injecting additional linker paths used by any recipe.
+        # This can also be modified by recipes (like the librt recipe)
+        # to make sure that some sort of global resource is available &
+        # linked for all others.
+        self.extra_global_link_paths = []
 
     def __str__(self):
         return self.arch
@@ -56,7 +63,13 @@ class Arch(object):
             cflags.append('-gcc-toolchain {}'.format(toolchain))
 
         env['CFLAGS'] = ' '.join(cflags)
-        env['LDFLAGS'] = ' '
+
+        # Link the extra global link paths first before anything else
+        # (such that overriding system libraries with them is possible)
+        env['LDFLAGS'] = ' ' + " ".join([
+            "-L" + shlex.quote(l)
+            for l in self.extra_global_link_paths
+        ]) + ' '
 
         sysroot = join(self.ctx._ndk_dir, 'sysroot')
         if exists(sysroot):
--- a/pythonforandroid/recipes/librt/__init__.py
+++ b/pythonforandroid/recipes/librt/__init__.py
@@ -1,5 +1,7 @@
+from os import makedirs, remove
+from os.path import exists, join
 import sh
-from os.path import join
+
 from pythonforandroid.recipe import Recipe
 from pythonforandroid.logger import shprint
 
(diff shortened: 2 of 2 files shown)


AndreMiras:
Why handling it like an usual recipe, e.g. it has it's own build dir (not temporary) and we leave the "build" artifact and just link to it?

JonasT:
@AndreMiras ah lol yeah that would make a lot more sense. Is `ctx.build_dir` the private/recipe-specific one? I'll have a look and change it accordingly

inclement:
I agree with the general value dislike of adding this type of thing, but it seems pretty reasonable and I've thought about similar options before. Ultimately, we probably do want a generic way for recipes to say 'other recipes may want to include/link with me'.

pythonforandroid/archs.py
```diff
@@ -56,7 +62,13 @@ def get_env(self, with_flags_in_cc=True, clang=False):
             cflags.append('-gcc-toolchain {}'.format(toolchain))
 
         env['CFLAGS'] = ' '.join(cflags)
-        env['LDFLAGS'] = ' '
+
+        # Link the extra global link paths first before anything else
+        # (such that overriding system libraries with them is possible)
+        env['LDFLAGS'] = ' ' + " ".join([
+            "-L'" + l.replace("'", "'\"'\"'") + "'"
```
inclement: I'm not completely clear, what is this fixing?

JonasT: @inclement this is just adding these new global link options as proposed, and adding them first so they take precedence. (actually now that you say it I'm not sure the `-L` order matters so maybe the comment is nonsense and it could have been added at another point in that function, but then again I still don't see a reason to move it somewhere else) The replace is there because it should be, it does proper quoting in all cases for paths including space, single quotes, ...

inclement: I think the -L order does matter, just wasn't clear what the effect of the escaping is. I'm happy to believe it does the right thing, but maybe it could go in a different `quote_path` function to be clear?

JonasT: @inclement this escaping is the ultimate(TM) shell escaping which should be used *everywhere* - if it was, then we'd have no issue with spaces in folders and all these other things.

It's pretty simple as well, if you think about it: single quotes basically disable everything that could do anything (`$`, space chars, ...) - except for terminating single quotes. So the single quotes themselves are escaped by `'"'"'`, spaced out this is: `'  "  '  "  '`. Explanation: it terminates the current single quote block, then adds a single quote in double escapes (to avoid starting a new block), then starts a new single quote block directly attached.

I got this escaping method from `shlex.quote` actually, I believe?! I forgot. Maybe we could actually use that one, I tend to forget it exists. Anyway, my point was I didn't make this up myself but smarter people did, so the escaping method should be solid :smile:

inclement: Sure, I understand what it's doing and I totally believe it's correct, it's just hard to judge it without knowing the rules of quoting offhand. It sounds like using shlex.quote would be the best option.

JonasT: @inclement you're right, it should just be `shlex.quote` :+1:

JonasT:
build ran through fine for me with the `shlex.quote` version!

inclement:
Build failed because shlex.quote isn't present yet in Python 2, maybe that's why it wasn't used in the first place?

That was the only issue though, so merging :)

JonasT:
Wait isn't it kinda bad if it fails under python 2? Should I make another pull request to revert it to the explicit quote `replace` again?

inclement:
For now it only means the boost recipe fails under python2, because otherwise there's nothing in extra_global_link_paths. Also, I'm getting tired of making things harder to support python2 :p

If you want to change it back to the explicit replace, that would be okay though, just maybe comment that it's only there because shlex.quote isn't available in py2.

JonasT:
@inclement it also affects reportlab and greenlets I believe, so the fallout isn't *that* tiny. I'll make a pull request to change it back with a comment

PR #1624, opened 2019-01-29
Fix ctypes.util.find_library() not finding any libraries on Android

JonasT:
This is finally my long-delayed followup to #1517 to fix `ctypes.util.find_library()` for Python 3. It should fix the last somewhat showstopper bug to using Python 3 & ctypes out of the box on p4a master :heart: 

done & tested (with real world python3 app that uses ctypes)

### DIFF
--- a/pythonforandroid/recipes/python3/__init__.py
+++ b/pythonforandroid/recipes/python3/__init__.py
@@ -21,6 +21,8 @@ class Python3Recipe(GuestPythonRecipe):
     url = 'https://www.python.org/ftp/python/{version}/Python-{version}.tgz'
     name = 'python3'
 
+    patches = ["patches/fix-ctypes-util-find-library.patch"]
+
     depends = ['hostpython3']
     conflicts = ['python3crystax', 'python2', 'python2legacy']
 
--- a/pythonforandroid/recipes/python3/patches/fix-ctypes-util-find-library.patch
+++ b/pythonforandroid/recipes/python3/patches/fix-ctypes-util-find-library.patch
@@ -1,0 +1,23 @@
+diff --git a/Lib/ctypes/util.py b/Lib/ctypes/util.py
+--- a/Lib/ctypes/util.py
++++ b/Lib/ctypes/util.py
+@@ -67,4 +67,19 @@
+                 return fname
+         return None
+ 
++# This patch overrides the find_library to look in the right places on
++# Android
++if True:
++    def find_library(name):
++        # Check the user app libs and system libraries directory:
++        app_root = os.path.normpath(os.path.abspath('../../'))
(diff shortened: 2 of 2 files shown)


inclement:
Just wondering, does ctypes respect $LD_LIBRARY_PATH on Android? I've used this successfully for other purposes but not sure if python's logic will use it correctly.

If not it isn't a problem to add a patch.

JonasT:
@inclement very interesting idea. So your suggestion is to try to set LD_LIBRARY_PATH on startup e.g. with some one-liner in `start.c` or similar?

inclement:
@JonasT Exactly. For ctypes that should work the same on any Android version, as long as ctypes checks LD_LIBRARY_PATH. I checked the source and it looks like it does on the desktop, so that should work unless it gets messed up by not recognising what platform it's running on.

I guess this could solve some problems for other library loading as well, I didn't think about that before. Maybe it could obselete the explicit library loading in PythonUtil.java, but not sure.

JonasT:
@inclement neat idea! I am marking this WIP for now until I investigated the LD_LIBRARY_PATH alternative

JonasT:
@inclement okay, so unless I screwed it up totally, relative paths don't work in `LD_LIBRARY_PATH`. (the environment was set for sure, I see it printed at runtime in adb/logcat) Now while absolute paths are possible, this would need to be combined in `start.c` and would be a little ugly. I'm actually not sure I would prefer that over this relatively separate & clean patch, since `start.c` is one of the few files I would actually want to not add more functionality to - C is simply much more annoying to maintain in the long run.

What are your thoughts on this? Should I also give the absolute path a try, or are we going to go with this patch similar to how the Python 2 recipe already does it?

inclement:
I would be interested to know if the absolute path would work. I suspect not, I'm pretty sure nothing that reads LD_LIBRARY_PATH would really care about it.

If you can check it then fine, we can merge this patch, I'll look into making python get the paths in a better way some other time.

JonasT:
@inclement ok hang on, I'll hack it into my own `main.py` startup code before the imports and check what happens. That should have the same effects, right? Should work well enough to see if it works with absolute paths without me bothering with a proper `start.c` patch

inclement:
That's fine, I'm really wondering just in case it would work at all for future reference.

inclement:
Thanks!

JonasT:
@inclement ok, I tested this & printed out the absolute path & resulting env & even added `os.listdir()` output to be very sure I specified the correct folder and `LD_LIBRARY_PATH` was actually set, and it **doesn't work**. So no dice with `LD_LIBRARY_PATH` even with absolute path, it seems (unless I made some other fundamental mistake I'm not aware of :smile: but I can't think of anything)

PR #1625, opened 2019-01-29
[WIP] Run project's setup.py if present, unless --ignore-setup-py was set

JonasT:
Run project's setup.py if present, unless --ignore-setup-py was set.

**What exactly does this do:**

Changes to `p4a`'s behavior:

1. If a `setup.py` is present in the `--private` code directory and `--use-setup-py` is set,
    all dependencies that map to recipes will be implicitly added to `--requirements`
2. After all modules in `--requirements` have been built, the `setup.py` is run (again only if
   `--use-setup-py` was specified)

This has the following consequences:

1. You can now use a `setup.py` and `--use-setup-py`, and either omit `--requirements`
    (If you use Python 3) or otherwise just specify `--requirements=python2
2. For projects without a `setup.py` file, *nothing changes*
3. For projects that for some reason have a `setup.py` but don't want it to run for p4a, there will be a big fat warning printed that in the future, `--ignore-setup-py` must be specified (right now the default is still not to run the `setup.py` unless an explicit `--use-setup-py` was given)

**Why this pull request is more than just a neat feature addition:**

This functionality solves fundamental issues in python-for-android::

1. Currently, projects that need to be installed to work are a huge pain to get working. (They need to be copied to another place and then that folder needs to be added as an additional `--requirements` entry) This will no longer be needed
2. `--requirements` is kind of annoying and violating DRY for projects that already use `setup.py` for desktop
3. Currently, Cython projects that do cross-package `.pxd` imports are **impossible to install** without notable modifications to the packages themselves (e.g. manually copying the `.pxd` files around from one package into the other and set them up to be importable, which may not be much easier than fully merging the packages into one). This is because p4a assumes the install order of all the non-recipe dependencies doesn't matter, which is wrong for this case.

    With this pull request, this won't be *fully* fixed: however, if the affected Cython libraries are regular enough or p4a-aware enough that they work without a special CythonRecipe (since basic Cython now works in p4a master without any recipe), then the project's `setup.py` can install them and proper install order will be preserved - so for these libraries, it will be fixed.

### DIFF
--- a/Dockerfile.py3
+++ b/Dockerfile.py3
@@ -95,7 +95,8 @@ ENV WORK_DIR="${HOME_DIR}" \
 
 # install system dependencies
 RUN ${RETRY} apt -y install -qq --no-install-recommends \
-        python3 virtualenv python3-pip wget lbzip2 patch sudo \
+        python3 virtualenv python3-pip python3-venv \
+        wget lbzip2 patch sudo \
     && apt -y autoremove
 
 # build dependencies
--- a/doc/source/quickstart.rst
+++ b/doc/source/quickstart.rst
@@ -298,6 +298,14 @@ your own Kivy branch you might set::
 The specified directory will be copied into python-for-android instead
 of downloading from the normal url specified in the recipe.
 
+setup.py file (experimental)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If your application is also packaged for desktop using `setup.py`,
+you may want to use your `setup.py` instead of the
+``--requirements`` option to avoid specifying things twice.
+For that purpose, check out :doc:`distutils`
+
 Going further
 ~~~~~~~~~~~~~
 
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -9,6 +9,7 @@ import glob
 import sys
 import re
 import sh
+import shutil
 import subprocess
 
 from pythonforandroid.util import (
@@ -17,6 +18,7 @@ from pythonforandroid.util import (
 )
 from pythonforandroid.logger import (info, warning, info_notify, info_main, shprint)
 from pythonforandroid.archs import ArchARM, ArchARMv7_a, ArchAarch_64, Archx86, Archx86_64
+from pythonforandroid.pythonpackage import get_package_name
 from pythonforandroid.recipe import CythonRecipe, Recipe
 from pythonforandroid.recommendations import (
     check_ndk_version, check_target_api, check_ndk_api,
@@ -42,6 +44,9 @@ class Context(object):
     # but before being placed in dists
     libs_dir = None
     aars_dir = None
+
+    # Whether setup.py or similar should be used if present:
+    use_setup_py = False
 
     ccache = None  # whether to use ccache
     cython = None  # the cython interpreter name
@@ -486,21 +491,10 @@ class Context(object):
             if not os.path.exists(name):
                 # Non-existing dir, cannot look this up.
                 return False
-            if os.path.exists(os.path.join(name, "setup.py")):
-                # Get name from setup.py:
-                name = subprocess.check_output([
-                    sys.executable, "setup.py", "--name"],
-                    cwd=name)
-                try:
-                    name = name.decode('utf-8', 'replace')
-                except AttributeError:
-                    pass
-                name = name.strip()
-                if len(name) == 0:
-                    # Failed to look up any meaningful name.
-                    return False
-            else:
-                # A folder with whatever, cannot look this up.
+            try:
+                name = get_package_name(os.path.abspath(name))
+            except ValueError:
+                # Failed to look up any meaningful name.
                 return False
 
         # Try to look up recipe by name:
@@ -523,7 +517,9 @@ class Context(object):
         return not self.has_package(name, arch)
 
 
-def build_recipes(build_order, python_modules, ctx):
+def build_recipes(build_order, python_modules, ctx, project_dir,
+                  ignore_project_setup_py=False
+                 ):
     # Put recipes in correct build order
     info_notify("Recipe build order is {}".format(build_order))
     if python_modules:
@@ -578,22 +574,56 @@ def build_recipes(build_order, python_modules, ctx):
             recipe.postbuild_arch(arch)
 
     info_main('# Installing pure Python modules')
-    run_pymodules_install(ctx, python_modules)
+    run_pymodules_install(
+        ctx, python_modules, project_dir,
+        ignore_setup_py=ignore_project_setup_py
+    )
 
     return
 
 
-def run_pymodules_install(ctx, modules):
+def project_has_setup_py(project_dir):
+    if project_dir is not None and \
+            (os.path.exists(os.path.join(project_dir,
+                            "setup.py")) or
+             os.path.exists(os.path.join(project_dir,
+                            "pyproject.toml"))
+            ):
+        return True
+    return False
+
+
+def run_pymodules_install(ctx, modules, project_dir, ignore_setup_py=False):
+    """ This function will take care of all non-recipe things, by:
+
+        1. Processing them from --requirements (the modules argument)
+           and installing them
+
+        2. Installing the user project/app itself via setup.py if
+           ignore_setup_py=True
+
+    """
+
+    info('*** PYTHON PACKAGE / PROJECT INSTALL STAGE ***')
     modules = list(filter(ctx.not_has_package, modules))
 
-    if not modules:
-        info('There are no Python modules to install, skipping')
+    # Bail out if no python deps and no setup.py to process:
+    if not modules and (
+            ignore_setup_py or
+            not project_has_setup_py(project_dir)
+            ):
+        info('No Python modules and no setup.py to process, skipping')
         return
 
-    info('The requirements ({}) don\'t have recipes, attempting to install '
-         'them with pip'.format(', '.join(modules)))
-    info('If this fails, it may mean that the module has compiled '
-         'components and needs a recipe.')
+    # Output messages about what we're going to do:
+    if modules:
+        info('The requirements ({}) don\'t have recipes, attempting to '
+             'install them with pip'.format(', '.join(modules)))
+        info('If this fails, it may mean that the module has compiled '
+             'components and needs a recipe.')
+    if project_has_setup_py(project_dir) and not ignore_setup_py:
+        info('Will process project install, if it fails then the '
+             'project may not be compatible for Android install.')
 
     venv = sh.Command(ctx.virtualenv)
     with current_directory(join(ctx.build_dir)):
@@ -604,16 +634,6 @@ def run_pymodules_install(ctx, modules):
                     ),
                 'venv'
                )
-
-        info('Creating a requirements.txt file for the Python modules')
-        with open('requirements.txt', 'w') as fileh:
-            for module in modules:
-                key = 'VERSION_' + module
-                if key in environ:
-                    line = '{}=={}\n'.format(module, environ[key])
-                else:
-                    line = '{}\n'.format(module)
-                fileh.write(line)
 
         # Prepare base environment and upgrade pip:
         base_env = copy.copy(os.environ)
@@ -638,13 +658,6 @@ def run_pymodules_install(ctx, modules):
         env = copy.copy(base_env)
         env.update(recipe_env)
 
-        info('Installing Python modules with pip')
-        info('IF THIS FAILS, THE MODULES MAY NEED A RECIPE. '
-             'A reason for this is often modules compiling '
-             'native code that is unaware of Android cross-compilation '
-             'and does not work without additional '
-             'changes / workarounds.')
-
         # Make sure our build package dir is available, and the virtualenv
         # site packages come FIRST (so the proper pip version is used):
         env["PYTHONPATH"] += ":" + ctx.get_site_packages_dir()
@@ -653,12 +666,140 @@ def run_pymodules_install(ctx, modules):
             "python" + ctx.python_recipe.major_minor_version_string,
             "site-packages")) + ":" + env["PYTHONPATH"]
 
-        # Do actual install:
-        shprint(sh.bash, '-c', (
-            "venv/bin/pip " +
-            "install -v --target '{0}' --no-deps -r requirements.txt"
-        ).format(ctx.get_site_packages_dir().replace("'", "'\"'\"'")),
-                _env=copy.copy(env))
+        # Install the manually specified requirements first:
+        if not modules:
+            info('There are no Python modules to install, skipping')
+        else:
+            info('Creating a requirements.txt file for the Python modules')
+            with open('requirements.txt', 'w') as fileh:
+                for module in modules:
+                    key = 'VERSION_' + module
+                    if key in environ:
+                        line = '{}=={}\n'.format(module, environ[key])
+                    else:
+                        line = '{}\n'.format(module)
+                    fileh.write(line)
+
+            info('Installing Python modules with pip')
+            info('IF THIS FAILS, THE MODULES MAY NEED A RECIPE. '
+                 'A reason for this is often modules compiling '
+                 'native code that is unaware of Android cross-compilation '
+                 'and does not work without additional '
+                 'changes / workarounds.')
+
+            shprint(sh.bash, '-c', (
+                "venv/bin/pip " +
+                "install -v --target '{0}' --no-deps -r requirements.txt"
+            ).format(ctx.get_site_packages_dir().replace("'", "'\"'\"'")),
+                    _env=copy.copy(env))
+
+        # Afterwards, run setup.py if present:
+        if project_has_setup_py(project_dir) and not ignore_setup_py:
+            with current_directory(project_dir):
+                info('got setup.py or similar, running project install. ' +
+                     '(disable this behavior with --ignore-setup-py)')
+
+                # Compute & output the constraints we will use:
+                info('Contents that will be used for constraints.txt:')
+                constraints = subprocess.check_output([
+                    join(
+                        ctx.build_dir, "venv", "bin", "pip"
+                    ),
+                    "freeze"
+                ], env=copy.copy(env))
+                try:
+                    constraints = constraints.decode("utf-8", "replace")
+                except AttributeError:
+                    pass
+                info(constraints)
+
+                # Make sure all packages found are fixed in version
+                # by writing a constraint file, to avoid recipes being
+                # upgraded & reinstalled:
+                with open('constraints.txt', 'wb') as fileh:
+                    fileh.write(constraints.encode("utf-8", "replace"))
+
+                info('Populating venv\'s site-packages with '
+                     'ctx.get_site_packages_dir()...')
+
+                # Copy dist contents into site-packages for discovery.
+                # Why this is needed:
+                # --target is somewhat evil and messes with discovery of
+                # packages in PYTHONPATH if that also includes the target
+                # folder. So we need to use the regular virtualenv
+                # site-packages folder instead.
+                # Reference:
+                # https://github.com/pypa/pip/issues/6223
+                ctx_site_packages_dir = os.path.normpath(
+                    os.path.abspath(ctx.get_site_packages_dir())
+                )
+                venv_site_packages_dir = os.path.normpath(os.path.join(
+                    ctx.build_dir, "venv", "lib", [
+                        f for f in os.listdir(os.path.join(
+                            ctx.build_dir, "venv", "lib"
+                        )) if f.startswith("python")
+                    ][0], "site-packages"
+                ))
+                copied_over_contents = []
+                for f in os.listdir(ctx_site_packages_dir):
+                    full_path = os.path.join(ctx_site_packages_dir, f)
+                    if not os.path.exists(os.path.join(
+                                venv_site_packages_dir, f
+                            )):
+                        if os.path.isdir(full_path):
+                            shutil.copytree(full_path, os.path.join(
+                                venv_site_packages_dir, f
+                            ))
+                        else:
+                            shutil.copy2(full_path, os.path.join(
+                                venv_site_packages_dir, f
+                            ))
+                        copied_over_contents.append(f)
+
+                # Get listing of virtualenv's site-packages, to see the
+                # newly added things afterwards & copy them back into
+                # the distribution folder / build context site-packages:
+                previous_venv_contents = os.listdir(venv_site_packages_dir)
+
+                # Actually run setup.py:
+                info('Launching package install...')
+                shprint(sh.bash, '-c', (
+                    "'" + join(
+                        ctx.build_dir, "venv", "bin", "pip"
+                    ).replace("'", "'\"'\"'") + "' " +
+                    "install -c constraints.txt -v ."
+                ).format(ctx.get_site_packages_dir().replace("'", "'\"'\"'")),
+                        _env=copy.copy(env))
+
+                # Go over all new additions and copy them back:
+                info('Copying additions resulting from setup.py back ' +
+                     'into ctx.get_site_packages_dir()...')
+                new_venv_additions = []
+                for f in (set(os.listdir(venv_site_packages_dir)) -
+                          set(previous_venv_contents)):
+                    new_venv_additions.append(f)
+                    full_path = os.path.join(venv_site_packages_dir, f)
+                    if os.path.isdir(full_path):
+                        shutil.copytree(full_path, os.path.join(
+                            ctx_site_packages_dir, f
+                        ))
+                    else:
+                        shutil.copy2(full_path, os.path.join(
+                            ctx_site_packages_dir, f
+                        ))
+
+                # Undo all the changes we did to the venv-site packages:
+                info('Reverting additions to virtualenv\'s site-packages...')
+                for f in set(copied_over_contents + new_venv_additions):
+                    full_path = os.path.join(venv_site_packages_dir, f)
+                    if os.path.isdir(full_path):
+                        shutil.rmtree(full_path)
+                    else:
+                        os.remove(full_path)
+        elif not ignore_setup_py:
+            info("No setup.py found in project directory: " +
+                 str(project_dir)
+                )
 
         # Strip object files after potential Cython or native code builds:
         standard_recipe.strip_object_files(ctx.archs[0], env,
--- a/pythonforandroid/pythonpackage.py
+++ b/pythonforandroid/pythonpackage.py
@@ -1,0 +1,693 @@
+""" This module offers highlevel functions to get package metadata
+    like the METADATA file, the name, or a list of dependencies.
+
+    Usage examples:
+
+       # Getting package name from pip reference:
+       from pytonforandroid.pythonpackage import get_package_name
+       print(get_package_name("pillow"))
+       # Outputs: "Pillow" (note the spelling!)
+
+       # Getting package dependencies:
+       from pytonforandroid.pythonpackage import get_package_dependencies
+       print(get_package_dependencies("pep517"))
+       # Outputs: "['pytoml']"
+
+       # Get package name from arbitrary package source:
+       from pytonforandroid.pythonpackage import get_package_name
+       print(get_package_name("/some/local/project/folder/"))
+       # Outputs package name
+
+    NOTE:
+
+    Yes, this module doesn't fit well into python-for-android, but this
+    functionality isn't available ANYWHERE ELSE, and upstream (pip, ...)
+    currently has no interest in taking this over, so it has no other place
+    to go.
+    (Unless someone reading this puts it into yet another packaging lib)
+
+    Reference discussion/upstream inclusion attempt:
+
+    https://github.com/pypa/packaging-problems/issues/247
+
+"""
+
+
+from io import open  # needed for python 2
+import os
+from pep517.envbuild import BuildEnvironment
+from pep517.wrappers import Pep517HookCaller
+import pytoml
+import shutil
+import subprocess
+import sys
+import tarfile
+import tempfile
+import textwrap
+import time
+try:
+    from urllib.parse import urlparse
+    from urllib.parse import unquote as urlunquote
+except ImportError:  # Python 2...
+    from urlparse import urlparse
+    from urlparse import unquote as urlunquote
+import zipfile
+
+
+def transform_dep_for_pip(dependency):
+    if dependency.find("@") > 0:
+        # WORKAROUND FOR UPSTREAM BUG:
+        # https://github.com/pypa/pip/issues/6097
+        # (Please REMOVE workaround once that is fixed & released upstream!)
+        #
+        # Basically, setup_requires() can contain a format pip won't install
+        # from a requirements.txt (PEP 508 URLs).
+        # To avoid this, translate to an #egg-name= reference:
+        url = (dependency.partition("@")[2].strip() +
+               "#egg-name=" +
+               dependency.partition("@")[0].strip()
+              )
+        return url
+    return dependency
+
+
+def extract_metainfo_files_from_package(
+        package,
+        output_folder,
+        debug=False
+        ):
+    """ Extracts metdata files from the given package to the given folder,
+        which may be referenced in any way that is permitted in
+        a requirements.txt file or install_requires=[] listing.
+
+        Current supported metadata files that will be extracted:
+
+        - pytoml.yml  (only if package wasn't obtained as wheel)
+        - METADATA
+    """
+
+    if package is None:
+        raise ValueError("package cannot be None")
+
+    if not os.path.exists(output_folder) or os.path.isfile(output_folder):
+        raise ValueError("output folder needs to be existing folder")
+
+    # A temp folder for making a package copy in case it's a local folder,
+    # because extracting metadata might modify files
+    # (creating sdists/wheels...)
+    temp_folder = tempfile.mkdtemp(prefix="pythonpackage-package-copy-")
+    try:
+        # Package is indeed a folder! Get a temp copy to work on:
+        if is_filesystem_path(package):
+            shutil.copytree(
+                parse_as_folder_reference(package),
+                os.path.join(temp_folder, "package")
+            )
+            package = os.path.join(temp_folder, "package")
+
+        # Because PEP517 can be noisy and contextlib.redirect_* fails to
+        # contain it, we will run the actual analysis in a separate process:
+        try:
+            subprocess.check_output([
+                sys.executable,
+                "-c",
+                "import importlib\n"
+                "import json\n"
+                "import os\n"
+                "import sys\n"
+                "sys.path = [os.path.dirname(sys.argv[3])] + sys.path\n"
+                "m = importlib.import_module(\n"
+                "    os.path.basename(sys.argv[3]).partition('.')[0]\n"
+                ")\n"
+                "m._extract_metainfo_files_from_package_unsafe("
+                "    sys.argv[1],"
+                "    sys.argv[2],"
+                ")",
+                package, output_folder, os.path.abspath(__file__)],
+                stderr=subprocess.STDOUT,  # make sure stderr is muted.
+                cwd=os.path.join(os.path.dirname(__file__), "..")
+            )
+        except subprocess.CalledProcessError as e:
+            output = e.output.decode("utf-8", "replace")
+            if debug:
+                print("Got error obtaining meta info.")
+                print("Detail output:")
+                print(output)
+                print("End of Detail output.")
+            raise ValueError(
+                "failed to obtain meta info - "
+                "is '{}' a valid package? "
+                "Detailed output:\n{}".format(package, output)
+                )
+    finally:
+        shutil.rmtree(temp_folder)
+
+
+def _get_system_python_executable():
+    """ Returns the path the system-wide python binary.
+        (In case we're running in a virtualenv or venv)
+    """
+    # This function is required by get_package_as_folder() to work
+    # inside a virtualenv, since venv creation will fail with
+    # the virtualenv's local python binary.
+    # (venv/virtualenv incompatibility)
+
+    # Abort if not in virtualenv or venv:
+    if not hasattr(sys, "real_prefix") and (
+            not hasattr(sys, "base_prefix") or
+            os.path.normpath(sys.base_prefix) ==
+            os.path.normpath(sys.prefix)):
+        return sys.executable
+
+    # Extract prefix we need to look in:
+    if hasattr(sys, "real_prefix"):
+        search_prefix = sys.real_prefix  # virtualenv
+    else:
+        search_prefix = sys.base_prefix  # venv
+
+    def python_binary_from_folder(path):
+        def binary_is_usable(python_bin):
+            try:
+                filenotfounderror = FileNotFoundError
+            except NameError:  # Python 2
+                filenotfounderror = OSError
+            try:
+                subprocess.check_output([
+                    os.path.join(path, python_bin), "--version"
+                ], stderr=subprocess.STDOUT)
+                return True
+            except (subprocess.CalledProcessError, filenotfounderror):
+                return False
+
+        python_name = "python" + sys.version
+        while (not binary_is_usable(python_name) and
+               python_name.find(".") > 0):
+            # Try less specific binary name:
+            python_name = python_name.rpartition(".")[0]
+        if binary_is_usable(python_name):
+            return os.path.join(path, python_name)
+        return None
+
+    # Return from sys.real_prefix if present:
+    result = python_binary_from_folder(search_prefix)
+    if result is not None:
+        return result
+
+    # Check out all paths in $PATH:
+    bad_candidates = []
+    good_candidates = []
+    ever_had_nonvenv_path = False
+    for p in os.environ.get("PATH", "").split(":"):
+        # Skip if not possibly the real system python:
+        if not os.path.normpath(p).startswith(
+                os.path.normpath(search_prefix)
+                ):
+            continue
+
+        # First folders might be virtualenv/venv we want to avoid:
+        if not ever_had_nonvenv_path:
+            sep = os.path.sep
+            if ("system32" not in p.lower() and "usr" not in p) or \
+                    {"home", ".tox"}.intersection(set(p.split(sep))) or \
+                    "users" in p.lower():
+                # Doesn't look like bog-standard system path.
+                if (p.endswith(os.path.sep + "bin") or
+                        p.endswith(os.path.sep + "bin" + os.path.sep)):
+                    # Also ends in "bin" -> likely virtualenv/venv.
+                    # Add as unfavorable / end of candidates:
+                    bad_candidates.append(p)
+                    continue
+            ever_had_nonvenv_path = True
+
+        good_candidates.append(p)
+
+    # See if we can now actually find the system python:
+    for p in good_candidates + bad_candidates:
+        result = python_binary_from_folder(p)
+        if result is not None:
+            return result
+
+    raise RuntimeError("failed to locate system python in: " +
+                       sys.real_prefix)
+
+
+def get_package_as_folder(dependency):
+    """ This function downloads the given package / dependency and extracts
+        the raw contents into a folder.
+
+        Afterwards, it returns a tuple with the type of distribution obtained,
+        and the temporary folder it extracted to. It is the caller's
+        responsibility to delete the returned temp folder after use.
+
+        Examples of returned values:
+
+        ("source", "/tmp/pythonpackage-venv-e84toiwjw")
+        ("wheel", "/tmp/pythonpackage-venv-85u78uj")
+
+        What the distribution type will be depends on what pip decides to
+        download.
+    """
+
+    venv_parent = tempfile.mkdtemp(
+        prefix="pythonpackage-venv-"
+    )
+    try:
+        # Create a venv to install into:
+        try:
+            if int(sys.version.partition(".")[0]) < 3:
+                # Python 2.x has no venv.
+                subprocess.check_output([
+                    sys.executable,  # no venv conflict possible,
+                                     # -> no need to use system python
+                    "-m", "virtualenv",
+                    "--python=" + _get_system_python_executable(),
+                    os.path.join(venv_parent, 'venv')
+                ], cwd=venv_parent)
+            else:
+                # On modern Python 3, use venv.
+                subprocess.check_output([
+                    _get_system_python_executable(), "-m", "venv",
+                    os.path.join(venv_parent, 'venv')
+                ], cwd=venv_parent)
+        except subprocess.CalledProcessError as e:
+            output = e.output.decode('utf-8', 'replace')
+            raise ValueError(
+                'venv creation unexpectedly ' +
+                'failed. error output: ' + str(output)
+            )
+        venv_path = os.path.join(venv_parent, "venv")
+
+        # Update pip and wheel in venv for latest feature support:
+        try:
+            filenotfounderror = FileNotFoundError
+        except NameError:  # Python 2.
+            filenotfounderror = OSError
+        try:
+            subprocess.check_output([
+                os.path.join(venv_path, "bin", "pip"),
+                "install", "-U", "pip", "wheel",
+            ])
+        except filenotfounderror:
+            raise RuntimeError(
+                "venv appears to be missing pip. "
+                "did we fail to use a proper system python??\n"
+                "system python path detected: {}\n"
+                "os.environ['PATH']: {}".format(
+                    _get_system_python_executable(),
+                    os.environ.get("PATH", "")
+                )
+            )
+
+        # Create download subfolder:
+        os.mkdir(os.path.join(venv_path, "download"))
+
+        # Write a requirements.txt with our package and download:
+        with open(os.path.join(venv_path, "requirements.txt"),
+                  "w", encoding="utf-8"
+                 ) as f:
+            def to_unicode(s):  # Needed for Python 2.
+                try:
+                    return s.decode("utf-8")
+                except AttributeError:
+                    return s
+            f.write(to_unicode(transform_dep_for_pip(dependency)))
+        try:
+            subprocess.check_output(
+                [
+                    os.path.join(venv_path, "bin", "pip"),
+                    "download", "--no-deps", "-r", "../requirements.txt",
+                    "-d", os.path.join(venv_path, "download")
+                ],
+                stderr=subprocess.STDOUT,
+                cwd=os.path.join(venv_path, "download")
+            )
+        except subprocess.CalledProcessError as e:
+            raise RuntimeError("package download failed: " + str(e.output))
+
+        if len(os.listdir(os.path.join(venv_path, "download"))) == 0:
+            # No download. This can happen if the dependency has a condition
+            # which prohibits install in our environment.
+            # (the "package ; ... conditional ... " type of condition)
+            return (None, None)
+
+        # Get the result and make sure it's an extracted directory:
+        result_folder_or_file = os.path.join(
+            venv_path, "download",
+            os.listdir(os.path.join(venv_path, "download"))[0]
+        )
+        dl_type = "source"
+        if not os.path.isdir(result_folder_or_file):
+            # Must be an archive.
+            if result_folder_or_file.endswith((".zip", ".whl")):
+                if result_folder_or_file.endswith(".whl"):
+                    dl_type = "wheel"
+                with zipfile.ZipFile(result_folder_or_file) as f:
+                    f.extractall(os.path.join(venv_path,
+                                              "download", "extracted"
+                                             ))
+                    result_folder_or_file = os.path.join(
+                        venv_path, "download", "extracted"
+                    )
+            elif result_folder_or_file.find(".tar.") > 0:
+                # Probably a tarball.
+                with tarfile.open(result_folder_or_file) as f:
+                    f.extractall(os.path.join(venv_path,
+                                              "download", "extracted"
+                                             ))
+                    result_folder_or_file = os.path.join(
+                        venv_path, "download", "extracted"
+                    )
+            else:
+                raise RuntimeError(
+                    "unknown archive or download " +
+                    "type: " + str(result_folder_or_file)
+                )
+
+        # If the result is hidden away in an additional subfolder,
+        # descend into it:
+        while os.path.isdir(result_folder_or_file) and \
+                len(os.listdir(result_folder_or_file)) == 1 and \
+                os.path.isdir(os.path.join(
+                    result_folder_or_file,
+                    os.listdir(result_folder_or_file)[0]
+                )):
+            result_folder_or_file = os.path.join(
+                result_folder_or_file,
+                os.listdir(result_folder_or_file)[0]
+            )
+
+        # Copy result to new dedicated folder so we can throw away
+        # our entire virtualenv nonsense after returning:
+        result_path = tempfile.mkdtemp()
+        shutil.rmtree(result_path)
+        shutil.copytree(result_folder_or_file, result_path)
+        return (dl_type, result_path)
+    finally:
+        shutil.rmtree(venv_parent)
+
+
+def _extract_metainfo_files_from_package_unsafe(
+        package,
+        output_path
+        ):
+    # This is the unwrapped function that will
+    # 1. make lots of stdout/stderr noise
+    # 2. possibly modify files (if the package source is a local folder)
+    # Use extract_metainfo_files_from_package_folder instead which avoids
+    # these issues.
+
+    clean_up_path = False
+    path_type = "source"
+    path = parse_as_folder_reference(package)
+    if path is None:
+        # This is not a path. Download it:
+        (path_type, path) = get_package_as_folder(package)
+        if path_type is None:
+            # Download failed.
+            raise ValueError(
+                "cannot get info for this package, " +
+                "pip says it has no downloads (conditional dependency?)"
+            )
+        clean_up_path = True
+
+    try:
+        build_requires = []
+        metadata_path = None
+        if path_type != "wheel":
+            # We need to process this first to get the metadata.
+
+            # Ensure pyproject.toml is available (pep517 expects it)
+            if not os.path.exists(os.path.join(path, "pyproject.toml")):
+                with open(os.path.join(path, "pyproject.toml"), "w") as f:
+                    f.write(textwrap.dedent(u"""\
+                    [build-system]
+                    requires = ["setuptools", "wheel"]
+                    build-backend = "setuptools.build_meta"
+                    """))
+
+            # Copy the pyproject.toml:
+            shutil.copyfile(
+                os.path.join(path, 'pyproject.toml'),
+                os.path.join(output_path, 'pyproject.toml')
+            )
+
+            # Get build backend from pyproject.toml:
+            with open(os.path.join(path, 'pyproject.toml')) as f:
+                build_sys = pytoml.load(f)['build-system']
+                backend = build_sys["build-backend"]
+
+            # Get a virtualenv with build requirements and get all metadata:
+            env = BuildEnvironment()
+            metadata = None
+            with env:
+                hooks = Pep517HookCaller(path, backend)
+                env.pip_install([transform_dep_for_pip(req) for req in build_requires])
+                reqs = hooks.get_requires_for_build_wheel({})
+                env.pip_install([transform_dep_for_pip(req) for req in reqs])
+                try:
+                    metadata = hooks.prepare_metadata_for_build_wheel(path)
+                except Exception:  # sadly, pep517 has no good error here
+                    pass
+            if metadata is not None:
+                metadata_path = os.path.join(
+                    path, metadata, "METADATA"
+                )
+        else:
+            # This is a wheel, so metadata should be in *.dist-info folder:
+            metadata_path = os.path.join(
+                path,
+                [f for f in os.listdir(path) if f.endswith(".dist-info")][0],
+                "METADATA"
+            )
+
+        # Copy the metadata file:
+        shutil.copyfile(metadata_path, os.path.join(output_path, "METADATA"))
+    finally:
+        if clean_up_path:
+            shutil.rmtree(path)
+
+
+def is_filesystem_path(dep):
+    """ Convenience function around parse_as_folder_reference() to
+        check if a dependency refers to a folder path or something remote.
+
+        Returns True if local, False if remote.
+    """
+    return (parse_as_folder_reference(dep) is not None)
+
+
+def parse_as_folder_reference(dep):
+    """ See if a dependency reference refers to a folder path.
+        If it does, return the folder path (which parses and
+        resolves file:// urls in the process).
+        If it doesn't, return None.
+    """
+    # Special case: pep508 urls
+    if dep.find("@") > 0 and (
+            (dep.find("@") < dep.find("/") or "/" not in dep) and
+            (dep.find("@") < dep.find(":") or ":" not in dep)
+            ):
+        # This should be a 'pkgname @ https://...' style path, or
+        # 'pkname @ /local/file/path'.
+        return parse_as_folder_reference(dep.partition("@")[2].lstrip())
+
+    # Check if this is either not an url, or a file URL:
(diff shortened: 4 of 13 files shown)


JonasT:
Just to be super clear, this pull request doesn't implement any auto-detecting recipes in the `setup.py`, and no mapping of them to the actual recipes. It also doesn't implement any recipe or dependency blacklisting. But I have plans to expand on most of this :smile: :+1:

JonasT:
**Update: tested & finished:**

I tested this in a nontrivial `python3` app, including the conflict pinning to avoid touching installed recipes (at least those recognized by pip as installed which were all I had as deps, not sure if some are done in such a hackish way that they might not be) and including not specifying any `--requirements` other than the recipe ones with letting the `setup.py` pull in everything else.

My app also installs itself and imports itself from the site-packages (and won't work if not installed), so I also tested that self-import works properly.

**So feel free to review, test, and give me feedback!** :+1: :slightly_smiling_face:

doc/source/quickstart.rst
```diff
@@ -149,18 +149,34 @@ You have the possibility to configure on any command the PATH to the SDK, NDK an
...
+
+**Note on `--requirements`: you must add all
+libraries/dependencies your app needs to run.**
+Example: `--requirements=python3,kivy,vispy`. For an SDL2 app,
+`kivy` is not needed, but you need to add any wrappers you might
+use (e.g. `pysdl2`).
 
-    p4a apk --private $HOME/code/myapp --package=org.example.myapp --name "My application" --version 0.1 --bootstrap=sdl2 --requirements=python2,kivy
+This `p4a apk ...` command builds a distribution with `python3`,
+`kivy`, and everything else you specifeid in the requirements.
```
AndreMiras: Minor typo: `s/specifeid/specified/`

doc/source/quickstart.rst
```diff
@@ -283,6 +288,45 @@ include such as::
...
+- `--requirements` still must contain all the dependencies that
+  need a python-for-android recipe to work.
+  If you aren't sure which ones do, put all of them there as usual
+  or try it out.
+  (recipes without dependencies *can* be omitted, however!)
+
+- The `setup.py` *can* specify any dependencies that are also
+  in `--requirements` (no matter if recipe'd or not). However, it *can't*
+  enforce incompatible versions for `--requirements`-listed packages -
+  it is best not to enforce anything for these packages beyoond a
```
AndreMiras: typo `s/beyoond/beyond/`

pythonforandroid/build.py
```diff
@@ -650,22 +653,22 @@ def build_recipes(build_order, python_modules, ctx):
...
 
-    if not modules:
-        info('There are no Python modules to install, skipping')
-        return
-
-    info('The requirements ({}) don\'t have recipes, attempting to install '
-         'them with pip'.format(', '.join(modules)))
-    info('If this fails, it may mean that the module has compiled '
-         'components and needs a recipe.')
+    if len(modules) > 0:
```
AndreMiras: Minor: I think the style guide actually suggest to use `if not module:` like it was before to check empty list, see https://www.python.org/dev/peps/pep-0008/

pythonforandroid/build.py
```diff
@@ -709,25 +703,157 @@ def run_pymodules_install(ctx, modules):
...
+        # Install the manually specified requirements first:
+        if not modules:
+            info('There are no Python modules to install, skipping')
+        else:
+            info('Creating a requirements.txt file for the Python modules')
+            with open('requirements.txt', 'w') as fileh:
+                for module in modules:
+                    key = 'VERSION_' + module
+                    if key in environ:
+                        line = '{}=={}\n'.format(module, environ[key])
```
AndreMiras: Out of context since you just moved these lines around, but I was curious why this is written that way :thinking:

pythonforandroid/build.py
```diff
@@ -709,25 +703,157 @@ def run_pymodules_install(ctx, modules):
...
+                # Make sure all packages found are fixed in version
+                # by writing a constraint file, to avoid recipes being
+                # upgraded & reinstalled:
+                with open('constraints.txt', 'wb') as fileh:
+                    output = subprocess.check_output([
+                        "bash", "-c",
+                        "source '" + join(
+                            ctx.build_dir, "venv", "bin", "activate"
+                        ).replace("'", "'\"'\"'") + "' && " +
+                        "pip freeze"
```
AndreMiras: Minor DRY: can't we run that command once and save the output, then once we print it, and the other time we write it to file.
Or simply you write it to that `constraints.txt` file as you did and then you just print the file

pythonforandroid/build.py
```diff
@@ -709,25 +703,157 @@ def run_pymodules_install(ctx, modules):
...
+                    )):
+            with current_directory(project_dir):
+                info('setup.py is present, running project install. ' +
+                     '(disable this behavior with --ignore-setup-py)')
+
+                # Output some debug info:
+                info('Contents that will be used for constraints.txt:')
+                shprint(sh.bash, '-c', (
+                    "source '" + join(
+                        ctx.build_dir, "venv", "bin", "activate"
```
AndreMiras: to make it shorter, you don't actually need to `source` the `activate` script, you can directly invoke the `pip` command via `./venv/bin/pip freeze` that saves us the `source` command :sunglasses:

pythonforandroid/build.py
```diff
@@ -709,25 +703,157 @@ def run_pymodules_install(ctx, modules):
...
+                # Get listing of virtualenv's site-packages, to see the
+                # newly added things afterwards & copy them back into
+                # the distribution folder / build context site-packages:
+                previous_venv_contents = os.listdir(venv_site_packages_dir)
+
+                # Actually run setup.py:
+                info('Launch setup.py...')
+                shprint(sh.bash, '-c', (
+                    "source '" + join(
+                        ctx.build_dir, "venv", "bin", "activate"
```
AndreMiras: Minor: same here, if you're lazy and want to save some lines, you can skip the `source` part and just go:
```python
./venv/bin/pip install -c constraints.txt -v .
````

File: pythonforandroid/build.py [hunk 8]
```diff
@@ -650,22 +653,22 @@ def build_recipes(build_order, python_modules, ctx):
...
+        ctx, python_modules, project_dir,
+        ignore_setup_py=ignore_project_setup_py
+    )
 
     return
 
 
-def run_pymodules_install(ctx, modules):
+def run_pymodules_install(ctx, modules, project_dir, ignore_setup_py=False):
     modules = list(filter(ctx.not_has_package, modules))
```
AndreMiras: This method starts to be pretty big, do you think that would make sense to split it a bit more by concern or that's too painful?

AndreMiras:
Impressive work as usual :clap: 
Few remarks before I actually test it.

1. I would have intuitively turned the feature off by default and make it possible to enable with a flag. That way we're sure we don't introduce breaking changes
2. This may be working now, but of course it introduces more complexity, which means more things to think of when introducing new changes, more checks to do, more lines to maintain. Plus this is not unit tested nor integration tested. This is probably not easy to automate it anyway. But that's a friendly reminder to keep in mind every time we introduce code to the core. This code will have to be maintained
3. Did you see the python2 CI is now failing, do you know why? https://travis-ci.org/kivy/python-for-android/builds/486715046

pythonforandroid/build.py
JonasT: I vaguely recall that John Carmack wrote a nice article why splitting up big functions without actually making use of the inter-dependencies can actually be harmful. (because you get hidden dependencies between the split up functions increasing the risk later to mess things up when changing stuff, and the code flow is less obvious for no benefit)

I agree with this sentiment, and don't think there is any use in splitting it up just because it's long. However, if you think one part in there might be also needed to be called from somewhere else then that sounds like a good idea to me

pythonforandroid/build.py
JonasT: I don't think there's a good reason. It is stored in the env in `toolchain.py` and dug out of the env here, it should be a member on the `Context()` instead. Pretty sure this is just a spontaneous idea gone meh which is worth refactoring, but this pull request is already big enough

JonasT:
I had an issue with `graph.py` (and other recipe code) being case sensitive, while pip is not. So I had to fix that as well to get everything working which blew up the pull request a little, sorry about that :fearful:

AndreMiras:
Thanks @JonasT for that PR!
You know how I like huge PR, don't you? :stuck_out_tongue: 
Specially since this one may have controversial changes. Would you mind to split it by concerns?

AndreMiras:
Thank you so much @JonasT ! :beer:

Dockerfile.py3
```diff
@@ -96,7 +96,7 @@ ENV WORK_DIR="${HOME_DIR}" \
 # install system dependencies
 RUN ${RETRY} apt -y install -qq --no-install-recommends \
         python3 virtualenv python3-pip wget lbzip2 patch sudo \
-    && apt -y autoremove
+    && apt -y autoremove python3-venv
```
AndreMiras: indeed wrong line :joy:

pythonforandroid/build.py
AndreMiras: Well sometimes when the split by concern makes obvious sense. And sometimes it also simplifies unit testing it.
But it was just a suggestion, I didn't dig into it myself and I will not for now

JonasT: makes sense. for what it's worth, this function is the core install-the-whole-bunch place so I don't think it is very useful to be unit tested. (it's mainly relevant for the integration test where the non-split up code doesn't make a difference) so good point @ unit tests, but IMHO this should be ok

pythonforandroid/pythonpackage.py
```diff
@@ -0,0 +1,664 @@
...
+        a requirements.txt file or install_requires=[] listing.
+
+        Current supported metadata files that will be extracted:
+
+        - pytoml.yml  (only if package wasn't obtained as wheel)
+        - METAINFO
+    """
+
+    if package is None:
+        raise ValueError("package cannot be None")
```
AndreMiras: I'm curious why don't you let it crash later with `AttributeError` rather than doing this?

pythonforandroid/pythonpackage.py
```diff
@@ -0,0 +1,664 @@
...
+                print("Detail output:")
+                print(output)
+                print("End of Detail output.")
+            raise ValueError(
+                "failed to obtain meta info - "
+                "is '{}' a valid package? "
+                "Detailed output:\n{}".format(package, output)
+                )
+    finally:
+        shutil.rmtree(temp_folder)
```
AndreMiras: some thoughts: this is why we use context managers. It's surprising Python doesn't provide one built-in for the tempfile use case. I may build that package one day...

pythonforandroid/pythonpackage.py
```diff
@@ -0,0 +1,664 @@
...
+
+    # Abort if not in virtualenv or venv:
+    if not hasattr(sys, "real_prefix") and (
+            not hasattr(sys, "base_prefix") or
+            os.path.normpath(sys.base_prefix) ==
+            os.path.normpath(sys.prefix)):
+        return sys.executable
+
+    # Extract prefix we need to look in:
+    if hasattr(sys, "real_prefix"):
```
AndreMiras: looks fishy, but I'll take it

pythonforandroid/pythonpackage.py
```diff
@@ -0,0 +1,664 @@
...
+    for p in good_candidates + bad_candidates:
+        result = python_binary_from_folder(p)
+        if result is not None:
+            return result
+
+    raise RuntimeError("failed to locate system python in: " +
+                       sys.real_prefix)
+
+
+def get_package_as_folder(dependency, major_python_version=3):
```
AndreMiras: Thanks for writing some docstrings :smile: 
But this looks so complex :fearful: I hope it's properly covered by tests:grimacing:  You know I'll probably check the corner cases and look to simply things :sweat_smile:

pythonforandroid/pythonpackage.py
```diff
@@ -0,0 +1,664 @@
...
+                'venv creation unexpectedly ' +
+                'failed. error output: ' + str(output)
+            )
+        venv_path = os.path.join(venv_parent, "venv")
+
+        # Update pip and wheel in venv for latest feature support:
+        try:
+            filenotfounderror = FileNotFoundError
+        except NameError:  # Python 2.
+            filenotfounderror = OSError
```
AndreMiras: ouch, I think `six` is already in our deps, maybe we should use it here

pythonforandroid/pythonpackage.py
```diff
@@ -0,0 +1,664 @@
...
+                for entry in metadata_entries
+                if entry.startswith("Requires-Dist")
+            ]
+
+            return list(set(requirements))  # remove duplicates
+    finally:
+        shutil.rmtree(output_folder)
+
+
+package_name_cache = dict()
```
AndreMiras: is this all that slow that we want to implement our own caching?

pythonforandroid/pythonpackage.py
```diff
@@ -0,0 +1,664 @@
...
+                      "to package name: " + str(package_dep))
+            package = get_package_name(package_dep)
+            if package.lower() in packages_processed:
+                continue
+            if verbose:
+                print("get_package_dependencies: processing package: " +
+                      str(package),
+                      )
+                print("get_package_dependencies: Packages seen so far: " +
+                      str(packages_processed),
```
AndreMiras: minor: don't we find "{}".format() more elegant than str concatenation?

pythonforandroid/pythonpackage.py
```diff
@@ -0,0 +1,664 @@
...
+
+
+def get_package_dependencies(package, recursive=False, verbose=False):
+    """ Obtain the dependencies from a package. Please note this
+        function is possibly SLOW, especially if you enable
+        the recursive mode.
+    """
+    packages_processed = set()
+    package_queue = [package]
+    reqs = []
```
AndreMiras: why don't we use a set from the beginning then if at the end we want it to act as a set?

pythonforandroid/pythonpackage.py
JonasT: if you know a context manager for this I'd happily add it here!

pythonforandroid/pythonpackage.py
JonasT: The meaning of these is explained here: https://stackoverflow.com/questions/1871549/determine-if-python-is-running-inside-virtualenv

pythonforandroid/pythonpackage.py
JonasT: The `_get_system_python_executable` one? Yeah I added multiple test cases that actually create a virtualenv, see `tests/test_pythonpackage.py`

pythonforandroid/toolchain.py
```diff
@@ -588,6 +676,25 @@ def warn_on_deprecated_args(self, args):
...
 
+        # Output warning if setup.py is present and neither --ignore-setup-py
+        # nor --use-setup-py was specified.
+        if getattr(args, "private", None) is not None and \
+                (os.path.exists(os.path.join(args.private, "setup.py") or
+                 os.path.exists(os.path.join(args.private, "pyproject.toml")))
+                ):
+            if not getattr(args, "use_setup_py", False) and \
+                    not getattr(args, "ignore_setup_py", False):
+                warning("  **** FUTURE BEHAVIOR CHANGE WARNING ****")
```
AndreMiras: Feels like I already saw that somewhere else :wink:

setup.py
```diff
@@ -21,8 +21,10 @@
 # https://github.com/kivy/buildozer/issues/722
 install_reqs = [
     'appdirs', 'colorama>=0.3.3', 'jinja2', 'six',
-    'enum34; python_version<"3.4"', 'sh>=1.10; sys_platform!="nt"'
+    'enum34; python_version<"3.4"', 'sh>=1.10; sys_platform!="nt"',
+    'pep517', 'pytoml', 'virtualenv'
```
AndreMiras: :sob:

tests/test_pythonpackage.py
```diff
@@ -0,0 +1,133 @@
...
+import shutil
+import sys
+import subprocess
+import tempfile
+
+from pythonforandroid.pythonpackage import (
+    get_package_dependencies,
+    get_dep_names_of_package,
+    get_package_name,
+    _get_system_python_executable,
```
AndreMiras: Minor: we import `_get_system_python_executable` why is it `protected`?

pythonforandroid/pythonpackage.py
JonasT: @AndreMiras fun fact, to determine the name of a python package you need to install everything in its `setup_requires` or `build_requires`. So yes, for my project this actually runs like a minute long :nauseated_face:  since the name is queried multiple times I added this cache

tests/test_pythonpackage.py
```diff
@@ -0,0 +1,133 @@
...
+    # but we can't test this inside tox's virtualenv:
+    pybin = _get_system_python_executable()
+
+    # The python binary needs to match our major version to be correct:
+    pyversion = subprocess.check_output([
+        pybin, "-c", "import sys; print(sys.version)"
+    ], stderr=subprocess.STDOUT)
+    try:
+        pyversion = pyversion.decode("utf-8", "replace")
+    except AttributeError:
```
AndreMiras: I'm curious about that line?

JonasT: the decode? just makes sure it's a unicode string, `subprocess.check_output()` returns bytes sometimes (I have to admit I can't say off the top of my head under which)

setup.py
JonasT: :grimacing:

pythonforandroid/pythonpackage.py
JonasT: oh yeah that was silly of me, changing it now :+1:

pythonforandroid/pythonpackage.py
JonasT: yup we do I just keep forgetting :joy:

pythonforandroid/pythonpackage.py
JonasT: meh I usually don't do this but this function is a bit bigger and already does stuff in the filesystem before that would first cause an error, and I once passed None from another function and was confused :smile: no *good* reasons but it feels more appropriate to catch it early for this one :woman_shrugging: (I'm happy to remove it if you think otherwise)

AndreMiras:
Thanks for putting so much effort again to make another advanced pull request :+1: 
I think the feature you're trying to push is nice, but my fear is that this code is pretty complex and nobody, but you will be willing to maintain. But if eventually someone else or even you make subsequent fixes, it's very likely that we would break things.
So I'm really going to try to push for simplifying, documenting and unit/integration testing things as much as possible. Even though what you bring here is probably tricky to properly test.

JonasT:
@AndreMiras for what it's worth this is like after 4 rewrites, it's really tricky to get these packaging things into a good shape. By which I don't mean to say in any way I'm tired of working on it, but that it's complex: therefore, by all means I'm very interested in ideas how to simplify it!

As for tests, I would think I didn't to the most incomplete job in `tests/test_pythonpackage.py` but let me know what you think :smile:

pythonforandroid/pythonpackage.py
JonasT: changed :heavy_check_mark:

pythonforandroid/pythonpackage.py
JonasT: changed :heavy_check_mark:

tests/test_pythonpackage.py
JonasT: mostly because it's a function nobody should ever need to use and which isn't particularly awesome :grimacing:

AndreMiras:
Yes I agree with you it's complex and we'll not rush it into the tree, but eventually little by little we may get there.
I like that some of it is at least somewhat isolated into a module.
Also I think it's good to remember sometimes feature should not be integrated if they involve impact the complexity too much or if they don't cover enough cases. It's mainly an arbitrary choice (even though complexity and usage can be measured somehow). For instance see the CPython GIL, even though having actual multithreading would be awesome it would make the code more complex so they did the GIL. I'm not saying it's good or bad choice, but at some point we make arbitrary choice taking complexity into account.

brvier:
> I think the feature you're trying to push is nice, but my fear is that this code is pretty complex and nobody, but you will be willing to maintain. But if eventually someone else or even you make subsequent fixes, it's very likely that we would break things.

I m clearly interested in this PR. Also, not sure i could maintain it in future.

JonasT:
While I get the notion @ complexity, we are basically speaking of "you have a setup.py for all platforms and it just works". I would argue this is an absolute killer feature, but of course I'm biased :joy:

Also for what it's worth I don't plan to get hit by a truck tomorrow, but you never know I guess

pythonforandroid/pythonpackage.py
AndreMiras: Yes I think that's a very fair point :+1:  maybe just add a comment stating so.
```
# sanity check to fail fast rather than in the middle of the process of touching the file system
```
With a unit test that prove the point it would obviously be even better. The test would not simply show that `ValueError` is being raised. But also that some critical part of the code (e.g. the ones touching the file system) are not being reached. You can use unittest.mock to achieve this.

And as a general note, anything that doesn't look natural in the code should have at least a comment if not a unit test

pythonforandroid/pythonpackage.py
AndreMiras: Thanks! Yes I was assuming something like this :smiley: 
That's usually how I end up writing some one-liner methods :joy: because it's can be more self explanatory

pythonforandroid/pythonpackage.py
AndreMiras: Thanks! I'll play with it more in depth if I ever get time

tests/test_pythonpackage.py
AndreMiras: Yes keep it that way, I like it being protected, but when I saw it being imported I got confused. Since it's only imported in the unit test, it's actually fine. It was nitpik anyway

tests/test_pythonpackage.py
AndreMiras: No about the try/catch in a unit test. In general try catching in unit tests is very fishy

JonasT: @AndreMiras as always I'm happy to remove it if you think I should, *but* I am pretty sure this is the agreed upon "pythonic" way to decode a unicode bytearray everywhere, so I'm really not sure what the issue is...?

I do also use try catch in another place to throw the error again with a better description, and I really don't think there is any issue with it.

(or the other place where I explained in a comment after checking for "pep517" that the test cannot reasonably run since that is not available in the host python)

I agree with the general notion if there isn't a really good reason for it, but in this particular case I'm not sure I can understand your dislike :question: nevertheless, you want me to get it out then I'll remove it and use `type()` or whatever

JonasT: TL;DR: not a hill I want to die on, but I think this is the best way to write it with my limited knowledge

pythonforandroid/pythonpackage.py
JonasT: (I discussed this with pip / setuptools folks btw, it really just is like that, inherently. so it's not something we're doing wrong, there is no other way due to how much code injection python packaging scripts allow)

tests/test_pythonpackage.py
AndreMiras: Well if you can tell me on what circonstances it would raise an `AttributeError` then maybe I would probably better understand why we need a try/catch here in this unit test :thinking: 
Because I didn't find any obvious reason so far:
```sh
python2 -c 'import subprocess; print(subprocess.check_output(["uname"]).decode("utf-8"))'
Linux

python3 -c 'import subprocess; print(subprocess.check_output(["uname"]).decode("utf-8"))'
Linux
```
In general try/catching in unit test is a code smell, of course there are exception, but they probably deserve a code comment to be justified.
So if this `subprocess.check_output()` call sometimes returns something unexpected I would definitely love to have a clear comment explaining why and under what conditions.
The best way to find out is maybe to drop this try/catch and see if it's actually catching something, if it does then we can document it. If not, great we simplified the code a bit further

tests/test_pythonpackage.py
```diff
@@ -0,0 +1,133 @@
...
+        p1 = os.path.normpath(run__get_system_python_executable(pybin))
+        p2 = os.path.normpath(pybin)
+        assert(p1 == p2)
+    except RuntimeError as e:
+        if "pep517" in str(e.args):
+            # System python probably doesn't have pep517 available.
+            # Not much we can do in that case.
+            pass
+        else:
+            raise e
```
AndreMiras: Minor: in Python2 doing a `raise e` would swallow the original source. I would recommend using just `raise` instead.
Anyway I'm not sure what do we do that exception handling thing here :thinking:

tests/test_pythonpackage.py
```diff
@@ -0,0 +1,133 @@
...
+        try:
+            return result.decode("utf-8", "replace").strip()
+        except AttributeError:
+            return result.strip()
+
+    # Check that in system-wide python, the system-wide python is returned:
+    try:
+        p1 = os.path.normpath(run__get_system_python_executable(pybin))
+        p2 = os.path.normpath(pybin)
+        assert(p1 == p2)
```
AndreMiras: assert parenthesis is misleading, it's not a function but a keyword

tests/test_pythonpackage.py
```diff
@@ -0,0 +1,133 @@
...
+            ")\n"
+            "print(m._get_system_python_executable())",
+            os.path.join(os.path.dirname(__file__), "..",
+                         "pythonforandroid", "pythonpackage.py"),
+        ]
+        # Actual call to python:
+        try:
+            result = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as e:
+            raise RuntimeError("call failed, with output: " + str(e.output))
```
AndreMiras: why masking the `CalledProcessError` exception with another one? I'm not sure what purpose it addresses?

tests/test_pythonpackage.py
```diff
@@ -0,0 +1,133 @@
...
+        ])
+        subprocess.check_output([
+            os.path.join(test_dir, "virtualenv", "bin", "pip"),
+            "install", "-U", "pep517"
+        ])
+        assert(os.path.normpath(run__get_system_python_executable(
+                                os.path.join(test_dir, "virtualenv",
+                                             "bin", "python"))) ==
+               os.path.normpath(pybin))
+        if int(sys.version.partition(".")[0]) >= 3:
```
AndreMiras: Conditional statements in a unit test with different code path kinda hurt me by default.
Not sure if it would make more sense or be more elegant to use `unittest.skipIf` for Python2. At the end it's probably the same because it using a conditional statement, but it makes it maybe more explicit

tests/test_pythonpackage.py
```diff
@@ -0,0 +1,133 @@
...
+                "install", "-U", "pip"
+            ])
+            subprocess.check_output([
+                os.path.join(test_dir, "venv", "bin", "pip"),
+                "install", "-U", "pep517"
+            ])
+            assert(os.path.normpath(run__get_system_python_executable(
+                                    os.path.join(test_dir, "virtualenv",
+                                                 "bin", "python"))) ==
+                   os.path.normpath(pybin))
```
AndreMiras: It's pretty hard to say what we're testing and what we're not testing here. It's definitely more an integration test here going as far as touching the filesystem, hitting the network and stuff. Plus we use this helper wrapping function that may also mask things from the actual method we wanted to test.

tests/test_pythonpackage.py
JonasT: > In general try/catching in unit test is a code smell

Have to disagree here. How is`try .. catch` not *always* code smell if it's not for a purpose or too wide, and how is it ever smelly if it's narrow with a clear purpose?

The main problem here seems to be that you missed why I use it here, so let me explain:

> Well if you can tell me on what circonstances it would raise an AttributeError

https://docs.python.org/3/library/subprocess.html#subprocess.check_output

As you can read here, it can return bytes, *or* an str. depends on arguments. I would consider it more of a code smell to hardcode the decode as you do, because people tend to adjust parameters and it will blow up then - unless you write it as I did. I suppose you could argue let it blow up if people change the arguments and fix it, but why waste the future maintainers time with that if it can be written in a way that doesn't have that problem?

> The best way to find out is maybe to drop this try/catch and see if it's actually catching something,

Disagree very much. That'd be guessing around a bog standard function which is `subprocess.check_output` of which the behavior of it is well understood and documented. (again see doc link above)

I have to admit I had to look this up and did as you brought it up, because I only vaguely remembered that it could maybe return both types - but it really can, so I think the code makes perfect sense as-is

> but they probably deserve a code comment to be justified.

I disagree again, with this instance specifically. 100% standard approach of "make this a unicode string if it's not already one". I would have to explain this every time I use `subprocess.check_output`, that doesn't sound like a useful thing to do if we want to have a good signal to noise ratio in the code comments

> Again this is in the context of unit testing [...] But specially have one in a unit test looks weird to me

Another disagree :grimacing: to me that's not a good approach/not a useful criteria. Either it has a use or not, and here it has one

These are all just my personal opinions of course. As usual I'll go with what the team decides, and I can be wrong about everything (and I often am), and this can for sure also be written without `try...catch`. However, this is quite a regular Pythonic approach *as far as I'm aware*, by using the object with the intended behavior (decode) and catch the unusual cases (already string, can't decode -> AttributeError) instead of using `type()` hardcoding any of the type checks.

But if you have a better code suggestion let me know! :+1: I think your `subprocess(...).decode` isn't much better though. I suppose you could say if someone changes the arguments and it blows up they'll notice, but I don't see why it shouldn't be written in the first place such that it doesn't matter...?

tests/test_pythonpackage.py
JonasT: haha, sorry about that. i do these all the time because subconsciously I think it's a wrong decision/ugly on python's part (should have been turned to a function with python 3 as `print` was) but I know about the confusing behavior if a message is appended. I'll remove it :+1:

tests/test_pythonpackage.py
JonasT: This code path is very likely to be hit with a very wide array of errors, so this just improves the backtrace standard output (which for a subprocess error by default won't print the failed process output, which usually is vital to understand why things went wrong at this particular location)

tests/test_pythonpackage.py
JonasT: Open for anything, I have to admit I'm not very familiar with them @ conditional statement/tests, so I just didn't think about it :grimacing:

tests/test_pythonpackage.py
JonasT: Multiple things here: 1. you're right I should do more fine-grained tests of the important functions below, will add them 2. hitting the network: unavoidable since the code path for downloading is quite different so this needs to be tested :woman_shrugging: and it can't really be mocked well because you'd need to mock `pip` download things, that seems more trouble than it's worth (open for ideas though), 3. yea you're right @ wrapping, my tests don't go deep enough

will add more tests tomorrow, stay tuned :+1:

tests/test_pythonpackage.py
AndreMiras: > As you can read here, it can return bytes, or an str. depends on arguments. I would consider it more of a code smell to hardcode the decode as you do, because people tend to adjust parameters and it will blow up then - unless you write it as I did. I suppose you could argue let it blow up if people change the arguments and fix it, but why waste the future maintainers time with that if it can be written in a way that doesn't have that problem?

Everything lies in this statement... You say "because people tend to adjust parameters and it will blow up then" but that's the point, it's an unit test, so you know its entry upfront, this is what you communicate with your test. You say "I know that with input X it gives Y". If your test says "I'm not sure what X is and also not sure what Y would give" then what you're actually testing is not clear...

A good way is to simply refer to some "official" implementation, test or documentation. Here's the cpython one...
https://github.com/python/cpython/blob/v3.7.2/Lib/test/test_subprocess.py#L149

JonasT:
As written on discord, you lost me completely. Why would whether bytes or str is returned be ANYTHING that  is tested here ever, or be interesting to the test reader when they want to know whether pythonpackage.py works? So how is "what you are testing" even remotely applying to this specific detail? I just don't follow (and at this point I feel a little dense, but I honestly don't get it)

tests/test_pythonpackage.py
AndreMiras: Another reference, Django, widely used framework. All the references to subprocess in tests:

- https://github.com/django/django/blob/2.1.7/tests/i18n/test_compilation.py
- https://github.com/django/django/blob/2.1.7/tests/runtests.py
- https://github.com/django/django/blob/2.1.7/tests/admin_scripts/tests.py
- https://github.com/django/django/blob/2.1.7/tests/dbshell/test_postgresql_psycopg2.py

None of them do try/catch in here. Again try/catch or code branching is not forbidden, but often a code smell and should always be justified.
In this case I would just drop it

JonasT:
I suppose "hey nobody else gives a sh*t about this, stop wasting lines for it" is a fair point :D I'll remove it once I get back to the pc. I still think this pattern has value, but if nobody else does then it's probably just me being weird

tests/test_pythonpackage.py
AndreMiras: My idea is to simplify the code and flow control in general. If a branch provide just little value, I'd rather kill it.
As often I like to apply the Pareto principle to coding or [YANGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).
So basically this code path seam to serve only the case where people are updating p4a code and break something, then they would somehow see the trace from this line and somehow it would help them more than seeing a `CalledProcessError`. So I'd say, let's drop it

JonasT: YANGNI doesn't apply when I needed it :smile: 

Seriously, I put this in because I hit this multiple times. The error output really is *extremely* useless with this. Are you **sure** you want this gone?

JonasT: "just little value" - I got the strong feeling you haven't actually checked the effect, because I'm not sure how that would apply. Let me show you:

**Default output:**
`E               subprocess.CalledProcessError: Command '['/usr/bin/python3.7', '-c', "import importlib\nimport json\nbla()\nimport os\nimport sys\nsys.path = [os.path.dirname(sys.argv[1])] + sys.path\nm = importlib.import_module(\n    os.path.basename(sys.argv[1]).partition('.')[0]\n)\nprint(m._get_system_python_executable())", '/home/jonas/Develop/python-for-android/tests/../pythonforandroid/pythonpackage.py']' returned non-zero exit status 1.`

*(seriously, can you tell anything about what went wrong from this?)*

**Altered output:**
```
>           raise RuntimeError("call failed, with output: " + str(e.output))`
E           RuntimeError: call failed, with output: Traceback (most recent call last):
E             File "<string>", line 3, in <module>
E           NameError: name 'bla' is not defined
```
*(gives traceback + exact line for the externally run code for the nested error that occured)*

Now if you play around with this test as I did when writing it and you **still** want me to remove this, I'll happily do so. Let me know what you think

tests/test_pythonpackage.py
JonasT: I added more in-depth comments to explain it. Let me know what you think

tests/test_pythonpackage.py
AndreMiras: it's just that the more you wrap you test with some bloated helper method, the more you're not testing what you are supposed to test.
Let's leave it for now because it's a detail compared to the general test structure. Just didn't have a chance to properly wrap my head around it to suggest better way to address.
Just a pointer: have you played with unittest.mock in the past? Maybe it's time to take a look

JonasT: @AndreMiras hmmmm... yes I used mock, but this function gets the system python and this tests that it works in different  python environments. How would you mock this? I mean the point is to see if it works in different environments, how would you mock an entire environment with all intricate details that could interfere?

Open for other approaches to testing this, I just don't see any

JonasT: Summed up better, this is definitely an integration test. (and inherently needs to be I'm quite sure, it depends too much on the environment that I see any useful way of mocking it) Would it help if I added a comment that makes this more clear?

AndreMiras: Yes it makes sense that you cannot mock if you play with system python, subprocess :+1: I was imagining so but I was hoping not :stuck_out_tongue: 
Like I said I haven't started digging seriously into what you're doing yet. So I'm just challenging you and throwing ideas in the hope you suddenly find a genius idea to make the whole think much more simple without me having to wrap my head into it somehow :sweat_smile:

tests/test_pythonpackage.py
AndreMiras: Yes I also think it's a sad feature of Python3. I'm not sure why they fixed the `print()` and not `assert`. I hope there're good reasons at least.
Thanks for addressing anyay

tests/test_pythonpackage.py
JonasT: Sorry, I never posted a summary of our followup chat discussion:

I came to the conclusion that while I personally think the bytes/str difference shouldn't be important, it is still *there* and to people who know exactly what it returns, the try/catch conditional decode could be highly confusing (in a way like it was to @AndreMiras ). So while I think I had good motivations, they're probably not worth it in terms of the potential confusion of readers and I've come to agree with @AndreMiras that it should be removed

tests/test_pythonpackage.py
JonasT: Oh right that makes sense. I tend to somehow assume you already got a perfect improved verison laid out in your head and sometimes I'm a bit stumped then that I didn't see it yet :joy: lol. Keep on challenging :+1:

tests/test_pythonpackage.py
```diff
@@ -0,0 +1,138 @@
...
+            "python-for-android", recursive=True
+        )
+        if "MarkupSafe" in dep
+    ]
+    # Check that colorama is returned without the install condition when
+    # just getting the names:
+    assert "colorama" in get_dep_names_of_package("python-for-android")
+
+
+def test__get_system_python_executable():
```
AndreMiras: I would definitely split this test in 2-3 parts.
So here I would make it a test class instead where you add your warning comment as a class docstring.
1. First test would stop here: `assert pyversion.strip() == sys.version.strip()`
2. Second test would start in `test_dir = tempfile.mkdtemp()` and stop in `assert os.path.normpath(sys_python_path) == os.path.normpath(pybin)`
3. Third test would probably be the Python3 dedicated one with a conditional/explicit skipif or something like that

JonasT:
@AndreMiras split up the tests and added way more, both as you asked for, but now there's a new problem:

The tests take **really long** to run through - like, *a couple of minutes...* reasons I suspect are: 1. there are like 5+ tests with PyPI references (necessarily, since that also needs to be tested), 2. for many operations there's a venv/virtualenv created with `setuptools` and other minor things installed which is usually somewhat quick (probably less than 10 or even 5 seconds in most cases but also necessary/unavoidable), and with this large amount of tests it adds up significantly.

So that's not good for our CI/travis run. Can tox be told to exclude a certain test file? Any other suggestions what we could do about that?

JonasT:
I'm just thinking of some caching opportunities e.g. for the metadata files extraction and I'm pretty confident they will somewhat speed up things, *however* even if we're optimistic and I cut it in half, the tests probably still run like a full minute at least :grimacing:

AndreMiras:
Thanks for the tests! I haven't reviewed yet, but it seems to look good!
As for its speed I think a full minute is OK-ish still even though I'm concerned about its potential "flakyness" :smile: 
As for tox ignoring some tests, yes you can pass params to the `pytest` command to tell it to ignore some. For instance some tests could go to a `tests/integration/` folder and we could run them only on py3. I don't know

JonasT:
I'm not sure our travis integration details should affect the folder structure. I would prefer just configuring this with e.g. tox arguments put into `.travis.yml` so that to anyone else, the full test set & test folder structure would not be influenced whatever compromise we do due to our build time needs

> Even though I'm concerned about its potential "flakyness"

Should mostly depend on PyPI being reliable & reachable and the `python-for-android` package being in a no-nonsense state on there

> I think a full minute is OK-is

haven't timed it yet, but right now it's probably more like five... :cry:

JonasT:
@AndreMiras for what it's worth the *additional* caching I tried I gave up on because it made less than 10%ish of a difference and that didn't justify the additional complexity IMHO. while it is possible I get some ideas how to speed this up, I think for now we should assume it won't get significantly faster. So we need to decide whether this is worth merging at all (I obviously think it is, but you all need to decide) and whether there should be something done to skip some of the tests in travis in one way or another since they're taking something around 5 minutes to run in addition to the current build time.

AndreMiras:
>  oh right, you mean just for the tests that are more like integration tests by their nature. that would be almost all of them though, so I'm not sure that's a useful criteria for what to run on travis joy

Ahah yes you're right, we need to think about it more deeply, but there're definitely some tests that are lot more integration than other. I mean the one playing directly with the system python and writing to the file system for instance

> for what it's worth the additional caching I tried I gave up on because it made less than 10%ish of a difference and that didn't justify the additional complexity IMHO

Thanks for taking another look and taking the decision to drop it for the sake of simplicity :pray: 

> So we need to decide whether this is worth merging at all (I obviously think it is, but you all need to decide)

That's the important question and I honestly don't know yet personally.

> whether there should be something done to skip some of the tests in travis in one way or another since they're taking something around 5 minutes to run in addition to the current build time

Yes it's technically possible and it's easy, by checking the `CI` or `TRAVIS` environment variable. But then we don't automatically check for regressions. But yes 5 more minutes to that already long & flaky build is something

JonasT:
@AndreMiras I separated into two files now and it only adds slightly above 3 minutes now :grimacing: by specifying an option in `.travis.yml` to skip the less essential file. (I didn't add a check for this in the test itself or make `tox` per default run less since I think special tricks like this to speed up travis should be kept inside `.travis.yml` if possible, such that a manually executed `tox` run is as comprehensive & "normal" as it can be)

I'm not sure what else could be reasonably done, so maybe now would be a good point to see if we can come to some sort of plan/decision of where to take this feature/pull request, if anywhere at all?

Unless you got something else to be addressed of course

JonasT:
oops, still found a bug with the future behavior change warning which is now fixed. but apart from that, I think the only thing I can at least personally see that is still missing is some sort of verdict/decision :smile:

tests/test_pythonpackage.py
JonasT: this is a test class now, so I'll resolve this :+1:

tests/test_pythonpackage.py
JonasT: I added more tests now, feel free to protest if still any are missing but for now I think this can be resolved :+1:

tests/test_pythonpackage.py
JonasT: this is now a pytest skip thing! so I'll resolve this

tests/test_pythonpackage.py
JonasT: resolving this since it is now hopefully explained better! :+1: (feel free to protest if not)

pythonforandroid/build.py
JonasT: hopefully I remember this correctly, but I do have a `constraints` variable now so I think this should be covered! if not poke me, but for now it seems like this is resolved :+1:

tests/test_pythonpackage.py
JonasT: for what it's worth, this is how it looks like now: https://github.com/kivy/python-for-android/blob/02c7780c15f35dd94eff2cf12c172c3308251ccf/tests/test_pythonpackage_basic.py#L176

I still have that indirect-raise in because I found the additional output useful, but this shouldn't be the roadblock (= I could also remove it) if we want to move things ahead

Let me know if you consider this conversation resolved, or still want the `raise` indirection out!

setup.py
JonasT: I am wondering, virtualenv was previously already used in other parts but this was never an install req. It seems to me like that was actually an oversight and it should be there, or am I missing something? Should this go away?

I mean I'm not actually ever directly importing it, but I am calling it via python -m virtualenv (and most of the other code via the virtualenv executable, which I believe actually should be refactored into the same call style because it can break with user local installs, but that's another issue)

pythonforandroid/toolchain.py
JonasT: oh wow I just noticed this and took it out. yeah it was in twice :laughing: resolving since that is fixed now!

pythonforandroid/pythonpackage.py
JonasT: resolving for now since I don't see a better way, but feel free to poke/unresolve if you still got some ideas on this :+1:

pythonforandroid/pythonpackage.py
JonasT: not sure? I actually experimented around a bit and tried to assign FileNotFoundError directly on python 2, but I couldn't get it to work. (it would lead to an odd NameError, I must have confused the scoping somehow by messing with the built-in)

will resolve for now since you haven't shared a good idea and I agree it's ugly but I don't have another either and it does seem fairly minor. but feel free to poke/slap and/or unresolve if you come up with something! :+1:

JonasT:
For what it's worth I mocked some more of the tests now that I'm better at it, and it only adds ~100 seconds now (still not good but almost down to half from before). Mocking more is probably not very feasible since it uses pip to download, and I think that should remain tested to have some basic idea that everything as a whole still operates.

Any updates on what I should still change or whether we'll possibly merge this or not?

pythonforandroid/pythonpackage.py
```diff
@@ -0,0 +1,684 @@
...
+            )
+
+        # Create download subfolder:
+        os.mkdir(os.path.join(venv_path, "download"))
+
+        # Write a requirements.txt with our package and download:
+        with open(os.path.join(venv_path, "requirements.txt"),
+                  "w", encoding="utf-8"
+                 ) as f:
+            def to_unicode(s):  # Needed for Python 2.
```
AndreMiras: Oh do we?

pythonforandroid/pythonpackage.py
```diff
@@ -0,0 +1,684 @@
...
+        # Get the result and make sure it's an extracted directory:
+        result_folder_or_file = os.path.join(
+            venv_path, "download",
+            os.listdir(os.path.join(venv_path, "download"))[0]
+        )
+        dl_type = "source"
+        if not os.path.isdir(result_folder_or_file):
+            # Must be an archive.
+            if result_folder_or_file.endswith(".zip") or \
+                    result_folder_or_file.endswith(".whl"):
```
AndreMiras: Minor: you can pass a tuple and skip the `or`
```python
result_folder_or_file.endswith((".zip", ".whl"))
```

pythonforandroid/pythonpackage.py
JonasT: well I didn't prefix all strings with `u` so in python 2 they are all bytes, but I use the python 3 `open()` since the python 2 `open()` as almost everything else in py 2 is kind of annoying to me :joy: so in that case this needs to be converted

or what exactly is your question?

pythonforandroid/pythonpackage.py
```diff
@@ -0,0 +1,684 @@
...
+
+        # Copy the metadata file:
+        shutil.copyfile(metadata_path, os.path.join(output_path, "METADATA"))
+    finally:
+        if clean_up_path:
+            shutil.rmtree(path)
+
+
+def is_filesystem_path(dep):
+    if dep.startswith("/") or dep.startswith("file://") or (
```
AndreMiras: same minor thing:
```python
dep.startswith(("/", "file://"))
```

JonasT: I feel pretty dumb for not knowing that pattern yet :joy:

pythonforandroid/pythonpackage.py
```diff
@@ -0,0 +1,684 @@
...
+            dep.find("/") > 0 and
+            dep.find("://") < 0 and
+            (dep.find("@") < 0 or
+             dep.find("@") > dep.find("/"))):
+        return True
```
AndreMiras: Minor: why don't you directly return the whole expression? I mean:
``` python
return my_boolean_expression
```
Rather than:
```python
if my_boolean_expression:
    return True
else:
    return False
```

pythonforandroid/pythonpackage.py
```diff
@@ -0,0 +1,684 @@
...
+    return False
+
+
+def parse_as_folder_reference(dep):
+    """ See if a dependency reference refers to a folder path.
+        If it does, return the folder path (which parses and
+        resolves file:// urls in the process).
+        If it doesn't, return None.
+    """
+    if dep.startswith("/") or dep.startswith("file://") or (
```
AndreMiras: `dep.startswith(("/", "file://"))`

pythonforandroid/toolchain.py
```diff
@@ -179,6 +180,7 @@ def build_dist_from_args(ctx, dist, args):
             ctx, dist.recipes, bs,
             blacklist=blacklist
         ))
+    assert(len(set(build_order).intersection(set(python_modules))) == 0)
```
AndreMiras: - ahah  you're still using parenthesis on `assert`
- also I would rather skip the `len` check and just do `== set()`

tests/test_pythonpackage_basic.py
```diff
@@ -0,0 +1,310 @@
...
+        dep_names = get_dep_names_of_package(
+            test_fake_package, recursive=False,
+            keep_version_pins=False, verbose=True,
+            include_build_requirements=True
+            )
+        assert len(
+            {"buildozer", "python-for-android", "setuptools"}.intersection(
+                dep_names
+            )
+        ) == 3  # all three must be included
```
AndreMiras: then why not directly:
```python
assert {"buildozer", "python-for-android", "setuptools"} == dep_names
```

pythonforandroid/pythonpackage.py
AndreMiras: Yes this is pretty much my question, I just want to kill eventually not so necessary code branching

pythonforandroid/pythonpackage.py
JonasT: done!

pythonforandroid/pythonpackage.py
JonasT: done!

pythonforandroid/pythonpackage.py
JonasT: I kinda just factored that function away since for some mindboggling reason I missed it was kinda doing the same thing as the one below it :joy:

pythonforandroid/toolchain.py
JonasT: lol! done

tests/test_pythonpackage_basic.py
JonasT: Ok so this is difficult. The short answer is because dep_names is `{"buildozer", "python-for-android", "setuptools", "wheel"}`.

The long answer is, any outside entity (as which I pretend to be testing, I'm not saying that is reasonable but that was the idea) should approach this with "I use setuptools to build, so `setuptools` should be in there" (with `include_build_requirements=True`), and then there's of course the other direct deps `buildozer` and `python-for-android`. For items that **I care about** (as a user) that would be `{"buildozer", "python-for-android", "setuptools"}` which is what the test verifies.

Now that `wheel` is added is due to the way pep517 operates which essentially needs it for building the metadata - that knowledge doesn't fall from the sky though, that is done manually in `pythonforandroid/pythonpackage.py`, *but* it feels like knowledge I just felt shouldn't be something the outside user worries about or something I would want to need to keep in sync as such an internal detail, and therefore I didn't want to test it here.

So that is why this is a subset test and the full set is not hardcoded. I'm sure you could argue for just writing the full practical result but it makes more sense to me the way it is, due to the explanation given above

pythonforandroid/pythonpackage.py
JonasT: It might be possible to avoid by littering `u` prefixes all over the code. (but conversions in other places might then pop up, I'm not sure, too much code to tell on a whim) But even if that worked, once we eventually drop python 2 we'd need to remove countless `u` prefixes instead of taking out that `to_unicode` inline function. That kind of makes more sense to me, but I'm happy to leave the decision to you

JonasT: (not that `u` in python 3 does any harm so I suppose we could just leave it in forever - but IMHO it is a little ugly given in python 3 it's just completely unnecessary and often just left away. Not sure what the more common approach is here usually...? But my intuition votes no `u` everywhere)

inclement: Is this what `from __future__ import unicode_literals` is for?

JonasT:
>  but if you specify the root folder with --private then won't the setup.py install the folder and p4a will copy the folder into your app?

Hmmm, good point. Yes, it ends up both in the root, and in the site-packages, and in the end it still runs `main.py` as before. Maybe I should change it only copy the `main.py` in, and none of the other files anymore? Do you have any suggestions? For my app that would work since my `main.py` just imports the package from site packages and runs it and I don't have large amounts of data files, but I'm not 100% sure what the usual user expectation would be

> With that in mind, would you be able to add a trivial testapp with the folder structure you have in mind and a setup.py that just does some trivial package install?

Just add a `setup.py` to your `--private` folder root (where your `main.py` is). Then launch `p4a apk --use-setup-py ...` and it will automatically scan the deps from it, and call it during the python package install phase such that your app gets installed into the site-packages inside the python bundle.

> but I guess it should also work fine if you use the same setup.py file for both purposes? That would be nice and neat.

Yes that should work fine, because the handling I am adding really doesn't care that your `setup.py` also has the `apk` method (or whether it was used to indirectly launch it). All it does is simply invoke `setup.py install` or the pep517 mechanisms to list the dependencies, it shouldn't care any more about that additional `apk` action than normally running `setup.py install` does

> ). I still think that point 2 (--requirements is annoying) is relatively trivial, and if this DRY is a problem then you've already reached python packaging nirvana :p

I disagree @ trivial! For small projects maybe, but just think about version pins or really long indirect deps lists! And now imagine you'll keep them in sync, always. Since that always works out so well. (It probably won't...!)

But I find the way more compelling argument that it's much easier to start with p4a with "use your setup.py as before", instead of "read about --requirements, realize you need all deps INCLUDING indirect ones in there, now go find out what your indirect deps even are, then realize you need to specify some differently so the according recipe is found". This process is much more complicated for newbies than us veterans may realize, and you don't need to do any of it with this pull request

pythonforandroid/toolchain.py
```diff
@@ -588,6 +668,25 @@ def warn_on_deprecated_args(self, args):
...
+        if getattr(args, "private", None) is not None and \
+                (os.path.exists(os.path.join(args.private, "setup.py")) or
+                 os.path.exists(os.path.join(args.private, "pyproject.toml"))
+                ):
+            if not getattr(args, "use_setup_py", False) and \
+                    not getattr(args, "ignore_setup_py", False):
+                warning("  **** FUTURE BEHAVIOR CHANGE WARNING ****")
+                warning("Your project appears to contain a setup.py file.")
+                warning("Currently, these are ignored by default.")
+                warning("This will CHANGE in an upcoming version!")
```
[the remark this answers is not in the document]
JonasT: Haha, I guess I did suggest that it will change in the future :joy: of course if most of you think that's a bad idea I'll just take it out. But my suggestion would be, just for the sake of having even easier defaults, that eventually this is changed such that `--use-setup-py` is the default

pythonforandroid/pythonpackage.py
JonasT: ah nice find! I just tested it though, and it seems pep517 and other external stuff will still pass in python-2-`str` even with that added, which would need conversion in a couple of places. I think in the end it might be less special handling to just have that `to_unicode` there

JonasT:
Marked as WIP because I think @inclement 's double-copy remark should be given some thought. I think the best approach I can think of is to copy only `main.py` when a `setup.py` is used, and make it very clear that the `setup.py install` needs to fully install all needed data files as well.

But I'd like to hear your ideas for this. Maybe it should be an option? Although I am inclined to push people to write proper `setup.py` files if they want to use this which would IMHO for sure mean to copy everything needed into site-packages without exceptions

JonasT:
I was impatient as always and implemented it like that :joy: :joy: :joy: might still be broken lol, but in theory it works :grimacing:

inclement:
> Just add a setup.py to your --private folder root (where your main.py is). Then launch p4a apk --use-setup-py ... and it will automatically scan the deps from it, and call it during the python package install phase such that your app gets installed into the site-packages inside the python bundle.

I get this, but I'd like to have a testapp as much for documentation purposes as anything else. It's valuable to be able to say not just 'look at the docs' but 'look at this example' when a user asks how to include their cython code.

> I disagree @ trivial! For small projects maybe, but just think about version pins or really long indirect deps lists! And now imagine you'll keep them in sync, always.

I appreciate the point, but I still think that it's a relatively niche problem to actually have, and that in any case it's hardly unresolvable - you don't need to keep multiple lists in sync, just read them from the same place.

> But I consider the way more compelling argument that it's much easier to start with p4a with "use your setup.py as before

I strongly agree with this point, but this really isn't what I'd mean by "use your setup.py as before" for most purposes. From my point of view `python setup.py apk` does is much more in that vein, since it integrates p4a as a natural syntax alongside other distribution options, and makes an attempt to use the setup.py config (app name etc.) to set all the p4a options.

I don't mean that this is in conflict with your change, clearly the ability to install the user's python module in the target environment and to compile cython is quite reasonable, I'd just like both to work in a cohesive way. I'm still thinking through what this needs to imply.

>  instead of "read about --requirements, realize you need all deps INCLUDING indirect ones in there, now go find out what your indirect deps even are, then realize you need to specify some differently so the according recipe is found"

This is multiple issues:

- read about --requirements
- realise you need all deps
- realise you need to specify some differently so the according recipe is found

Of these, the second two are mostly not being resolved by anything to do with the user's setup.py, but by your neat code to resolve the dependencies for them. The first one is the one I consider relatively trivial, I think that anyone using a packaging tool should be prepared to expect an argument explicitly specifying what to package.

To be clear though, I'm not saying you should remove this feature, there's nothing fundamentally wrong with it and I don't believe that it causes any problems. I'm just more personally interested in maybe integrating this functionality into the normal requirements resolution, or especially to the `python setup.py apk` syntax for which using the install_requires is clearly expected (and not currently done!).

> But I'd like to hear your ideas for this. Maybe it should be an option? Although I am inclined to push people to write proper setup.py files if they want to use this

I don't think I have a clear vision about this, I'm still thinking about it. I'll try to post some thoughts tomorrow, or on discord if riot ever comes back up!

JonasT:
>>  But I consider the way more compelling argument that it's much easier to start with p4a with "use your setup.py as before
> I strongly agree with this point, but this really isn't what I'd mean by "use your setup.py as before" for most purposes. 

I think that view is very flawed. The options and avoided duplication and easy calling are only one side of the coin, which `apk` covers, fair enough. But the much more important one is **actually running the install routine.** I think you are under-estimating how much easier that is for porting larger projects, e.g. you just *can't* run my app without installing it and working around that would be quite complicated (since the install routine exists for a reason)

As for your other suggestions regarding name resolution and `apk` integration:

I'm a bit torn whether the name-to-recipe resolution should be automatically used without a `setup.py` as well, simply because it's quite slow. Otherwise I would agree that it's a useful thing to do. As for making `apk` install_requires(), that should be possible but it sounds more like a distinct problem (I wouldn't expect the necessary code changes to be in any of the parts I changed)

inclement:
> But the much more important one is actually running the install routine. I think you are under-estimating how much easier that is for porting larger projects, e.g. you just can't run my app without installing it and working around that would be quite complicated (since the install routine exists for a reason)

Maybe I wasn't clear, this is the part I think is a great feature and am 100% behind.

JonasT:
Ah, very well. I agree all the other points are not that major. And I think it makes sense that `setup.py apk` would learn to scan for install_requires() as well, since then these two features could be combined. So that sounds like a reasonable idea. I'm just not sure if it makes sense to also put it into this pull request or rather a followup one

inclement:
No need to put it in this PR, I know it isn't part of the core change and there's no need to have it immediately!

JonasT:
Alrighty then let me put together a quick test app. What about maybe converting one of the ones that are already run anyway for travis? Then we wouldn't need an additional travis test run but this would still be tested

inclement:
Yep, using one of the existing testapps sounds great, I'm really just thinking of something where someone can be told 'copy this basic file/folder structure to start'. Thanks.

JonasT:
Hmmm the only test app currently run right now for python 3 is `testapp_python3_sqlite_openssl_googlendk`, but that one would be a little less minimal, right? I could convert that one anyway and add another additional minimal example into the docs (with actual setup.py contents) or something, or what would you suggest?

inclement:
I guess what I'd really like would be an example that builds a trivial cython file, e.g. a cython function that calls libc.math.sqrt and a main.py that just runs that function. The main.py wouldn't need to have a gui or anything, just nice to have a demonstration of how you would package an app with a cython file. I realise this isn't hard to work out from the point of view of knowing how these things work, but I think it's quite valuable for users who often are not that familiar with how cython or python packaging really fits together.

JonasT:
@inclement ok I am just encountering a problem: to run the `setup.py`, it needs to be 1. named `setup.py`, 2. located in the root of `--private` such that the package can still be installed (that also means as a consequence `main.py` would usually be an **outside** file, not inside a submodule folder). Both is not the case for every single test app in `testapps` right now

I also find this pretty counter-intuitive in general, apart from this being a problem for testing this pull request. Should I refactor them all to be properly contained in a folder with a regular `setup.py`? Or what is the reason they're separated out like that? I've never seen it done like that before

inclement:
I think the simplest way to fit with the current folder structure would be to make a new folder `testapp_with_setup` (better name welcome!) that contains the setup.py and main.py and any other files/folders you want, then a script `build_testapp_with_setup.sh` at the root level containing the p4a command to build that folder.

The current structure is organised by app code (in the folders) and setup.py files intended for `python setup_some_details.py` that build specific folders with specific p4a options. The reason for separating things this way is that multiple setup.py files often share the same app code. I've never perceived any issues with this structure, but I'm not clear what change you're proposing exactly.

JonasT:
> but I'm not clear what change you're proposing exactly.

I proposed making them all separate subfolders with `setup.py`!

Oh they share the app code? Hm, what about using symlinks then? The thing is it's IMHO kind of an unusual organization (not necessarily bad, just not very common) and now if I make some with subfolder with `setup.py` inside it's going to be an even weirder mix, right? Some folders being submodules to the setup-py-crowd in the folder above, some bringing their own `setup.py` inside, that doesn't sound too great to me

inclement:
I've thought about that before, but I ended up thinking that the end result would be just the same but with unnecessary nesting: instead of multiple files `setup_{X,Y,Z}.py` you'd have multiple folders `build_app_{X,Y,Z}` each containing a single file `setup.py`. I didn't think that was really any better.

How about making another issue for this folder structure if you think it's worth looking at? I'm happy to discuss it, but probably not worth blocking this big PR on it.

JonasT:
Sure I agree it wasn't any better up to current master, so I see where you're coming from. But with this PR this structure will be *required* for at least one test app, so I'd rather have them all like that personally

But okay, I can leave it varying/inconsistent for now, it just bothered me a little since its ugly to me :grimacing:

JonasT:
@inclement sorry I was also just sidetracked with CSS in my toolkit and forgot to add a test app :joy: it's complete and I use this daily, so apart from the missing test app (which is why I left `WIP` in the title) I don't see anything else that's missing

inclement:
@JonasT The thing I wanted to do but didn't get around to was to try to think through fully how the python-for-android api should work - I kind of feel that both the existing setup.py stuff, and the way your additions work, are imperfect interfaces. I was thinking of trying to write some ideal documentation for how I'd like python-for-android to work for people, as a way to try to see what we could change.

I think this might lead to changing the recommended api, maybe even a breaking change, but I guess it's better to just roll with it and have this code in than to think too hard about changes I don't have time for right now anyway!

So, thanks for the PR :)

nitanmarcel:
@JonasT Could you take a look at my issue #1808 ? I think is linked with your pull requests. Thanks

JonasT:
@nitanmarcel indeed that was me :grimacing: fix is here: https://github.com/kivy/python-for-android/pull/1809/ someone should get around to review and merge it in the next hours, then things should hopefully work again with buildozer

PR #1627, opened 2019-01-29
Clean-up LDSHARED, fixes #1360

### LINKED ISSUE #1360: Fix/clean-up LDSHARED

AndreMiras:
Since #1361 was merged, we can drop LDSHARED override.

### DIFF
--- a/pythonforandroid/recipes/pycryptodome/__init__.py
+++ b/pythonforandroid/recipes/pycryptodome/__init__.py
@@ -6,11 +6,5 @@ class PycryptodomeRecipe(PythonRecipe):
     url = 'https://github.com/Legrandin/pycryptodome/archive/v{version}.tar.gz'
     depends = ['setuptools', 'cffi']
 
-    def get_recipe_env(self, arch=None, with_flags_in_cc=True):
-        env = super(PycryptodomeRecipe, self).get_recipe_env(arch, with_flags_in_cc)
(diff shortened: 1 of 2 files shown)


inclement:
@AndreMiras Looks like this travis build is failing with the "Cannot locate <crypt.h>" error.

AndreMiras:
Thanks @inclement build is green now :smiley:

PR #1632, opened 2019-01-30
[WIP] Experimental: an attempt to fix library loading with LD_LIBRARY_PATH

JonasT:
Based on the idea of @inclement this is a test to see whether `ctypes.util.find_library` can be fixed by just setting `LD_LIBRARY_PATH` accordingly. Will report back on the results after testing


JonasT:
since it doesn't work, closing it

PR #1635, opened 2019-01-30
Revert use of shlex.quote to avoid problems with python 2

JonasT:
Revert use of shlex.quote introduced in `pythonforandroid/archs.py` to avoid problems with python 2

tested & (my own real app) build ran through, ready for merge (or change requests) as far I'm concerned


inclement:
Thanks!

PR #1640, opened 2019-01-31
Added separate module for checking user SDK, NDK, API etc.

inclement:
I fixed the ndk version check (the user parameter is removed entirely as it has no purpose), and moved various different checks into their own functions in a new recommendations module. This should make it simpler to update p4a's warnings and recommendations in the future.

Also fixed a few bugs found while testing.


inclement:
@AndreMiras Can you approve again and merge? One test failed only due to travis flakiness, the code changes seem fine.

Issue #589, opened 2016-01-12
VERSION_recipename env var functionality is not documented


inclement:
Issue opened as a reminder.
##

PR #1641, opened 2019-01-31
Remove old toolchain doc and add short note about overriding recipe sources

inclement:
Now is as good a time as any to ditch the old toolchain doc.

Adding the doc note fixes https://github.com/kivy/python-for-android/issues/589.


inclement:
@AndreMiras Can you approve and merge this? It doesn't touch any code/tests.

Issue #703, opened 2016-04-11
Need to set locale env variable for python3 package recipes


emillynge:
When calling `python3.5 setup.py` using sh.py, the setup can fail because the spawned process doesn't correctly set the locale to use UTF-8.

In the case of plyer, a changelog is read during setup that contains non-ascii symbols. Without correct locale, the python interpreter will try to decode the file upon  `open('changelog.md')` using ascii codec and raise an exception.

Setting the `LANG` enviroment variable to system locale solves the issue.

```
class PlyerRecipe(PythonRecipe):
    version = '1.2.5dev'
    url = 'https://github.com/emillynge/plyer/archive/browser.tar.gz'
    depends = ['pyjnius', 'python3crystax']

    def get_recipe_env(self, *args, **kwargs):
        env = super().get_recipe_env(*args, **kwargs)
        env['LANG'] = "en_GB.UTF-8"
        return env
```

Maybe this environment variable should be set by default by `PythonRecipe`?
##

inclement:
I don't see a problem with doing this, but I'll leave it a little while in case anyone raises any issues as I've seen this kind of encoding setting have unexpected side effects before.

PR #1642, opened 2019-01-31
Set $LANG in PythonRecipe

inclement:
Seems pretty benign so no reason not to do it.

Fixes https://github.com/kivy/python-for-android/issues/703.

### DIFF
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -764,6 +764,10 @@ class PythonRecipe(Recipe):
 
         env['PYTHONNOUSERSITE'] = '1'
 
+        # Set the LANG, this isn't usually important but is a better default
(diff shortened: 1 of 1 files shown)


pythonforandroid/recipe.py
```diff
@@ -764,6 +764,10 @@ def get_recipe_env(self, arch=None, with_flags_in_cc=True):
 
         env['PYTHONNOUSERSITE'] = '1'
 
+        # Set the LANG, this isn't usually important but is a better default
+        # as it occasionally matters how Python e.g. reads files
+        env['LANG'] = "en_GB.UTF-8"
```
AndreMiras: Why not `en_GB.UTF-8` :trollface:

PR #1646, opened 2019-01-31
Made the activity launch mode default to singleTask

inclement:
Fixes https://github.com/kivy/python-for-android/issues/1271.

Using singleTask seems appropriate to make sure that our activity is not recreated, so new intents are passed to the existing activity if possible. Doc at https://developer.android.com/guide/topics/manifest/activity-element.


inclement:
@AndreMiras Can you approve this? Although the tests failed due to flakiness, it doesn't touch anything that should affect the tests (unless some trivial syntax error).

PR #1651, opened 2019-02-01
Speed up Python and NumPy compilation process

AndreMiras:
Uses `-j` flag on target/host python as well as numpy to run compilation
in parallel. In Python an optimum number of jobs is guessed if the flag
is used alone. In NumPy it must be explicitly set.

### DIFF
--- a/pythonforandroid/python.py
+++ b/pythonforandroid/python.py
@@ -4,6 +4,7 @@ build our python3 and python2 recipes and his corresponding hostpython recipes.
 '''
 
 from os.path import dirname, exists, join
+from multiprocessing import cpu_count
 from shutil import copy2
 from os import environ
 import subprocess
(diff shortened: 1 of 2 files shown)


pythonforandroid/python.py
```diff
@@ -250,7 +250,7 @@ def build_arch(self, arch):
                 py_version = self.major_minor_version_string
                 if self.major_minor_version_string[0] == '3':
                     py_version += 'm'
-                shprint(sh.make, 'all',
+                shprint(sh.make, 'all', '-j',
```
KeyWeeUsr: Don't we need to add the `cpu_count` as an argument for `-j` flag?

pythonforandroid/python.py
```diff
@@ -420,7 +420,7 @@ def build_arch(self, arch):
                 shprint(sh.cp, join('Modules', 'Setup.dist'),
                         join(build_dir, 'Modules', 'Setup'))
 
-                result = shprint(sh.make, '-C', build_dir)
+                result = shprint(sh.make, '-j', '-C', build_dir)
```
KeyWeeUsr: Same here?

pythonforandroid/python.py
AndreMiras: As I wrote in my commit description, the number of jobs can be omitted in Python `make`, see for instance https://devguide.python.org/
But I will make it explicit if we prefer it more that way :+1:

pythonforandroid/python.py
AndreMiras: See my comment above

PR #1652, opened 2019-02-01
Speed up Docker chown via COPY parameter

AndreMiras:
For some reason running the `chown` command explicitly is hanging on
Docker. Using the `--chown` parameter in the `COPY` command is faster.
However that new method has a couple of drawbacks:
1) it was introduced in a recent Docker version 17+
2) it doesn't accept variables as argument, only hardcoded values

Also note that we don't need to chmod `ANDROID_SDK_HOME` anymore.


AndreMiras:
Build is green :confetti_ball: merging

PR #1653, opened 2019-02-01
WIP Update broken recipes list, refs #1514

AndreMiras:
Removed gevent, cffi and pycryptodome recipes from it since they now
compile fine. Also bumped few versions.


AndreMiras:
Merged via https://github.com/kivy/python-for-android/pull/1656

PR #1662, opened 2019-02-02
Stable

AndreMiras:


AndreMiras:
We close it for now, will see later if we kill stable

PR #1667, opened 2019-02-02
Fix setup.py install breaking due to unicode characters in README.md on Python 3

JonasT:
Fix setup.py install breaking due to unicode characters in README.md on Python 3

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,10 +1,11 @@
 
-from setuptools import setup, find_packages
+import glob
+from io import open  # for open(..,encoding=...) parameter in python 2
 from os import walk
 from os.path import join, dirname, sep
 import os
-import glob
 import re
+from setuptools import setup, find_packages
 
 # NOTE: All package data should also be set in MANIFEST.in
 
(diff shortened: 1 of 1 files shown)


AndreMiras:
Thanks for giving it a try!
But I think the fix actually is https://github.com/kivy/python-for-android/pull/1666 right?
Have you tried running `tox` on your changes locally?

JonasT:
@AndreMiras the fix for what? This fixes `setup.py` blowing up if unicode characters are in `README.md`, which they are right now

JonasT:
Ok, fix works for me (tested both Python 3 and Python 2 install). Since I forgot to post the backtrace and this came up in chat, this is what it fixes:

```
Collecting https://github.com/kivy/python-for-android/archive/master.zip
  Downloading https://github.com/kivy/python-for-android/archive/master.zip
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-gpvwn6h5/setup.py", line 56, in <module>
        long_description = fileh.read()
      File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
        return codecs.ascii_decode(input, self.errors)[0]
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xf0 in position 5032: ordinal not in range(128)
```

AndreMiras:
Thanks for clarifying and thanks for simplifying the initial PR

PR #1669, opened 2019-02-03
Fix various issues with graphs and recipes

JonasT:
- fix graph depending on capitalization of dependency/recipe names when pip doesn't
- fix graph giving no useful errors for conflicts (now it gives them sometimes)
- fix graph entering infinite loop when no bootstrap is found
- fix recipe not being obtainable irregardless of capitalization
- fix recipe IOError-on-doesn't-exist not distinguishable from
  recipe-exists-but-code-messes-up-and-causes-unintentional-IOError.
  a non-existent recipe will now cause a ValueError
- fix the bootstraps' own dependencies being wrong and unused
- change preferred default from `python2` + `sdl2` to `python3` + `sdl2`
- possibly other minor fixes I forgot about

(some of these fixes are required for #1625 / project's `setup.py` to be reasonably used)

### DIFF
--- a/pythonforandroid/bootstraps/service_only/__init__.py
+++ b/pythonforandroid/bootstraps/service_only/__init__.py
@@ -9,7 +9,9 @@ class ServiceOnlyBootstrap(Bootstrap):
 
     name = 'service_only'
 
-    recipe_depends = ['genericndkbuild', ('python2', 'python3', 'python3crystax')]
+    recipe_depends = list(
+        set(Bootstrap.recipe_depends).union({'genericndkbuild'})
+    )
 
     def run_distribute(self):
         info_main('# Creating Android project from build and {} bootstrap'.format(
--- a/pythonforandroid/bootstraps/webview/__init__.py
+++ b/pythonforandroid/bootstraps/webview/__init__.py
@@ -7,7 +7,9 @@ import sh
 class WebViewBootstrap(Bootstrap):
     name = 'webview'
 
-    recipe_depends = ['genericndkbuild', ('python2', 'python3', 'python3crystax')]
+    recipe_depends = list(
+        set(Bootstrap.recipe_depends).union({'genericndkbuild'})
+    )
 
     def run_distribute(self):
         info_main('# Creating Android project from build and {} bootstrap'.format(
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -11,7 +11,10 @@ import re
 import sh
 import subprocess
 
-from pythonforandroid.util import (ensure_dir, current_directory, BuildInterruptingException)
+from pythonforandroid.util import (
+    current_directory, ensure_dir, get_virtualenv_executable,
+    BuildInterruptingException
+)
 from pythonforandroid.logger import (info, warning, info_notify, info_main, shprint)
 from pythonforandroid.archs import ArchARM, ArchARMv7_a, ArchAarch_64, Archx86, Archx86_64
 from pythonforandroid.recipe import CythonRecipe, Recipe
@@ -305,13 +308,7 @@ class Context(object):
 
         check_ndk_api(ndk_api, self.android_api)
 
-        virtualenv = None
-        if virtualenv is None:
-            virtualenv = sh.which('virtualenv2')
-        if virtualenv is None:
-            virtualenv = sh.which('virtualenv-2.7')
-        if virtualenv is None:
-            virtualenv = sh.which('virtualenv')
+        virtualenv = get_virtualenv_executable()
         if virtualenv is None:
             raise IOError('Couldn\'t find a virtualenv executable, '
                           'you must install this to use p4a.')
@@ -509,7 +506,7 @@ class Context(object):
         # Try to look up recipe by name:
         try:
             recipe = Recipe.get_recipe(name, self)
-        except IOError:
+        except ValueError:
             pass
         else:
             name = getattr(recipe, 'site_packages_name', None) or name
@@ -618,14 +615,15 @@ def run_pymodules_install(ctx, modules):
                     line = '{}\n'.format(module)
                 fileh.write(line)
 
+        # Prepare base environment and upgrade pip:
         base_env = copy.copy(os.environ)
         base_env["PYTHONPATH"] = ctx.get_site_packages_dir()
-
         info('Upgrade pip to latest version')
         shprint(sh.bash, '-c', (
             "source venv/bin/activate && pip install -U pip"
         ), _env=copy.copy(base_env))
 
+        # Install Cython in case modules need it to build:
         info('Install Cython in case one of the modules needs it to build')
         shprint(sh.bash, '-c', (
             "venv/bin/pip install Cython"
@@ -648,15 +646,17 @@ def run_pymodules_install(ctx, modules):
              'changes / workarounds.')
 
         # Make sure our build package dir is available, and the virtualenv
-        # site packages come FIRST (for the proper pip version):
+        # site packages come FIRST (so the proper pip version is used):
         env["PYTHONPATH"] += ":" + ctx.get_site_packages_dir()
         env["PYTHONPATH"] = os.path.abspath(join(
             ctx.build_dir, "venv", "lib",
             "python" + ctx.python_recipe.major_minor_version_string,
             "site-packages")) + ":" + env["PYTHONPATH"]
+
+        # Do actual install:
         shprint(sh.bash, '-c', (
-            "source venv/bin/activate && " +
-            "pip install -v --target '{0}' --no-deps -r requirements.txt"
+            "venv/bin/pip " +
+            "install -v --target '{0}' --no-deps -r requirements.txt"
         ).format(ctx.get_site_packages_dir().replace("'", "'\"'\"'")),
                 _env=copy.copy(env))
 
--- a/pythonforandroid/graph.py
+++ b/pythonforandroid/graph.py
@@ -7,17 +7,31 @@ from pythonforandroid.bootstrap import Bootstrap
 from pythonforandroid.util import BuildInterruptingException
 
 
-class RecipeOrder(dict):
+def fix_deplist(deps):
+    """ Turn a dependency list into lowercase, and make sure all entries
+        that are just a string become a tuple of strings
+    """
+    deps = [
+        ((dep.lower(),)
+         if not isinstance(dep, (list, tuple))
+         else tuple([dep_entry.lower()
+                     for dep_entry in dep
+                    ]))
+        for dep in deps
+    ]
+    return deps
 
+
+class RecipeOrder(dict):
     def __init__(self, ctx):
         self.ctx = ctx
 
-    def conflicts(self, name):
+    def conflicts(self):
         for name in self.keys():
             try:
                 recipe = Recipe.get_recipe(name, self.ctx)
-                conflicts = recipe.conflicts
-            except IOError:
+                conflicts = [dep.lower() for dep in recipe.conflicts]
+            except ValueError:
                 conflicts = []
 
             if any([c in self for c in conflicts]):
@@ -25,31 +39,50 @@ class RecipeOrder(dict):
         return False
 
 
-def recursively_collect_orders(name, ctx, all_inputs, orders=[]):
+def get_dependency_tuple_list_for_recipe(recipe, blacklist=[]):
+    """ Get the dependencies of a recipe with filtered out blacklist, and
+        turned into tuples with fix_deplist()
+    """
+    if recipe.depends is None:
+        dependencies = []
+    else:
+        # Turn all dependencies into tuples so that product will work
+        dependencies = fix_deplist(recipe.depends)
+
+        # Filter out blacklisted items and turn lowercase:
+        dependencies = [
+            deptuple for deptuple in dependencies
+            if not set(deptuple).intersection(set(blacklist))
+        ]
+    return dependencies
+
+
+def recursively_collect_orders(
+        name, ctx, all_inputs, orders=[], blacklist=[]
+        ):
     '''For each possible recipe ordering, try to add the new recipe name
     to that order. Recursively do the same thing with all the
     dependencies of each recipe.
 
     '''
+    name = name.lower()
     try:
         recipe = Recipe.get_recipe(name, ctx)
-        if recipe.depends is None:
-            dependencies = []
-        else:
-            # make all dependencies into lists so that product will work
-            dependencies = [([dependency] if not isinstance(
-                dependency, (list, tuple))
-                            else dependency) for dependency in recipe.depends]
+        dependencies = get_dependency_tuple_list_for_recipe(
+            recipe, blacklist=blacklist
+        )
 
         # handle opt_depends: these impose requirements on the build
         # order only if already present in the list of recipes to build
-        dependencies.extend([[d] for d in recipe.get_opt_depends_in_list(all_inputs)])
+        dependencies.extend(fix_deplist(
+            [[d] for d in recipe.get_opt_depends_in_list(all_inputs)]
+        ))
 
         if recipe.conflicts is None:
             conflicts = []
         else:
-            conflicts = recipe.conflicts
-    except IOError:
+            conflicts = [dep.lower() for dep in recipe.conflicts]
+    except ValueError:
         # The recipe does not exist, so we assume it can be installed
         # via pip with no extra dependencies
         dependencies = []
@@ -61,7 +94,7 @@ def recursively_collect_orders(name, ctx, all_inputs, orders=[]):
         if name in order:
             new_orders.append(deepcopy(order))
             continue
-        if order.conflicts(name):
+        if order.conflicts():
             continue
         if any([conflict in order for conflict in conflicts]):
             continue
@@ -99,17 +132,131 @@ def find_order(graph):
                 bset.discard(result)
 
 
-def get_recipe_order_and_bootstrap(ctx, names, bs=None):
-    recipes_to_load = set(names)
-    if bs is not None and bs.recipe_depends:
-        recipes_to_load = recipes_to_load.union(set(bs.recipe_depends))
+def obvious_conflict_checker(ctx, name_tuples, blacklist=[]):
+    """ This is a pre-flight check function that will completely ignore
+        recipe order or choosing an actual value in any of the multiple
+        choice tuples/dependencies, and just do a very basic obvious
+        conflict check.
+    """
+    deps_were_added_by = dict()
+    deps = set()
 
-    possible_orders = []
+    # Add dependencies for all recipes:
+    to_be_added = [(name_tuple, None) for name_tuple in name_tuples]
+    while len(to_be_added) > 0:
+        current_to_be_added = list(to_be_added)
+        to_be_added = []
+        for (added_tuple, adding_recipe) in current_to_be_added:
+            assert(type(added_tuple) == tuple)
+            if len(added_tuple) > 1:
+                # No obvious commitment in what to add, don't check it itself
+                # but throw it into deps for later comparing against
+                # (Remember this function only catches obvious issues)
+                deps.add(added_tuple)
+                continue
+
+            name = added_tuple[0]
+            recipe_conflicts = set()
+            recipe_dependencies = []
+            try:
+                # Get recipe to add and who's ultimately adding it:
+                recipe = Recipe.get_recipe(name, ctx)
+                recipe_conflicts = {c.lower() for c in recipe.conflicts}
+                recipe_dependencies = get_dependency_tuple_list_for_recipe(
+                    recipe, blacklist=blacklist
+                )
+            except ValueError:
+                pass
+            adder_first_recipe_name = adding_recipe or name
+
+            # Collect the conflicts:
+            triggered_conflicts = []
+            for dep_tuple_list in deps:
+                # See if the new deps conflict with things added before:
+                if set(dep_tuple_list).intersection(
+                       recipe_conflicts) == set(dep_tuple_list):
+                    triggered_conflicts.append(dep_tuple_list)
+                    continue
+
+                # See if what was added before conflicts with the new deps:
+                if len(dep_tuple_list) > 1:
+                    # Not an obvious commitment to a specific recipe/dep
+                    # to be added, so we won't check.
+                    # (remember this function only catches obvious issues)
+                    continue
+                try:
+                    dep_recipe = Recipe.get_recipe(dep_tuple_list[0], ctx)
+                except ValueError:
+                    continue
+                conflicts = [c.lower() for c in dep_recipe.conflicts]
+                if name in conflicts:
+                    triggered_conflicts.append(dep_tuple_list)
+
+            # Throw error on conflict:
+            if triggered_conflicts:
+                # Get first conflict and see who added that one:
+                adder_second_recipe_name = "'||'".join(triggered_conflicts[0])
+                second_recipe_original_adder = deps_were_added_by.get(
+                    (adder_second_recipe_name,), None
+                )
+                if second_recipe_original_adder:
+                    adder_second_recipe_name = second_recipe_original_adder
+
+                # Prompt error:
+                raise BuildInterruptingException(
+                    "Conflict detected: '{}'"
+                    " inducing dependencies {}, and '{}'"
+                    " inducing conflicting dependencies {}".format(
+                        adder_first_recipe_name,
+                        (recipe.name,),
+                        adder_second_recipe_name,
+                        triggered_conflicts[0]
+                    ))
+
+            # Actually add it to our list:
+            deps.add(added_tuple)
+            deps_were_added_by[added_tuple] = adding_recipe
+
+            # Schedule dependencies to be added
+            to_be_added += [
+                (dep, adder_first_recipe_name or name)
+                for dep in recipe_dependencies
+                if dep not in deps
+            ]
+    # If we came here, then there were no obvious conflicts.
+    return None
+
+
+def get_recipe_order_and_bootstrap(ctx, names, bs=None, blacklist=[]):
+    # Get set of recipe/dependency names, clean up and add bootstrap deps:
+    names = set(names)
+    if bs is not None and bs.recipe_depends:
+        names = names.union(set(bs.recipe_depends))
+    names = fix_deplist([
+        ([name] if not isinstance(name, (list, tuple)) else name)
+        for name in names
+    ])
+    blacklist = [bitem.lower() for bitem in blacklist]
+
+    # Remove all values that are in the blacklist:
+    names_before_blacklist = list(names)
+    names = []
+    for name in names_before_blacklist:
+        cleaned_up_tuple = tuple([
+            item for item in name if item not in blacklist
+        ])
+        if cleaned_up_tuple:
+            names.append(cleaned_up_tuple)
+
+    # Do check for obvious conflicts (that would trigger in any order, and
+    # without comitting to any specific choice in a multi-choice tuple of
+    # dependencies):
+    obvious_conflict_checker(ctx, names, blacklist=blacklist)
+    # If we get here, no obvious conflicts!
 
     # get all possible order graphs, as names may include tuples/lists
     # of alternative dependencies
-    names = [([name] if not isinstance(name, (list, tuple)) else name)
-             for name in names]
+    possible_orders = []
     for name_set in product(*names):
         new_possible_orders = [RecipeOrder(ctx)]
         for name in name_set:
@@ -128,13 +275,14 @@ def get_recipe_order_and_bootstrap(ctx, names, bs=None):
             continue
         orders.append(list(order))
 
-    # prefer python2 and SDL2 if available
+    # prefer python3 and SDL2 if available
     orders = sorted(orders,
-                    key=lambda order: -('python2' in order) - ('sdl2' in order))
+                    key=lambda order: -('python3' in order) - ('sdl2' in order))
 
     if not orders:
         raise BuildInterruptingException(
-            'Didn\'t find any valid dependency graphs. This means that some of your '
+            'Didn\'t find any valid dependency graphs. '
+            'This means that some of your '
             'requirements pull in conflicting dependencies.')
 
     # It would be better to check against possible orders other
@@ -151,6 +299,11 @@ def get_recipe_order_and_bootstrap(ctx, names, bs=None):
 
     if bs is None:
         bs = Bootstrap.get_bootstrap_from_recipes(chosen_order, ctx)
+        if bs is None:
+            # Note: don't remove this without thought, causes infinite loop
+            raise BuildInterruptingException(
+                "Could not find any compatible bootstrap!"
+            )
         recipes, python_modules, bs = get_recipe_order_and_bootstrap(
             ctx, chosen_order, bs=bs)
     else:
@@ -161,7 +314,7 @@ def get_recipe_order_and_bootstrap(ctx, names, bs=None):
             try:
                 recipe = Recipe.get_recipe(name, ctx)
                 python_modules += recipe.python_depends
-            except IOError:
+            except ValueError:
                 python_modules.append(name)
             else:
                 recipes.append(name)
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -574,6 +574,7 @@ class Recipe(with_metaclass(RecipeMeta)):
     @classmethod
     def get_recipe(cls, name, ctx):
         '''Returns the Recipe with the given name, if it exists.'''
+        name = name.lower()
         if not hasattr(cls, "recipes"):
             cls.recipes = {}
         if name in cls.recipes:
@@ -581,20 +582,28 @@ class Recipe(with_metaclass(RecipeMeta)):
 
         recipe_file = None
         for recipes_dir in cls.recipe_dirs(ctx):
-            recipe_file = join(recipes_dir, name, '__init__.py')
-            if exists(recipe_file):
+            if not exists(recipes_dir):
+                continue
+            # Find matching folder (may differ in case):
+            for subfolder in listdir(recipes_dir):
+                if subfolder.lower() == name:
+                    recipe_file = join(recipes_dir, subfolder, '__init__.py')
+                    if exists(recipe_file):
+                        name = subfolder  # adapt to actual spelling
+                        break
+                    recipe_file = None
+            if recipe_file is not None:
                 break
-            recipe_file = None
 
         if not recipe_file:
-            raise IOError('Recipe does not exist: {}'.format(name))
+            raise ValueError('Recipe does not exist: {}'.format(name))
 
         mod = import_recipe('pythonforandroid.recipes.{}'.format(name), recipe_file)
         if len(logger.handlers) > 1:
             logger.removeHandler(logger.handlers[1])
         recipe = mod.recipe
         recipe.ctx = ctx
-        cls.recipes[name] = recipe
+        cls.recipes[name.lower()] = recipe
         return recipe
 
 
--- a/tests/test_graph.py
+++ b/tests/test_graph.py
@@ -1,24 +1,74 @@
 from pythonforandroid.build import Context
-from pythonforandroid.graph import get_recipe_order_and_bootstrap
+from pythonforandroid.graph import (
+    fix_deplist, get_recipe_order_and_bootstrap, obvious_conflict_checker,
+)
 from pythonforandroid.bootstrap import Bootstrap
+from pythonforandroid.recipe import Recipe
 from pythonforandroid.util import BuildInterruptingException
 from itertools import product
 
+import mock
 import pytest
-
 
 ctx = Context()
 
 name_sets = [['python2'],
              ['kivy']]
 bootstraps = [None,
-              Bootstrap.get_bootstrap('pygame', ctx),
               Bootstrap.get_bootstrap('sdl2', ctx)]
 valid_combinations = list(product(name_sets, bootstraps))
 valid_combinations.extend(
     [(['python3crystax'], Bootstrap.get_bootstrap('sdl2', ctx)),
-     (['kivy', 'python3crystax'], Bootstrap.get_bootstrap('sdl2', ctx))])
-invalid_combinations = [[['python2', 'python3crystax'], None]]
+     (['kivy', 'python3crystax'], Bootstrap.get_bootstrap('sdl2', ctx)),
+     (['flask'], Bootstrap.get_bootstrap('webview', ctx)),
+     (['pysdl2'], None),  # auto-detect bootstrap! important corner case
+    ]
+)
+invalid_combinations = [
+    [['python2', 'python3crystax'], None],
+    [['pysdl2', 'genericndkbuild'], None],
+]
+invalid_combinations_simple = list(invalid_combinations)
+# NOTE !! keep in mind when setting invalid_combinations_simple:
+#
+# This is used to test obvious_conflict_checker(), which only
+# catches CERTAIN conflicts:
+#
+# This must be a list of conflicts where the conflict is ONLY in
+# non-tuple/non-ambiguous dependencies, e.g.:
+#
+#     dependencies_1st = ["python2", "pillow"]
+#     dependencies_2nd = ["python3crystax", "pillow"]
+#
+# This however won't work:
+#
+#     dependencies_1st = [("python2", "python3"), "pillow"]
+#     dependencies_2nd = [("python2legacy", "python3crystax"), "pillow"]
+#
+# (This is simply because the conflict checker doesn't resolve this to
+# keep the code ismple enough)
+
+
+def get_fake_recipe(name, depends=None, conflicts=None):
+    recipe = mock.Mock()
+    recipe.name = name
+    recipe.get_opt_depends_in_list = lambda: []
+    recipe.get_dir_name = lambda: name
+    recipe.depends = list(depends or [])
+    recipe.conflicts = list(conflicts or [])
+    return recipe
+
+
+def register_fake_recipes_for_test(monkeypatch, recipe_list):
+    _orig_get_recipe = Recipe.get_recipe
+
+    def mock_get_recipe(name, ctx):
+        for recipe in recipe_list:
+            if recipe.name == name:
+                return recipe
+        return _orig_get_recipe(name, ctx)
+    # Note: staticmethod() needed for python ONLY, don't ask me why:
+    monkeypatch.setattr(Recipe, 'get_recipe', staticmethod(mock_get_recipe))
 
 
 @pytest.mark.parametrize('names,bootstrap', valid_combinations)
@@ -30,16 +80,114 @@ def test_valid_recipe_order_and_bootstrap(names, bootstrap):
 def test_invalid_recipe_order_and_bootstrap(names, bootstrap):
     with pytest.raises(BuildInterruptingException) as e_info:
         get_recipe_order_and_bootstrap(ctx, names, bootstrap)
-    assert e_info.value.message == (
-        "Didn't find any valid dependency graphs. "
-        "This means that some of your requirements pull in conflicting dependencies."
+    assert "conflict" in e_info.value.message.lower()
+
+
+@pytest.mark.parametrize('names,bootstrap', valid_combinations)
+def test_valid_obvious_conflict_checker(names, bootstrap):
+    # Note: obvious_conflict_checker is stricter on input
+    # (needs fix_deplist) than get_recipe_order_and_bootstrap!
+    obvious_conflict_checker(ctx, fix_deplist(names))
+
+
+@pytest.mark.parametrize('names,bootstrap',
+                         invalid_combinations_simple  # see above for why this
+                        )                             # is a separate list
+def test_invalid_obvious_conflict_checker(names, bootstrap):
+    # Note: obvious_conflict_checker is stricter on input
+    # (needs fix_deplist) than get_recipe_order_and_bootstrap!
+    with pytest.raises(BuildInterruptingException) as e_info:
+        obvious_conflict_checker(ctx, fix_deplist(names))
+    assert "conflict" in e_info.value.message.lower()
+
+
+def test_misc_obvious_conflict_checker(monkeypatch):
+    # Check that the assert about wrong input data is hit:
+    with pytest.raises(AssertionError) as e_info:
+        obvious_conflict_checker(
+            ctx,
+            ["this_is_invalid"]
+            # (invalid because it isn't properly nested as tuple)
+        )
+
+    # Test that non-recipe dependencies work in overall:
+    obvious_conflict_checker(
+        ctx, fix_deplist(["python3", "notarecipelibrary"])
     )
+
+    # Test that a conflict with a non-recipe dependency works:
+    # This is currently not used, so we need a custom test recipe:
+    # To get that, we simply modify one!
+    with monkeypatch.context() as m:
+        register_fake_recipes_for_test(m, [
+            get_fake_recipe("recipe1", conflicts=[("fakelib")]),
+        ])
+        with pytest.raises(BuildInterruptingException) as e_info:
+            obvious_conflict_checker(ctx, fix_deplist(["recipe1", "fakelib"]))
+        assert "conflict" in e_info.value.message.lower()
+
+    # Test a case where a recipe pulls in a conditional tuple
+    # of additional dependencies. This is e.g. done for ('python3',
+    # 'python2', ...) but most recipes don't depend on this anymore,
+    # so we need to add a manual test for this case:
+    with monkeypatch.context() as m:
+        register_fake_recipes_for_test(m, [
+            get_fake_recipe("recipe1", depends=[("libffi", "Pillow")]),
+        ])
+        obvious_conflict_checker(ctx, fix_deplist(["recipe1"]))
+
+
+def test_indirectconflict_obvious_conflict_checker(monkeypatch):
+    # Test a case where there's an indirect conflict, which also
+    # makes sure the error message correctly blames the OUTER recipes
+    # as original conflict source:
+    with monkeypatch.context() as m:
+        register_fake_recipes_for_test(m, [
+            get_fake_recipe("outerrecipe1", depends=["innerrecipe1"]),
+            get_fake_recipe("outerrecipe2", depends=["innerrecipe2"]),
+            get_fake_recipe("innerrecipe1"),
+            get_fake_recipe("innerrecipe2", conflicts=["innerrecipe1"]),
+        ])
+        with pytest.raises(BuildInterruptingException) as e_info:
+            obvious_conflict_checker(
+                ctx,
+                fix_deplist(["outerrecipe1", "outerrecipe2"])
+            )
+        assert ("conflict" in e_info.value.message.lower() and
+                "outerrecipe1" in e_info.value.message.lower() and
+                "outerrecipe2" in e_info.value.message.lower())
+
+
+def test_multichoice_obvious_conflict_checker(monkeypatch):
+    # Test a case where there's a conflict with a multi-choice tuple:
+    with monkeypatch.context() as m:
+        register_fake_recipes_for_test(m, [
+            get_fake_recipe("recipe1", conflicts=["lib1", "lib2"]),
+            get_fake_recipe("recipe2", depends=[("lib1", "lib2")]),
+        ])
+        with pytest.raises(BuildInterruptingException) as e_info:
+            obvious_conflict_checker(
+                ctx,
+                fix_deplist([("lib1", "lib2"), "recipe1"])
+            )
+        assert "conflict" in e_info.value.message.lower()
 
 
 def test_bootstrap_dependency_addition():
     build_order, python_modules, bs = get_recipe_order_and_bootstrap(
         ctx, ['kivy'], None)
     assert (('hostpython2' in build_order) or ('hostpython3' in build_order))
+
+
+def test_graph_deplist_transformation():
+    test_pairs = [
+        (["Pillow", ('python2', 'python3')],
+         [('pillow',), ('python2', 'python3')]),
+        (["Pillow", ('python2',)],
+         [('pillow',), ('python2',)]),
+    ]
+    for (before_list, after_list) in test_pairs:
+        assert fix_deplist(before_list) == after_list
 
 
 def test_bootstrap_dependency_addition2():
--- a/pythonforandroid/bootstrap.py
+++ b/pythonforandroid/bootstrap.py
@@ -49,7 +49,10 @@ class Bootstrap(object):
     dist_name = None
     distribution = None
 
-    recipe_depends = ['sdl2']
+    # All bootstraps should include Python in some way:
+    recipe_depends = [
+        ("python2", "python2legacy", "python3", "python3crystax"),
+    ]
 
     can_be_chosen_automatically = True
     '''Determines whether the bootstrap can be chosen as one that
@@ -167,7 +170,7 @@ class Bootstrap(object):
                 for recipe in recipes:
                     try:
                         recipe = Recipe.get_recipe(recipe, ctx)
-                    except IOError:
+                    except ValueError:
                         conflicts = []
                     else:
                         conflicts = recipe.conflicts
@@ -175,7 +178,7 @@ class Bootstrap(object):
                             for conflict in conflicts]):
                         ok = False
                         break
-                if ok:
+                if ok and bs not in acceptable_bootstraps:
                     acceptable_bootstraps.append(bs)
         info('Found {} acceptable bootstraps: {}'.format(
             len(acceptable_bootstraps),
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -566,7 +566,8 @@ tools directory of the Android SDK.
 
     # --private is required unless for sdl2, where there's also --launcher
     ap.add_argument('--private', dest='private',
-                    help='the dir of user files',
+                    help='the directory with the app source code files' +
+                         ' (containing your main.py entrypoint)',
                     required=(get_bootstrap_name() != "sdl2"))
     ap.add_argument('--package', dest='package',
                     help=('The name of the java package the project will be'
--- a/pythonforandroid/bootstraps/sdl2/__init__.py
+++ b/pythonforandroid/bootstraps/sdl2/__init__.py
@@ -8,7 +8,9 @@ import sh
 class SDL2GradleBootstrap(Bootstrap):
     name = 'sdl2'
 
-    recipe_depends = ['sdl2']
+    recipe_depends = list(
+        set(Bootstrap.recipe_depends).union({'sdl2'})
+    )
 
     def run_distribute(self):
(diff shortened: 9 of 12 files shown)


pythonforandroid/build.py
```diff
@@ -305,13 +306,7 @@ def prepare_build_environment(self,
...
         check_ndk_api(ndk_api, self.android_api)
 
-        virtualenv = None
-        if virtualenv is None:
-            virtualenv = sh.which('virtualenv2')
-        if virtualenv is None:
-            virtualenv = sh.which('virtualenv-2.7')
-        if virtualenv is None:
-            virtualenv = sh.which('virtualenv')
+        virtualenv = get_virtualenv_executable()
```
AndreMiras: I like that you moved this somewhere else :+1:

AndreMiras:
Thanks for making a dedicated one for it. We'll review it more seriously later on.
Did you see that the `tox` test broke?

pythonforandroid/bootstraps/webview/__init__.py
```diff
@@ -7,7 +7,9 @@
 class WebViewBootstrap(Bootstrap):
     name = 'webview'
 
-    recipe_depends = ['genericndkbuild', ('python2', 'python3', 'python3crystax')]
+    recipe_depends = list(
+        set(Bootstrap.recipe_depends).union(set(['genericndkbuild']))
```
AndreMiras: won't it be a little bit more concise/Pythonic to use literals and operators?
Something like:
```python
recipe_depends = list(set(Bootstrap.recipe_depends) | {'genericndkbuild'})
```

JonasT:
@AndreMiras thanks I had to fix some of the unit tests, it should work now!

pythonforandroid/bootstraps/webview/__init__.py
AndreMiras: I haven't checked if `|` is recommended over the `union`, it was just a matter of taste.
So feel free to use `union` if you prefer until I prove PEP8 says otherwise :wink:

pythonforandroid/build.py
```diff
@@ -587,7 +581,7 @@ def build_recipes(build_order, python_modules, ctx):
     return
 
 
-def run_pymodules_install(ctx, modules):
+def run_pymodules_install(ctx, modules, project_dir):
```
AndreMiras: with that new signature, the integration test are failing at the moment. Could you update to match?
```
  File "setup_testapp_python3_sqlite_openssl.py", line 23, in <module>
    package_data={'testapp_sqlite_openssl': ['*.py', '*.png']}
  File "/usr/lib/python3.6/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/usr/lib/python3.6/distutils/dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python3.6/distutils/dist.py", line 974, in run_command
    cmd_obj.run()
  File "/home/user/pythonforandroid/bdistapk.py", line 80, in run
    main()
  File "/home/user/pythonforandroid/toolchain.py", line 1051, in main
    ToolchainCL()
  File "/home/user/pythonforandroid/toolchain.py", line 560, in __init__
  File "/home/user/pythonforandroid/toolchain.py", line 151, in wrapper_func
  File "/home/user/pythonforandroid/toolchain.py", line 192, in build_dist_from_args
  File "/home/user/pythonforandroid/build.py", line 579, in build_recipes
    run_pymodules_install(ctx, python_modules)
TypeError: run_pymodules_install() missing 1 required positional argument: 'project_dir'
travis_time:end:0195f4fa:start=1549223130899523065,finish=1549223892350143515,duration=761450620450
[0K[31;1mThe command "docker run p4a /bin/sh -c "$COMMAND"" exited with 1.
```

JonasT: @AndreMiras oops, a leftover from the setup.py part I forgot to remove

pythonforandroid/bootstraps/service_only/__init__.py
```diff
@@ -9,7 +9,9 @@ class ServiceOnlyBootstrap(Bootstrap):
 
     name = 'service_only'
 
-    recipe_depends = ['genericndkbuild', ('python2', 'python3', 'python3crystax')]
+    recipe_depends = list(
+        set(Bootstrap.recipe_depends).union({'genericndkbuild'})
```
AndreMiras: wait that's not the same right?
```python
>>> list(set(['python2', 'python3', 'python3crystax']).union({'genericndkbuild'}))
['python2', 'python3crystax', 'python3', 'genericndkbuild']
```
Before we had the following result:
```python
['genericndkbuild', ('python2', 'python3', 'python3crystax')]
```
See how it was a two dimensional list I guess it was on purpose. Did you purposely flatten it? Won't it break some features then?

pythonforandroid/graph.py
```diff
@@ -7,49 +7,79 @@
...
+def fix_deplist(deps):
+    """ Turn a dependency list into lowercase, and make sure all entries
+        that are just a string become a tuple of strings
+    """
+    deps = [
+        ((dep.lower(),) if not isinstance(dep, (list, tuple))
+         else tuple([dep_entry.lower() for dep_entry in dep]))
+        for dep in deps
+    ]
+    return deps
```
AndreMiras: Minor: these isolated utils functions are easy to unit test. Even if they may look trivial, I think next time they deserve some tests, what do you think? Please skip it for now, but I'll slowly try to increase the coverage of p4a and may introduce some coverage check at some point :smiling_imp:

pythonforandroid/build.py
```diff
@@ -11,7 +11,7 @@
 import sh
 import subprocess
 
-from pythonforandroid.util import (ensure_dir, current_directory, BuildInterruptingException)
+from pythonforandroid.util import (current_directory, ensure_dir, get_virtualenv_executable, BuildInterruptingException)
```
KeyWeeUsr: Pep8:
```
import (
    stuff
)
```

pythonforandroid/graph.py
```diff
@@ -7,49 +7,79 @@
...
+            except ValueError:
                 conflicts = []
 
             if any([c in self for c in conflicts]):
                 return True
         return False
 
 
-def recursively_collect_orders(name, ctx, all_inputs, orders=[]):
+def get_dependency_tuple_list_for_recipe(recipe, blacklist=[]):
```
AndreMiras: Using mutable default parameters is discouraged in Python https://docs.python-guide.org/writing/gotchas/
Could you double check we're safe here and other places? As a rule of thumb I would just recommend to eventually use e.g. `blacklist=None`

pythonforandroid/build.py
```diff
@@ -649,15 +644,17 @@ def run_pymodules_install(ctx, modules):
...
         env["PYTHONPATH"] = os.path.abspath(join(
             ctx.build_dir, "venv", "lib",
             "python" + ctx.python_recipe.major_minor_version_string,
             "site-packages")) + ":" + env["PYTHONPATH"]
+
+        # Do actual install:
         shprint(sh.bash, '-c', (
-            "source venv/bin/activate && " +
-            "pip install -v --target '{0}' --no-deps -r requirements.txt"
+            "venv/bin/pip " +
```
KeyWeeUsr: `source` sets paths for Python via the activating scripts. Why is the removal necessary? It doesn't seem correct to me.

pythonforandroid/graph.py
```diff
@@ -7,49 +7,79 @@
...
 
 
-class RecipeOrder(dict):
+def fix_deplist(deps):
+    """ Turn a dependency list into lowercase, and make sure all entries
+        that are just a string become a tuple of strings
+    """
+    deps = [
+        ((dep.lower(),) if not isinstance(dep, (list, tuple))
+         else tuple([dep_entry.lower() for dep_entry in dep]))
```
KeyWeeUsr: Instead of a single space use this syntax, which is more readable:

```
stuff
if cond
else stuff
for val in iter
```

pythonforandroid/graph.py
```diff
@@ -7,49 +7,79 @@
...
+    if recipe.depends is None:
+        dependencies = []
+    else:
+        # Turn all dependencies into tuples so that product will work
+        dependencies = fix_deplist(recipe.depends)
+
+        # Filter out blacklisted items and turn lowercase:
+        dependencies = [
+            deptuple for deptuple in dependencies
+            if len(set(deptuple).intersection(set(blacklist))) == 0
```
KeyWeeUsr: `if not len(...)` or even `if not ...` only.

pythonforandroid/graph.py
```diff
@@ -99,17 +129,125 @@ def find_order(graph):
...
+                continue
+
+            name = added_tuple[0]
+            recipe_conflicts = []
+            recipe_dependencies = []
+            try:
+                # Get recipe to add and who's ultimately adding it:
+                recipe = Recipe.get_recipe(name, ctx)
+                recipe_conflicts = [c.lower() for c in recipe.conflicts]
+                recipe_dependencies = \
```
KeyWeeUsr: Python is not bash. :)

pythonforandroid/graph.py
```diff
@@ -99,17 +129,125 @@ def find_order(graph):
...
+                    get_dependency_tuple_list_for_recipe(
+                        recipe, blacklist=blacklist
+                    )
+            except ValueError:
+                pass
+            adder_recipe_1_name = adding_recipe or name
```
KeyWeeUsr: Number in a variable name? Don't do that :D

pythonforandroid/graph.py
```diff
@@ -99,17 +129,125 @@ def find_order(graph):
...
+                    conflicts = [c.lower() for c in
+                                 Recipe.get_recipe(dep_tuple_list[0], ctx).
+                                 conflicts]
+                    if name in conflicts:
+                        triggered_conflicts.append(dep_tuple_list)
+                except ValueError:
+                    pass
+
+            # Throw error on conflict:
+            if len(triggered_conflicts) > 0:
```
AndreMiras: Minor: I think the pythonic way to (consistently) check for non-empty list is to be implicit `if triggered_conflicts:`

pythonforandroid/graph.py
```diff
@@ -99,17 +129,125 @@ def find_order(graph):
...
+                    )
+            except ValueError:
+                pass
+            adder_recipe_1_name = adding_recipe or name
+
+            # See if this recipe itself conflicts with what was added before:
+            triggered_conflicts = []
+            for dep_tuple_list in deps:
+                if len(set(dep_tuple_list).intersection(
+                        set(recipe_conflicts))) == len(dep_tuple_list):
```
KeyWeeUsr: Multiple evaluation of recipe_conflicts is not necessary, bake it before for loop and then you might have a condition on a single line.

pythonforandroid/graph.py
```diff
@@ -99,17 +129,125 @@ def find_order(graph):
...
+                     adder_recipe_2_name)  # might be in dict, but None
+                # Prompt error:
+                raise BuildInterruptingException(
+                    "Conflict detected:" +
+                    " '" + str(adder_recipe_1_name) + "'" +
+                    " inducing dependencies " + str((recipe.name,)) +
+                    " and '" + str(adder_recipe_2_name) + "'" +
+                    " inducing conflicting dependencies " +
+                    str(triggered_conflicts[0])
+                )
```
AndreMiras: Minor: isn't `format()` producing more beautiful/readable code or is it more a matter of taste?

pythonforandroid/graph.py
```diff
@@ -99,17 +129,125 @@ def find_order(graph):
...
+
+            # See if what was added before conflicts with this recipe:
+            for dep_tuple_list in deps:
+                if len(dep_tuple_list) > 1:
+                    # Not an obvious commitment, so we won't check.
+                    # (remember this function only catches obvious issues)
+                    continue
+                try:
+                    conflicts = [c.lower() for c in
+                                 Recipe.get_recipe(dep_tuple_list[0], ctx).
```
KeyWeeUsr: Put the get_recipe value to variable and make the comprehension simpler. :)

pythonforandroid/graph.py
```diff
@@ -99,17 +129,125 @@ def find_order(graph):
...
+                                 conflicts]
+                    if name in conflicts:
+                        triggered_conflicts.append(dep_tuple_list)
+                except ValueError:
+                    pass
```
KeyWeeUsr: No. Check beforehand instead of evaluating and waiting for exception.

pythonforandroid/graph.py
KeyWeeUsr: if triggered_conflicts ...

pythonforandroid/graph.py
```diff
@@ -99,17 +129,125 @@ def find_order(graph):
...
+        recipe order or choosing an actual value in any of the multiple
+        choice tuples/dependencies, and just do a very basic obvious
+        conflict check.
+    """
+    deps_were_added_by = dict()
+    deps = set()
+
+    # Add dependencies for all recipes:
+    to_be_added = [(name_tuple, None) for name_tuple in name_tuples]
+    while len(to_be_added) > 0:
```
AndreMiras: Minor: that loop is insanely large AND deep :cry: 
That function looks pretty complex in terms of cyclomatic complexity for "just do a very basic obvious conflict check"

pythonforandroid/graph.py
```diff
@@ -99,17 +129,125 @@ def find_order(graph):
...
+                    if name in conflicts:
+                        triggered_conflicts.append(dep_tuple_list)
+                except ValueError:
+                    pass
+
+            # Throw error on conflict:
+            if len(triggered_conflicts) > 0:
+                # Get first conflict and see who added that one:
+                adder_recipe_2_name = "'||'".join(triggered_conflicts[0])
+                adder_recipe_2_name =\
```
KeyWeeUsr: Python is not bash :)

pythonforandroid/graph.py
```diff
@@ -99,17 +129,125 @@ def find_order(graph):
...
+                    (deps_were_added_by.get(adder_recipe_2_name,
+                                            adder_recipe_2_name) or
+                     adder_recipe_2_name)  # might be in dict, but None
+                # Prompt error:
+                raise BuildInterruptingException(
+                    "Conflict detected:" +
```
KeyWeeUsr: Use a proper formatting with format()

pythonforandroid/graph.py
```diff
@@ -99,17 +129,125 @@ def find_order(graph):
...
 
-    possible_orders = []
+    # Remove all values that are in the blacklist:
+    names_before_blacklist = list(names)
+    names = []
+    for name in names_before_blacklist:
+        cleaned_up_tuple = tuple([
+            item for item in name if item not in blacklist
+        ])
+        if len(cleaned_up_tuple) > 0:
```
AndreMiras: Minor: here and other place let's be Pythonic, for once the PEP8 allows us to be implicit, let's be :wink: 
```python
if not cleaned_up_tuple:
```

pythonforandroid/graph.py
```diff
@@ -133,13 +271,14 @@ def get_recipe_order_and_bootstrap(ctx, names, bs=None):
             raise
         orders.append(list(order))
 
-    # prefer python2 and SDL2 if available
+    # prefer python3 and SDL2 if available
     orders = sorted(orders,
-                    key=lambda order: -('python2' in order) - ('sdl2' in order))
+                    key=lambda order: -('python3' in order) - ('sdl2' in order))
```
AndreMiras: Nice! I've been waiting this one for a long time and never dared to touch it :smile: 
I remember I had to update the default in the conditional rebuild script however 92fde71d1621c4428b509731cf4d78ddb61b7acd

pythonforandroid/graph.py
```diff
@@ -133,13 +271,14 @@ def get_recipe_order_and_bootstrap(ctx, names, bs=None):
...
 
     if not orders:
         raise BuildInterruptingException(
-            'Didn\'t find any valid dependency graphs. This means that some of your '
+            'Didn\'t find any valid dependency graphs. '
+            'This means that some of your '
             'requirements pull in conflicting dependencies.')
```
AndreMiras: Thanks for sneaking `E501` on some of your pull requests :pray:
I actually also respect it on my projects, but p4a had different history so far :smile:

pythonforandroid/recipe.py
```diff
@@ -574,27 +574,36 @@ def list_recipes(cls, ctx):
     @classmethod
     def get_recipe(cls, name, ctx):
         '''Returns the Recipe with the given name, if it exists.'''
+        name = name.lower()
         if not hasattr(cls, "recipes"):
             cls.recipes = {}
-        if name in cls.recipes:
+        if name in cls.recipes.keys():
```
AndreMiras: Minor open question: is there any PEP recommandation that recommends being explicit on that specific case or is it just a generic "explicit over implicit"? :thinking:

tests/test_graph.py
```diff
@@ -12,28 +12,34 @@
...
+    try:
+        get_recipe_order_and_bootstrap(ctx, names, bootstrap)
+    except Exception as e:
+        # pytest output is too terse, so do something better here:
+        print(
+            "test_valid_recipe_order_and_bootstrap FAILED " +
+            "with these specific parameters: " +
+            str((ctx, names, bootstrap))
+        )
+        raise e
```
AndreMiras: That test looks a bit confusing now isn't it? Is this all just to make the failing test output easier to read on crash?
I would prefer to avoid that, specially being that broad catching `Exception` what if the code as a `SyntaxError` for instance...

tests/test_graph.py
```diff
@@ -12,28 +12,34 @@
...
 
 @pytest.mark.parametrize('names,bootstrap', invalid_combinations)
 def test_invalid_recipe_order_and_bootstrap(names, bootstrap):
     with pytest.raises(BuildInterruptingException) as e_info:
         get_recipe_order_and_bootstrap(ctx, names, bootstrap)
-    assert e_info.value.message == (
-        "Didn't find any valid dependency graphs. "
-        "This means that some of your requirements pull in conflicting dependencies."
-    )
+    assert e_info.value.message.lower().find("conflict") >= 0
```
AndreMiras: I understand you changed it so we don't have to update it too often whenever we change the wordings here, but well I think it's better to stay explicit or at least increase the context. Because just grepping `conflict` when it fails won't help that much and also reading the test feels a bit like "what's happening?"
And by the way why using `find("conflict")` over `"conflict" in` notation?

pythonforandroid/build.py
AndreMiras: I suggested it because it works perfect and it's way more concise, but well you guys do as you want then

pythonforandroid/build.py
AndreMiras: I haven't enforced that one in tox yet (via `isort` + `E501`), but I do have it on my personal projects

pythonforandroid/graph.py
AndreMiras: I actually like it. Isn't Python usually pushing for `EAFP` https://docs.python.org/3/glossary.html ?
But that's up to you guys :+1:

pythonforandroid/recipe.py
hackalog: PEP 234 does. https://www.python.org/dev/peps/pep-0234/#dictionary-iterators

the only time I'd use the explicit `keys()` is if I was looping over the keys and modifying somehting (`for x in list(dictname.keys())`), but in this case, the shorter version is notionally clearer, PEP compliant, and very likely faster.

pythonforandroid/build.py
KeyWeeUsr: Pip  can use the environment. If the environment is wrong, pip will behave wrong, that's why. :)

pythonforandroid/graph.py
KeyWeeUsr: It's in 3 loops and calls a loop. Any such code shape will slow everything down. Check for what you expect to be wrong and then move to the other iteration.

pythonforandroid/build.py
hackalog: I remember seeing someone mention recently that invoking the python binary inside the venv is the same as invoking the venv first then running python. Could it also be true for pip?

Reference here: https://news.ycombinator.com/item?id=18613891

pythonforandroid/graph.py
hackalog: FWIW, my personal preference is`format()`. `format` is a nice migration path to fstrings (the *most* readable option), for the day when python 3.5 and earlier support is dropped.

If there are too many arguments to format to make it clear what is what, I'll sometimes pass an unpacked dictionary (e.g. `**format_opts`)and use the key names explicitly, e.g.   `'and {key1} inducing conlicting dependencies {key2}'.format(**fmt_opts)`

pythonforandroid/graph.py
```diff
@@ -99,17 +129,125 @@ def find_order(graph):
                 bset.discard(result)
 
 
-def get_recipe_order_and_bootstrap(ctx, names, bs=None):
-    recipes_to_load = set(names)
+def obvious_conflict_checker(ctx, name_tuples, blacklist=[]):
```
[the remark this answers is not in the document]
hackalog: Breakage in graph resolution are insidious bugs (witness the recent openssl/requirements ordering issues). I'd be far more comfortable if any new code came with tests

pythonforandroid/graph.py
hackalog: I'd say it's more than discouraged. Please don't ever use `[]` as a default argument.

See, e.g. : https://docs.python-guide.org/writing/gotchas/

pythonforandroid/build.py
AndreMiras: Yes I've always been doing this way and never had the issue the most import thing we're sourcing from the `activate` script is the `PATH` the rest is mostly playing with the `PS1`. But if @KeyWeeUsr prefers it the verbose way I'm also fine

AndreMiras: It looks like @tshirtman would also agree https://github.com/kivy/python-for-android/commit/8f1c25fac7c5aa24721d3e004b8f2dfaf74571f1 :stuck_out_tongue_winking_eye:

tests/test_graph.py
JonasT: > Because just grepping conflict when it fails won't help that much and also reading the test feels a bit like "what's happening"?

maybe I should just add a comment then?

Adding a fixed message is simply no longer possible, I could if at all grep for a specific sentence, but the function will no longer return a static text

>  And by the way why using find("conflict") over "conflict" in notation?

Because I'm silly :smile: I'll change it :+1:

AndreMiras: Yes a comment is also fine :+1: 
I'm missing where the message is no longer static, it looks pretty static to me or it's another message?
https://github.com/kivy/python-for-android/blob/a0ef2a09523ce74b877190ba4957e977234d962c/pythonforandroid/graph.py#L279

pythonforandroid/graph.py
JonasT: @AndreMiras @hackalog the parameter is never changed, it is a read-only parameter everywhere except for the initial `get_recipe_order_and_bootstrap` where I lowercase it at the top. `blacklist=None` would need me to add conditionals to all the subfunctions. I don't mind if that is really what you want, but I don't think that is actually a better solution and would look rather ugly

tests/test_graph.py
AndreMiras: As you wish, it just feel not right/hacky to have something like this in the unit test and I feel like it could confuse more than it would help. But that's maybe because I didn't see what you have seen.
You can ignore my comment then

pythonforandroid/graph.py
JonasT: @KeyWeeUsr this line isn't `>80` but pretty close without the backslash, and I considered it quite less readable that way. That is why this is one of the few places where I used it to increase readability. As usual, I'm happy to take it out if y'all want me to

pythonforandroid/graph.py
JonasT: @AndreMiras *shrug* I don't think that is a surprise or even a problem. It basically builds the graph in a stupid manner, so it has the same runtime complexity (obviously) as the regular graph. However, it does it in a much simpler way in terms of how it deals with multiple possibilities, which is why it is a much nicer place for error handling

Think about the runtime in a maths fashion: it's pretty much "just" double. It's not exponentially larger or anything, and there is definitely no obvious way to get a better result unless you want to do something rather different than the actual graph code itself.

So I'm not sure what your comment intends to achieve. What would you rather have me do instead?

pythonforandroid/graph.py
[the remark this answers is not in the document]
JonasT: @AndreMiras @hackalog fear not, I don't mind adding a test. actually I remember thinking about this for this function and somewhat forgot, silly me. I'll add one

pythonforandroid/graph.py
AndreMiras: I think he meant to use parentheses over backslash

pythonforandroid/graph.py
JonasT: this is really just looking so weird because I didn't assign `Recipe.get_recipe` outside of the loop which @KeyWeeUsr already reasonably suggested. With that changed this will be implicitly fixed :+1:

pythonforandroid/graph.py
JonasT: @KeyWeeUsr again, looks IMHO considerably more cramped and less readable without it (I'm hot adding these for fun), but I'm happy to change it. also, would you mind properly writing out the remark next time? I don't think that is the most obvious or pleasant way to word it, but maybe that's just me :grimacing:

pythonforandroid/graph.py
JonasT: yea it should be format. was just me farting on my keyboard, there is no good reason it looks ugly like that

pythonforandroid/bootstraps/service_only/__init__.py
JonasT: @AndreMiras it does the same thing as before, this is what it actually compares to `list(set([('python2', 'python3', 'python3crystax')]).union({'genericndkbuild'}))`. It's not flattened

JonasT: Try to treat `Bootstrap.recipe_depends` as a black box. It can be a full regular flat list, it can contain tuple members as before. It's just whatever previous dependencies were around. As for `('python2', 'python3', ...)` that will be a tuple entry inside that list as before, and remain one during conversion to a set. There's really not much of a difference

tests/test_graph.py
JonasT: @AndreMiras it calls `obvious_conflict_checker` first which produces a different, non-static message for all the obvious conflicts. so it is indeed no longer static

tests/test_graph.py
JonasT: @AndreMiras I tried to clarify it some more. basically if the `names=...` set is too big, pytest just cuts off the members which is very annoying trying to find out what was problematic about the dependency combinations for which the test failed. I pretty much just added this to make sure one can always see the full dependency set tested even if pytest decided to cut it off

pythonforandroid/build.py
JonasT: @KeyWeeUsr I would *expect* that pip inside a venv would be wired to use the proper paths. However, I'm not 100% sure - but I would be very surprised if not

pythonforandroid/graph.py
JonasT: you mean additional (syntactically non-significant) parentheses? hm, I guess that might be an idea, let me give it a try

JonasT: I don't know, that looks quite ugly and confusing to me:
```
                recipe_dependencies = (
                    get_dependency_tuple_list_for_recipe(
                        recipe, blacklist=blacklist
                    ))
```
Again this is one of the few instances where I think a backslash would actually make sense. And `pep8`/`tox` don't complain, so while I get it's a general rule to avoid it if possible, it's still meant to be used *when it makes sense*, right?

@KeyWeeUsr can you propose what you'd actually write?

pythonforandroid/graph.py
JonasT: changed!

pythonforandroid/graph.py
JonasT: changed it

pythonforandroid/graph.py
JonasT: changed it! thanks :+1:

pythonforandroid/graph.py
JonasT: changed :heart:

pythonforandroid/graph.py
JonasT: changed it now :+1:

pythonforandroid/graph.py
JonasT: changed it, thanks :+1:

pythonforandroid/graph.py
JonasT: @KeyWeeUsr I changed it but it's a bit finicky, do you like how it looks now?

pythonforandroid/build.py
JonasT: @KeyWeeUsr changed it! can you check if you actually like how it looks now?

pythonforandroid/graph.py
JonasT: fixed, thanks :+1:

pythonforandroid/graph.py
JonasT: I used format now! are you happy with how it looks like, or should I use a dictionary with parameter names?

pythonforandroid/build.py
JonasT: @KeyWeeUsr this seems to suggest it's fine to just call pip directly: https://virtualenv.readthedocs.io/en/latest/userguide/#usage

(see how it says `This associated pip can be run from ENV/bin/pip.` without any remarks that this might be problematic)

pythonforandroid/graph.py
KeyWeeUsr: @JonasT The brackets approach doesn't mean you need to add more brackets :) Use already present code. Python is built on blocks with indentation, so if you see a bracket, it means you can do a block. From stuff such as
```python
int(
    1
)
```

up to your lines:
```python
                recipe_dependencies = get_dependency_tuple_list_for_recipe(
                    recipe, blacklist=blacklist
                )
```
< 79, no backslash, hopefully fully pep8 compliant. Backslash makes absolutely no sense here and it's ugly imo. I can't find an example where the backslash is pretty right now, however I can find e.g. >79 char line which is more readable than broken into multiple lines - https://github.com/kivy/pyjnius/blob/5f6aa738b2b7ae496ec43bfadbf9933ee640bb7b/jnius/jni.pxi#L117 because it's almost clean C and a function pointer + a table-like structuring of a file to easily find return type, etc.

pythonforandroid/graph.py
```diff
@@ -99,17 +132,126 @@ def find_order(graph):
...
+                    continue
+                conflicts = [c.lower() for c in dep_recipe.conflicts]
+                if name in conflicts:
+                    triggered_conflicts.append(dep_tuple_list)
+
+            # Throw error on conflict:
+            if triggered_conflicts:
+                # Get first conflict and see who added that one:
+                adder_second_recipe_name = "'||'".join(triggered_conflicts[0])
+                adder_second_recipe_name =\
```
KeyWeeUsr:
```suggestion
                adder_second_recipe_name = (
                    deps_were_added_by.get(
                        adder_second_recipe_name,
                        adder_second_recipe_name
                    ) or adder_second_recipe_name
                )  # might be in dict, but None
```

pythonforandroid/graph.py
JonasT: @KeyWeeUsr I find that to be quite less readable, however. Because it's indented so much and `get_dependency_tuple_list_for_recipe` is such a long name, it ends up looking quite cramped. Again if you really want me to change it I'll do it, but I *personally* think it's uglier / less readable here.

So what's your final call?

JonasT: @KeyWeeUsr changed it now as per your final call in chat :+1: this is really not a hill I want to die on so that's fine with me, it was really just what I legitimately thought looked better but that is obviously very subjective :grimacing:

pythonforandroid/graph.py
JonasT: @KeyWeeUsr I removed this backslash now by just breaking apart the code a little more. I was probably just trying to do too much in a single line :man_shrugging:

pythonforandroid/build.py
KeyWeeUsr: Ok, in that case, feel free to use the direct path. :)

pythonforandroid/graph.py
JonasT: Added a unit test! :+1:

pythonforandroid/graph.py
JonasT: (also compare `orders=[]` which was already there, where the same applies)

pythonforandroid/graph.py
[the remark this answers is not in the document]
JonasT: Added a test now! @AndreMiras @hackalog can you check if you like it?

pythonforandroid/graph.py
JonasT: changed it, thanks!

pythonforandroid/graph.py
JonasT: I hope you're fine with me resolving this, since I added in your suggestion :+1: and it was an insightful discussion for sure!

pythonforandroid/graph.py
JonasT: Resolving this one since the code looks different now and you didn't seem to hate it :smile: (and the suggestion no longer applies to the new code)

pythonforandroid/graph.py
JonasT: thanks :heart: I'm resolving this conversation so we get a better view of what's still left to do

pythonforandroid/graph.py
JonasT: resolving since the backslash use is gone :sunglasses: :+1:

pythonforandroid/bootstraps/service_only/__init__.py
JonasT: did that convince you, or do you still need anything else?

pythonforandroid/build.py
JonasT: resolving since this is fixed now :+1: thanks for pointing out

pythonforandroid/build.py
JonasT: thanks :heart: I'm resolving this conversation so we get a better view of what's still left to do

tests/test_graph.py
JonasT: should I still change something? did that clear up your questions?

pythonforandroid/graph.py
JonasT: let me know what your final decision is. should I still change it?

pythonforandroid/bootstraps/service_only/__init__.py
AndreMiras: OK cool so basically `Bootstrap.recipe_depends` is already a two dimension list? That's great then, I was assuming `Bootstrap.recipe_depends` was simply a list
Thanks :+1:

tests/test_graph.py
```diff
@@ -12,28 +14,82 @@
...
+#     dependencies_1st = ["python2", "pillow"]
+#     dependencies_2nd = ["python3crystax", "pillow"]
+#
+# This however won't work:
+#
+#     dependencies_1st = [("python2", "python3"), "pillow"]
+#     dependencies_2nd = [("python2legacy", "python3crystax"), "pillow"]
+#
+# (This is simply because the conflict checker doesn't resolve this to
+# keep the code ismple enough)
```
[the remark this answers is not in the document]
JonasT: @AndreMiras there is a test below, this just explains why it has a separate data set definition :slightly_smiling_face:

tests/test_graph.py
AndreMiras: Thanks for adding that comment, that may help future maintainers to understand why we have something weird here.
So you said if `names=...` is too long. But I don't see it too long in any of the tests here. Plus when a test fail dev put a breakpoint before the exception is raised anyway. So why don't we keep things simple?
Really it's hard for me to see the value it provides, it just feels like confusion. I tried without the try/catch block and the code just runs the same on conflicting case. I've only created a small `names` list, but that's what are test have anyway isn't it?

pythonforandroid/graph.py
AndreMiras: No no you don't need to change the CI

pythonforandroid/graph.py
AndreMiras: If you definitely think your option is better and will not generate headache later, then let's leave it. I'll just pray for nobody to do something silly in here :pray:

pythonforandroid/graph.py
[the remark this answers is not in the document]
AndreMiras: That's a good start, that's all I ask. Next time we find a bug in this code, we can just cover it improving the test.
Thanks for that :+1:

tests/test_graph.py
JonasT: > But I don't see it too long in any of the tests here.

Oh it is, I added it because it was cut off and that wasted my time. I don't like it when test output wastes my time :smile: but same applies here, if you all vote against it I remove it. But I like my tests to always produce useful output, not only 90% of the time

>  I've only created a small `names` list, but that's what are test have anyway isn't it?

There's at least one that has a longer one, otherwise it wouldn't have been cut off :woman_shrugging: it certainly wasn't due to a *hypothetical* cut off, if that is what you mean

pythonforandroid/graph.py
```diff
@@ -99,17 +132,129 @@ def find_order(graph):
...
+            recipe_dependencies = []
+            try:
+                # Get recipe to add and who's ultimately adding it:
+                recipe = Recipe.get_recipe(name, ctx)
+                recipe_conflicts = {c.lower() for c in recipe.conflicts}
+                recipe_dependencies = get_dependency_tuple_list_for_recipe(
+                    recipe, blacklist=blacklist
+                )
+            except ValueError:
+                pass
```
AndreMiras: in which case do we raise the `ValueError`? Is it on `Recipe.get_recipe()` e.g. when the recipe doesn't exist? If so it looks a bit weird to just `pass` and not `continue` for instance. I'm probably missing something. Nonetheless that code path probably deserves a test case :grimacing: what do you think?

pythonforandroid/graph.py
```diff
@@ -99,17 +132,129 @@ def find_order(graph):
...
+            adder_first_recipe_name = adding_recipe or name
+
+            # Collect the conflicts:
+            triggered_conflicts = []
+            for dep_tuple_list in deps:
+                # See if the new deps conflict with things added before:
+                if len(set(dep_tuple_list).intersection(
+                       recipe_conflicts)) == len(dep_tuple_list):
```
AndreMiras: hm why am I feeling checking only the `len` is not exactly what we want to do here?
I'm not 100% sure what you're achieving, don't you want this instead?:
```python
if set(dep_tuple_list) & recipe_conflicts == set(dep_tuple_list):
```
If this is what you wanted to do, not only it's more concise, but also safer. It's just that this `len` thing is confusing me, I can't think of when it would be valid thing to do. But I might be totally off I'm not sure

pythonforandroid/graph.py
```diff
@@ -99,17 +132,129 @@ def find_order(graph):
...
+        names = names.union(set(bs.recipe_depends))
+    names = fix_deplist([
+        ([name] if not isinstance(name, (list, tuple)) else name)
+        for name in names
+    ])
+    blacklist = [bitem.lower() for bitem in blacklist]
 
-    possible_orders = []
+    # Remove all values that are in the blacklist:
+    names_before_blacklist = list(names)
```
AndreMiras: Minor: using list slicing `[:]` is usually faster. But who cares :wink:

tests/test_graph.py
```diff
@@ -42,6 +98,17 @@ def test_bootstrap_dependency_addition():
     assert (('hostpython2' in build_order) or ('hostpython3' in build_order))
 
 
+def test_graph_deplist_transformation():
+    test_pairs = [
```
AndreMiras: Minor: I'm actually not a big fan of `pytest.mark.parametrize` (and `pytest` in general), but since you started using it above, why didn't you continue?

pythonforandroid/graph.py
JonasT: I don't believe there is any reasonable way to reduce it. Or can you suggest one? 

As for breaking it down: splitting it up into multiple functions without actually reusing the code elsewhere IMHO  would actually **reduce** readability of the overall algorithm/code flow, with a function of this size which I actually still find straightforward enough - it's pretty much just a queue processing loop with relatively granular, commented steps inside, it's not that hard to understand or that complex in nesting. Littering that code into multiple functions wouldn't make it any simpler to read, again, IMHO. So I think that is a bad idea, unless you can show me a split up that actually looks decent & better to read, but I have a hard time imagining any

So the only other way to reduce it would be to remove functionality, unless there is some miracle approach I have missed. Or of course it could be removed and we could go back to dumb error messages.

What exactly did you have in mind?

pythonforandroid/graph.py
```diff
@@ -99,17 +132,129 @@ def find_order(graph):
...
+
+            # Throw error on conflict:
+            if triggered_conflicts:
+                # Get first conflict and see who added that one:
+                adder_second_recipe_name = "'||'".join(triggered_conflicts[0])
+                second_recipe_original_adder = deps_were_added_by.get(
+                    adder_second_recipe_name, None
+                )
+                if second_recipe_original_adder:
+                    adder_second_recipe_name = second_recipe_original_adder
```
AndreMiras: not covered by the tests

pythonforandroid/graph.py
```diff
@@ -99,17 +132,129 @@ def find_order(graph):
...
+                # See if what was added before conflicts with the new deps:
+                if len(dep_tuple_list) > 1:
+                    # Not an obvious commitment to a specific recipe/dep
+                    # to be added, so we won't check.
+                    # (remember this function only catches obvious issues)
+                    continue
+                try:
+                    dep_recipe = Recipe.get_recipe(dep_tuple_list[0], ctx)
+                except ValueError:
+                    continue
```
AndreMiras: not covered by a unit test

pythonforandroid/graph.py
```diff
@@ -99,17 +132,129 @@ def find_order(graph):
...
+    to_be_added = [(name_tuple, None) for name_tuple in name_tuples]
+    while len(to_be_added) > 0:
+        current_to_be_added = list(to_be_added)
+        to_be_added = []
+        for (added_tuple, adding_recipe) in current_to_be_added:
+            assert(type(added_tuple) == tuple)
+            if len(added_tuple) > 1:
+                # No obvious commitment in what to add, ignore it.
+                # (Remember this function only catches obvious issues)
+                continue
```
AndreMiras: not covered by a unit test

pythonforandroid/graph.py
JonasT: Just to be clearer/more concrete: I think putting 4-line code chunks (or even 5-6 lines) into a separate function is in general a bad idea unless it is actually reused somewhere. It creates more friction scrolling to that function and figuring out what it does in detail than you gain in readability, it's IMHO better to just put a comment over the chunk that says what it does (which I already did)

pythonforandroid/graph.py
```diff
@@ -99,17 +132,129 @@ def find_order(graph):
...
+                       recipe_conflicts)) == len(dep_tuple_list):
+                    triggered_conflicts.append(dep_tuple_list)
+                    continue
+
+                # See if what was added before conflicts with the new deps:
+                if len(dep_tuple_list) > 1:
+                    # Not an obvious commitment to a specific recipe/dep
+                    # to be added, so we won't check.
+                    # (remember this function only catches obvious issues)
+                    continue
```
AndreMiras: not covered by a test

pythonforandroid/graph.py
JonasT: http://number-none.com/blow/john_carmack_on_inlined_code.html CTRL+F for "I have historically used" to jump to the section where the benefits of not splitting up at least super tiny things are explained, if you're curious about some smarter writing than what I could ever argue on this topic

pythonforandroid/graph.py
JonasT: @AndreMiras oops nice catch. yeah your code is what it should be, the `len()` isn't breaking things *in this specific scenario* but useless and error-prone. I'll change it

tests/test_graph.py
JonasT: @AndreMiras I don't like it either, and all fully new code I added doesn't use it. I just reused it where the test case was based on a copy&paste of another one :smile:

pythonforandroid/graph.py
JonasT: the function? or what exactly are you referring to?

that error is covered by `def test_invalid_obvious_conflict_checker(names, bootstrap)`

pythonforandroid/graph.py
JonasT: what did you have in mind? this code path is hit as soon as you have a single non-recipe dep. I'm not sure if the tests cover that, but it shouldn't be hard to add if that is what you have in mind

pythonforandroid/graph.py
```diff
@@ -99,17 +132,129 @@ def find_order(graph):
...
+    deps_were_added_by = dict()
+    deps = set()
+
+    # Add dependencies for all recipes:
+    to_be_added = [(name_tuple, None) for name_tuple in name_tuples]
+    while len(to_be_added) > 0:
+        current_to_be_added = list(to_be_added)
+        to_be_added = []
+        for (added_tuple, adding_recipe) in current_to_be_added:
+            assert(type(added_tuple) == tuple)
```
AndreMiras: Minor: why would we `assert` that? why not covering the case in a test then?
Also the `assert` statement is better without parenthesis because you can fall into a common pitfall using them

pythonforandroid/graph.py
JonasT: covered in what way? tests that hit this code path are executed, what exactly do you want to be tested? (the very ubiquitous `('python2', 'python3', ..)` tuple hits this code path, as an example)

pythonforandroid/graph.py
JonasT: It is just an input data check. I put it there because I was stupid and put in wrong input data, I probably forgot to remove it (but it's not per se harmful)

Basically `obvious_conflict_checker` doesn't accept the mixed string/tuple lists as the overall graph function but only already tuple transformed lists and at some point I forgot. This assert hits if you screw that up, which never happens normally since nobody is supposed to call the function like that (it's only called from the graph code that already does this cleanup previously)

So similarly as above, what test case would you have in mind? If I remove the assert it's also going to fail, so the test wouldn't really verify the assert. The assert is really just there to catch it earlier

pythonforandroid/graph.py
AndreMiras: Again I'm suggesting, as I said `break it down per concern when possible`. So it means what it means, if you think it's OK as it is then fine

pythonforandroid/graph.py
AndreMiras: No that code path is NOT executed by a test

pythonforandroid/graph.py
JonasT: @AndreMiras hm alright. maybe I misinterpreted your first remark, sorry :smile: I can't think of a much better way to write it, but you know, I'm very biased since I wrote it :joy:  so if you happen to come up with a more concrete change, feel free to let me know

pythonforandroid/graph.py
AndreMiras: The test case I have in mind is simply the test that shows this assert raising

JonasT: hm fair enough, I suppose I could check for `AssertionError`. I'll put something in :+1:

pythonforandroid/graph.py
AndreMiras: I don't think we have conflict for non-recipe thing... So I don't really think that part make sense and probably we want to also cover that code path with a test, you might see some silly thing going on with the output log :smile:

pythonforandroid/graph.py
JonasT: removed it so I hope it's okay if I resolve this one for now :+1:

pythonforandroid/graph.py
AndreMiras: Yes this is exactly what I have in mind, good thing if it's not that complex. This big/complex function definitely deserve a nice coverage not only for not breaking things in the future and document the feature, but also to prove all the code path make sense

pythonforandroid/graph.py
AndreMiras: I'm saying this specific line `201` is never reached by any of the test. Put a breakpoint and see it yourself

tests/test_graph.py
AndreMiras: yep I was just curious sorry, I definitely looked it up and didn't find it.
I do also think not wasting time is very import just as code readability.
I won't waste yours more you already did an awesome job, so leave it. But to be clear, yep it definitely not convinced that try/catch provide value but rather the opposite, at least to me

pythonforandroid/graph.py
JonasT: added a test case! thanks for the poke :+1:

pythonforandroid/graph.py
JonasT: ah, fascinating! okay let me think up a test case where it will be hit, I'll add one

AndreMiras: ahah now we're getting to the point :smile:
So if you can't think of a test, how do you expect another maintainer to understand this code path?
And if there're no possible code path, then it's dead code and you can delete it.
I'm not pretending that I know if there is a path or not. Just pointing that sometimes things are not as obvious as we thought in the first place, that's why we have tests

tests/test_graph.py
AndreMiras: it makes sense that you used what was already in place, I would probably have done the same actualy

pythonforandroid/graph.py
JonasT: no the problem is just that the available recipes are really way too simple for this. you need to build up some indirection, it's not that the setup per se is complicated, but the current tests quite frankly suck a little in that regard... (no code anywhere to set up some more complex recipe dependencies than the ones the actual recipes use, which just don't cut it)

> then it's dead code

it's not dead code

pythonforandroid/graph.py
JonasT: added a test! :heart: hope it's okay if I resolve this, otherwise please slap me

pythonforandroid/graph.py
JonasT: I added a test case now and verified via print debugging :joy: to see that it's hit. curious, you have to tell me some day how you saw in tox that it's not hit, because I don't actually know how to check that :smile: (I only know how to use independently of tox)

tests/test_graph.py
AndreMiras: OK that's fine if it really bugs me, I'll do the change myself.
Basically I would have added a bit more context using the static part. e.g. `Conflict detected:` or `conflicting dependencies`. But it's fine as it is still way better than current master :+1:

pythonforandroid/graph.py
JonasT: Okay, this is now also hit by at least one of the tests! hope you don't mind me resolving :+1:

pythonforandroid/graph.py
AndreMiras: Sure :+1:

pythonforandroid/graph.py
JonasT: You're right, as usual :joy: the problem was that `('python2', 'python3', ...)` are part of the initial set usually and not added in as dependency themselves. I set up a case for that now and at least according to my testing it enters that code path :+1:

AndreMiras: Nice well done! Well I also put breakpoint or asserts to see it, just like you did.
One day I'll setup code coverage so we can see it from Travis. I do think it matter and I'm very happy you were willing to improve our coverage like you did :beer:

pythonforandroid/graph.py
AndreMiras: Thanks :+1:

pythonforandroid/graph.py
JonasT: @AndreMiras well I am lazy but when I'm poked like that I gotta fix it :joy: so thanks for the poking

tests/test_graph.py
```diff
@@ -12,36 +15,151 @@
...
+    obvious_conflict_checker(
+        ctx, fix_deplist(["python3", "notarecipelibrary"])
     )
 
+    # Test that a conflict with a non-recipe dependency works:
+    # This is currently not used, so we need a custom test recipe:
+    # To get that, we simply modify one!
+    r = Recipe.get_recipe("python3", ctx)
+    old_conflicts = list(r.conflicts)
+    try:
```
AndreMiras: why, why is that?! why do we try in unit test like that? :sob: 
if the test fail it fails, right don't try/except in test please please :pray:

JonasT: @AndreMiras I'm letting it fail there is no `except` there :smile: it's just to get the recipe back into its old state no matter the test outcome

> don't try/except in test

i didn't :open_mouth:

pythonforandroid/graph.py
AndreMiras: Well now here's another bug, I'm surprise why the test didn't catch it. Can you see the bug introduced? In that same line?

pythonforandroid/graph.py
```diff
@@ -99,17 +132,129 @@ def find_order(graph):
...
+            except ValueError:
+                pass
+            adder_first_recipe_name = adding_recipe or name
+
+            # Collect the conflicts:
+            triggered_conflicts = []
+            for dep_tuple_list in deps:
+                # See if the new deps conflict with things added before:
+                if set(dep_tuple_list).intersection(
+                       recipe_conflicts) == dep_tuple_list:
```
AndreMiras: Now there's a new bug, can you spot it?
Which means coverage is not quite there yet :grimacing:

tests/test_graph.py
AndreMiras: ahah OK I'll try to buy that then :smile: 
I'll not ask you to write a context manager just for 2 tests  :wink: 
Thanks!

pythonforandroid/graph.py
JonasT: oh fascinating! I don't think this is possible to be hit because `get_dependency_tuple_list_for_recipe` applies `fix_deplist` which ensures all instances of `dep_tuple_list` would already be a set. (so I don't think this is actually possible to trigger via a test?) but yeah, it is kind of nonsensical how it is written now, I'll fix it

JonasT: oh nevermind they're `tuple`s not `set`s so it is a true bug :smile:

JonasT: Okay, I think I might be able to write a test case that hits this. let me give it a try (disclaimer: I'm a bit sleep deprived right now so I'm not promising anything :joy: )

AndreMiras: I was more talking about the type mismatch `tuple` the other one is `set`
I'm happy unit tests and coverage test can prove some point even for a "simple" function like that one :wink:

JonasT: @AndreMiras wrote a test `test_multichoice_obvious_conflict_checker` that actually hits the bug (gives error before the fix, happy after!) thanks :+1:

JonasT: resolving since hopefully fixed :slightly_smiling_face:

pythonforandroid/graph.py
JonasT: I actually added a test now, `test_indirectconflict_obvious_conflict_checker`. and it had a small bug :joy: only affected the error message, but still!

resolving since it's covered & fixed now

tests/test_graph.py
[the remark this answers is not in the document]
JonasT: hope it's okay to resolve this since with the test additions, hopefully this should be cleared up! :+1:

tests/test_graph.py
JonasT: @AndreMiras thinking about this more, I think you're right. I removed it, it was only for one of the pairs and a pytest output option maybe would even have fixed it...? it seems odd to hack this in for a single corner case. I removed it again :+1: 

hope it's okay to resolve this since hopefully this should be cleared up now :heart:

AndreMiras: Thank you I can sleep tight now! :relaxed:

tests/test_graph.py
```diff
@@ -1,24 +1,74 @@
...
+    recipe = mock.Mock()
+    recipe.name = name
+    recipe.get_opt_depends_in_list = lambda: []
+    recipe.get_dir_name = lambda: name
+    recipe.depends = list(depends or [])
+    recipe.conflicts = list(conflicts or [])
+    return recipe
+
+
+def register_fake_recipes_for_test(monkeypatch, recipe_list):
```
AndreMiras: Remark/Taste: Since I'm a `pytest` racist I like core modules more, I would have written this using the core `unittest.mock.patch` module:
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

JonasT: it was pretty much the first google result I got working :smile: I have no strong opinions on this

AndreMiras: I understand, leave it, there're no known official best practice that say `don't use pytest fanciness`. It's just me that like to stick to core modules as much as I can :+1:

pythonforandroid/graph.py
AndreMiras: Yes you wrote it as I imagined thanks

tests/test_graph.py
```diff
@@ -30,18 +80,116 @@ def test_valid_recipe_order_and_bootstrap(names, bootstrap):
...
+
+    # Test a case where a recipe pulls in a conditional tuple
+    # of additional dependencies. This is e.g. done for ('python3',
+    # 'python2', ...) but most recipes don't depend on this anymore,
+    # so we need to add a manual test for this case:
+    with monkeypatch.context() as m:
+        register_fake_recipes_for_test(m, [
+            get_fake_recipe("recipe1", depends=[("libffi", "Pillow")]),
+        ])
+        obvious_conflict_checker(ctx, fix_deplist(["recipe1"]))
```
AndreMiras: This test won't say if this feature is working or not, it currently just check that the code is not crashing, right? There're few other tests like that here and there. Can you think of if it's something we can improve?

JonasT: @AndreMiras it checks more because `obvious_conflict_checker` can raise `BuildInterruptingException`. So if the test uses that `with` construct to expect that it tests/expects a conflict result, otherwise it tests/expects a non-conflict result. So it's definitely "evaluating" the result at least in the most basic way

AndreMiras: Yes this is what I mean by crashing. I mean that this test just verify some code path is not raising any exception, nothing more.
So my point is instead of doing that check can you think of a way to prove maybe another point with this test? Basically which code path did you have in mind and can we prove it went through. That's my point

JonasT: Maybe it could check more at some point but I'm pretty okay with what it does now. I can't think at least of an obvious way to check it more granular, although surely there is always one way or another to improve it

AndreMiras: Yes I'm just challenging you, we covered quite some cases with that PR, so kudos to you!

PR #1676, opened 2019-02-04
Recipe class unit tests

AndreMiras:
Starts unit testing trivial cases for the `Recipe` class.
We will continue increasing the coverage as we encounter bugs.

### DIFF
--- a/tests/test_recipe.py
+++ b/tests/test_recipe.py
@@ -1,0 +1,43 @@
+import types
+import unittest
+from pythonforandroid.build import Context
+from pythonforandroid.recipe import Recipe
+
+
+class TestRecipe(unittest.TestCase):
+
+    def test_recipe_dirs(self):
+        """
+        Trivial `recipe_dirs()` test.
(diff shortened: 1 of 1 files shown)


opacam:
Ei, @AndreMiras, why do you initialize a `Context()` in each test? Maybe we could create `ctx` class variable and reuse that context for each test?

```
class TestRecipe(unittest.TestCase):
    def setUp(self):
        self.ctx = Context()
```

JonasT:
@opacam it decreases the risk of side effects though to keep it separate, and it probably has like almost zero performance impact. I would personally recommend not to share test resources unless there's a benefit to it, just to keep things more independent

AndreMiras:
Yes you both have good point :+1:  It's often hard to balance between reusing, decoupling and readability.
Since it was just a little line of code this time I went for creating each time, but yes I could have used the `setUp()` method just like I did in `tests/recipes/test_gevent.py`.
Let's see as the boiler plate gets bigger and the number of tests increase maybe we will end up using `setUp()` here too

PR #1677, opened 2019-02-05
Fix zope_interface and add python3 compatibility

opacam:
The `zope.interface` module lacks of the `__init__.py` file in one of his folders, that leads into:

`ImportError: No module named zope.interface`

### DIFF
--- a/pythonforandroid/recipes/zope_interface/__init__.py
+++ b/pythonforandroid/recipes/zope_interface/__init__.py
@@ -1,5 +1,6 @@
 from pythonforandroid.recipe import PythonRecipe
 from pythonforandroid.toolchain import current_directory
+from os.path import join
 import sh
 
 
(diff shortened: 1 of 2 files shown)


KeyWeeUsr:
Perhaps a better idea would be to just create an empty patch for that file, so that we don't do more tricky stuff within the recipe. If/after they fix it, we will get a conflict for applying patch and will know that everything is ok on their side.

opacam:
Oh yes, @keyWeeUsr, a lot better with a patch, thanks!!
:smile:

36Boxes:
Tried this patch with python2 and it worked for me!

pythonforandroid/recipes/zope_interface/__init__.py
```diff
@@ -9,10 +10,18 @@ class ZopeInterfaceRecipe(PythonRecipe):
...
     patches = ['no_tests.patch']
 
+    def build_arch(self, arch):
+        super(ZopeInterfaceRecipe, self).build_arch(arch)
+        # The zope.interface module lacks of the __init__.py file in one of his
+        # folders (once is installed), that leads into an ImportError.
+        # Here we intentionally apply a patch to solve that, so, in case that
+        # this is solved in the future an error will be triggered
+        zope_install = join(self.ctx.get_site_packages_dir(arch.arch), 'zope')
+        self.apply_patch('fix-init.patch', arch.arch, build_dir=zope_install)
```
AndreMiras: Minor: I'm not sure I understand why we're applying the patch that way rather than adding it to the `patches` list?

AndreMiras: ahah fun, thanks for the explanation!
Then is it not the manifest file or something like that we want to fix with a patch?
I mean the way you did seems fine, I'm just challenging to see if there're no better ways

PR #1680, opened 2019-02-05
Delete kivent recipes

inclement:
I know we talked about this, but they still seem to be there for now :p

For general reference, these are to be removed because they don't work and have no prospect of working because we're unlikely to spend any time on it and kovak has become busy with other things. If someone submits new recipes that do work, that would be welcome.


inclement:
https://github.com/kivy/python-for-android/pull/1686

PR #1682, opened 2019-02-06
[WIP] Fix `configure` for python's recipes when building from macOS

opacam:
**This is a long shot**

This pr is made in order to try to solve the linkage problems that we have when using a macOS as a host system.

Here I migrate a macOS's patch from python2legacy's recipe that we need in order to successfully link when building from a macOS. I suspect that we still may need that for the new python recipes...so ...I think that this pr may solve the linkage problems that we currently have, described in  here #1647.

Unfortunately I couldn't test this (I haven't a macOS to test it)...so...

**this should not be merged unless we can confirm that actually solves the problem**

### DIFF
--- a/pythonforandroid/recipes/python2/__init__.py
+++ b/pythonforandroid/recipes/python2/__init__.py
@@ -2,6 +2,7 @@ from os.path import join, exists
 from pythonforandroid.recipe import Recipe
 from pythonforandroid.python import GuestPythonRecipe
 from pythonforandroid.logger import shprint
+from pythonforandroid.patching import is_darwin
 import sh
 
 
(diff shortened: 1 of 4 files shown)


inclement:
Thanks @opacam, this is great to have to point people at.

dpProjectBuzz:
I am busy testing  the solution now.

The error which I am receiving is the same as before and after applying the solution.

Although I must admit that I am very new to c++ and .mk syntax

OS=Mojave
Python_m=3

--------------------CONSOLE DUMP--------------
The ++clang missing link error

Some warnings and things.

  STDOUT:
Android NDK: WARNING
                                 /buildozer/android/platform/build/build/bootstrap_builds/sdl2-python3/jni/application/src/Android.mk:main: non-system libraries in linker flags: -lpython3.7m    
Android NDK:     This is likely to result in incorrect builds. Try using LOCAL_STATIC_LIBRARIES    
Android NDK:     or LOCAL_SHARED_LIBRARIES instead to list the library dependencies of the    
Android NDK:     current module

PR #1683, opened 2019-02-06
Implement --blacklist option and include more modules/recipes by default

JonasT:
This is another subpart split out of my misguidedly oversized #1625 (`setup.py`) pull request.

Changes:

- Adds `--blacklist` option that prevents recipes/packages from being added even though they are in the `depends` of recipes or otherwise added
- Makes the following modules included per default: `android` (via `bootstrap.py` dependency of all bootstraps), `openssl`/`libffi`/`sqlite3` (via `python3` recipe dependency)
- Documents `--blacklist` option and that `android` is now included by default
- Cleans up the packaging kivy/sdl2 apps part in the Quickstart section a little

### DIFF
--- a/pythonforandroid/graph.py
+++ b/pythonforandroid/graph.py
@@ -39,10 +39,13 @@ class RecipeOrder(dict):
         return False
 
 
-def get_dependency_tuple_list_for_recipe(recipe, blacklist=[]):
+def get_dependency_tuple_list_for_recipe(recipe, blacklist=None):
     """ Get the dependencies of a recipe with filtered out blacklist, and
         turned into tuples with fix_deplist()
     """
+    if blacklist is None:
+        blacklist = set()
+    assert(type(blacklist) == set)
     if recipe.depends is None:
         dependencies = []
     else:
@@ -51,14 +54,15 @@ def get_dependency_tuple_list_for_recipe(recipe, blacklist=[]):
 
         # Filter out blacklisted items and turn lowercase:
         dependencies = [
-            deptuple for deptuple in dependencies
-            if not set(deptuple).intersection(set(blacklist))
+            tuple(set(deptuple) - blacklist)
+            for deptuple in dependencies
+            if tuple(set(deptuple) - blacklist)
         ]
     return dependencies
 
 
 def recursively_collect_orders(
-        name, ctx, all_inputs, orders=[], blacklist=[]
+        name, ctx, all_inputs, orders=None, blacklist=None
         ):
     '''For each possible recipe ordering, try to add the new recipe name
     to that order. Recursively do the same thing with all the
@@ -66,6 +70,10 @@ def recursively_collect_orders(
 
     '''
     name = name.lower()
+    if orders is None:
+        orders = []
+    if blacklist is None:
+        blacklist = set()
     try:
         recipe = Recipe.get_recipe(name, ctx)
         dependencies = get_dependency_tuple_list_for_recipe(
@@ -75,7 +83,8 @@ def recursively_collect_orders(
         # handle opt_depends: these impose requirements on the build
         # order only if already present in the list of recipes to build
         dependencies.extend(fix_deplist(
-            [[d] for d in recipe.get_opt_depends_in_list(all_inputs)]
+            [[d] for d in recipe.get_opt_depends_in_list(all_inputs)
+             if d.lower() not in blacklist]
         ))
 
         if recipe.conflicts is None:
@@ -106,7 +115,9 @@ def recursively_collect_orders(
             dependency_new_orders = [new_order]
             for dependency in dependency_set:
                 dependency_new_orders = recursively_collect_orders(
-                    dependency, ctx, all_inputs, dependency_new_orders)
+                    dependency, ctx, all_inputs, dependency_new_orders,
+                    blacklist=blacklist
+                )
 
             new_orders.extend(dependency_new_orders)
 
@@ -132,7 +143,7 @@ def find_order(graph):
                 bset.discard(result)
 
 
-def obvious_conflict_checker(ctx, name_tuples, blacklist=[]):
+def obvious_conflict_checker(ctx, name_tuples, blacklist=None):
     """ This is a pre-flight check function that will completely ignore
         recipe order or choosing an actual value in any of the multiple
         choice tuples/dependencies, and just do a very basic obvious
@@ -140,6 +151,8 @@ def obvious_conflict_checker(ctx, name_tuples, blacklist=[]):
     """
     deps_were_added_by = dict()
     deps = set()
+    if blacklist is None:
+        blacklist = set()
 
     # Add dependencies for all recipes:
     to_be_added = [(name_tuple, None) for name_tuple in name_tuples]
@@ -227,7 +240,7 @@ def obvious_conflict_checker(ctx, name_tuples, blacklist=[]):
     return None
 
 
-def get_recipe_order_and_bootstrap(ctx, names, bs=None, blacklist=[]):
+def get_recipe_order_and_bootstrap(ctx, names, bs=None, blacklist=None):
     # Get set of recipe/dependency names, clean up and add bootstrap deps:
     names = set(names)
     if bs is not None and bs.recipe_depends:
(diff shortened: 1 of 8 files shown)


JonasT:
@inclement this changes by the way what we talked about, that it will now include all core modules per default and the people who want a more trimmed down build need to actively specify `--blacklist` to reduce it. I added a doc section for it though, so it shouldn't be too hard to figure out for anyone who wants to do that

pythonforandroid/graph.py
```diff
@@ -51,14 +52,15 @@ def get_dependency_tuple_list_for_recipe(recipe, blacklist=[]):
 
         # Filter out blacklisted items and turn lowercase:
         dependencies = [
-            deptuple for deptuple in dependencies
-            if not set(deptuple).intersection(set(blacklist))
+            tuple(set(deptuple) - blacklist)
+            for deptuple in dependencies
+            if tuple(set(deptuple) - blacklist)
```
AndreMiras: Minor/troll: so you prefer to use `intersect()` over `&` but still want to use `-` over `difference()` :thinking: 
Honestly I prefer to use the operator for all operations

pythonforandroid/toolchain.py
```diff
@@ -301,6 +309,13 @@ def __init__(self):
                   'Python modules'),
             default='')
 
+        generic_parser.add_argument(
+            '--blacklist',
+            help=('Blacklist an internal recipe from use. Allows '
+                  'disabling Python 3 core modules to save size'),
+            dest="blacklist",
+            default='')
```
AndreMiras: why isn't the default `None` or `[]`?

tests/test_graph.py
```diff
@@ -83,6 +84,41 @@ def test_invalid_recipe_order_and_bootstrap(names, bootstrap):
...
+def test_blacklist():
+    # First, get order without blacklist:
+    build_order, python_modules, bs = get_recipe_order_and_bootstrap(
+        ctx, ["python3", "kivy"], None
+    )
+    # Now, obtain again with blacklist:
+    build_order_2, python_modules_2, bs_2 = get_recipe_order_and_bootstrap(
+        ctx, ["python3", "kivy"], None, blacklist=["libffi"]
+    )
+    assert "libffi" not in set(build_order_2)
```
AndreMiras: Minor: why enforcing the `set()` here? Creating the `set()` probably is `O(n)` anyway so why bothering?

tests/test_graph.py
```diff
@@ -83,6 +84,41 @@ def test_invalid_recipe_order_and_bootstrap(names, bootstrap):
...
+    assert set(build_order_2).union({"libffi"}) == set(build_order)
```
AndreMiras: Minor: I know you can use the `|` operator now that I saw you using the `-` operator. So I know you're capable of :trollface:

pythonforandroid/graph.py
```diff
@@ -39,10 +39,11 @@ def conflicts(self):
         return False
 
 
-def get_dependency_tuple_list_for_recipe(recipe, blacklist=[]):
+def get_dependency_tuple_list_for_recipe(recipe, blacklist=set()):
```
AndreMiras: I keep thinking that I hate default mutable args and it should be a practice we want to avoid as much as possible

hackalog: Yes. Please. `blacklist=None`. Add the two lines of code in the function body to avoid the insidious bug down the road...

pythonforandroid/toolchain.py
JonasT: @AndreMiras it's a string (comma separated list)

pythonforandroid/graph.py
JonasT: done :+1:

JonasT:
@AndreMiras commit details added, also I just tested this with a real world app to verify that `--blacklist=openssl` indeed makes it unavailable and it does. As far as I'm concerned this is ready for merge :+1:

inclement:
As discussed before, I still don't really see the value of doing this, but I'll defer to the clear approval of everyone else.

However, doesn't this break the existing --blacklist argument?

AndreMiras:
Ouch good point @inclement 
Too bad, that one right? https://github.com/kivy/python-for-android/blob/6660b9e/pythonforandroid/bootstraps/common/build/build.py#L634
Not nice :cry: 
Also I didn't know you discussed that previously, I should have aligned with you to double check.
Now what I guess first thing is to verify if indeed the other `--blacklist` is being masked by this one. In this case it would require using another option name that doesn't clash with that one.
Or do we want to completely revert that change?

inclement:
> Now what I guess first thing is to verify if indeed the other `--blacklist` is being masked by this one

I didn't directly check, but I don't see how it could not be.

Let's at least rename it to something like `--blacklist-requirements`. I might personally prefer to avoid the word blacklist, partly because I don't like how broad the documented procedure is when actually I can't see what it would ever be used for except this one specific thing. Maybe I'm missing some other use cases though.

Is there a strong reason to implement an argument for this rather than the other suggestion of having a MinimalPython3 recipe that omits the opt_depends from the hard requirements?

AndreMiras:
True story, the argument I see is just to stick to most of the default python distributions we know. But yes then I agree it then requires to add the "blacklist" case hence more code, more noise... :confused: 
We can also drop the blacklist part of this PR and if dev want to reduce size of their apk even further they use the usual `--blacklist` param to blacklist via path. So that would simplify the code/doc while keeping the `sqlite3` feature included by default

inclement:
> True story, the argument I see is just to stick to most of the default python distributions we know

This is fine, but a generic blacklist isn't the only way to deal with it. The suggestion mentioned above (also discussed on discord I think?) was to have a MinimalPython3 recipe that's identical to the Python3 one except that it places all the not-really-dependencies in opt_depends rather than depends. The result would be that anyone taking the extra step to optimise the apk replaces python3 with minimalpython3 (or other name as appropriate), then adds any optional dependencies explicitly.

> We can also drop the blacklist part of this PR and if dev want to reduce size of their apk even further they use the usual `--blacklist` param to blacklist via path

I don't think this is good enough, it's too indirect a method for something that a significant fraction of users will ultimately want to do.

AndreMiras:
I like the MinimalPython3 idea and I don't at the same time.
I like because it seems a more elegant solution to that problem. But I don't because even though it feels pretty simple it also add some "complexity" over the prio-pr1683-state. I mean I can already foresee that recipe would either be a lot of copy past, hence we would have two recipe to maintain python3 (e.g. configure flag update, version bump). Or we would go with fancy inheritance/composition to honor DRY and then it still adds complexity.

I do agree that if we care more about minimal apk size than we care about "the ease" to use p4a, then maybe p4a is better without #1683. Then dev who need `openssl` and `sqlite3` support would just go as before, by simply providing it. It's unexpected Python behaviour but not that cumbersome after all

inclement:
> But I don't because even though it feels pretty simple it also add some "complexity" over the prio-pr1683-state.

So does the whole new blacklisting system just introduced - that's basically my objection here, I think it's far more complex than a minimal python recipe but for not much benefit since it doesn't have any other use that I can see.

>  I mean I can already foresee that recipe would either be a lot of copy past, hence we would have two recipe to maintain python3 (e.g. configure flag update, version bump). Or we would go with fancy inheritance/composition to honor DRY and then it still adds complexity.

One of the recipes would be a trivial subclass of the other, the only difference is the depends/opt_depends.

I guess another alternative for the same effect would be a `--minimal-python` argument of some sort, but I think the recipe would be fine.

AndreMiras:
> So does the whole new blacklisting system just introduced - that's basically my objection here, I think it's far more complex than a minimal python recipe but for not much benefit since it doesn't have any other use that I can see.

Yes agree that's why I was suggesting if at the end it annoys more than it helps we can revert it, period.

JonasT:
@AndreMiras @inclement oops @ option, feel free to revert I can reintroduce it with a different name as a new pull request. Sorry about that, feel kind of silly for not noticing that :cry: 

@inclement I personally think it is nicer to disable what you don't want rather than either have everything, or minimal/nothing with everything added in, it just seems a little strange to me. Also, I had the hope of being able to one day use this argument to block dependencies pulled in by a `setup.py` although maybe that will never be possible since `pip` appears to have no easy way to completely blacklist things... so maybe ignore that

JonasT:
(nevertheless if you both prefer a minimal recipe then maybe that's the way to go. That's just the reasons why I preferred a substractive rather than an additive approach)

JonasT:
I made #1689 in case you decide to keep the functionality but with a proper parameter name

PR #1684, opened 2019-02-06
Added option to support custom shared libraries with <uses-library> tag

pax0r:
Support for: 
https://developer.android.com/guide/topics/manifest/uses-library-element

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -590,14 +590,16 @@ tools directory of the Android SDK.
         ap.add_argument('--launcher', dest='launcher', action='store_true',
                         help=('Provide this argument to build a multi-app '
                               'launcher, rather than a single app.'))
-    ap.add_argument('--permission', dest='permissions', action='append',
+    ap.add_argument('--permission', dest='permissions', action='append', default=[],
                     help='The permissions to give this app.', nargs='+')
-    ap.add_argument('--meta-data', dest='meta_data', action='append',
+    ap.add_argument('--meta-data', dest='meta_data', action='append', default=[],
                     help='Custom key=value to add in application metadata')
+    ap.add_argument('--uses-library', dest='android_used_libs', action='append', default=[],
+                    help='Used shared libraries included using <uses-library> tag in AndroidManifest.xml')
     ap.add_argument('--icon', dest='icon',
                     help=('A png file to use as the icon for '
                           'the application.'))
-    ap.add_argument('--service', dest='services', action='append',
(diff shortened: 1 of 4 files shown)


pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -749,6 +751,9 @@ def _read_configuration():
     if args.meta_data is None:
         args.meta_data = []
 
+    if args.android_used_libs is None:
+        args.android_used_libs = []
```
inclement: Very minor thing: is it enough to just write `default=[]` in the `ap.add_argument(...)` instead?

pax0r: I think it should be, I just wanted it to be consistent with existing code. If you like it I can refactor rest of it to use defaults.

pax0r: fixed

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -740,18 +742,9 @@ def _read_configuration():
               'deprecated and does nothing.')
         args.sdk_version = -1  # ensure it is not used
 
-    if args.permissions is None:
-        args.permissions = []
-    elif args.permissions:
-        if isinstance(args.permissions[0], list):
+    if args.permissions and isinstance(args.permissions[0], list):
             args.permissions = [p for perm in args.permissions for p in perm]
```
AndreMiras: Thanks for updating that PR so many times and also thanks for squashing your commits!
We have one more here:
`pythonforandroid/bootstraps/common/build/build.py:746:13: E117 over-indented`
Don't kill us now :grimacing: 
You can always run `tox` locally on your dev env to verify test and linting is OK if you don't want to wait for Travis to complain

pax0r: Sorry, is obviously my fault for not running pep8 locally ;)  I mostly rely on pycharm for catching formatting errors but it failed me at this one ;)

AndreMiras:
@inclement you have some ongoing "requested changes" that I think got addressed. Do you mind doing a final check and merging?

AndreMiras:
@pax0r since @inclement seems to be afk I think we can merge soon. Have you been using this change for a while without issues? Do you think it's OK for a merge now?

pax0r:
Yes, we build application using some 3rd party library and had no issues. After merge I'll also provide patch for buildozer with support for this option.

AndreMiras:
Merged thanks!
@inclement comments were addressed, but the his review was still in "request changes" and that was blocking the merge, so I had to dismiss them

PR #1690, opened 2019-02-07
Use nativeSetenv() provided by SDL2 and cleanups

j-devel:
Since 16b07eda (bumping SDL2 version from 2.0.4 to 2.0.9),  `pythonforandroid/recipes/sdl2/add_nativeSetEnv.patch` that provides `nativeSetEnv()` is obsolete (in terms of patching location).  Furthermore, we don't even need this patch now because SDL2-2.0.9 itself provides the same functionality called `nativeSetenv()` (NOT `nativeSetEnv()`, beware the character's case!!) as follows:
```
$ ack nativeSetenv SDL2-2.0.9
SDL2-2.0.9/android-project/app/src/main/java/org/libsdl/app/SDLActivity.java
724:    public static native void nativeSetenv(String name, String value);
947:                    nativeSetenv(name, value);

SDL2-2.0.9/src/core/android/SDL_android.c
135:JNIEXPORT void JNICALL SDL_JAVA_INTERFACE(nativeSetenv)(
884:JNIEXPORT void JNICALL SDL_JAVA_INTERFACE(nativeSetenv)(
```

This PR is to

- remove the obsolete patch file, and
- replace the usages of the old `nativeSetEnv()` with `nativeSetenv()` of SDL2-2.0.9.

Note: the occurrences of `nativeSetEnv()` in `bootstraps/pygame/build/src/org/renpy/android/SDLSurfaceView.java` are intact since this is code is based on SDL1.2 which is now a legacy (and so is pygame...).


AndreMiras:
Thanks for the PR! I can't remember if it was @JonasT or @opacam who did some huge effort updating SDL2. It would be great if one of the two could take a look at that PR :pray:

opacam:
Ok, I took a look and it seems fine to me (also I have tested it for python2 and python3...and all goes well). Still I prefer that @JonasT or @inclement approve this changes (they made the latest  sdl2 related changes in `python-for-android` and `kivy`...his knowledge of this is better than mine :wink:)

pythonforandroid/bootstraps/service_only/build/src/main/java/org/kivy/android/PythonActivity.java
```diff
@@ -404,7 +404,7 @@ public static void stop_service() {
     }
 
 
-    public static native void nativeSetEnv(String j_name, String j_value);
+    public static native void nativeSetenv(String name, String value);
```
JonasT: Is this required given this is already defined in the parent class `SDLActivity`? I'm not 100% brushed up on my java knowledge but this looks like it might be redundant to me

JonasT: oh right nevermind this is in the non-SDL2 bootstrap, I missed that for a second. all fine :+1:

inclement:
Great, thanks, especially unifying the other bootstraps :)

Issue #1687, opened 2019-02-07
kivy==master Window undefined


Ham-Merhead:
### Versions

* Python:  3.6
* OS:         Ubuntu 18.04
* Kivy:       master
* Cython:   29.1

### Description

Using python-for-android (master) downloaded 2/6/2019 (same with 2/2/2019)
Issue with kivy==master
This example crashes on Android load at main.py line 16
Log shows Window as NoneType object.

```
from kivy.app import App
from kivy.uix.label import Label
from kivy.graphics import Color, Rectangle
from kivy.core.window import Window

class Demo(App):
    def set_background_color(self):
        with self.label.canvas.before:
            Color(0.5, 0.0, 0.5)
            Rectangle(size=Window.size) 
       
    def _resize_handler(self,obj,size):
        self.set_background_color()

    def build(self):
        foo = Window.size                      ##### <<<<< Line 16
        Window.bind(size=self._resize_handler)
        self.label = Label(text = 'Greetings Earthlings')
        self.set_background_color()
        return self.label

if __name__ == '__main__':
    Demo().run()

```
### buildozer.spec

Command:
```sh

1) Kivy==master produces crash:
p4a clean_all
p4a apk --requirements=python3,kivy==master

2) Kivy default produces no crash
p4a clean_all
p4a apk --requirements=python3,kivy

3) Kivy==master as a change to default (so no clean_all before) produces no crash
p4a apk --requirements=python3,kivy==master
```

.p4a file:
```
--dist_name=ex
--private .
--package=com.example.ex
--name Ex
--arch=armeabi-v7a
--orientation sensor
--sdk_dir /home/bobf/androidtools/sdk
--ndk_dir /home/bobf/androidtools/android-ndk-r17c
--android_api 28
--ndk-api 21
--version 0.0.0

```

### Logs

```
02-06 13:01:54.057 26517 26548 I python  :    File "/mnt/c/users/bobf/documents/pm/tst2/main.py", line 16, in build
02-06 13:01:54.057 26517 26548 I python  :  AttributeError: 'NoneType' object has no attribute 'size'
02-06 13:01:54.057 26517 26548 I python  : Python for android ended.

```

OptimusGREEN:
These appear to be linked/same issue 
https://github.com/kivy/kivy/issues/6136

PR #1692, opened 2019-02-07
SDL2_image update to 2.0.4 and set kivy's version to master

opacam:
Resolves: #1687 and kivy/kivy#6136


pythonforandroid/recipes/kivy/__init__.py
```diff
@@ -6,8 +6,7 @@
 
 
 class KivyRecipe(CythonRecipe):
-    # post kivy==1.10.1 with SDL 2.0.9 fixes
-    version = 'b47f669'
+    version = 'master'
```
AndreMiras: why unpinning rather than linking to a "stable" commit hash e.g. the last one https://github.com/kivy/kivy/commit/775cd9e6e39809ad52de10280a5aa8d76c56a24c ?
```python
version = '775cd9e'
```
I personally prefer reproducible builds, but I know this can be rather controversial

OptimusGREEN:
This has fixed the issue for me. thanks @opacam

pythonforandroid/recipes/kivy/__init__.py
JonasT: :+1: on the pin

opacam: @AndreMiras and @JonasT, I totally agree with both of you...so ...`pin` it's better...thanks @AndreMiras :smile:

¡¡I will change that!!

opacam:
Ok changes made.

**kivy's version pinned in kivy/kivy@a95d67f: `Fixes SDL2 image loading (jpg)`**

PR #1705, opened 2019-02-12
Allow to override downloaded filename in recipie

pax0r:
Sometimes filename is not present in url but it's set using Content-Disposition header.
As we need filename before http request this change allows to set custom filename for downloaded file.

For example zips generated by github are in form: https://codeload.github.com/kivy/python-for-android/zip/master and cannot be unpacked as they do not have `.zip` in name.

### DIFF
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -102,6 +102,8 @@ class Recipe(with_metaclass(RecipeMeta)):
 
     archs = ['armeabi']  # Not currently implemented properly
 
+    download_filename = None
+
     @property
     def version(self):
         key = 'VERSION_' + self.name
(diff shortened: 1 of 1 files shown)


AndreMiras:
Thanks @pax0r for the PR!
Yes you're right about this use case not being directly covered yet. However it's not that common, as we could always overcome it, by different means so far. How many recipe would currently need this feature for instance?
See that URL you shared could be https://github.com/kivy/python-for-android/archive/master.zip instead and then that would work. When it doesn't we can always override one of the Recipe methods to tacle it right? I personally always want to be careful and think twice before touching the core code, because less is more. Specially since this feature you're proposing doesn't come with a unit test, so next p4a update/refactor may break it without noticing.

pax0r:
Actually I use it for many custom recipes from github. On the other hand I didn't know that github works with adding the `.zip` extension to url.
I can write some simple unittests for it or you can reject this PR if you feel it's not required.

AndreMiras:
Thanks @pax0r for the feedback. I'm happy that the way around worked for you :smiley: 
Even if the feature makes total sense, I like the "it's urgent to do nothing" approach as well as the Pareto principe applied to code development. It more or less says "don't implement features that will not be used 80% of time".
That again is my personal vision, so I'd like to have other dev advice on it, @inclement for instance.
If we @inclement wants us to integrate this feature, then yes please provide a unit test for it.
Thanks again for the contribution!

JonasT:
@pax0r what about using the `https://github.com/kivy/python-for-android/archive/master.zip` URL style? does that give you a proper downloaded filename?

inclement:
I'm inclined to not merge this since the github use case is the only example and as discussed it doesn't seem necessary. Open to disagreement though.

PR #1707, opened 2019-02-12
Additional hook after final apk created

OptimusGREEN:
A post completion hook:

Example Usage: Rename or copy/move final apk


OptimusGREEN:
local env issue caused masked results on this and as a result I am withdrawing the pr.

PR #1708, opened 2019-02-13
[WIP] Fix landscape orientation

opacam:
We exchange the orientation's landscape values  because they don't match what the user expects

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/libsdl/app/SDLActivity.java
+++ b/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/libsdl/app/SDLActivity.java
@@ -762,9 +762,9 @@ public class SDLActivity extends Activity implements View.OnSystemUiVisibilityCh
         if (hint.contains("LandscapeRight") && hint.contains("LandscapeLeft")) {
             orientation = ActivityInfo.SCREEN_ORIENTATION_SENSOR_LANDSCAPE;
         } else if (hint.contains("LandscapeRight")) {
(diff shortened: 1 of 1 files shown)


JonasT:
@opacam @inclement @tito just as a heads up, we should **never ever** change `SDLActivity.java` directly, it's a 1:1 copy of the upstream source code. The correct way to modify it is to change this file: https://github.com/kivy/python-for-android/issues/1709

otherwise we would go back to the horrible historic state where an outdated file is indistinguishable from random custom changes we made, which would make any future upgrade way more costly to do

inclement:
@JonasT Don't worry, I wouldn't have done this anyway, although I didn't know we had a patch for applying such changes. I'm pretty sure the actual root issue is the one I opened at https://github.com/kivy/kivy/pull/6163/files anyway, Kivy itself was just using the wrong default landscape direction.

JonasT:
@inclement yeah I added that patch mechanism with that bigger SDL2 upgrade since the messy mix of outdated & changed for our own purposes annoyed me too much :smile:

AndreMiras:
Thanks for the head up @JonasT I didn't realise it. So maybe it would make some sense to not track this file at all, but download it from upstream during the build process?

JonasT:
@AndreMiras that would certainly be possible to do, but I'm not sure how easy this would be (the bootstrap is a recipe after all so it should be possible to use the regular download mechanism somehow, but this would need to be coordinated properly with the copy-things-with-common-bootstrap-files-into-one-place mechanism)

PR #1712, opened 2019-02-17
Added warning for arguments containing carriage returns. This happens…

Aralox:
… if a shell script was written in Windows but executed in Linux, as may be the case with bash for Windows.

Example of undefined behaviour: 

- sh script to invoke p4a is written in windows and run in windows bash.
- In this script, the last argument to p4a is `--requirements=python3,kivy,numpy`, which causes the last package to be `numpy\r` due to the CRLF at the end of the line. This might get through to pip but not have any associated p4a recipes, leading to many red herrings and wild goose chases.

The solution to this was straightforward (change my text editor to use LF instad of CRLF), but this behaviour was difficult and time-consuming to debug. Although this problem is not due to python-for-android, I decided to open this PR regardless as it would improve developer quality of life.

### DIFF
--- a/pythonforandroid/toolchain.py
+++ b/pythonforandroid/toolchain.py
@@ -229,6 +229,7 @@ class ToolchainCL(object):
     def __init__(self):
 
         argv = sys.argv
+        self.warn_on_carriage_return_args(argv)
         # Buildozer used to pass these arguments in a now-invalid order
         # If that happens, apply this fix
         # This fix will be removed once a fixed buildozer is released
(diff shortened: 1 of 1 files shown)


pythonforandroid/toolchain.py
```diff
@@ -575,6 +576,12 @@ def add_parser(subparsers, *args, **kwargs):
         # Each subparser corresponds to a method
         getattr(self, args.subparser_name.replace('-', '_'))(args)
 
+    def warn_on_carriage_return_args(self, args):
+        for check_arg in args:
+            if '\r' in check_arg:
+                warning("Argument '" + str(check_arg.replace('\r', '')) + "' contains a carriage return (\\r).")
```
AndreMiras: Minor: maybe we should use `format()` rather than string concatenation?

pythonforandroid/toolchain.py
```diff
@@ -575,6 +576,12 @@ def add_parser(subparsers, *args, **kwargs):
         # Each subparser corresponds to a method
         getattr(self, args.subparser_name.replace('-', '_'))(args)
 
+    def warn_on_carriage_return_args(self, args):
```
AndreMiras: Minor: probably what you want is a static method

Aralox:
Thank you for your comments Andre, and also for telling me about the 50/72 rule. I'm going to start using that at work!

PR #1714, opened 2019-02-18
Fix setenv() of non-SDL2 bootstrap

### LINKED ISSUE #1702: Webview build can't find `SDL_setenv`.
I'm attempting to build the Python3 webview test app from the `master` branch using NDK 19 and an arm64-v8a target. However linking `libmain.so` fails with `undefined reference to 'SDL_setenv'`.

I see [start.c](https://github.com/kivy/python-for-android/blob/master/pythonforandroid/bootstraps/common/build/jni/application/src/start.c) references `SDL_setenv` in `Java_org_kivy_android_PythonActivity_nativeSetenv()` which is compiled because `BOOTSTRAP_NAME_WEBVIEW` is defined. Is it intended to call `SDL_setenv` given that `SDL.h` has not been included (since a webview build defines `BOOTSTRAP_USES_NO_SDL_HEADERS`)??

j-devel:
Fixes #1702.  This is regression related to #1690.  For non-SDL2 bootstrap, we should not use `SDL_setenv()` here; instead, use the plain `setenv()` as previously.  Sorry that I missed testing this bootstrap case.

Before fix:
```
/home/a/.local/share/python-for-android/build/bootstrap_builds/webview-python2/jni/application/src/start.c:471: error: undefined reference to 'SDL_setenv'
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```
After fix: No compile errors regarding this issue.

### DIFF
--- a/pythonforandroid/bootstraps/common/build/jni/application/src/start.c
+++ b/pythonforandroid/bootstraps/common/build/jni/application/src/start.c
@@ -468,7 +468,7 @@ void Java_org_kivy_android_PythonActivity_nativeSetenv(
     const char *utfname = (*env)->GetStringUTFChars(env, name, NULL);
     const char *utfvalue = (*env)->GetStringUTFChars(env, value, NULL);
 
-    SDL_setenv(utfname, utfvalue, 1);
+    setenv(utfname, utfvalue, 1);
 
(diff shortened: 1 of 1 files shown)


AndreMiras:
Yes we definitely need some runtime automated testing. Thank you for addressing that regression!

PR #1716, opened 2019-02-19
Consider arch when looking for compatible dist.

Aralox:
Cache self.dist so it isn't calculated multiple times (e.g. wrapper_func in toolchain).

Current problem scenario:
- I build for arm
- I build the same project for x86. p4a uses existing distribution and packages arm libraries into the apk instead of recompiling project for x86

I also modified _dist() as it was getting called multiple times, causing the dist search (and log output) twice.

### DIFF
--- a/pythonforandroid/distribution.py
+++ b/pythonforandroid/distribution.py
@@ -44,6 +44,7 @@ class Distribution(object):
     @classmethod
     def get_distribution(cls, ctx, name=None, recipes=[],
                          ndk_api=None,
+                         archs=None,
                          force_build=False,
                          extra_dist_dirs=[],
                          require_perfect_match=False,
@@ -62,8 +63,10 @@ class Distribution(object):
             exists, it will be used.
         recipes : list
             The recipes that the distribution must contain.
-        force_download: bool
-            If True, only downloaded dists are considered.
+        ndk_api : str
+            NDK API level used
+        archs : list
+            List of architectures to be built
         force_build : bool
             If True, the dist is forced to be built locally.
         extra_dist_dirs : list
@@ -92,15 +95,13 @@ class Distribution(object):
         # 1) Check if any existing dists meet the requirements
         _possible_dists = []
         for dist in possible_dists:
-            if (
-                ndk_api is not None and dist.ndk_api != ndk_api
-            ) or dist.ndk_api is None:
+            if ((ndk_api is not None and dist.ndk_api != ndk_api)
+                or dist.ndk_api is None or dist.archs is None):
                 continue
-            for recipe in recipes:
-                if recipe not in dist.recipes:
-                    break
-            else:
+            if (set(recipes).issubset(set(dist.recipes))
+                and set(archs).issubset(set(dist.archs))):
                 _possible_dists.append(dist)
+
         possible_dists = _possible_dists
 
         if possible_dists:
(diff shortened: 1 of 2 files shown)


pythonforandroid/toolchain.py
```diff
@@ -792,9 +793,9 @@ def export_dist(self, args):
 
     @property
     def _dist(self):
-        ctx = self.ctx
-        dist = dist_from_args(ctx, self.args)
-        return dist
+        if (not hasattr(self, 'dist') or self.dist is None):
+            self.dist = dist_from_args(self.ctx, self.args)
+        return self.dist
```
AndreMiras: I have to say this feels unnatural somehow:
1) I would simplify the initialisation of the attribute by assigning it to `None` within the constructor
2) It feels weird to have the property being "self._protected" while the member "self.public". Maybe we want to call the member  `__dist` at least? I don't know

pythonforandroid/distribution.py
```diff
@@ -92,15 +95,13 @@ def get_distribution(cls, ctx, name=None, recipes=[],
...
-            ) or dist.ndk_api is None:
+            if ((ndk_api is not None and dist.ndk_api != ndk_api)
+                or dist.ndk_api is None or dist.archs is None):
                 continue
-            for recipe in recipes:
-                if recipe not in dist.recipes:
-                    break
-            else:
+            if (set(recipes).issubset(set(dist.recipes))
+                and set(archs).issubset(set(dist.archs))):
```
AndreMiras: Linter was complaining about these lines:
```
pythonforandroid/distribution.py:99:5: E129 visually indented line with same indent as next logical line
pythonforandroid/distribution.py:102:17: E129 visually indented line with same indent as next logical line
```
Could you fix so Travis finish the build in order to check eventual regressions?

AndreMiras:
Thank you for your contribution :beer: 
I didn't get time to really wrap my head around it and play with it yet. Just left a couple of comments

Aralox:
This pr isn't urgent. Workaround is to run `p4a clean_dists` between compiling for different architectures. I think the other pr currently in progress for upgrading to ndk 19 might already take this into account.

AndreMiras:
Hi @Aralox is it still relevant after changes from https://github.com/kivy/python-for-android/pull/1722 ?
If I understood correctly, we can verify this by compiling for two different archs without running a clean between and checking if the APKs run OK on both. If the issue is still here, the second arch we build for would have compiled dists from the previous one hence should crash, correct?

Aralox:
Hi Andre yes I think that's exactly it. I haven't really kept up with the progress of this project so I'm not sure if the other PR addresses this (it probably does), but yeah that would be the verification. Feel free to close this PR if unnecessary. Cheers :)

inclement:
@AndreMiras As far as I can see, #1722 doesn't affect this and it's still an issue, am I missing something?

opacam:
Yeah, agree with @inclement, I don't see any conflict with the NDK-r19 migration thing...but I just now realize that this will probably conflict with #1926 (as the author of this last one, no problem at all, I will fix conflicts if convenient :wink:).

inclement:
I'm dealing with this in a new PR.

opacam:
@inclement, good to know :+1:, feel free to use the commits I made in  at #1926, related with this arch problem, I see two:
  - [Check android_api and archs when reusing dist](https://github.com/kivy/python-for-android/pull/1926/commits/1e7d0d4112606b30bae1674f3a8352837b558358)
  - [Add an arch suffix to our dists](https://github.com/kivy/python-for-android/pull/1926/commits/096e3e5874ba5ff908b3c705270eb26112e6b97c)

Of course, feel free to implement it whatever way you want, you may have a better idea :smile:

inclement:
@opacam I'm feeling inclined to make a simpler PR that addresses just this for now, in a slightly different way, with #1926 left to merge for the next release. I don't want to merge #1926 quite yet because I'm concerned it may be a little overcomplicated.

opacam:
@inclement, no problem at all, I agree that #1926 complicates things... if we can slim it in some way better for all of us :wink:

so go for this new PR :+1:!

inclement:
Closing as I think this was resolved.

PR #1718, opened 2019-02-19
Fix sdl2_image compile error when arch is x86

j-devel:
Building an apk with `--bootstrap=sdl2` and `--arch=x86` breaks due to failing `pythonforandroid/recipes/sdl2_image/disable_jpg.patch`.  This error is minor regression of #1692 where `disable_jpg.patch` was not upgraded for  SDL2_image 2.0.4.
```
$ p4a apk --private ./myapp --package=org.example.myapp --name myapp --version 0.1 --bootst
rap=sdl2 --requirements=python2 --arch=x86 --debug
```
```
  RAN: /usr/bin/patch -t -d /home/a/.local/share/python-for-android/build/bootstrap_builds/sdl2-python2/jni/SDL2_image -p1 -i /media/sf_vb-share-xubuntu-1804/python-for-android-j-fixSdlImageCompileX86-2/pythonforandroid/recipes/sdl2_image/disable_jpg.patch

  STDOUT:
patching file Android.mk
Hunk #1 FAILED at 6.
1 out of 1 hunk FAILED -- saving rejects to file Android.mk.rej
```
**Fix**
I think this `disable_jpg.patch` is a legacy that was introduced when multi-arch support was being added to the project (f8eb278); and ~~we can safely void it now~~ (I guess it would be the best to 
 ask for a comment from @inclement ).  (I tested compiling without this patch for `x86` and everything works fine including runtime tests.)


inclement:
Thanks, learning that it works without the patch is even better :)

PR #1720, opened 2019-02-19
Generate android version code accounting for arch and min sdk

OptimusGREEN:
Android version code is generated to account for the architecture it is built for as well as the minimum sdk.

The new version format is arch-minsdk-app version

This will allow apps of the same version name built for different architectures be uploaded to the play store and will provide separation for overlap.

example:
app version: 1.0.5
min sdk: 21

old style version code would be 10005 for all architectures

new version codes would be:
x86: 62110005
arm7: 72110005
arm8: 82110005

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -360,10 +360,17 @@ main.py that loads it.''')
 
     version_code = 0
     if not args.numeric_version:
+        # Set version code in format (arch-minsdk-app_version)
+        with open(join(dirname(__file__), 'dist_info.json'), 'r') as dist_info:
+            dist_data = json.load(dist_info)
+        arch = dist_data["archs"][0]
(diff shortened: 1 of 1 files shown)


j-devel:
@OptimusGREEN I just found that apk build for `x86_64` is failing:
```
  File "/home/a/.local/share/python-for-android/dists/unnamed_dist_1/build.py", line 368, in make_package
    arch_code = arch_dict[arch]
KeyError: u'x86_64'
```
Will `x86_64` case be supported? 🤔

OptimusGREEN:
Hmmm, I wasn't aware that x86_64 was supported as Buildozer spec only offers the other 3 as arch options. `The Android arch to build for, choices: armeabi-v7a, arm64-v8a, x86` @inclement we could always add this in if supported?

inclement:
@OptimusGREEN Go ahead, updating the buildozer.spec would be welcome. Thanks!

OptimusGREEN:
updated this to include x86_64 here #1733 and also spec here https://github.com/kivy/buildozer/pull/839

PR #1723, opened 2019-02-20
Fixes psycopg2 lib install dir, closes #1711

### LINKED ISSUE #1711: Psycopg2 error after the apk installation.
erro psycopg

I got this error while debugging the android apk. I associate this to Buildozer because I specified into the buildozer.spec requirements the psycopg2 library. It means that is not working.

How do I fix it? I know that is a recipe for psycopg2 here: https://github.com/kivy/python-for-android/blob/master/pythonforandroid/recipes/psycopg2/__init__.py

How can I add this recipe to my project, to buildozer compile it successfully ?

AndreMiras:
Also note this recipe seems to depend on `nl_langinfo` which is only
available from ANDROID_API 26.

### DIFF
--- a/pythonforandroid/recipes/psycopg2/__init__.py
+++ b/pythonforandroid/recipes/psycopg2/__init__.py
@@ -6,6 +6,9 @@ import sh
 class Psycopg2Recipe(PythonRecipe):
     """
     Requires `libpq-dev` system dependency e.g. for `pg_config` binary.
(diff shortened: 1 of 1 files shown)


AndreMiras:
Yes you're right @inclement I actually was thinking of something like this, or eventually trying to make it possible to still run it even below 26 somehow.
Thanks for the review/merge

PR #1725, opened 2019-02-24
Zworkb services

zworkb:
minor changes for service_only bootstrap:

- service_only was recjeted by build.py (fixed that thanks to @AndreMiras 's hint)
- removed dep on crystax in setup.py in testapps/setup_testapp_service.py
- create own notification channel in PythonService.java (needed for android 8+)
- catch the case that kivy is not in the requirements and thus some args are missing

service_only bootstrap now works with android (7 and 8 tested)


pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -297,9 +297,12 @@ def make_package(args):
...
     env_vars_tarpath = tempfile.mkdtemp(prefix="p4a-extra-env-")
     with open(os.path.join(env_vars_tarpath, "p4a_env_vars.txt"), "w") as f:
-        f.write("P4A_IS_WINDOWED=" + str(args.window) + "\n")
-        if hasattr(args, "orientation"):
-            f.write("P4A_ORIENTATION=" + str(args.orientation) + "\n")
+        try:
+            f.write("P4A_IS_WINDOWED=" + str(args.window) + "\n")
+            if hasattr(args, "orientation"):
+                f.write("P4A_ORIENTATION=" + str(args.orientation) + "\n")
+        except Exception as e:
```
tshirtman: Don't we have a more precise exception to catch? It's always better to be sure the except clause catch the precise error that we are expecting, to avoid mismanaging other errors.

pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonService.java
```diff
@@ -18,6 +18,12 @@
 
 import org.renpy.android.Hardware;
 
+//new imports for channel definition
```
tshirtman: they are new now, but won't be anymore in some time, we can keep the part about them being needed for channel definition.

pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonService.java
```diff
@@ -116,7 +122,31 @@ protected void doStartForeground(Bundle extras) {
...
+            Log.v("python service","android <8: start foregrund withut new channel");
+            startForeground(1, notification);
+        }
+    }
+
+    /**
+     * https://stackoverflow.com/questions/47531742/startforeground-fail-after-upgrade-to-android-8-1
+     *
+     * this change is necessary because from Android 8 on we need our own channel
+     */
```
tshirtman: ditto, it's a change now, but i'll just be part of the code later, which will make the comment weird, just remove "change" from the sentence.

pythonforandroid/bootstraps/common/build/build.py
zworkb: oops sure, we need AttributeError here

pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonService.java
zworkb: god point

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -297,9 +297,12 @@ def make_package(args):
...
     with open(os.path.join(env_vars_tarpath, "p4a_env_vars.txt"), "w") as f:
-        f.write("P4A_IS_WINDOWED=" + str(args.window) + "\n")
-        if hasattr(args, "orientation"):
-            f.write("P4A_ORIENTATION=" + str(args.orientation) + "\n")
+        try:
+            f.write("P4A_IS_WINDOWED=" + str(args.window) + "\n")
+            if hasattr(args, "orientation"):
+                f.write("P4A_ORIENTATION=" + str(args.orientation) + "\n")
+        except AttributeError as e:
+            print ('kivy is not in the requirements')
```
AndreMiras: Nice! Just some minor suggestions, we could simplify that code by killing the `hasattr()` check since we're catching `AttributeError` anyway. Also you can remove the `as e` local variable since you're not using it.
So I suggest the following then:
```python
        try:
            f.write("P4A_IS_WINDOWED=" + str(args.window) + "\n")
            f.write("P4A_ORIENTATION=" + str(args.orientation) + "\n")
        except AttributeError:
            print ('kivy is not in the requirements')
```

pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonService.java
```diff
@@ -116,7 +122,31 @@ protected void doStartForeground(Bundle extras) {
...
+     * necessary because from Android 8 on we need our own channel
+     */
+    private void startForegroundWithNewChannel(int id, Notification notification){
+        String NOTIFICATION_CHANNEL_ID = "org.bd.pureservice"; //TODO: dynamically
+        String channelName = "Pureservice";
+        NotificationChannel chan = new NotificationChannel(NOTIFICATION_CHANNEL_ID, channelName, NotificationManager.IMPORTANCE_NONE);
+        chan.setLightColor(Color.BLUE);
+        chan.setLockscreenVisibility(Notification.VISIBILITY_PRIVATE);
+        NotificationManager manager = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
+        assert manager != null;
```
AndreMiras: Minor question: why would we `assert` this rather than doing something else?
https://developer.android.com/reference/android/content/Context.html#getSystemService(java.lang.String)
> The service name or null if the class is not a supported system service.

So just wondering if we couldn't do something better about it. But the overall change is a great improvement anyway, so let's not bother much more if we can't think about something clever.

pythonforandroid/bootstraps/common/build/build.py
zworkb: ..or I make an explicit check for args.window, perhaps thats cleaner

AndreMiras: Up to you, but I think it's more Pythonic to `EAFP`, see:
https://stackoverflow.com/a/11360880
In any case it's better to not mix both approaches indeed

AndreMiras:
Looks very good overall, thanks for making p4a a better place! We left a couple of comments if you could review/address.
Cheers!

pythonforandroid/bootstraps/common/build/build.py
zworkb: normally I also tend to EAFP, but here we have also to cover the case that we have 'window' and no 'orientatin', then the message would be misleading

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -297,7 +297,8 @@ def make_package(args):
     # Add extra environment variable file into tar-able directory:
     env_vars_tarpath = tempfile.mkdtemp(prefix="p4a-extra-env-")
     with open(os.path.join(env_vars_tarpath, "p4a_env_vars.txt"), "w") as f:
-        f.write("P4A_IS_WINDOWED=" + str(args.window) + "\n")
+        if hasattr(args, 'window'):
```
AndreMiras: nitpik: Please don't change it, but I just have to say it, we're inconsistent here with single and double quotes

pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonService.java
AndreMiras: Resolving as "addressed" thanks!

pythonforandroid/bootstraps/common/build/build.py
zworkb: yes I see that quote styles are quite mixed 50/50 between single and double quotes ;)

pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonService.java
```diff
@@ -116,7 +122,31 @@ protected void doStartForeground(Bundle extras) {
             builder.setSmallIcon(context.getApplicationInfo().icon);
             notification = builder.build();
         }
-        startForeground(1, notification);
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O){
+            Log.v("python service","android 8: start foregrund with new channel");
+            startForegroundWithNewChannel(1, notification);
+        }else{
+            Log.v("python service","android <8: start foregrund withut new channel");
```
AndreMiras: minor typo `s/withut/without/`

AndreMiras:
Hey @zworkb we're almost there! Do you mind taking a final look at the couple of comments not yet addressed so we can test and merge?

pythonforandroid/recipes/android/__init__.py
```diff
@@ -40,7 +40,7 @@ def prebuild_arch(self, arch):
 
         is_sdl2 = bootstrap_name in ('sdl2', 'sdl2python3', 'sdl2_gradle')
         is_pygame = bootstrap_name in ('pygame',)
-        is_webview = bootstrap_name in ('webview',)
+        is_webview = bootstrap_name in ('webview', 'service_only',)
```
JonasT: shouldn't there rather be a "is_serviceonly" added? right now the way it reads doesn't make much sense, which will confuse future coders fiddling with this code

zworkb: @JonasT:

thanks for the hint, i have now added an 'is_service_only' flag as proposed, so the PR should be ready for merge

AndreMiras:
Thanks again for all that hard work. Would you mind rebasing/squashing your changes. I cannot really review since the diff is very confusing and hard to read.  There're changes from other PR that ended up in here somehow.
image

pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonService.java
```diff
@@ -18,6 +18,12 @@
 
 import org.renpy.android.Hardware;
 
+//imports for channel definition
+import android.app.NotificationManager;
+import android.app.NotificationChannel;
+import android.content.Context;
```
AndreMiras: Minor: Already defined line 10

pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonService.java
```diff
@@ -18,6 +18,12 @@
...
+import android.graphics.Color;
+import android.os.Build;
```
AndreMiras: Minor: already defined line 3

pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonService.java
```diff
@@ -90,13 +96,13 @@ public int onStartCommand(Intent intent, int flags, int startId) {
...
         String serviceDescription = extras.getString("serviceDescription");
-
         Notification notification;
         Context context = getApplicationContext();
         Intent contextIntent = new Intent(context, PythonActivity.class);
         PendingIntent pIntent = PendingIntent.getActivity(context, 0, contextIntent,
             PendingIntent.FLAG_UPDATE_CURRENT);
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.HONEYCOMB) {
+
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.O) {
```
AndreMiras: Why are we not checking against `HONEYCOMB`? Is it on purpose or by mistake?

pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonService.java
```diff
@@ -109,7 +115,20 @@ protected void doStartForeground(Bundle extras) {
...
+            // https://stackoverflow.com/questions/47531742/startforeground-fail-after-upgrade-to-android-8-1
+            String NOTIFICATION_CHANNEL_ID = "org.kivy.p4a";    //TODO: make this configurable
+            String channelName = "PythonSerice";                //TODO: make this configurable
+            NotificationChannel chan = new NotificationChannel(NOTIFICATION_CHANNEL_ID, channelName, 
+                NotificationManager.IMPORTANCE_NONE);
+            
+            chan.setLightColor(Color.BLUE);
+            chan.setLockscreenVisibility(Notification.VISIBILITY_PRIVATE);
+            NotificationManager manager = (NotificationManager) getSystemService(Context.NOTIFICATION_SERVICE);
+            assert manager != null;
```
AndreMiras: Minor: If null it would crash the next line with a rather explicit message right? Why not removing this assert then?

pythonforandroid/bootstraps/service_only/build/templates/Service.tmpl.java
```diff
@@ -17,7 +17,7 @@
      * {@inheritDoc}
      */
     @Override
-    public int getStartType() {
+    public int startType() {
```
AndreMiras: I think I've seen this change breaking things in the past. I haven't tried on device (Android 7) yet, but are we sure-ish about it?

AndreMiras:
Thanks for rebasing/squashing. It looks pretty good overall! :clap: 
I still want to test it on Android 7 to see if we don't have regression. My schedule is pretty awful lately, but I'll try asap.

pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonService.java
zworkb: that was on purpose. It is true that HONEYCOMB(value:11)  and O(value:26) stand for Android 8, but only O can be used to query for version ranges.
see [Build.java](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/os/Build.java).
I have tried it and if I use HONEYCOMB for coparison it fails on a device runnng android 7

AndreMiras:
@JonasT I think your comment was addressed, could you review/approve if things are OK to you too?

JonasT:
Looks good :+1: 

---
Unrelated/not merge relevant: *surprise mega confusion* wait what, what's that `sdl2python3` in the code? is that some super old relic? not something that is to be fixed by this pull request of course since it's in the code part left alone, but I've never seen that and I rearranged all the bootstraps so that's weird :smile:

PR #1727, opened 2019-02-24
Add ci_mode to disable download progress

mkg20001:
Otherwise Travis CI complains about log size

Ref kivy/buildozer#823

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,4 +32,4 @@ before_script:
 
 script:
   - docker build --tag=p4a --file Dockerfile.py3 .
-  - docker run p4a /bin/sh -c "$COMMAND"
+  - docker run -e CI p4a /bin/sh -c "$COMMAND"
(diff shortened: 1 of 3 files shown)


AndreMiras:
Very nice, thank you! May I suggest to directly use something from TravisCI?
https://docs.travis-ci.com/user/environment-variables/#default-environment-variables
What I mean is to replace `CI_MODE` directly by `CI` so we don't have to bother setting it in `.travis.yml` since it's already automatically set by Travis. In a same fashion there's a predefined variable called `TRAVIS`, but then it feels a bit coupled. If you prefer you can also use the predefined `CONTINUOUS_INTEGRATION` variable rather than `CI`, I leave that to you

mkg20001:
@AndreMiras I've switched to the `$CI` variable. Both in this pr and https://github.com/kivy/buildozer/pull/833

mkg20001:
I've just noticed: The android SDK installs still continue to spam the logs

screenshot

Would somebody want to look into that?

AndreMiras:
Thanks for addressing the comments!
Yes the logs are still showing it because we still need to pass the environment variable from the host to the Docker container. So basically in `.travis.yml` you can edit the `docker run` command to add `-e CI`, so final command should be something like:
`docker run -e CI p4a /bin/sh -c "$COMMAND"`

PR #1733, opened 2019-02-26
Add missing arch to version code generator.

OptimusGREEN:
x86_64 missing from version code generator.

PR #1734, opened 2019-02-26
Fixes object is not subscriptable error, refs #1733

AndreMiras:
Follow up fix for #1733 (untested)


OptimusGREEN:
Thanks. Was a stressful day.  😁

PR #1738, opened 2019-03-03
fix groestlcoin_hash recipe

tshirtman:
apparently since the move to python3, the recipes can/should disable call_hostpython_via_targetpython.

### DIFF
--- a/pythonforandroid/recipes/groestlcoin_hash/__init__.py
+++ b/pythonforandroid/recipes/groestlcoin_hash/__init__.py
@@ -5,7 +5,6 @@ class GroestlcoinHashRecipe(CythonRecipe):
     version = '1.0.1'
     url = 'https://github.com/Groestlcoin/groestlcoin-hash-python/archive/{version}.tar.gz'
     depends = []
-    call_hostpython_via_targetpython = False
     cythonize = False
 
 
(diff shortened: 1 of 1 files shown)


pythonforandroid/recipes/groestlcoin_hash/__init__.py
```diff
@@ -5,7 +5,7 @@ class GroestlcoinHashRecipe(CythonRecipe):
     version = '1.0.1'
     url = 'https://github.com/Groestlcoin/groestlcoin-hash-python/archive/{version}.tar.gz'
     depends = []
-    call_hostpython_via_targetpython = True
+    call_hostpython_via_targetpython = False
```
AndreMiras: Thanks for the fix!
Why don't we drop that line directly since this is `CythonRecipe` default anyway?
https://github.com/kivy/python-for-android/blob/0.7.0/pythonforandroid/recipe.py#L958
Approving, let's wait to see if the build turns green before merging :pray:

tshirtman: Oh, hadn't seen that it was set to False here, had only checked in the base Recipe, where it's True by default, will update.

PR #1741, opened 2019-03-06
Requests runtime permissions list, fixes #1704

### LINKED ISSUE #1704: Match official requestPermissions interface
The Android API for requesting runtime permission has the following signature:
```java
void requestPermissions (Activity activity,
                String[] permissions,
                int requestCode)
```
Currently via our `PythonActivity.java` class, we're providing:
```java
public void requestNewPermission(String permission)
```
See https://github.com/kivy/python-for-android/blob/02519cc/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonActivity.java#L608
It would be great if our interface was also somewhat matching the official name/signature.

- official documentation [ActivityCompat#requestPermissions()](https://developer.android.com/reference/android/support/v4/app/ActivityCompat#requestPermissions(android.app.Activity,%20java.lang.String[],%20int))
- PR discussion https://github.com/kivy/python-for-android/pull/1528#discussion_r255345553

This would be a non-breaking change since we can keep our current `requestNewPermission()` method and just add the new `requestPermissions()` one.

AndreMiras:
Matches the official Android API via, making it possible to request
multiple runtime permissions at once.

Keeps the previous `android.permissions.request_permission()` interface
for compatibility.

Bumps pyjnius to post 1.2.0 release that solves required method
resolution issues. On 1.2.0, the error was:
```
jnius.jnius.JavaException: No methods matching your arguments, available: []
```

### DIFF
--- a/doc/source/apis.rst
+++ b/doc/source/apis.rst
@@ -15,8 +15,8 @@ to access the SD card, the camera, and other things.
 This can be done through the `android` module which is *available per default*
 unless you blacklist it. Use it in your app like this::
 
-      from android.permissions import request_permission, Permission
-      request_permission(Permission.WRITE_EXTERNAL_STORAGE)
+      from android.permissions import request_permissions, Permission
+      request_permissions([Permission.WRITE_EXTERNAL_STORAGE])
 
 The available permissions are listed here:
 
--- a/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonActivity.java
+++ b/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonActivity.java
@@ -604,17 +604,16 @@ public class PythonActivity extends SDLActivity {
     }
 
     /**
-     * Used by android.permissions p4a module to request a permission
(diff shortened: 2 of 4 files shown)


AndreMiras:
Tested with:
```python
request_permissions([Permission.CAMERA, Permission.WRITE_EXTERNAL_STORAGE])
```
As we can see in the attached screenshot (:es: sorry), it did popup a page with two permission requests.

image
image

JonasT:
Looks good and backwards compatible :+1: do you want me to test it as well with my app?

AndreMiras:
Thanks @JonasT for the review. Yes feel free to test it. Keep in mind it requires pyjnius update otherwise it would crash

JonasT:
@AndreMiras oh huh I am just thinking, the docs example might also need an update, right? the one given here: https://github.com/kivy/python-for-android/blob/master/doc/source/apis.rst#runtime-permissions

AndreMiras:
Thanks @JonasT for the review and the heads-up, I've updated the PR with the documentation changes

PR #1742, opened 2019-03-07
Updated sdl2_mixer version to 2.0.4

misl6:
This PR updates sdl2_mixer version to 2.0.4

Tested on a real-world app w/ python3

### DIFF
--- a/pythonforandroid/recipes/sdl2_mixer/toggle_modplug_mikmod_smpeg_ogg.patch
+++ b/pythonforandroid/recipes/sdl2_mixer/toggle_modplug_mikmod_smpeg_ogg.patch
@@ -1,28 +1,25 @@
-diff -Naur SDL2_mixer/Android.mk SDL2_mixer-1/Android.mk
---- SDL2_mixer/Android.mk	2019-03-07 21:07:42.149457400 +0100
-+++ SDL2_mixer-1/Android.mk	2019-03-07 21:05:06.851338800 +0100
-@@ -3,20 +3,20 @@
+--- orig/Android.mk	2018-10-31 15:58:59.000000000 +0100
++++ patched/Android.mk	2019-03-08 12:11:35.271258500 +0100
+@@ -3,7 +3,7 @@
 
 
  # Enable this if you want to support loading FLAC music with libFLAC
--SUPPORT_FLAC := false
-+SUPPORT_FLAC ?= true
+-SUPPORT_FLAC ?= true
++SUPPORT_FLAC := false
  FLAC_LIBRARY_PATH := external/flac-1.3.2
 
  # Enable this if you want to support loading OGG Vorbis music via Tremor
--SUPPORT_OGG := true
-+SUPPORT_OGG ?= true
- OGG_LIBRARY_PATH := external/libogg-1.3.2
+@@ -12,11 +12,11 @@
  VORBIS_LIBRARY_PATH := external/libvorbisidec-1.2.1
 
  # Enable this if you want to support loading MP3 music via MPG123
--SUPPORT_MP3_MPG123 := false
-+SUPPORT_MP3_MPG123 ?= true
+-SUPPORT_MP3_MPG123 ?= true
++SUPPORT_MP3_MPG123 := false
  MPG123_LIBRARY_PATH := external/mpg123-1.25.6
 
  # Enable this if you want to support loading MOD music via modplug
--SUPPORT_MOD_MODPLUG := false
-+SUPPORT_MOD_MODPLUG ?= true
+-SUPPORT_MOD_MODPLUG ?= true
++SUPPORT_MOD_MODPLUG := false
  MODPLUG_LIBRARY_PATH := external/libmodplug-0.8.9.0
 
  # Enable this if you want to support TiMidity
--- a/pythonforandroid/recipes/sdl2_mixer/toggle_modplug_mikmod_smpeg_ogg.patch
+++ b/pythonforandroid/recipes/sdl2_mixer/toggle_modplug_mikmod_smpeg_ogg.patch
@@ -1,31 +1,28 @@
 diff -Naur SDL2_mixer/Android.mk SDL2_mixer-1/Android.mk
---- SDL2_mixer/Android.mk	2019-03-07 16:14:49.468477800 +0100
-+++ SDL2_mixer-1/Android.mk	2019-03-07 16:10:22.371142900 +0100
-@@ -3,24 +3,24 @@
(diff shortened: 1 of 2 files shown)


pythonforandroid/recipes/sdl2_mixer/toggle_modplug_mikmod_smpeg_ogg.patch
```diff
@@ -1,29 +1,31 @@
...
- # Enable this if you want to support loading MOD music via mikmod
- # The library path should be a relative path to this directory.
--SUPPORT_MOD_MIKMOD ?= true
-+SUPPORT_MOD_MIKMOD := false
- MIKMOD_LIBRARY_PATH := external/libmikmod-3.1.12
- 
- # Enable this if you want to support loading MP3 music via SMPEG
- # The library path should be a relative path to this directory.
--SUPPORT_MP3_SMPEG ?= true
-+SUPPORT_MP3_SMPEG := false
```
AndreMiras: any reason why we don't want to support this things anymore? Is it for licensing issues?

misl6: @AndreMiras , after a double-check, SUPPORT_MOD_MODPLUG, SUPPORT_MOD_MIKMOD, SUPPORT_MP3_SMPEG were already disabled in 2.0.1 .

Speaking about ogg and Timidity, nice catch, i was thinking that We weren't supporting these features.
If needed, i could test the sdl2_mixer implem with support for timitidy and ogg and push a fix for that.

AndreMiras: well yes basically if the pull request is just about bumping the version, by default I would not toggle enable/disable things. I mean I would just leave it as it is

JonasT: Also just posted this in discord: SMPEG and MIKMOD seem LGPL, it is really important to always also check licensing issues. LGPL really is problematic not because it's infectious but because of the induced requirements (providing the source and a way to swap it out) which many p4a users will forget about it. So my vote would be to disable it, at least per default

JonasT: @misl6 it looks to me like `SUPPORT_MP3_SMPEG := false` is still removed in your pull request. Is that intentional? Did your flags revert go wrong?

pythonforandroid/recipes/sdl2_mixer/toggle_modplug_mikmod_smpeg_ogg.patch
```diff
@@ -1,29 +1,28 @@
...
- OGG_LIBRARY_PATH := external/libogg-1.3.1
+-SUPPORT_OGG := true
++SUPPORT_OGG ?= true
+ OGG_LIBRARY_PATH := external/libogg-1.3.2
  VORBIS_LIBRARY_PATH := external/libvorbisidec-1.2.1
- 
+
+ # Enable this if you want to support loading MP3 music via MPG123
+-SUPPORT_MP3_MPG123 := false
++SUPPORT_MP3_MPG123 ?= true
```
JonasT: this library is LGPL. I don't think it should be included per default, I think that's a problematic change

pythonforandroid/recipes/sdl2_mixer/toggle_modplug_mikmod_smpeg_ogg.patch
```diff
@@ -1,29 +1,28 @@
...
- # Enable this if you want to support loading MOD music via modplug
- # The library path should be a relative path to this directory.
--SUPPORT_MOD_MODPLUG ?= true
-+SUPPORT_MOD_MODPLUG := false
- MODPLUG_LIBRARY_PATH := external/libmodplug-0.8.8.4
- 
- # Enable this if you want to support loading MOD music via mikmod
- # The library path should be a relative path to this directory.
--SUPPORT_MOD_MIKMOD ?= true
-+SUPPORT_MOD_MIKMOD := false
```
JonasT: libmikmod also appears to be LGPL. It would therefore also be reasonable not to be included by default, but your pull request seems to remove the `SUPPORT_MOD_MIKMOD := false`. Is that intentional?

pythonforandroid/recipes/sdl2_mixer/toggle_modplug_mikmod_smpeg_ogg.patch
JonasT: @AndreMiras for what it's worth, this is currently set to `false`/disabled on master. So removing this diff probably **enables** it in this pull request - which I think would be problematic because of the licensing. I think it's better left out, LGPL in `.apk` isn't an unsolvable problem but can bite developers who are unaware this will be included

JonasT: @misl6 @AndreMiras oops, poked around some more and SMPEG was just removed entirely from SDL2_Mixer, so false alarm :smile: sorry for the comment spam regarding this particular flag

pythonforandroid/recipes/sdl2_mixer/toggle_modplug_mikmod_smpeg_ogg.patch
misl6: SUPPORT_MOD_MIKMOD flag is not available anymore cause mikmod library is been removed since sdl2_mixer release 2.0.2

pythonforandroid/recipes/sdl2_mixer/toggle_modplug_mikmod_smpeg_ogg.patch
JonasT: @misl6 can you explain why you marked this conversation as resolved? Unless it's a fairly obvious case (in which it would be still somewhat expected you spell it out why you resolved it), please don't resolve things brought up by others because people may then forget to keep track of them

pythonforandroid/recipes/sdl2_mixer/toggle_modplug_mikmod_smpeg_ogg.patch
JonasT: right, another one of the removed ones I missed, makes sense :+1: thanks for the info

inclement:
Thanks!

Issue #1093, opened 2017-08-21
"android" recipe isn't compatible with Python 3


germn:
apk built with "android" recipe and Python 3 will crash after starup:

> 06-05 12:45:47.101 30382 30533 I python  :    File "/data/user/0/org.voctrain.voctrain/files/app/crystax_python/site-packages/kivy/base.py", line 210, in remove_android_splash
> 06-05 12:45:47.111 30382 30533 I python  :      from android import remove_presplash
> 06-05 12:45:47.111 30382 30533 I python  :    File "/data/user/0/org.voctrain.voctrain/files/app/crystax_python/site-packages/android/__init__.py", line 8, in <module>
> 06-05 12:45:47.111 30382 30533 I python  :      from android._android import *
> 06-05 12:45:47.111 30382 30533 I python  :    File "android/_android.pyx", line 178, in init android._android (android/_android.c:6886)
> 06-05 12:45:47.111 30382 30533 I python  :  TypeError: can't concat bytes to str
> 06-05 12:45:47.391 30382 30533 I python  : Python for android ended.

Originally reported [here](https://groups.google.com/forum/#!topic/kivy-users/CgGTCN1TG2w), I faced same issue.

inclement:
Thanks for the report, I'll look into it, I guess it shouldn't be hard to fix.

For what it's worth, the android module isn't essential, and most of its functionality can be accomplished easily using pyjnius, so you may be able to use that as a workaround. That is probably why I haven't noticed this before.

inclement:
I looked into this but couldn't reproduce it. Am I right in thinking that the crash should occur immediately on import of the android module?

germn:
@inclement I didn't import `android` module at all. I add `android` recipe as dependency only [to avoid](https://github.com/kivy/python-for-android/issues/963) logging message "Could not remove android presplash".

I also use this code inside app's build method, it's probably related:

```
if platform == "android":
    from jnius import autoclass
    activity = autoclass('org.kivy.android.PythonActivity').mActivity
    activity.removeLoadingScreen()
```

Sorry for not mentioning it's earlier.

inclement:
I see, there shouldn't be anything special there (in fact, I remove the presplash this way myself). There are no disadvantages to just not including the android module and letting Kivy give the warning.

Looking at the code, I don't actually see why this error should occur, as neither of the strings being combined looks like bytes.

germn:
[...]

PR #1747, opened 2019-03-10
decode JAVA_NAMESPACE in _android.pyx

tshirtman:
not sure why i need that if others don't, but i can't build anything depending on the android recipe without it.

### DIFF
--- a/pythonforandroid/recipes/android/src/android/_android.pyx
+++ b/pythonforandroid/recipes/android/src/android/_android.pyx
@@ -175,7 +175,7 @@ api_version = autoclass('android.os.Build$VERSION').SDK_INT
 version_codes = autoclass('android.os.Build$VERSION_CODES')
 
 
-python_act = autoclass(JAVA_NAMESPACE + u'.PythonActivity')
+python_act = autoclass(JAVA_NAMESPACE.decode('utf8') + u'.PythonActivity')
 Rect = autoclass(u'android.graphics.Rect')
 mActivity = python_act.mActivity
 if mActivity:


misl6:
I'm also for that. (actually I'm patching it every time manually)
Never did a PR to fix it cause once I've read issue #1093,  and I was overthinking that the problem was just mine.

AndreMiras:
FWIW a similar issue got addressed here https://github.com/kivy/python-for-android/pull/1475

misl6:
> LGTM thanks!
> It probably happens happens only on host python3 and `pygame` bootstrap, right?
> https://github.com/kivy/python-for-android/blob/0.7.0/pythonforandroid/recipes/android/__init__.py#L51

@AndreMiras , also on python3+sdl2

tshirtman:
yeah, i'm pretty sure sdl2 is what's used in my build. And yeah i'm confused as well because the code for sdl2 for this declaration seems to make it an unicode string, but i still get the issue.

tshirtman:
merged, we'll see if that cause issues to others and learn more when that happens i guess.

idevo89:
I gett an AttributeError after the change with python3

03-10 21:40:33.812  7532  7554 I python  :    File "android/_android.pyx", line 178, in init android._android
03-10 21:40:33.812  7532  7554 I python  :  AttributeError: 'str' object has no attribute 'decode'

AndreMiras:
Looks like we're gonna "learn more when that happens" soon :smile: 
So maybe it's time to dig into it further since it fixes for some people, but breaks for other.
@tshirtman could you trace on your side what happens there https://github.com/kivy/python-for-android/blob/0.7.0/pythonforandroid/recipes/android/__init__.py#L51
Because basically `JAVA_NAMESPACE.decode('utf8')` should only work if it's `bytes` on python3 correct? But from the code shared above we see it's not supposed to be `bytes` on `sdl2` but `str` right?
So one thing I'm curious about is what code path makes your `JAVA_NAMESPACE` being a `bytes`.
Could you take a look? @misl6 can you also verify the same.
As for @idevo89 from what I see yes we definitely seemed to have overlooked it and your error given the current status is the one that definitely makes more sense on `sdl2` and python3.

tito:
This is a very bad idea. If this is broken, you can expect many others stuff to be broken as well (anybody that actually does the same concatenation). if JAVA_NAMESPACE is bytes, then it's just wrong no?

I bet it's just about cython running with -3 or not

gabrielbrise:
I'm stuck at this for the whole day, it doesn't seem to fix by changing the _android.pyx (both occurrences) that I found inside the /.buildozer that is created inside my project's folder.
What needs to be done other than changing the files and executing buildozer to compile a new apk? I was thinking maybe some cached stuff that was created when .decode was still there could cause it.

`python  :    File "android/_android.pyx", line 178, in init android._android
python  :  AttributeError: 'str' object has no attribute 'decode'`

Yes, I am new to programming in general so I could have made a simple mistake or said some non-sense above.

gabrielbrise:
> I'm stuck at this for the whole day, it doesn't seem to fix by changing the _android.pyx (both occurrences) that I found inside the /.buildozer that is created inside my project's folder.
> What needs to be done other than changing the files and executing buildozer to compile a new apk? I was thinking maybe some cached stuff that was created when .decode was still there could cause it.
> 
> `python : File "android/_android.pyx", line 178, in init android._android python : AttributeError: 'str' object has no attribute 'decode'`
> 
> Yes, I am new to programming in general so I could have made a simple mistake or said some non-sense above.

Just needed to use `buildozer android clean`
After that the fix worked and I was able to compile my first succesful apk!

PR #1756, opened 2019-03-15
[xeddsa] fixed shared library copying

goffi-contrib:
The copy of the build shared library was not working anymore after the
update of python2 build chain. This patch fixes it by using
self.ctx.get_site_packages_dir().

### DIFF
--- a/pythonforandroid/recipes/xeddsa/__init__.py
+++ b/pythonforandroid/recipes/xeddsa/__init__.py
@@ -24,12 +24,7 @@ class XedDSARecipe(CythonRecipe):
                 hostpython, 'ref10/build.py',
                 _env=env
             )
-            python_version = self.ctx.python_recipe.version[0:3]
-            site_packages_dir = 'lib/python{python_version}/site-packages'.format(
-                python_version=python_version)
-            site_packages = join(self.ctx.get_python_install_dir(),
-                                 site_packages_dir)
-            shprint(sh.cp, '_crypto_sign.so', site_packages)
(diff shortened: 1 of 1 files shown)


AndreMiras:
It still seems to be failing with that same `_crypto_sign.so` error. I haven't tried to reproduce it locally yet. 
If this works for you locally then we may have an issue with our Docker env.
Let me know if you want me to take a look

goffi-contrib:
@AndreMiras yes it does work for me locally. Would be nice if you could have a look indeed, thanks :)

AndreMiras:
OK I've played with it in Docker.
```
(venv) user@f58394473c86:~/testapps$ find ~/ -name "*crypto_sign*.so*"
/home/user/.local/share/python-for-android/build/other_builds/xeddsa-python3/armeabi-v7a__ndk_target_21/xeddsa/ref10/bin/libcrypto_sign_dynamic.so
/home/user/.local/share/python-for-android/build/other_builds/xeddsa-python3/armeabi-v7a__ndk_target_21/xeddsa/_crypto_sign.cpython-37m-x86_64-linux-gnu.so
(venv) user@f58394473c86:~/testapps$
```
So the library actually is `_crypto_sign.cpython-37m-x86_64-linux-gnu.so` on Ubuntu. So I guess we should cp with a wildcard. I'll give it a try locally, update the PR and merge if it builds

AndreMiras:
Conditional recipe build is green https://travis-ci.org/kivy/python-for-android/jobs/512985553
Relevant part of the log: `copy ./_crypto_sign.cpython-37m-x86_64-linux-gnu.so`
Merging, thanks again for the contrib!

PR #1760, opened 2019-03-16
[omemo-backend-signal] updated to 0.2.3

goffi-contrib:
"wireformat.patch" has been removed because it is not needed anymore since the recent update of protobuf recipe.


AndreMiras:
Thanks for the PR, the build is currently failing because of the `DoubleRatchet` download as you were mentioning. I'd like us to spend some time to investigate this too before we can merge

AndreMiras:
Now it's failing because the `xeddsa` recipe error `/bin/cp: cannot stat '_crypto_sign.so': No such file or directory`.
This probably needs to be addressed in https://github.com/kivy/python-for-android/pull/1756

AndreMiras:
Build is now passing, merging, thanks for the contrib :smile:

PR #1761, opened 2019-03-16
[cryptography] updated to 2.6.1

goffi-contrib:


AndreMiras:
LGTM, but conditional build is failing. I think it's time to add `libssl-dev` dependency to our Docker image.
Could you add the dep to that line https://github.com/kivy/python-for-android/blob/9736243/Dockerfile.py3#L114 for instance?

AndreMiras:
So it should actually reference to the local `openssl` recipe and for some reason it doesn't.
If I compile `openssl` recipe explicitly first then it works, otherwise it doesn't seem to properly reference to it. I'd like to investigate further

AndreMiras:
So it looks like a bug with the dependency resolution, we'll be tracking it here https://github.com/kivy/python-for-android/issues/1764

goffi-contrib:
@AndreMiras done

AndreMiras:
Awesome work, it builds now. I've merged it, can you rebase your other pull requests on master so we have that fix in place?
By the way thanks for always squashing your commits :pray:

PR #1762, opened 2019-03-16
[doubleratchet] removed this recipe

goffi-contrib:


pythonforandroid/recipes/doubleratchet/__init__.py
```diff
@@ -3,7 +3,7 @@
 
 class DoubleRatchetRecipe(PythonRecipe):
     name = 'doubleratchet'
-    version = '0.4.0'
+    version = '0.6.0'
     url = 'https://pypi.python.org/packages/source/D/DoubleRatchet/DoubleRatchet-{version}.tar.gz'
```
AndreMiras: I did experience the issue you're talking about regarding the slow download on that package.
Maybe we want to give a try to the github release instead.
`https://github.com/Syndace/python-doubleratchet/archive/v0.6.0-beta.tar.gz`
What do you think?
Otherwise that won't probably compile in our CI, so you could add it to the list of "broken recipe" here? https://github.com/kivy/python-for-android/blob/6c31e10/ci/constants.py

goffi-contrib:
@AndreMiras I've been testing without the recipe, it's working (DoubleRatchet is pure python module). 

I've tried to remove `omemo-bakend-signal` too, but when it's installed as pure python, dependencies are not installed correctly (notably DoubleRatchet is not present in the final distribution).

PR #1765, opened 2019-03-17
Fixes libglob recipe, closes #1399

### LINKED ISSUE #1399: Broken libglob recipe
`libglob` recipe compilation fails for the exact same reason problem as ifaddrs.
See details https://github.com/kivy/python-for-android/issues/1398

AndreMiras:
- makes sure the compiler is being used rather than ccache alone
- do not manually copy over header files


AndreMiras:
I'm not using this recipe, but broken recipes make me sad so I had to fix it

PR #1772, opened 2019-03-25
Properly search native lib dir in ctypes, fixes #1770

### LINKED ISSUE #1770: ctypes.util.find_library() doesn't work on arch arm64-v8a
### Versions

* Python: 3.6.7
* OS: Ubuntu 18.10
* Kivy: N/A
* Cython: 0.29.6

### Description

Build for arch `arm64-v8a` succeeds but raises a `FileNotFoundError` during execution *(see logcat below)*

According to @inclement, this might be related to  #1701 

### buildozer.spec

*Not using buildozer.*

**setup.py**
```python
from distutils.core import setup
from setuptools import find_packages


options = {'apk': {'debug': None,
                   'bootstrap': 'sdl2',
                   'requirements': 'libffi,sdl2,sdl2_image,python3,pysdl2,Pillow',
                   'package': 'org.some.put',
                   'android-api': 28,
                   'arch': 'arm64-v8a',
                   'dist-name': 'bdisttest',
                   }}

setup(
    name='foo',
    version='0.1',
    description='bar',
    author='me',
    author_email='foo@bar.com',
    packages=find_packages(),
    options=options,
    package_data={'foobar': ['*.py', 'gui/*.py', 'assets/images/*.png']}
)
```

### Logs

**adb logcat | grep python**
```
03-24 20:29:10.692  2554  2554 V pythonutil: Checking pattern libcrystax\.so against libpython3.7m.so
03-24 20:29:10.692  2554  2554 V pythonutil: Checking pattern libcrystax\.so against libssl1.1.so
03-24 20:29:10.692  2554  2554 V pythonutil: Checking pattern libcrystax\.so against libSDL2_mixer.so
03-24 20:29:10.692  2554  2554 V pythonutil: Checking pattern libcrystax\.so against libSDL2_ttf.so
03-24 20:29:10.692  2554  2554 V pythonutil: Checking pattern libcrystax\.so against libffi.so
03-24 20:29:10.692  2554  2554 V pythonutil: Checking pattern libcrystax\.so against libmain.so
03-24 20:29:10.692  2554  2554 V pythonutil: Checking pattern libcrystax\.so against libSDL2.so
03-24 20:29:10.692  2554  2554 V pythonutil: Checking pattern libcrystax\.so against libhidapi.so
03-24 20:29:10.693  2554  2554 V pythonutil: Checking pattern libcrystax\.so against libsqlite3.so
03-24 20:29:10.693  2554  2554 V pythonutil: Checking pattern libcrystax\.so against libcrypto1.1.so
03-24 20:29:10.693  2554  2554 V pythonutil: Checking pattern libcrystax\.so against libSDL2_image.so
03-24 20:29:10.693  2554  2554 V pythonutil: Checking pattern libsqlite3\.so against libpython3.7m.so
03-24 20:29:10.693  2554  2554 V pythonutil: Checking pattern libsqlite3\.so against libssl1.1.so
03-24 20:29:10.694  2554  2554 V pythonutil: Checking pattern libsqlite3\.so against libSDL2_mixer.so
03-24 20:29:10.694  2554  2554 V pythonutil: Checking pattern libsqlite3\.so against libSDL2_ttf.so
03-24 20:29:10.694  2554  2554 V pythonutil: Checking pattern libsqlite3\.so against libffi.so
03-24 20:29:10.694  2554  2554 V pythonutil: Checking pattern libsqlite3\.so against libmain.so
03-24 20:29:10.694  2554  2554 V pythonutil: Checking pattern libsqlite3\.so against libSDL2.so
03-24 20:29:10.694  2554  2554 V pythonutil: Checking pattern libsqlite3\.so against libhidapi.so
03-24 20:29:10.694  2554  2554 V pythonutil: Checking pattern libsqlite3\.so against libsqlite3.so
03-24 20:29:10.694  2554  2554 V pythonutil: Pattern libsqlite3\.so matched file libsqlite3.so
03-24 20:29:10.694  2554  2554 V pythonutil: Checking pattern libsqlite3\.so against libcrypto1.1.so
03-24 20:29:10.695  2554  2554 V pythonutil: Checking pattern libsqlite3\.so against libSDL2_image.so
03-24 20:29:10.695  2554  2554 V pythonutil: Checking pattern libffi\.so against libpython3.7m.so
03-24 20:29:10.695  2554  2554 V pythonutil: Checking pattern libffi\.so against libssl1.1.so
03-24 20:29:10.696  2554  2554 V pythonutil: Checking pattern libffi\.so against libSDL2_mixer.so
03-24 20:29:10.696  2554  2554 V pythonutil: Checking pattern libffi\.so against libSDL2_ttf.so
03-24 20:29:10.696  2554  2554 V pythonutil: Checking pattern libffi\.so against libffi.so
03-24 20:29:10.696  2554  2554 V pythonutil: Pattern libffi\.so matched file libffi.so
03-24 20:29:10.696  2554  2554 V pythonutil: Checking pattern libffi\.so against libmain.so
03-24 20:29:10.696  2554  2554 V pythonutil: Checking pattern libffi\.so against libSDL2.so
03-24 20:29:10.696  2554  2554 V pythonutil: Checking pattern libffi\.so against libhidapi.so
03-24 20:29:10.696  2554  2554 V pythonutil: Checking pattern libffi\.so against libsqlite3.so
03-24 20:29:10.696  2554  2554 V pythonutil: Checking pattern libffi\.so against libcrypto1.1.so
03-24 20:29:10.696  2554  2554 V pythonutil: Checking pattern libffi\.so against libSDL2_image.so
03-24 20:29:10.697  2554  2554 V pythonutil: Checking pattern libssl.*\.so against libpython3.7m.so
03-24 20:29:10.697  2554  2554 V pythonutil: Checking pattern libssl.*\.so against libssl1.1.so
03-24 20:29:10.697  2554  2554 V pythonutil: Pattern libssl.*\.so matched file libssl1.1.so
03-24 20:29:10.697  2554  2554 V pythonutil: Checking pattern libssl.*\.so against libSDL2_mixer.so
03-24 20:29:10.697  2554  2554 V pythonutil: Checking pattern libssl.*\.so against libSDL2_ttf.so
03-24 20:29:10.697  2554  2554 V pythonutil: Checking pattern libssl.*\.so against libffi.so
03-24 20:29:10.698  2554  2554 V pythonutil: Checking pattern libssl.*\.so against libmain.so
03-24 20:29:10.698  2554  2554 V pythonutil: Checking pattern libssl.*\.so against libSDL2.so
03-24 20:29:10.698  2554  2554 V pythonutil: Checking pattern libssl.*\.so against libhidapi.so
03-24 20:29:10.698  2554  2554 V pythonutil: Checking pattern libssl.*\.so against libsqlite3.so
03-24 20:29:10.698  2554  2554 V pythonutil: Checking pattern libssl.*\.so against libcrypto1.1.so
03-24 20:29:10.698  2554  2554 V pythonutil: Checking pattern libssl.*\.so against libSDL2_image.so
03-24 20:29:10.698  2554  2554 V pythonutil: Checking pattern libcrypto.*\.so against libpython3.7m.so
03-24 20:29:10.699  2554  2554 V pythonutil: Checking pattern libcrypto.*\.so against libssl1.1.so
03-24 20:29:10.699  2554  2554 V pythonutil: Checking pattern libcrypto.*\.so against libSDL2_mixer.so
03-24 20:29:10.699  2554  2554 V pythonutil: Checking pattern libcrypto.*\.so against libSDL2_ttf.so
03-24 20:29:10.699  2554  2554 V pythonutil: Checking pattern libcrypto.*\.so against libffi.so
03-24 20:29:10.699  2554  2554 V pythonutil: Checking pattern libcrypto.*\.so against libmain.so
03-24 20:29:10.699  2554  2554 V pythonutil: Checking pattern libcrypto.*\.so against libSDL2.so
03-24 20:29:10.699  2554  2554 V pythonutil: Checking pattern libcrypto.*\.so against libhidapi.so
03-24 20:29:10.699  2554  2554 V pythonutil: Checking pattern libcrypto.*\.so against libsqlite3.so
03-24 20:29:10.700  2554  2554 V pythonutil: Checking pattern libcrypto.*\.so against libcrypto1.1.so
03-24 20:29:10.700  2554  2554 V pythonutil: Pattern libcrypto.*\.so matched file libcrypto1.1.so
03-24 20:29:10.700  2554  2554 V pythonutil: Checking pattern libcrypto.*\.so against libSDL2_image.so
03-24 20:29:10.700  2554  2554 V pythonutil: Loading library: sqlite3
03-24 20:29:10.702  2554  2554 V pythonutil: Loading library: ffi
03-24 20:29:10.703  2554  2554 V pythonutil: Loading library: SDL2
03-24 20:29:10.706  2554  2554 V pythonutil: Loading library: SDL2_image
03-24 20:29:10.708  2554  2554 V pythonutil: Loading library: SDL2_mixer
03-24 20:29:10.710  2554  2554 V pythonutil: Loading library: SDL2_ttf
03-24 20:29:10.712  2554  2554 V pythonutil: Loading library: ssl1.1
03-24 20:29:10.714  2554  2554 V pythonutil: Loading library: crypto1.1
03-24 20:29:10.715  2554  2554 V pythonutil: Loading library: python2.7
03-24 20:29:10.716  2554  2554 V pythonutil: Library loading error: dalvik.system.PathClassLoader[DexPathList[[zip file "/data/app/com.tizilogic.pyos-UUr8o1SWqdGHXQVCTO8uLQ==/base.apk"],nativeLibraryDirectories=[/data/app/com.tizilogic.pyos-UUr8o1SWqdGHXQVCTO8uLQ==/lib/arm64, /data/app/com.tizilogic.pyos-UUr8o1SWqdGHXQVCTO8uLQ==/base.apk!/lib/arm64-v8a, /system/lib64, /system/vendor/lib64]]] couldn't find "libpython2.7.so"
03-24 20:29:10.717  2554  2554 V pythonutil: Loading library: python3.5m
03-24 20:29:10.717  2554  2554 V pythonutil: Library loading error: dalvik.system.PathClassLoader[DexPathList[[zip file "/data/app/com.tizilogic.pyos-UUr8o1SWqdGHXQVCTO8uLQ==/base.apk"],nativeLibraryDirectories=[/data/app/com.tizilogic.pyos-UUr8o1SWqdGHXQVCTO8uLQ==/lib/arm64, /data/app/com.tizilogic.pyos-UUr8o1SWqdGHXQVCTO8uLQ==/base.apk!/lib/arm64-v8a, /system/lib64, /system/vendor/lib64]]] couldn't find "libpython3.5m.so"
03-24 20:29:10.717  2554  2554 V pythonutil: Loading library: python3.6m
03-24 20:29:10.718  2554  2554 V pythonutil: Library loading error: dalvik.system.PathClassLoader[DexPathList[[zip file "/data/app/com.tizilogic.pyos-UUr8o1SWqdGHXQVCTO8uLQ==/base.apk"],nativeLibraryDirectories=[/data/app/com.tizilogic.pyos-UUr8o1SWqdGHXQVCTO8uLQ==/lib/arm64, /data/app/com.tizilogic.pyos-UUr8o1SWqdGHXQVCTO8uLQ==/base.apk!/lib/arm64-v8a, /system/lib64, /system/vendor/lib64]]] couldn't find "libpython3.6m.so"
03-24 20:29:10.718  2554  2554 V pythonutil: Loading library: python3.7m
03-24 20:29:10.721  2554  2554 V pythonutil: Loading library: main
03-24 20:29:10.722  2554  2554 V pythonutil: Failed to load _io.so or unicodedata.so...but that's okay.
03-24 20:29:10.723  2554  2554 V pythonutil: Unsatisfied linker when loading ctypes
03-24 20:29:10.723  2554  2554 V pythonutil: Loaded everything!
03-24 20:29:10.810  2554  2588 I python  : Initializing Python for Android
03-24 20:29:10.810  2554  2588 I python  : Setting additional env vars from p4a_env_vars.txt
03-24 20:29:10.810  2554  2588 I python  : Changing directory to the one provided by ANDROID_ARGUMENT
03-24 20:29:10.810  2554  2588 I python  : /data/user/0/com.tizilogic.pyos/files/app
03-24 20:29:10.811  2554  2588 I python  : Preparing to initialize python
03-24 20:29:10.811  2554  2588 I python  : _python_bundle dir exists
03-24 20:29:10.811  2554  2588 I python  : calculated paths to be...
03-24 20:29:10.811  2554  2588 I python  : /data/user/0/com.tizilogic.pyos/files/app/_python_bundle/stdlib.zip:/data/user/0/com.tizilogic.pyos/files/app/_python_bundle/modules
03-24 20:29:10.811  2554  2588 I python  : set wchar paths...
03-24 20:29:10.849  2554  2588 I python  : Initialized python
03-24 20:29:10.849  2554  2588 I python  : AND: Init threads
03-24 20:29:10.849  2554  2588 I python  : testing python print redirection
03-24 20:29:10.850  2554  2588 I python  : Android path ['.', '/data/user/0/com.tizilogic.pyos/files/app/_python_bundle/stdlib.zip', '/data/user/0/com.tizilogic.pyos/files/app/_python_bundle/modules', '/data/user/0/com.tizilogic.pyos/files/app/_python_bundle/site-packages']
03-24 20:29:10.851  2554  2588 I python  : os.environ is environ({'PATH': '/sbin:/system/sbin:/system/bin:/system/xbin:/vendor/bin:/vendor/xbin', 'DOWNLOAD_CACHE': '/data/cache', 'ANDROID_BOOTLOGO': '1', 'ANDROID_ROOT': '/system', 'ANDROID_ASSETS': '/system/app', 'ANDROID_DATA': '/data', 'ANDROID_STORAGE': '/storage', 'EXTERNAL_STORAGE': '/sdcard', 'ASEC_MOUNTPOINT': '/mnt/asec', 'BOOTCLASSPATH': '/system/framework/core-oj.jar:/system/framework/core-libart.jar:/system/framework/conscrypt.jar:/system/framework/okhttp.jar:/system/framework/legacy-test.jar:/system/framework/bouncycastle.jar:/system/framework/ext.jar:/system/framework/framework.jar:/system/framework/telephony-common.jar:/system/framework/voip-common.jar:/system/framework/ims-common.jar:/system/framework/apache-xml.jar:/system/framework/org.apache.http.legacy.boot.jar:/system/framework/smartbondingservice.jar:/system/framework/sprengine.jar:/system/framework/android.hidl.base-V1.0-java.jar:/system/framework/android.hidl.manager-V1.0-java.jar:/system/framework/timakeystore.jar:/system/framework/fipstimakeystore.jar:/system/framework/ucmopensslenginehelper.jar:/system/framework/esecomm.jar:/system/framework/SemAudioThumbnail.jar:/system/framework/knoxsdk.jar:/system/framework/sec_edm.jar:/system/framework/sagearpolicymanager.jar:/system/framework/sec_sdp_sdk.jar:/system/framework/sec_sdp_hidden_sdk.jar:/system/framework/knoxvpnuidtag.jar', 'SYSTEMSERVERCLASSPATH': '/system/framework/services.jar:/system/framework/ethernet-service.jar:/system/framework/wifi-service.jar:/system/framework/samsung-services.jar:/system/framework/hqm.jar:/system/framework/dmf.jar:/system/framework/ssrm.jar', 'MC_AUTH_TOKEN_PATH': '/efs', 'KNOX_STORAGE': '/data/knox/ext_sdcard', 'ENC_EMULATED_STORAGE_TARGET': '/storage/enc_emulated', 'ANDROID_SOCKET_zygote': '8', 'ANDROID_ENTRYPOINT': 'main.pyc', 'ANDROID_ARGUMENT': '/data/user/0/com.tizilogic.pyos/files/app', 'ANDROID_APP_PATH': '/data/user/0/com.tizilogic.pyos/files/app', 'ANDROID_PRIVATE': '/data/user/0/com.tizilogic.pyos/files', 'ANDROID_UNPACK': '/data/user/0/com.tizilogic.pyos/files/app', 'PYTHONHOME': '/data/user/0/com.tizilogic.pyos/files/app', 'PYTHONPATH': '/data/user/0/com.tizilogic.pyos/files/app:/data/user/0/com.tizilogic.pyos/files/app/lib', 'PYTHONOPTIMIZE': '2', 'P4A_BOOTSTRAP': 'SDL2', 'PYTHON_NAME': 'python', 'P4A_IS_WINDOWED': 'False', 'P4A_ORIENTATION': 'portrait', 'P4A_NUMERIC_VERSION': 'None', 'P4A_MINSDK': '21', 'LC_CTYPE': 'C.UTF-8'})
03-24 20:29:10.851  2554  2588 I python  : Android kivy bootstrap done. __name__ is __main__
03-24 20:29:10.851  2554  2588 I python  : AND: Ran string
03-24 20:29:10.851  2554  2588 I python  : Run user program, change dir and execute entrypoint
03-24 20:29:10.985  2554  2588 I python  : Traceback (most recent call last):
03-24 20:29:10.985  2554  2588 I python  :   File "/home/userhome/workspace/build/bdist.android-arm64-v8a/pyos/main.py", line 6, in <module>
03-24 20:29:10.985  2554  2588 I python  :   File "/home/userhome/.local/share/python-for-android/build/python-installs/bdisttest/sdl2/__init__.py", line 2, in <module>
03-24 20:29:10.986  2554  2588 I python  :   File "/home/userhome/.local/share/python-for-android/build/python-installs/bdisttest/sdl2/dll.py", line 111, in <module>
03-24 20:29:10.986  2554  2588 I python  :   File "/home/userhome/.local/share/python-for-android/build/python-installs/bdisttest/sdl2/dll.py", line 49, in __init__
03-24 20:29:10.986  2554  2588 I python  :   File "/home/userhome/.local/share/python-for-android/build/python-installs/bdisttest/sdl2/dll.py", line 37, in _findlib
03-24 20:29:10.986  2554  2588 I python  :   File "/home/userhome/.local/share/python-for-android/build/other_builds/python3-libffi-openssl-sqlite3/arm64-v8a__ndk_target_21/python3/Lib/ctypes/util.py", line 78, in find_library
03-24 20:29:10.986  2554  2588 I python  : FileNotFoundError: [Errno 2] No such file or directory: '/data/data/com.tizilogic.pyos/lib'
03-24 20:29:10.986  2554  2588 I python  : Python for android ended.
```

JonasT:
This will change `ctypes.util.find_library()` to properly query & search native lib dir, which should fix #1770

Ran fine for me but I tested with an ARMv7 app. I asked the original reporter for an ARM64 test

### DIFF
--- a/pythonforandroid/recipes/python3/patches/fix-ctypes-util-find-library.patch
+++ b/pythonforandroid/recipes/python3/patches/fix-ctypes-util-find-library.patch
@@ -1,23 +1,15 @@
 diff --git a/Lib/ctypes/util.py b/Lib/ctypes/util.py
 --- a/Lib/ctypes/util.py
 +++ b/Lib/ctypes/util.py
-@@ -67,4 +67,19 @@
+@@ -67,4 +67,11 @@
                  return fname
          return None
  
 +# This patch overrides the find_library to look in the right places on
 +# Android
 +if True:
++    from android._ctypes_library_finder import find_library as _find_lib
 +    def find_library(name):
-+        # Check the user app libs and system libraries directory:
-+        app_root = os.path.normpath(os.path.abspath('../../'))
-+        lib_search_dirs = [os.path.join(app_root, 'lib'), "/system/lib"]
-+        for lib_dir in lib_search_dirs:
-+            for filename in os.listdir(lib_dir):
-+                if filename.endswith('.so') and (
-+                        filename.startswith("lib" + name + ".") or
-+                        filename.startswith(name + ".")):
-+                    return os.path.join(lib_dir, filename)
-+        return None
++        return _find_lib(name)
 +
  elif os.name == "posix" and sys.platform == "darwin":
--- a/tests/test_androidmodule_ctypes_finder.py
+++ b/tests/test_androidmodule_ctypes_finder.py
@@ -1,0 +1,124 @@
+
+import mock
+from mock import MagicMock
+import os
+import shutil
+import sys
+import tempfile
+
+
+# Import the tested android._ctypes_library_finder module,
+# making sure android._android won't crash us!
+# (since android._android is android-only / not compilable on desktop)
+android_module_folder = os.path.abspath(os.path.join(
+    os.path.dirname(__file__),
+    "..", "pythonforandroid", "recipes", "android", "src"
+))
+sys.path.insert(0, android_module_folder)
+sys.modules['android._android'] = MagicMock()
+import android._ctypes_library_finder
+sys.path.remove(android_module_folder)
+
+
+@mock.patch.dict('sys.modules', jnius=MagicMock())
+def test_get_activity_lib_dir():
+    import jnius  # should get us our fake module
+
+    # Short test that it works when activity doesn't exist:
+    jnius.autoclass = MagicMock()
+    jnius.autoclass.return_value = None
+    assert android._ctypes_library_finder.get_activity_lib_dir(
+        "JavaClass"
+    ) is None
+    assert mock.call("JavaClass") in jnius.autoclass.call_args_list
+
+    # Comprehensive test that verifies getApplicationInfo() call:
+    activity = MagicMock()
+    app_context = activity.getApplicationContext()
+    app_context.getPackageName.return_value = "test.package"
+    app_info = app_context.getPackageManager().getApplicationInfo()
+    app_info.nativeLibraryDir = '/testpath'
+
+    def pick_class(name):
+        cls = MagicMock()
+        if name == "JavaClass":
+            cls.mActivity = activity
+        elif name == "android.content.pm.PackageManager":
+            # Manager class:
+            cls.GET_SHARED_LIBRARY_FILES = 1024
+        return cls
+
+    jnius.autoclass = MagicMock(side_effect=pick_class)
+    assert android._ctypes_library_finder.get_activity_lib_dir(
+        "JavaClass"
+    ) == "/testpath"
+    assert mock.call("JavaClass") in jnius.autoclass.call_args_list
+    assert mock.call("test.package", 1024) in (
+        app_context.getPackageManager().getApplicationInfo.call_args_list
+    )
+
+
+@mock.patch.dict('sys.modules', jnius=MagicMock())
+def test_find_library():
+    test_d = tempfile.mkdtemp(prefix="p4a-android-ctypes-test-libdir-")
+    try:
+        with open(os.path.join(test_d, "mymadeuplib.so.5"), "w"):
+            pass
+        import jnius  # should get us our fake module
+
+        # Test with mActivity returned:
+        jnius.autoclass = MagicMock()
+        jnius.autoclass().mService = None
+        app_context = jnius.autoclass().mActivity.getApplicationContext()
+        app_info = app_context.getPackageManager().getApplicationInfo()
+        app_info.nativeLibraryDir = '/doesnt-exist-testpath'
+        assert android._ctypes_library_finder.find_library(
+            "mymadeuplib"
+        ) is None
+        assert mock.call("org.kivy.android.PythonActivity") in (
+            jnius.autoclass.call_args_list
+        )
+        app_info.nativeLibraryDir = test_d
+        assert os.path.normpath(android._ctypes_library_finder.find_library(
+            "mymadeuplib"
+        )) == os.path.normpath(os.path.join(test_d, "mymadeuplib.so.5"))
+
+        # Test with mService returned:
+        jnius.autoclass = MagicMock()
+        jnius.autoclass().mActivity = None
+        app_context = jnius.autoclass().mService.getApplicationContext()
+        app_info = app_context.getPackageManager().getApplicationInfo()
+        app_info.nativeLibraryDir = '/doesnt-exist-testpath'
+        assert android._ctypes_library_finder.find_library(
+            "mymadeuplib"
+        ) is None
+        app_info.nativeLibraryDir = test_d
+        assert os.path.normpath(android._ctypes_library_finder.find_library(
+            "mymadeuplib"
+        )) == os.path.normpath(os.path.join(test_d, "mymadeuplib.so.5"))
+    finally:
+        shutil.rmtree(test_d)
+
+
+def test_does_libname_match_filename():
+    assert android._ctypes_library_finder.does_libname_match_filename(
+        "mylib", "mylib.so"
+    )
+    assert not android._ctypes_library_finder.does_libname_match_filename(
+        "mylib", "amylib.so"
+    )
+    assert not android._ctypes_library_finder.does_libname_match_filename(
+        "mylib", "mylib.txt"
+    )
+    assert not android._ctypes_library_finder.does_libname_match_filename(
+        "mylib", "mylib"
+    )
+    assert android._ctypes_library_finder.does_libname_match_filename(
+        "mylib", "libmylib.test.so.1.2.3"
+    )
+    assert not android._ctypes_library_finder.does_libname_match_filename(
+        "mylib", "libtest.mylib.so"
+    )
+    assert android._ctypes_library_finder.does_libname_match_filename(
+        "mylib", "mylib.so.5"
+    )
--- a/pythonforandroid/recipes/android/src/android/_ctypes_library_finder.py
+++ b/pythonforandroid/recipes/android/src/android/_ctypes_library_finder.py
@@ -1,0 +1,63 @@
+
+import os
+
+
+def get_activity_lib_dir(activity_name):
+    from jnius import autoclass
+
+    # Get the actual activity instance:
+    activity_class = autoclass(activity_name)
+    if activity_class is None:
+        return None
+    activity = None
+    if hasattr(activity_class, "mActivity") and \
+            activity_class.mActivity is not None:
+        activity = activity_class.mActivity
+    elif hasattr(activity_class, "mService") and \
+            activity_class.mService is not None:
+        activity = activity_class.mService
+    if activity is None:
+        return None
+
+    # Extract the native lib dir from the activity instance:
+    package_name = activity.getApplicationContext().getPackageName()
+    manager = activity.getApplicationContext().getPackageManager()
+    manager_class = autoclass("android.content.pm.PackageManager")
+    native_lib_dir = manager.getApplicationInfo(
+        package_name, manager_class.GET_SHARED_LIBRARY_FILES
+    ).nativeLibraryDir
+    return native_lib_dir
+
+
+def does_libname_match_filename(search_name, file_path):
+    # Filter file names so given search_name="mymodule" we match one of:
+    #      mymodule.so         (direct name + .so)
+    #      libmymodule.so      (added lib prefix)
+    #      mymodule.arm64.so   (added dot-separated middle parts)
+    #      mymodule.so.1.3.4   (added dot-separated version tail)
+    #      and all above       (all possible combinations)
+    import re
+    file_name = os.path.basename(file_path)
+    return (re.match(r"^(lib)?" + re.escape(search_name) +
+                     r"\.(.*\.)?so(\.[0-9]+)*$", file_name) is not None)
+
+
+def find_library(name):
+    # Obtain all places for native libraries:
+    lib_search_dirs = ["/system/lib"]
+    lib_dir_1 = get_activity_lib_dir("org.kivy.android.PythonActivity")
+    if lib_dir_1 is not None:
+        lib_search_dirs.insert(0, lib_dir_1)
+    lib_dir_2 = get_activity_lib_dir("org.kivy.android.PythonService")
+    if lib_dir_2 is not None and lib_dir_2 not in lib_search_dirs:
+        lib_search_dirs.insert(0, lib_dir_2)
+
+    # Now scan the lib dirs:
+    for lib_dir in [l for l in lib_search_dirs if os.path.exists(l)]:
+        filelist = [
+            f for f in os.listdir(lib_dir)
+            if does_libname_match_filename(name, f)
+        ]
+        if len(filelist) > 0:
+            return os.path.join(lib_dir, filelist[0])
+    return None


JonasT:
This was now tested by me for ARMv7/python3 with a larger SDL2 app, and by the original bug reporter for ARM64 as explained here: https://github.com/kivy/python-for-android/issues/1770#issuecomment-477747743

Based on this I am confident to recommend an inclusion at the very least based on functionality, of course code quality remains to be judged :smile: happy for any sort of feedback!

pythonforandroid/recipes/python3/patches/fix-ctypes-util-find-library.patch
```diff
@@ -10,9 +10,15 @@ diff --git a/Lib/ctypes/util.py b/Lib/ctypes/util.py
...
-+        app_root = os.path.normpath(os.path.abspath('../../'))
-+        lib_search_dirs = [os.path.join(app_root, 'lib'), "/system/lib"]
-+        for lib_dir in lib_search_dirs:
++        from jnius import autoclass
++        activity = autoclass('org.kivy.android.PythonActivity').mActivity
++        package_name = activity.getApplicationContext().getPackageName()
++        manager = activity.getApplicationContext().getPackageManager()
++        nativeLibDir = manager.getApplicationInfo(
++            package_name, 1024  # PackageManager.GET_SHARED_LIBRARY_FILES
++        ).nativeLibraryDir
```
AndreMiras: Thanks for the contribution!
If I understand correctly you mainly changed the "hardcoded" `os.path.join(app_root, 'lib')` path for something more dynamic using pyjnius right?
What's the ouput of `nativeLibDir` then? I think we could still ditch the necessity if making it dynamic and just adding more "hardcoded" path the the `lib_search_dirs`. I think it's common even for things like gcc to have a list of hardcoded search path including some that don't exist. I think it would make the code easier to follow to just have these path hardcoded

JonasT: @AndreMiras on android these can all be in different paths like `/mnt/` so hardcoding an absolute path is impossible (as you probably already know) and ARM64 lacks the "lib" symlink inside the files dir where the current working directory is so you would need to descend out and into something else that depends on the app name for a relative path (which is impossible to hardcode without a dynamic component, and seems super risky given how much internal android dirs can move around between versions). In conclusion I agree it's a good point, but not practical anymore. Also for what it's worth, `/system/lib` is still hardcoded

Here is some related info: https://stackoverflow.com/questions/41709129/lib-directory-missing-after-installing-64-bit-android-application

AndreMiras: Useful info, thanks for sharing!

pythonforandroid/recipes/python3/patches/fix-ctypes-util-find-library.patch
```diff
@@ -10,9 +10,15 @@ diff --git a/Lib/ctypes/util.py b/Lib/ctypes/util.py
...
 +        # Check the user app libs and system libraries directory:
-+        app_root = os.path.normpath(os.path.abspath('../../'))
-+        lib_search_dirs = [os.path.join(app_root, 'lib'), "/system/lib"]
-+        for lib_dir in lib_search_dirs:
++        from jnius import autoclass
++        activity = autoclass('org.kivy.android.PythonActivity').mActivity
++        package_name = activity.getApplicationContext().getPackageName()
++        manager = activity.getApplicationContext().getPackageManager()
++        nativeLibDir = manager.getApplicationInfo(
++            package_name, 1024  # PackageManager.GET_SHARED_LIBRARY_FILES
```
AndreMiras: why don't you `manager.GET_SHARED_LIBRARY_FILES` rather than hardcoding that `1024` digit?
I guess you should be able to use that `manager` instance right? Otherwise maybe something like:
```python
PackageManager = autoclass('android.content.pm.PackageManager')
PackageManager.GET_SHARED_LIBRARY_FILES
```

pythonforandroid/recipes/python3/patches/fix-ctypes-util-find-library.patch
```diff
@@ -10,9 +10,15 @@ diff --git a/Lib/ctypes/util.py b/Lib/ctypes/util.py
...
 +    def find_library(name):
 +        # Check the user app libs and system libraries directory:
-+        app_root = os.path.normpath(os.path.abspath('../../'))
-+        lib_search_dirs = [os.path.join(app_root, 'lib'), "/system/lib"]
-+        for lib_dir in lib_search_dirs:
++        from jnius import autoclass
++        activity = autoclass('org.kivy.android.PythonActivity').mActivity
++        package_name = activity.getApplicationContext().getPackageName()
++        manager = activity.getApplicationContext().getPackageManager()
++        nativeLibDir = manager.getApplicationInfo(
```
AndreMiras: super minor: use python naming convention `s/nativeLibDir/native_lib_dir/` I understand it's on the edge because we use java code, but still :pray:

pythonforandroid/recipes/python3/patches/fix-ctypes-util-find-library.patch
JonasT: good point I'll test out a fix. Interestingly `manager.GET_SHARED_LIBRARY_FILES` is 0 which is probably a pyjnius bug, but using the class works

pythonforandroid/recipes/python3/patches/fix-ctypes-util-find-library.patch
```diff
@@ -10,9 +10,16 @@ diff --git a/Lib/ctypes/util.py b/Lib/ctypes/util.py
...
++        from jnius import autoclass
++        activity = autoclass('org.kivy.android.PythonActivity').mActivity
++        package_name = activity.getApplicationContext().getPackageName()
++        manager = activity.getApplicationContext().getPackageManager()
++        manager_class = autoclass("android.content.pm.PackageManager")
++        native_lib_dir = manager.getApplicationInfo(
++            manager_class.GET_SHARED_LIBRARY_FILES
++        ).nativeLibraryDir
++        lib_search_dirs = [native_lib_dir, "/system/lib"]
++        for lib_dir in [l for l in lib_search_dirs if os.path.exists(l)]:
```
AndreMiras: Out of context, but if you feel like it :grimacing: 
I feel like the code below could be more elegant using something like glob.
Using something like (untested):
```python
glob.glob("lib{}.so".format(name)) + glob.glob("{}.so".format(name)) 
```
Well since this is definitely out of context I'm happy that we don't address it in this PR.
I just wanted your advice

pythonforandroid/recipes/python3/patches/fix-ctypes-util-find-library.patch
```diff
@@ -10,9 +10,16 @@ diff --git a/Lib/ctypes/util.py b/Lib/ctypes/util.py
 +if True:
 +    def find_library(name):
 +        # Check the user app libs and system libraries directory:
-+        app_root = os.path.normpath(os.path.abspath('../../'))
-+        lib_search_dirs = [os.path.join(app_root, 'lib'), "/system/lib"]
-+        for lib_dir in lib_search_dirs:
++        from jnius import autoclass
```
AndreMiras: sorry to ask, but that should probably be an optional dep with a warning and fallback to previous behavior.
```python
try:
    from jnius import autoclass
    # same things as you did below
except ImportError:
    app_root = os.path.normpath(os.path.abspath('../../'))
    native_lib_dir = os.path.join(app_root, 'lib')
```
Well you get the idea. I don't like that my suggestion adds complexity, but I'm always worrying about adding new hard deps.

JonasT: @AndreMiras if ARM64 wasn't *the* main target for the future I would agree. but ARMv7 is going to grow less and less important anyway, right? (in which case making this dep optional won't help many people in the medium-term future) Do we want to carry around this additional complexity for the 2-3 years it still somewhat matters, hoping we don't forget to remove it later?

I personally would rather leave it out, but as always I leave the decision up to you guys

pythonforandroid/recipes/python3/patches/fix-ctypes-util-find-library.patch
JonasT: hmmm I think it would need to be like that:
```
files = [f for f in (glob.glob(os.path.join(lib_dir, "lib{}.*".format(name))) +
                     glob.glob(os.path.join(lib_dir, "{}.".format(name)))
         if f.endswith(".so")]
if len(files) > 0:
    return files[0]
```
I'm not sure I prefer that, it looks a bit unwieldy to me, and my first instinct reading `glob.glob` was to check the docs for how exactly it works, even though I was familiar with the purpose of the module. So if you just want my personal judgement on what I think of it, I would prefer the current way it is written

AndreMiras: yes I definitely don't like the way you wrote it above :joy: 
But I think it could be written in a much neat way. For instance using the `endswith()` doesn't make sense here. The point of using glob is to not do that.

Let's skip it for now since it's out of context anyway.

JonasT: Both search patterns allow for an overlap, you can't glob that. A regex would work though, but not sure how much more readable that would be?

AndreMiras: don't really why my initial suggestion would do that: `glob.glob("lib{}.so".format(name)) + glob.glob("{}.so".format(name)) ` it includes both prefix and suffix

JonasT: Does it? I don't think it does, there isn't even a `*` in that, how exactly would it glob anything?

JonasT: Just to clarify, the current code also matches e.g. `libsdl2.amd64.so` (when you try to load `sdl2.so`)

AndreMiras: That would return `libfoo.so` and `foo.so` for `name = "foo"` so yes it works. If you need some `*` of course glob can do that too.
No I'm suggesting to skip that ugly:
```
            for filename in os.listdir(lib_dir):
                if filename.endswith('.so') and (
                        filename.startswith("lib" + name + ".") or
                        filename.startswith(name + ".")):
```
I mean yes I would either go for a regex or for a glob. Basically the idea (I haven't tried) is to not only skip these `if and or` but eventually also the for look

JonasT: > yes it works

no it doesn't :joy: respective as for the same result

>  No I'm suggesting to skip that ugly:

I got that, but what you are suggesting simply doesn't do the same thing.

I just checked, there is no regex glob either so that would also get us back to list comprehension at the very least. Is that really better?

AndreMiras: yes that's what I mean, returning a least, just like your `listdir()` does actually. If the list is not empty you know you have it matching. That way you skip the for loop and the if. I think it's fairly readable comparing to that `if and or` thingy

JonasT: Well I realized it's not matching something like `libsdl2.so.3.2` anyway so a regex makes more sense. But I'm not sure I subscribe to the notion that this is any more readable :joy:

JonasT: @AndreMiras okay I tested it with regex now!

JonasT:
For what it's worth I tested the regex version and it appears to work fine! I also tested the regex itself on various test patterns on command line, I just wasn't sure how to ship a test for it in a non-awkward way but if you have an idea I'd be happy to do so. (Otherwise, at least take this as some sort of indication that the regex pattern isn't complete untested bollocks :joy: )

AndreMiras:
Yes I was wondering the same for the unit tests actually. It's pretty annoying because the function is probably simple enough to test, but it's in a patch file which makes it not easy testable.
So let's maybe just add some code comments explaining what we're expecting to match and trying to do in general in this function

JonasT:
Okay, added a comment & tested & deployed it again to check that the patch still works!

pythonforandroid/recipes/python3/patches/fix-ctypes-util-find-library.patch
```diff
@@ -10,14 +10,31 @@ diff --git a/Lib/ctypes/util.py b/Lib/ctypes/util.py
...
++        activity = autoclass("org.kivy.android.PythonActivity").mActivity
++        package_name = activity.getApplicationContext().getPackageName()
++        manager = activity.getApplicationContext().getPackageManager()
++        manager_class = autoclass("android.content.pm.PackageManager")
++        native_lib_dir = manager.getApplicationInfo(
++            package_name, manager_class.GET_SHARED_LIBRARY_FILES
++        ).nativeLibraryDir
++        lib_search_dirs = [native_lib_dir, "/system/lib"]
++        for lib_dir in [l for l in lib_search_dirs if os.path.exists(l)]:
++            # Filter file names so given name="mymodule.so" we match one of:
```
AndreMiras: Thanks for adding this comment.
I actually thought `find_library()` was more used with `name="mylib"` rather than `name="mylib.so"`

AndreMiras:
Thanks for you efforts! I'll try to test it on Sunday and merge if it doesn't introduce regression on armv7

pythonforandroid/recipes/python3/patches/fix-ctypes-util-find-library.patch
JonasT: oops yeah I guess that part of the comment is wrong :joy:

JonasT: fixed it now :joy:

AndreMiras: Thanks!

AndreMiras:
Ouch, I've just tested it and I found a regression with services.
```
03-31 19:00:34.400 29223 29239 I service :  Traceback (most recent call last):
03-31 19:00:34.400 29223 29239 I service :    File "/home/andre/workspace/EtherollApp/.buildozer/android/app/service/main.py", line 25, in <module>
03-31 19:00:34.401 29223 29239 I service :    File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/pyetheroll/etheroll.py", line 9, in <module>
03-31 19:00:34.402 29223 29239 I service :    File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/eth_account/__init__.py", line 1, in <module>
03-31 19:00:34.402 29223 29239 I service :    File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/eth_account/account.py", line 10, in <module>
03-31 19:00:34.403 29223 29239 I service :    File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/eth_keyfile/__init__.py", line 7, in <module>
03-31 19:00:34.403 29223 29239 I service :    File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/eth_keyfile/keyfile.py", line 6, in <module>
03-31 19:00:34.404 29223 29239 I service :    File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/Crypto/Cipher/__init__.py", line 78, in <module>
03-31 19:00:34.405 29223 29239 I service :    File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/Crypto/Cipher/_mode_ecb.py", line 46, in <module>
03-31 19:00:34.406 29223 29239 I service :    File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/Crypto/Util/_raw_api.py", line 168, in load_pycryptodome_raw_lib
03-31 19:00:34.406 29223 29239 I service :    File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/Crypto/Util/_raw_api.py", line 51, in load_lib
03-31 19:00:34.407 29223 29239 I service :    File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/cffi/api.py", line 141, in dlopen
03-31 19:00:34.408 29223 29239 I service :    File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/cffi/api.py", line 802, in _make_ffi_library
03-31 19:00:34.408 29223 29239 I service :    File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/python-installs/etheroll/cffi/api.py", line 788, in _load_backend_lib
03-31 19:00:34.409 29223 29239 I service :    File "/home/andre/workspace/EtherollApp/.buildozer/android/platform/build/build/other_builds/python3-libffi-openssl-sqlite3/armeabi-v7a__ndk_target_21/python3/Lib/ctypes/ut
il.py", line 78, in find_library
03-31 19:00:34.409 29223 29239 I service :  AttributeError: 'NoneType' object has no attribute 'getApplicationContext'
```

JonasT:
Oh yeah, that's bad @ services!

As I said in chat I'll follow your suggestion of moving it into the `android` module and add tests, and figure out how to make it work with the other bootstraps. (which should probably work fine with a simple check if the `autoclass` worked and trying all the posibilities or something) Stay tuned for an update :smile:

JonasT:
@AndreMiras done!! works for me so far, can you check out whether it works for you?

tests/test_androidmodule_ctypes_finder.py
```diff
@@ -0,0 +1,158 @@
...
+
+    return autoclass_returning_graphical_activity
+
+
+def test_get_activity_lib_dir():
+    import jnius  # should get us our fake module
+    jnius.autoclass = lambda name: None
+    assert(android._ctypes_library_finder.get_activity_lib_dir(
+        "org.kivy.android.PythonActivity"
+    ) is None)
```
AndreMiras: Please do not use parenthesis on your `assert` it's a keyword not a function

tests/test_androidmodule_ctypes_finder.py
```diff
@@ -0,0 +1,158 @@
...
+# Fake jnius module with autoclass function:
+import sys
+import types
+module_name = 'jnius'
+bogus_module = types.ModuleType(module_name)
+sys.modules[module_name] = bogus_module
+bogus_module.autoclass = lambda name: None
+
+
+def _get_autoclass_with_fake_activity(returning_test_path="/testpath",
```
AndreMiras: Thanks for adding a docstring to give some explanation here. The first feeling is that we're over complicating things. I didn't deep dive into it yet, but I wasn't expecting such complexity to test what we're doing

tests/test_androidmodule_ctypes_finder.py
```diff
@@ -0,0 +1,158 @@
...
+    with that returned object supporting .getApplicationInfo()
+    for PackageManager's GET_SHARED_LIBRARY_FILES info.
+
+    (That is the functionality the android._ctypes_library_finder module
+     uses to scan the application-specific library directories)
+    """
+    try:
+        from unittest.mock import MagicMock
+    except ImportError:  # Python 2.
+        from mock import MagicMock
```
AndreMiras: FYI `mock` is in are test deps, so even for python3 it gets installed. I know it feels weird since it's already in the core via unittest.mock. But I think it's a common practice still and then you can skip the try/except and go directly for:
```python
from mock import MagicMock
```
`tox` should pass because it `pip install mock`

tests/test_androidmodule_ctypes_finder.py
JonasT: I swear I don't do it on purpose, I'm just used to writing it like that so I keep forgetting :joy:

tests/test_androidmodule_ctypes_finder.py
JonasT: How else would you test `pyjnius` if not mock it to this extent? After all the android java api is otherwise not available

tests/test_androidmodule_ctypes_finder.py
JonasT: fixed it! thanks for the poke, it does look better that way :+1:

tests/test_androidmodule_ctypes_finder.py
JonasT: changed it now :+1: thanks for the poke, will resolve for now

pythonforandroid/recipes/python3/patches/fix-ctypes-util-find-library.patch
JonasT: resolving this since it uses regexes now with even separate tess :+1:

tests/test_androidmodule_ctypes_finder.py
```diff
@@ -0,0 +1,155 @@
+
+# Import android._ctypes_library_finder module which we need to do isolated.
+# (since android.__init__ will crash if imported not on android!)
```
AndreMiras: I think we usually patch top level module with `mock.patch.dict('sys.modules' ...`

JonasT: I actually asked in `#python` and nobody suggested that :joy: (since I never patched a module before), although is it really any different from what I'm doing? (I'm also just inserting an item into sys.modules)

tests/test_androidmodule_ctypes_finder.py
JonasT: Okay, so the `jnius` module itself is now placed a little nicer. But I'm not sure how to simplify specifically the `autoclass` mocking, I'm open for any suggestions!

tests/test_androidmodule_ctypes_finder.py
AndreMiras: OK so we have two things here. 
1) For the android module crashing, I think it's more conventional to just:
```python
sys.modules['android._android'] = mock.MagicMock()
```
Then the line above should work without crashing:
```python
from pythonforandroid.recipes.android.src.android import _ctypes_library_finder
```

2) For Python2 `src` not being a package, I would just:
```sh
touch pythonforandroid/recipes/android/src/__init__.py
```
I think it worth it.

These two together would reduce your 26 lines of code to just one `sys.modules['android._android'] = mock.MagicMock()` I think it's totally worth it

tests/test_androidmodule_ctypes_finder.py
AndreMiras: Yes I would mock it, but without implementing the methods. You just assert that the library was called with expected parameters because this is a unit test. So you assume the underlying library is already tested and behave consistently when called with some given params. So usually asserting on `call_args_list` is enough https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args_list

JonasT: Hmmmmm but I did find a bug implementing the methods, pretty sure I wouldn't have found it without that. But as usual your call, but it seems to me it was a useful thing to do & have around, because otherwise truly testing this means firing up an android device which is a huge resistance to anyone ever doing it

tests/test_androidmodule_ctypes_finder.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+# Import the tested android._ctypes_library_finder module,
+# making sure android._android won't crash us!
+# (since android._android is android-only / not compilable on desktop)
+android_module_folder = os.path.abspath(os.path.join(
+    os.path.dirname(__file__),
+    "..", "pythonforandroid", "recipes", "android", "src"
+))
+sys.path.insert(0, android_module_folder)
```
AndreMiras: would it really hurt to `pythonforandroid/recipes/android/src/__init__.py` so we can skip that thing above for Python2?

tests/test_androidmodule_ctypes_finder.py
JonasT: @AndreMiras right, great points! I find the direct import `pythonforandroid.recipes.android.src.android` with touch a bit ugly, so I did *almost* all of what you suggested but kept a bit of a `sys.path` hack but it's still way shorter than before. What do you think about how it looks now? Is that any good?

tests/test_androidmodule_ctypes_finder.py
JonasT: @AndreMiras dunno it just seems uglier to me, messing around in the filesystem etc. (and not in a temp folder) but that's just my personal impression of course

tests/test_androidmodule_ctypes_finder.py
AndreMiras: Yes it looks way better to me. Regarding the touch I realise I wasn't clear enough. I do mean to track it in git. Yes I agree touching dynamically would otherwise be too hacky

JonasT: as a permanent addition? not a fan, could confuse people wondering why that is a module when it doesn't look like one. I think containing this to the test itself is nicer. but as always, open for you/someone else making the final call on this

tests/test_androidmodule_ctypes_finder.py
AndreMiras: hm OK let's leave it like this for now then, will cry in silence over there :door:  :cry:

JonasT: @AndreMiras aww nooo don't be sad :cry: oh noo

AndreMiras: ahah I'm fine actually I agree the `__init__.py` is not super satisfying either so one hack or the other I'm fine

tests/test_androidmodule_ctypes_finder.py
AndreMiras: Let's leave it as you have for now then

tests/test_androidmodule_ctypes_finder.py
AndreMiras: Yes I would definitely not have too much logic in the tests and kill that `autoclass_returning_graphical_activity` helper. You don't need it to actually go through the code path.
The idea would be to instead magic mock `jnius.autoclass.__init__` and makes it returns the activity class that contains one time the mActivity and the other time the mService. Then you verify that the activity was called with the correct parameters. So it should pretty much test the same as you did, but without building an helper that big

JonasT: I don't think I follow. How would you verify the correct parameters were called without implementing something similar? Wouldn't you need to also have some function that tracks what it was called with? Or does mock have something here I missed?

Also, at least the final call that returns the path needs to return a meaningful value, because otherwise the entire test for `test_find_library`

You'll probably have to give me some pointers for what mock functionality to google or a small example to make me understand how you want me to implement this, I'm afraid, because I'm not fully following right now :grimacing:

AndreMiras: My idea is to play with Mock.call_args_list to verify your mock got called with the expected parameters https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.call_args_list
I agree it's tricky to grasp at first and specially this case in not simple for different reasons, e.g. the lazy import, the top level module import and the jnius module not even being installed. So yes it's three things that makes it tricky to test it a straight forward way. And actually pyjnius could easily be a test dep. But we can also deal without it I guess.

AndreMiras: Regardig the test path being meaningful, yes for an integration test, not for a unit test.
Do we like integration test? Yes sometimes :) Do we like missing up with the file system in tests, probably not

JonasT: @AndreMiras it's a *temp folder*, it's not messing up anything calm down :joy: it's a useful and very simple test

I just tested call_args_list/mock_calls/method_calls and they work well enough, **but** how do you suggest I return that path in that final relevant inner nested call? I still see no way to avoid this nesting, so I'm not sure that helps with anything

JonasT: I wasn't kidding by the way when I said this caught a bug I didn't see before. unless you really want me to, I really would prefer not to cut even one of the test cases

JonasT: ohhhhh neat it can just be set nested and it stays there, that is coooool ok wait a sec, I can simplify this

AndreMiras: It's still messing with the file system even if it's a temp folder. I trust you could find a bug with this test, that's also the purpose. I'm not surprise you didn't find how to make the mock useful here because as I said it's not straightforward in our case for the reason I gave above. That's why I'll give it a try when I get a chance to. I can take a look tonight, but it's tight

JonasT: Okay I managed to simplify it a bit realizing I can just assign to some nested things right away. It's still not super short but way shorter than before :smile:

tests/test_androidmodule_ctypes_finder.py
```diff
@@ -0,0 +1,127 @@
...
+    (That is the functionality the android._ctypes_library_finder module
+     uses to scan the application-specific library directories)
+    """
+
+    # This is a closure to bind the value of "returning_test_path":
+    def autoclass_returning_graphical_activity(name):
+        activity = MagicMock()
+        activity.getApplicationContext().getPackageName = (
+             lambda: "com.example.testpackage"
+        )
```
AndreMiras: it's fine to use lambda, but mock also has a built'in feature to do what you're doing here:
```python
activity.getApplicationContext().getPackageName.return_value = "com.example.testpackage"
```

JonasT: oh neat! don't think it'd be actually shorter here though, so should I change it? not sure if lambda is considered ugly here :grimacing: I really haven't seen much mock'ing code

AndreMiras: Lambda is completely fine :+1:

tests/test_androidmodule_ctypes_finder.py
AndreMiras: Nice we're getting there, thanks.
Here's an example of what I'm thinking of when unit testing `get_activity_lib_dir()` for instance. Here I test only the first `return None` I'll write some more later:
```
def test_get_activity_lib_dir_autoclass_init_none():
    """
    Makes sure get_activity_lib_dir() returns `None` when `autoclass()`
    returns `None`.
    """
    m_jnius = mock.MagicMock()
    m_autoclass = mock.MagicMock()
    m_autoclass.return_value = None
    m_jnius.attach_mock(m_autoclass, 'autoclass')
    with mock.patch.dict('sys.modules', jnius=m_jnius):
        assert android._ctypes_library_finder.get_activity_lib_dir(
            'JavaClass'
        ) is None
        assert m_autoclass.call_args_list == [mock.call('JavaClass')]
```

JonasT: Wouldn't that make everything much longer though? When switching out `jnius` in every test rather than once at the top

JonasT: I like that argument check though, I added that one in now!

AndreMiras: Yes it's surely longer if you write a unit test per code path of your tested function. For instance if I want to write two more code path I can do:
```python
def test_get_activity_lib_dir_mactivity():
    """
    By default the mActivity is being accessed to retrieve the native library dir
    through the application context.
    """
    m_jnius = mock.MagicMock()
    m_autoclass = mock.MagicMock()
    m_jnius.attach_mock(m_autoclass, 'autoclass')
    with mock.patch.dict('sys.modules', jnius=m_jnius):
        m_app_context = m_autoclass().mActivity.getApplicationContext()
        m_app_info = m_app_context.getPackageManager().getApplicationInfo()
        m_app_info.nativeLibraryDir = '/path/to/lib'
        assert android._ctypes_library_finder.get_activity_lib_dir(
            'JavaClass'
        ) == '/path/to/lib'

def test_get_activity_lib_dir_mservice():
    """
    If the mActivity is `None` we fallback to the mService.
    """
    m_jnius = mock.MagicMock()
    m_autoclass = mock.MagicMock()
    m_autoclass().mActivity = None
    m_jnius.attach_mock(m_autoclass, 'autoclass')
    with mock.patch.dict('sys.modules', jnius=m_jnius):
        m_app_context = m_autoclass().mService.getApplicationContext()
        m_app_info = m_app_context.getPackageManager().getApplicationInfo()
        m_app_info.nativeLibraryDir = '/path/to/lib'
        assert android._ctypes_library_finder.get_activity_lib_dir(
            'JavaClass'
        ) == '/path/to/lib'
```
But of course you could also group these. But anyway it's not about the test being long, it's more about it not adding too much side logic that could make it unclear the code path we're testing

JonasT: Well the jnius module itself never changes in any way beyond the obvious autoclass assignment that is local to every test, so personally I'm not too sure that looks any better. But sure if you prefer the local `mock.patch.dict` I can change it

AndreMiras: So this is a draft to show the idea, but it could be much more comprehensive while still not having any `if` branching. See how in the second test `test_get_activity_lib_dir_mservice` we just make the `mActivity = None` to test another branching of our `get_activity_lib_dir_mservice()` code. Plus see then later in this same test we retrieve the app context through `mService` via `m_app_context = m_autoclass().mService.getApplicationContext()` in order to patch the `nativeLibraryDir` return value. This prove we went through the correct code path for this case also. I should have added comments even if it's a draft to make this `mService` case more explicit.
Also note in the same fashion we could have covered `if activity is None:` case of `get_activity_lib_dir()` if we wanted, which is not currently covered by your tests.
Anyway I agree said mock/testing would hardly be straightforward, but I would definitely be more comfortable if skip utils functions with code branching in our tests. Tests should always be a dumb as possible

tests/test_androidmodule_ctypes_finder.py
AndreMiras: Anyway with the magic of `MagicMock` you don't have to define this. So you could skip these lines

tests/test_androidmodule_ctypes_finder.py
```diff
@@ -0,0 +1,131 @@
...
+            activity_class.mService = None
+            activity_class.mActivity = activity
+            return activity_class
+        elif name == "org.kivy.android.PythonService" and return_as_service:
+            activity_class.mActivity = None
+            activity_class.mService = activity
+            return activity_class
+        elif name == "android.content.pm.PackageManager":
+            manager_class = MagicMock()
+            manager_class.GET_SHARED_LIBRARY_FILES = 1024
```
AndreMiras: mocking explicitly is like documenting. Would the test fail if you remove that line? If not then it means it's not mandatory to prove the point of the interface you're unit testing

tests/test_androidmodule_ctypes_finder.py
```diff
@@ -0,0 +1,131 @@
...
+    ) is None
+    assert mock.call("JavaClass") in jnius.autoclass.call_args_list
+
+    jnius.autoclass = _get_autoclass_with_fake_activity("/testpath")
+    assert android._ctypes_library_finder.get_activity_lib_dir(
+        "org.kivy.android.PythonActivity"
+    ) == "/testpath"
+    assert android._ctypes_library_finder.get_activity_lib_dir(
+        "random.invalid.name"
+    ) is None
```
AndreMiras: This is a good example showing that this test is not testing your `get_activity_lib_dir()` code but it's testing the test util.
And this is exactly what I want to avoid with having code branching via utils functions in tests.
Do you see what I mean?

AndreMiras: Basically the fact that `autoclass` returns `None` on non-existing java class is not a feature of `get_activity_lib_dir` but it's a feature of `autoclass`, however our point is not to unit test `autoclass`, we assume `autoclass` is acting correctly. That's why a test like I wrote as a quick example `test_get_activity_lib_dir_autoclass_init_none()` would make more sense. Do you reckon that the way your write it makes it *feels* like it's a feature of `get_activity_lib_dir()` that it would behave differently depending on `activity_name`?

tests/test_androidmodule_ctypes_finder.py
AndreMiras: Well I usually prefer to patch as close as possible to where it's needed, just like I prefer to try/catch also as close as possible. But this wasn't the point of my test actually. My point is to show you can achieve what you were achieving without utils function containing if/ifelse/else which to me can make things look sketchy

JonasT: I revamped it a little according to your ideas, let me know what you think!

tests/test_androidmodule_ctypes_finder.py
JonasT: I added a check for this now in the updated test (the only part still involving a small `if` :grimacing: lol)

tests/test_androidmodule_ctypes_finder.py
JonasT: sure. it's revamped now to check the call args instead :+1:

tests/test_androidmodule_ctypes_finder.py
JonasT: it's skipped now in the revamped test except in the one where I actually check for it

tests/test_androidmodule_ctypes_finder.py
```diff
@@ -0,0 +1,124 @@
...
+        "JavaClass"
+    ) == "/testpath"
+    assert mock.call("JavaClass") in jnius.autoclass.call_args_list
+    assert mock.call("test.package", 1024) in (
+        app_context.getPackageManager().getApplicationInfo.call_args_list
+    )
+
+
+def test_find_library():
+    test_d = tempfile.mkdtemp(prefix="p4a-android-ctypes-test-libdir-")
```
AndreMiras: I was glad to see we actually have context manager for this. If you use it we can ditch the try/finally.
```
with tempfile.TemporaryDirectory(prefix='p4a-android-ctypes-test-libdir-') as test_d:
```

JonasT: @AndreMiras very nice find! sadly it appears to be python 3 only, I just tested. So maybe something to leave for our more later future when we go python 2 only?

AndreMiras: OK thanks I didn't realise, let's leave it as you have then :+1:

tests/test_androidmodule_ctypes_finder.py
```diff
@@ -0,0 +1,124 @@
+
+import mock
+from mock import MagicMock
+import os
+import shutil
+import sys
+import tempfile
+
+sys.modules["jnius"] = MagicMock()  # fake jnius used later
```
AndreMiras: I promise this is my last wish, but could you remove this line and patch it explicitly when we need it. At at method level with the decorator, e.g. `@mock.patch.dict('sys.modules', jnius=MagicMock())` :pray: 
I really prefer when things are as local as possible

tests/test_androidmodule_ctypes_finder.py
```diff
@@ -0,0 +1,124 @@
...
+    os.path.dirname(__file__),
+    "..", "pythonforandroid", "recipes", "android", "src"
+))
+sys.path.insert(0, android_module_folder)
+sys.modules['android._android'] = MagicMock()
+import android._ctypes_library_finder
+sys.path.remove(android_module_folder)
+
+
+def test_get_activity_lib_dir():
```
AndreMiras: `@mock.patch.dict('sys.modules', jnius=MagicMock())` please :pray:

tests/test_androidmodule_ctypes_finder.py
```diff
@@ -0,0 +1,124 @@
...
+    assert android._ctypes_library_finder.get_activity_lib_dir(
+        "JavaClass"
+    ) == "/testpath"
+    assert mock.call("JavaClass") in jnius.autoclass.call_args_list
+    assert mock.call("test.package", 1024) in (
+        app_context.getPackageManager().getApplicationInfo.call_args_list
+    )
+
+
+def test_find_library():
```
AndreMiras: `@mock.patch.dict('sys.modules', jnius=MagicMock())` please

tests/test_androidmodule_ctypes_finder.py
AndreMiras: Thanks!

PR #1777, opened 2019-03-31
[kivy] updated to 44a8a6f

goffi-contrib:


pythonforandroid/recipes/kivy/__init__.py
```diff
@@ -7,7 +7,7 @@
 
 class KivyRecipe(CythonRecipe):
     # post kivy==1.10.1, `fixes SDL2 image loading (jpg)`
```
AndreMiras: I guess this comment is not accurate anymore, would you mind updating?

AndreMiras:
Thanks for the pull request!
Could you make it clear in your commit title or body why is this update needed, what does it fix? I remember we talked about it on the chat, but having it in the commit history is even more important I think

goffi-contrib:
@AndreMiras sure, I've just updated it.

Issue #1783, opened 2019-04-05
Background Service is killed when app is swiped (close by user)


brvier:
### Versions

* Python: 2.7
* OS: Android
* Kivy: master
* Cython: recommended
* Pythonforandroid: master

### Description

When you close a application, the service is also closed. This is an unexpected behaviors, as service should run in background,  and not be killed then restarted when you kill the app.

This happen as when you remove task from task list, the method onTaskRemoved of the service is also called.

AndreMiras:
As far as I remember this is an Android feature. When the app is closed the service is also killed. Isn't it?
Quick googling lead me to:

- https://stackoverflow.com/q/42126979/185510
- https://stackoverflow.com/q/30525784/185510

brvier:
No it's not a feature, it's badly coded. 

Take a look at kde connect source code for example. And how the app so works. Or automate, ifttt

AndreMiras:
On KDE Connect seems to be using the `START_STICKY` bit. Unfortunately this is not yet working on the p4a implementation, see:

- https://github.com/kivy/python-for-android/pull/1374#issuecomment-421614772
- https://groups.google.com/forum/#!topic/kivy-users/1-zIEkI0snY

PR #1784, opened 2019-04-05
Do not stop service if task removed : #1783

brvier:
Comment the stopSelf(); in onTaskRemoved.

### DIFF
--- a/pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonService.java
+++ b/pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonService.java
@@ -137,7 +137,7 @@ public class PythonService extends Service implements Runnable {
     @Override
     public void onTaskRemoved(Intent rootIntent) {
         super.onTaskRemoved(rootIntent);
-        stopSelf();
+        //stopSelf();
     }
 
     @Override


AndreMiras:
Thanks @brvier for the pull request. Could you elaborate why this is needed?
This was introduced in https://github.com/kivy/python-for-android/pull/1374 for a specific reason described in the PR and in the method comment.
Why do you think it's not needed?
Also why just commenting it rather than deleting? Last if we end up adopting that change (still have to test it), we also want to update the method documentation

brvier:
Why commenting it ? Bad habits.

I don't understand why you want killing service and testing it when app is closed.

One example, if you have a Bluetooth connection in your service, you lost it, need to rescan, which could not be done at any moment in a background service, especially on some devices which prevent Bluetooth scan in background.

All other background service do the same. I'll found other example and try to understand why this was introduced in #1374

AndreMiras:
Stopping the service explicitly was a workaround to have it start again automatically when the app is killed actually. This is working in conjonction of `setAutoRestartService()` introduced in https://github.com/kivy/python-for-android/pull/643 and https://github.com/kivy/python-for-android/pull/1374

JonasT:
> I don't understand why you want killing service and testing it when app is closed.

From my own tests if the system kills the app at any point to preserve battery the service will actually continue to run. From this it appears that this service stop only happens if the user explicitly stops the app either via some integrated button calling `sys.exit()` (if you add any), or via recent apps wipe to get rid of it. The main purpose of a service to survive Android's auto-cleanup and keep vital functionality running appears to be preserved.

And isn't it kind of expected that everything stops when the user shuts down the app *explicitly*? Can you elaborate why you would want to not have that happen?

brvier:
> And isn't it kind of expected that everything stops when the user shuts down the app explicitly? Can you elaborate why you would want to not have that happen?

Not exactly, you should close user interface, but you should also keep background service communicating with external device working. There are many example of service running in background while user kill the app but it s mandatory to keep a permanent notification :
- Audio Apps : AntennaPods (A podcast player), Spotify
- Event based apps : KDE Connect, Syncthings, IFTTT
- App communicating with device in real time : Mi Fit (Xiaomi Mi Band), GadgetBridge (for pebble), 

There are many examples :)

TensorTom:
In our app, we are trying to have a service that runs in the background regardless of user interaction. Can we get this merged?

PR #1789, opened 2019-04-06
Ignores Python 2 import_recipe() warnings

AndreMiras:
Python 2 raises warnings on `imp.load_source()` when the module name contains dots. Ignoring this warning makes p4a output/logs more readable.
Note that for some reason the regression test will only catch the regression if ran isolated, e.g. via:
```sh
tox -e py27 -- tests/test_recipe.py::TestRecipe::test_import_recipe
```

### DIFF
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -1,5 +1,4 @@
 from os.path import basename, dirname, exists, isdir, isfile, join, realpath, split
-import importlib
 import glob
 from shutil import rmtree
 from six import PY2, with_metaclass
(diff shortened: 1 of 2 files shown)


pythonforandroid/recipe.py
```diff
@@ -24,8 +24,16 @@
...
     import imp
-    import_recipe = imp.load_source
+    import warnings
+
+    def import_recipe(module, filename):
+        with warnings.catch_warnings():
+            # ignores warnings raised by hierarchical module names
+            # (names containing dots) on Python 2
+            warnings.simplefilter("ignore", RuntimeWarning)
+            return imp.load_source(module, filename)
```
JonasT: why all the warning suppression when we could just transform to an absolute path?

**UNTESTED**, but dumping both the warnings and the odd branched definitions of `import_recipe` like this would IMHO be both shorter and more readable:

```
def import_recipe(module, filename):
    filename = os.path.join(os.path.abspath(os.path.dirname(__file__), "recipes", filename)
    if PY2:
        return imp.load_source(module, filename)
    else:
        import importlib.util
        if hasattr(importlib.util, 'module_from_spec'):
            # Python 3.5
            spec = importlib.util.spec_from_file_location(module, filename)
            mod = importlib.util.module_from_spec(spec)
            spec.loader.exec_module(mod)
            return mod
        else:
            # Python 3.3 and 3.4:
            from importlib.machinery import SourceFileLoader
            return SourceFileLoader(module, filename).load_module()
```
(unless I am missing something that wouldn't make this feasible)

JonasT: sorry, I guess the transformation of the path should be gated by a check of `os.path.isabs` - but I hope I illustrated the general idea well enough

AndreMiras: Thanks for the feedback! Yes I also think the way it's currently written is strange, but I didn't aim at refactoring it. And yes I would probably have written it the same way as you did.
Anyway I don't see how this would remove the warnings from the `imp.load_source(module, filename)` call :thinking: 
Could you elaborate?

JonasT: Oh it doesn't? Again I haven't tested it, but I thought the warnings were because it's a relative import. (rather than from an absolute path) or do they appear in any case when using `imp.load_source()`?

JonasT: Oops nevermind I don't know why I thought that :laughing: :laughing:

yeah, stick with the warning suppression then. but I would still suggest reformatting it into a single function, it's just a tad more readable like that IMHO

AndreMiras: I've refactored, I was really boiling to do it, but I didn't to mix concerns to ease reviewer task. But if you ask for it, there you have it :smile:

inclement:
Thanks!

JonasT:
@AndreMiras nice work :heart: for what it's worth, I get the mixed concerns but I'm also having the impression it's better to refactor it while you add something rather than not at all (which seems to be a bigger problem with a project of this size, just the nature of the economics) - so I think it's reasonable to put that in

PR #1791, opened 2019-04-08
Move ffmpeg download url to github repo

misl6:
Moved download url to github repo, in order to speed-up the download. (Current ffmpeg.org download link is freaking slow)

### DIFF
--- a/pythonforandroid/recipes/ffmpeg/__init__.py
+++ b/pythonforandroid/recipes/ffmpeg/__init__.py
@@ -5,6 +5,7 @@ import sh
 
 class FFMpegRecipe(Recipe):
     version = 'n3.4.5'
+    # Moved to github.com instead of ffmpeg.org to improve download speed
     url = 'https://github.com/FFmpeg/FFmpeg/archive/{version}.zip'
     depends = ['sdl2']  # Need this to build correct recipe order
     opts_depends = ['openssl', 'ffpyplayer_codecs']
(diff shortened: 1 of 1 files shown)


pythonforandroid/recipes/ffmpeg/__init__.py
```diff
@@ -4,8 +4,8 @@
 
 
 class FFMpegRecipe(Recipe):
-    version = '3.4.5'
-    url = 'http://ffmpeg.org/releases/ffmpeg-{version}.tar.bz2'
+    version = 'n3.4.5'
+    url = 'https://github.com/FFmpeg/FFmpeg/archive/{version}.zip'
```
AndreMiras: shouldn't we have the `n` here `https://github.com/FFmpeg/FFmpeg/archive/n{version}.zip` and then removed from the version?

misl6: Mmm, IMHO We should have the `n` in the version, so, if the user wants to use a specific version / hash from requirements he didn't have to edit also the ffmpeg recipe.

AndreMiras: OK :+1: 
Would you mind adding a comment above the `url` maybe, specifically stating we're using github over ffmpeg.org for performance reasons. That way we don't revert it later if we forget why it's like that

misl6: Sure! Added.

PR #1792, opened 2019-04-08
Update recipe.py

nikperotti:
Add ssl certificate

### DIFF
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -151,6 +151,8 @@ class Recipe(with_metaclass(RecipeMeta)):
             attempts = 0
             while True:
                 try:
+                    import ssl
+                    ssl._create_default_https_context = ssl._create_unverified_context
                     urlretrieve(url, target, report_hook)
                 except OSError as e:
                     attempts += 1


pythonforandroid/recipe.py
```diff
@@ -151,6 +151,8 @@ def report_hook(index, blksize, size):
             attempts = 0
             while True:
                 try:
+                    import ssl
+                    ssl._create_default_https_context = ssl._create_unverified_context
```
AndreMiras: I don't think we want to skip certificate verification by default. Also usually packages are served through valid cert. When it's not the case it's very likely that you're missing some default root cert in your setup

Thaodan: So he has to provide his instead of using the system cert?

AndreMiras: Well this is not providing any certificate to me, but says the client to not verify the server one which is dangerous

Thaodan: It is. But a way would be something to do like this:
https://github.com/sfackler/rust-native-tls/pull/96/commits/1bbea92db6fa081ab0734f15e43696edf45dace1

tshirtman: recipe.py doesn't run on android, it runs in the build system, when the distribution is being built, not checking the cert would open the door to build-time code injection, that would be hard to detect because it could happen in the code of any dependency of the app, imagine a wallet app being unable to trust any of its dependency has been downloaded without a MITM attach anymore? Publishing such a version of your app becomes way more stressful.

inclement:
Thanks for the PR, but it looks like we don't want to merge this for the reasons discussed in the review.

Issue #1793, opened 2019-04-11
error when compiling with flask_sqlalmechy and sqlamechy


spartero:
I'm doing an app on Flask. In this app I need to access the database, so I use flask_sqlalmechy. I understand that in the requirements section, I have to be able to both flask_sqlalmechy and sqlalmachy, correct me if I'm wrong. When packaging the app, I get an error.

How can i fix this?

- buildozer (0.40.dev0)
- ubuntu 18.04
- Cython (0.28.6)

```
File main.py

from flask import Flask
from flask import render_template
from flask_sqlalchemy import SQLAlchemy
import os

cabecera = ("descripcion","codio","pvp","T5","T6")
datos = (("automatico","27900","251,25€","25%","50%"),("automatico","27900","251,25€","25%","50%"),("automatico II/25A","a9k27900","251,25€","25%","50%"))


###############################
app = Flask(__name__)

############################
dbdir = "sqlite:///" + os.path.abspath(os.getcwd()) + "/static/db/database.db"

#dbdir ="sqlite:///./static/db/database.db"
app.config["SQLALCHEMY_DATABASE_URI"] = dbdir
app.config["SQLALCHEMY_TRACK_MODIFICATIONS"] = False
db = SQLAlchemy(app)

class tarifa(db.Model):
    id = db.Column(db.Integer, primary_key=True)
    descripcion = db.Column()
    codigo = db.Column()
    PVP = db.Column()
    T5 = db.Column()
    T6 = db.Column()

def consulta(tabla, tabla2):
    numero = 1
    resultado = []
    comando = 'SELECT count (id) FROM %s'%(tabla2)
    f = db.engine.execute(comando)
    for row in f:
        contador = row[0] + 1
    while numero < 3:
        cursor = tabla.query.filter_by(id=numero).first()
        resultado.append(cursor.descripcion)
        numero += 1
    return resultado



###########################
@app.route('/')
def index():
    return render_template('index.html')

@app.route('/automaticos')
def conductor():
    return render_template('automaticos.html', cabecera=cabecera, datos=datos)

@app.route('/hola')
def hola():
    return "hola"


@app.route('/buscar')
def buscar():
        buscar1 = consulta(tarifa,'tarifa')
        return render_template('test.html', buscar1=buscar1 )


if __name__ == '__main__':
    app.run(debug = False)

```

```
file buildozer

[app]

# (str) Title of your application
title = Sane

# (str) Package name
package.name = sane

[...]

PR #1794, opened 2019-04-16
Fix sqlalchemy recipe

sduenasg:
Fix sqlalchemy build errors and bump sqlalchemy from 1.0.9 to 1.3.3.


AndreMiras:
Thanks! Looks good, let's see if Travis is also happy about it.
This should hopefully fix #1793

PR #1807, opened 2019-04-30
Fix corner case of pip hack workaround we should get rid of anyway

JonasT:
This is just an adjustment for the workaround for
https://github.com/pypa/pip/issues/6097 to fix a corner case.
(The affected corner case was added to the unit test!)

So this code should really be thrown out once pip fixes that issue (so if we're wondering "should we even be dealing with this corner case?" - no, we shouldn't, pip should but sadly right now it doesn't)

---
Also I had this small fix lying around before #1625 was merged but forgot to include it :woman_facepalming:  :woman_facepalming:  :woman_facepalming: lol

### DIFF
--- a/pythonforandroid/pythonpackage.py
+++ b/pythonforandroid/pythonpackage.py
@@ -55,16 +55,21 @@ import zipfile
 
 
 def transform_dep_for_pip(dependency):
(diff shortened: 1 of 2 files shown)


inclement:
Thanks!

PR #1809, opened 2019-04-30
[WIP] Fix crashes when using other commands than 'apk'

JonasT:
I broke things with the `setup.py` pull request :joy: commands other than `p4a apk` appear to be broken, from my test this fix makes it work again, sorry :grimacing:

### DIFF
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -593,7 +593,8 @@ def project_has_setup_py(project_dir):
     return False
 
 
-def run_pymodules_install(ctx, modules, project_dir, ignore_setup_py=False):
+def run_pymodules_install(ctx, modules, project_dir=None,
+                          ignore_setup_py=False):
     """ This function will take care of all non-recipe things, by:
 
         1. Processing them from --requirements (the modules argument)
@@ -610,6 +611,7 @@ def run_pymodules_install(ctx, modules, project_dir, ignore_setup_py=False):
     # Bail out if no python deps and no setup.py to process:
     if not modules and (
             ignore_setup_py or
+            project_dir is None or
             not project_has_setup_py(project_dir)
             ):
         info('No Python modules and no setup.py to process, skipping')
@@ -621,7 +623,8 @@ def run_pymodules_install(ctx, modules, project_dir, ignore_setup_py=False):
              'install them with pip'.format(', '.join(modules)))
         info('If this fails, it may mean that the module has compiled '
              'components and needs a recipe.')
-    if project_has_setup_py(project_dir) and not ignore_setup_py:
+    if project_dir is not None and \
+            project_has_setup_py(project_dir) and not ignore_setup_py:
         info('Will process project install, if it fails then the '
              'project may not be compatible for Android install.')
 
@@ -694,7 +697,9 @@ def run_pymodules_install(ctx, modules, project_dir, ignore_setup_py=False):
                     _env=copy.copy(env))
 
         # Afterwards, run setup.py if present:
-        if project_has_setup_py(project_dir) and not ignore_setup_py:
+        if project_dir is not None and (
+                project_has_setup_py(project_dir) and not ignore_setup_py
+                ):
             with current_directory(project_dir):
                 info('got setup.py or similar, running project install. ' +
                      '(disable this behavior with --ignore-setup-py)')
--- a/pythonforandroid/toolchain.py
+++ b/pythonforandroid/toolchain.py
@@ -199,8 +199,11 @@ def build_dist_from_args(ctx, dist, args):
     if dist.needs_build:
         ctx.prepare_dist(ctx.dist_name)
 
-    build_recipes(build_order, python_modules, ctx, args.private,
-                  ignore_project_setup_py=args.ignore_setup_py,
+    build_recipes(build_order, python_modules, ctx,
+                  getattr(args, "private", None),
+                  ignore_project_setup_py=getattr(
+                      args, "ignore_setup_py", False
+                  ),
                  )
 
     ctx.bootstrap.run_distribute()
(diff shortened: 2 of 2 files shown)


nitanmarcel:
@JonasT This PR works but now there's a new error" Namespace object has no attribute private

JonasT:
@nitanmarcel can you give me either the backtrace or the source code line in which that error occurs? it's likely also something broken by the `setup.py` pull request

nitanmarcel:
@JonasT in the line 202 at build_recipes(...

JonasT:
did you use a non-sdl bootstrap by any chance? service or webview or something?

nitanmarcel:
> did you use a non-sdl bootstrap by any chance? service or webview or something?

Nope. Honestly I don't have any ideas what I'm doing :laughing: I'm using buildozer in a env with pyenv

nitanmarcel:
@JonasT The command buildoizer is trying to execute and fails:

```sehll
# Command failed: /usr/bin/python3.7 -m pythonforandroid.toolchain create --dist_name=rexteserdroid --bootstrap=sdl2 --requirements=python3,kivy,pillow,git+https://github.com/HeaTTheatR/KivyMD --arch armeabi-v7a --copy-libs --color=always --storage-dir="/home/alexandrumarcel/python_apps/rexdroid/.buildozer/android/platform/build" --ndk-api=21
```

nitanmarcel:
This little one liner fixed it for me :cake: 

```
    build_recipes(build_order, python_modules, ctx, args.private if hasattr(args, "private") else None,
                  ignore_project_setup_py=args.ignore_setup_py if hasattr(args, "ignore_setup_py") else False,
                 )
```

JonasT:
@nitanmarcel yeah I just tried something similar :+1: still testing whether it blows up the non-buildozer build, could you give the updated pull request a retry in buildozer?

nitanmarcel:
@JonasT yes, it didn't raised any exceptions so far so the latest fix worked

nitanmarcel:
Wasn't better to add the following arguments to the non APK ones? Or that would completely break p4a? @JonasT

pythonforandroid/build.py
```diff
@@ -694,7 +697,8 @@ def run_pymodules_install(ctx, modules, project_dir, ignore_setup_py=False):
                     _env=copy.copy(env))
 
         # Afterwards, run setup.py if present:
-        if project_has_setup_py(project_dir) and not ignore_setup_py:
+        if project_dir is not None and \
```
inclement: I would prefer using () instead of \ for line breaks.

pythonforandroid/toolchain.py
```diff
@@ -583,7 +586,7 @@ def add_parser(subparsers, *args, **kwargs):
             logger.setLevel(logging.DEBUG)
 
         self.ctx = Context()
-        self.ctx.use_setup_py = args.use_setup_py
+        self.ctx.use_setup_py = getattr(args, "use_setup_py", True)
```
inclement: I think it would be neater to move the --use-setup-py etc. arguments to generic_parser so that they're always in the args object, instead of using getattr in different places. It isn't a great api to have to stick non-generic stuff in generic_parser like this, but I think it's a failure of the way the abstraction is set up rather than something specific to this particular argument. There are plenty of other things in generic_parser for the same reason.

JonasT: I disagree, I think it shouldn't be in the generic_parser if it's not for the other commands. I'd rather have the special cases that are admittedly ugly but still be able to understand to which commands the arguments belong, BOTH as a coder and a user. Moving it into generic_parser seems like a cheap hack to me

However, as usual if you really think that's how it should be I will change it accordingly, so your call :+1:

pythonforandroid/build.py
JonasT: oops yea that's reasonable, changed it!

inclement:
Merging to get this fixed as it's a major issue, but I really dislike having hasattr/getattr spread around like this, I think it is a real code smell (arising from the argumentparser structure and how the arguments are used, not your mistake).

JonasT:
@inclement I agree some other solution might be nicer. but I think just putting many things into `generic_parser` is worse, and I'm not too sure what is better. maybe a wrapper object that will return None on non-existing attributes that don't start with `_` or something? I'm really not sure

JonasT:
Ok, another idea: what if all our parsers derive from a custom class instead of ArgumentParser which will somehow globally collect all the args, and overwrite `parse_args` to return some sort of wrapper objects that will return `None` for any attribute not present *only if provided by another known parser*?

inclement:
Well, that would be a neat but complex solution!

I was thinking that the simplest option, which I'd still consider an improvement, would be to have an explicit step in between `parser.parse_args` and everything else that sets default values for any attributes we expect to access but might not have been created.

I guess there are various other options using a more sophisticated ArgumentParser subclass, including what you've suggested.

JonasT:
@inclement I am willing to give that ArgumentParser subclass a shot if you think that might be worth the complexity. I would definitely personally prefer it over hardcoding args in some make-sure-they're-around function (that's essentially what you're proposing, right?)

inclement:
>  (that's essentially what you're proposing, right?)

Right, but it's not what I'm proposing, it's what we already have since I never fixed this since originally implementing it :p

> @inclement I am willing to give that ArgumentParser subclass a shot if you think that might be worth the complexity

I don't think it's worth actively focusing on unless you particularly want to, but I'd welcome a good rethink of p4a's argument parsing structure if it made the whole thing less smelly. Your proposal could be nice, but I don't have a strong vision of what it would end up looking like.

JonasT:
@inclement I was bored and gave it a try: https://github.com/kivy/python-for-android/pull/1812

PR #1814, opened 2019-05-05
Update pymunk to v5.5.0

viblo:


AndreMiras:
Thanks for the update.
I'm not sure why the build has been failing recently.
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```
I've tried restarting it 3-4 times and still the same story. I'm not sure what have changed, I'd like to investigate to keep green builds.

viblo:
Yes, I saw it as well. Strange, I can see that the master branch has the same problem, but there its the opposite builds that failed (and the latest commit there also shouldnt create any problems).. 

The build setup of pymunk have changed a little bit since the last version that was used here, so it would be good to have a green build to be sure it works.

viblo:
Just a note, I built a small test apk with the Dockerfile.py3 from this repo, and then tested it on a android phone and it seemed to work fine.

AndreMiras:
It eventually went green also on my other projects, will merge.
Thanks again!

viblo:
Great, thanks

PR #1825, opened 2019-05-26
Rework of freetype/harfbuzz recipes

opacam:
To fix cyclic dependency between `freetype` and `harfbuzz` and to produce shared libraries

And because we use those libraries in multiple recipes and building those libraries as shared libraries can reduce a little the apk size in some circumstances (when we build multiple recipes that depends on those libraries)

This changes are friendly with #1822 which needs shared libraries for `freetype` and `harfbuzz`

**Important note:** if `freetype` and `harfbuzz` are in `requirements`, then the cyclic dependency of those libraries will be solved otherwise the library will be build without linking with the other library (as expected)

**As a side note:** tested with a Pillow test app and arch armv7a (#1826)

### DIFF
--- a/pythonforandroid/recipes/freetype/__init__.py
+++ b/pythonforandroid/recipes/freetype/__init__.py
@@ -1,44 +1,120 @@
 from pythonforandroid.toolchain import Recipe
+from pythonforandroid.logger import shprint, info
 from pythonforandroid.util import current_directory
-from pythonforandroid.logger import shprint
-from os.path import exists, join, realpath
+from os.path import exists, join
+from multiprocessing import cpu_count
 import sh
 
 
 class FreetypeRecipe(Recipe):
+    """The freetype library it's special, because has cyclic dependencies with
+    harfbuzz library, so freetype can be build with harfbuzz support, and
+    harfbuzz can be build with freetype support. This complicates the build of
+    both recipes because in order to get the full set we need to compile those
+    recipes several times:
+        - build freetype without harfbuzz
+        - build harfbuzz with freetype
+        - build freetype with harfbuzz support
+
+    .. note::
+        To build freetype with harfbuzz support you must add `harfbuzz` to your
+        requirements, otherwise freetype will be build without harfbuzz
+
+    .. seealso::
+        https://sourceforge.net/projects/freetype/files/freetype2/2.5.3/
+    """
 
     version = '2.5.5'
     url = 'http://download.savannah.gnu.org/releases/freetype/freetype-{version}.tar.gz'  # noqa
 
-    depends = ['harfbuzz']
-
     def should_build(self, arch):
-        if exists(join(self.get_build_dir(arch.arch),
-                       'objs', '.libs', 'libfreetype.a')):
-            return False
-        return True
+        return not exists(
+            join(
+                self.get_build_dir(arch.arch),
+                'objs',
+                '.libs',
+                'libfreetype.so',
+            )
+        )
 
-    def build_arch(self, arch):
-        env = self.get_recipe_env(arch)
+    def get_recipe_env(self, arch=None, with_harfbuzz=False):
+        env = super(FreetypeRecipe, self).get_recipe_env(arch)
+        if with_harfbuzz:
+            harfbuzz_build = self.get_recipe(
+                'harfbuzz', self.ctx
+            ).get_build_dir(arch.arch)
+            freetype_install = join(self.get_build_dir(arch.arch), 'install')
+            env['CFLAGS'] = ' '.join(
+                [env['CFLAGS'], '-DFT_CONFIG_OPTION_USE_HARFBUZZ']
+            )
 
-        harfbuzz_recipe = Recipe.get_recipe('harfbuzz', self.ctx)
-        env['LDFLAGS'] = ' '.join(
-            [env['LDFLAGS'],
-             '-L{}'.format(join(harfbuzz_recipe.get_build_dir(arch.arch),
-                                'src', '.libs'))])
+            env['HARFBUZZ_CFLAGS'] = '-I{harfbuzz} -I{harfbuzz}/src'.format(
+                harfbuzz=harfbuzz_build
+            )
+            env['HARFBUZZ_LIBS'] = (
+                '-L{freetype}/lib -lfreetype '
+                '-L{harfbuzz}/src/.libs -lharfbuzz'.format(
+                    freetype=freetype_install, harfbuzz=harfbuzz_build
+                )
+            )
+        return env
 
+    def build_arch(self, arch, with_harfbuzz=False):
+        env = self.get_recipe_env(arch, with_harfbuzz=with_harfbuzz)
+
+        harfbuzz_in_recipes = 'harfbuzz' in self.ctx.recipe_build_order
+        prefix_path = self.get_build_dir(arch.arch)
+        if harfbuzz_in_recipes and not with_harfbuzz:
+            # This is the first time we build freetype and we modify `prefix`,
+            # because we will install the compiled library so later we can
+            # build harfbuzz (with freetype support) using this freetype
+            # installation
+            prefix_path = join(prefix_path, 'install')
+
+        # Configure freetype library
+        config_args = {
+            '--host={}'.format(arch.command_prefix),
+            '--prefix={}'.format(prefix_path),
+            '--without-zlib',
+            '--with-png=no',
+        }
+        if not harfbuzz_in_recipes:
+            info('Build freetype (without harfbuzz)')
+            config_args = config_args.union(
+                {'--disable-static', '--enable-shared', '--with-harfbuzz=no'}
+            )
+        elif not with_harfbuzz:
+            info('Build freetype for First time (without harfbuzz)')
+            # This time we will build our freetype library as static because we
+            # want that the harfbuzz library to have the necessary freetype
+            # symbols/functions, so we avoid to have two freetype shared
+            # libraries which will be confusing and harder to link with them
+            config_args = config_args.union(
+                {'--disable-shared', '--with-harfbuzz=no'}
+            )
+        else:
+            info('Build freetype for Second time (with harfbuzz)')
+            config_args = config_args.union(
+                {'--disable-static', '--enable-shared', '--with-harfbuzz=yes'}
+            )
+        info('Configure args are:\n\t-{}'.format('\n\t-'.join(config_args)))
+
+        # Build freetype library
         with current_directory(self.get_build_dir(arch.arch)):
             configure = sh.Command('./configure')
-            shprint(configure,
-                    '--host=arm-linux-androideabi',
-                    '--prefix={}'.format(realpath('.')),
-                    '--without-zlib',
-                    '--with-png=no',
-                    '--disable-shared',
-                    _env=env)
-            shprint(sh.make, '-j5', _env=env)
+            shprint(configure, *config_args, _env=env)
+            shprint(sh.make, '-j', str(cpu_count()), _env=env)
 
-            shprint(sh.cp, 'objs/.libs/libfreetype.a', self.ctx.libs_dir)
+            if not with_harfbuzz and harfbuzz_in_recipes:
+                info('Installing freetype (first time build without harfbuzz)')
+                # First build, install the compiled lib, and clean build env
+                shprint(sh.make, 'install', _env=env)
+                shprint(sh.make, 'distclean', _env=env)
+            else:
+                # Second build (or the first if harfbuzz not enabled), now we
+                # copy definitive libs to libs collection. Be sure to link your
+                # recipes to the definitive library, located at: objs/.libs
+                self.install_libs(arch, 'objs/.libs/libfreetype.so')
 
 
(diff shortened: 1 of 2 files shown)


pythonforandroid/recipes/harfbuzz/__init__.py
```diff
@@ -1,39 +1,89 @@
...
+    """The harfbuzz library it's special, because has cyclic dependencies with
+    freetype library, so freetype can be build with harfbuzz support, and
+    harfbuzz can be build with freetype support. This complicates the build of
+    both recipes because in order to get the full set we need to compile those
+    recipes several times:
+        - build freetype without harfbuzz
+        - build harfbuzz with freetype
+        - build freetype with harfbuzz support
+
+    .. seealso::
```
AndreMiras: Super minor: `see also`

pythonforandroid/recipes/harfbuzz/__init__.py
```diff
@@ -1,39 +1,89 @@
...
+        https://sourceforge.net/projects/freetype/files/freetype2/2.5.3/
+    """
     version = '0.9.40'
     url = 'http://www.freedesktop.org/software/harfbuzz/release/harfbuzz-{version}.tar.bz2'  # noqa
+    depends = ['freetype']
 
     def should_build(self, arch):
-        if exists(join(self.get_build_dir(arch.arch),
-                       'src', '.libs', 'libharfbuzz.a')):
+        if exists(
```
AndreMiras: Minor: return the bool expression directly:
```python
return not exists(...)
```

pythonforandroid/recipes/freetype/__init__.py
```diff
@@ -1,44 +1,118 @@
...
 
     version = '2.5.5'
     url = 'http://download.savannah.gnu.org/releases/freetype/freetype-{version}.tar.gz'  # noqa
 
-    depends = ['harfbuzz']
-
     def should_build(self, arch):
-        if exists(join(self.get_build_dir(arch.arch),
-                       'objs', '.libs', 'libfreetype.a')):
+        if exists(
```
AndreMiras: Minor: return the bool expression directly

pythonforandroid/recipes/freetype/__init__.py
```diff
@@ -1,44 +1,118 @@
...
+    recipes several times:
+        - build freetype without harfbuzz
+        - build harfbuzz with freetype
+        - build freetype with harfbuzz support
+
+    .. note::
+        To build freetype with harfbuzz support you must add `harfbuzz` to your
+        requirements, otherwise freetype will be build without harfbuzz
+
+    .. seealso::
```
AndreMiras: Super minor: `see also`

pythonforandroid/recipes/freetype/__init__.py
```diff
@@ -1,44 +1,118 @@
...
 import sh
 
 
 class FreetypeRecipe(Recipe):
+    """The freetype library it's special, because has cyclic dependencies with
+    harfbuzz library, so freetype can be build with harfbuzz support, and
+    harfbuzz can be build with freetype support. This complicates the build of
+    both recipes because in order to get the full set we need to compile those
+    recipes several times:
+        - build freetype without harfbuzz
```
AndreMiras: Do we do this part for the full set? It feels to me this one is only when we have `freetype` only in the requirements, right?
Thinking because of that line of code later `if not harfbuzz_in_recipes:`

pythonforandroid/recipes/freetype/__init__.py
inclement: I think seealso is a specific rst block specifier.

opacam:
Yesterday, I labeled this `WIP` because I wanted to take another look and also I wanted to o share a little app so this could be tested at runtime....unfortunately I could not finish the job...now I think that it's ready, but I don't have any problem in merging it after that @inclement test this with `matplotlib`...maybe even better.

The push I made today was made to apply the changes mentioned by @AndreMiras, thanks :smile:!!!

inclement:
I'm having a build issue on the matplotlib branch with harfbuzz not producing a .so output, probably not a major thing but just noting to not merge quite yet while I check it out.

inclement:
I haven't been able to resolve the issue I'm having. I'm testing by adding freetype and harfbuzz to the requirements of the basic python3 testapp, but it consistently fails at the following part of the build:
```

make[4]: warning: -jN forced in submake: disabling jobserver mode.
  CXX      libharfbuzz_la-hb-buffer-serialize.lo
  CXX      libharfbuzz_la-hb-blob.lo
  CXX      libharfbuzz_la-hb-buffer.lo
  CXX      libharfbuzz_la-hb-face.lo
  CXX      libharfbuzz_la-hb-common.lo
  CXX      libharfbuzz_la-hb-font.lo
  CXX      libharfbuzz_la-hb-ot-tag.lo
  CXX      libharfbuzz_la-hb-set.lo
  CXX      libharfbuzz_la-hb-shape.lo
  CXX      libharfbuzz_la-hb-shape-plan.lo
  CXX      libharfbuzz_la-hb-shaper.lo
  CXX      libharfbuzz_la-hb-unicode.lo
  CXX      libharfbuzz_la-hb-warning.lo
  CXX      libharfbuzz_la-hb-ot-font.lo
  CXX      libharfbuzz_la-hb-ot-layout.lo
  CXX      libharfbuzz_la-hb-ot-map.lo
  CXX      libharfbuzz_la-hb-ot-shape.lo
  CXX      libharfbuzz_la-hb-ot-shape-complex-arabic.lo
  CXX      libharfbuzz_la-hb-ot-shape-complex-default.lo
  CXX      libharfbuzz_la-hb-ot-shape-complex-hangul.lo
  CXX      libharfbuzz_la-hb-ot-shape-complex-hebrew.lo
  CXX      libharfbuzz_la-hb-ot-shape-complex-indic.lo
  CXX      libharfbuzz_la-hb-ot-shape-complex-indic-table.lo
  CXX      libharfbuzz_la-hb-ot-shape-complex-myanmar.lo
  CXX      libharfbuzz_la-hb-ot-shape-complex-sea.lo
  CXX      libharfbuzz_la-hb-ot-shape-complex-thai.lo
  CXX      libharfbuzz_la-hb-ot-shape-complex-tibetan.lo
  CXX      libharfbuzz_la-hb-ot-shape-normalize.lo
  CXX      libharfbuzz_la-hb-ot-shape-fallback.lo
  CXX      libharfbuzz_la-hb-fallback-shape.lo
  CXX      libharfbuzz_la-hb-ft.lo
  CXX      libharfbuzz_la-hb-ucdn.lo
  CXX      main-main.o
  CXX      test-test.o
  CXX      test_buffer_serialize-test-buffer-serialize.o
  CXX      test_size_params-test-size-params.o
  CXX      test_would_substitute-test-would-substitute.o
  GEN      harfbuzz.pc
  GEN      libharfbuzz.la
/home/sandy/android/android-ndk-r17c/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/../lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: cannot find -lharfbuzz
collect2: error: ld returned 1 exit status
make[4]: *** [Makefile:1094: libharfbuzz.la] Error 1
make[4]: Leaving directory '/home/sandy/.local/share/python-for-android/build/other_builds/harfbuzz/armeabi-v7a__ndk_target_21/harfbuzz/src'
make[3]: *** [Makefile:1723: all-recursive] Error 1
make[3]: Leaving directory '/home/sandy/.local/share/python-for-android/build/other_builds/harfbuzz/armeabi-v7a__ndk_target_21/harfbuzz/src'
make[2]: *** [Makefile:1015: all] Error 2
make[2]: Leaving directory '/home/sandy/.local/share/python-for-android/build/other_builds/harfbuzz/armeabi-v7a__ndk_target_21/harfbuzz/src'
make[1]: *** [Makefile:480: all-recursive] Error 1
make[1]: Leaving directory '/home/sandy/.local/share/python-for-android/build/other_builds/harfbuzz/armeabi-v7a__ndk_target_21/harfbuzz'
make: *** [Makefile:411: all] Error 2

```

Building with --disable-shared does work, but then of course the second freetype build fails.

@opacam Did you have any issue like this, any idea what's wrong?

opacam:
mmmm...this is strange...probably I faced this 3 years ago when I made the initial changes but not now...

I just did a build with the following requirements using `testapp_python3_sqlite_openssl`:

`harfbuzz,requests,peewee,sdl2,pyjnius,kivy,python3,freetype`

This builds fine on my system

Also tested with the basic `testapp_python3`:

`freetype,harfbuzz,sdl2,numpy,pyjnius,kivy,python3`

Also builds  fine for me...so...:thinking:

- did you remove all the files from the previous builds you made? (I always remove all the files from the `build` and `dist` folders when testing things...specially when playing with libraries)

AndreMiras:
> error: cannot find -lharfbuzz

@inclement is the harfbuzz shared lib generated somewhere at least? I mean could it be somewhere, but the linker doesn't find it because some `-L` flags are wrong? I wouldn't be surprised if for instance Arch and Ubuntu end up building the output lib to different directories.
What if you `find .buildozer/ -name "libharfbuzz*.so*" ? (I used buildozer, but adapt if you use p4a directly obviously)

inclement:
No, this happens even with a clean build, and harfbuzz does not generate any .so file. I've been trying to work out if it's supposed to do that before running this part of the build, or if this part of the build shouldn't be trying to link it in the first place. I've found it difficult to even work out where the -lharfbuzz is coming from.

inclement:
Would anyone be able to provide the full build_arch debug logs for freetype and harfbuzz in one of the testapps? I'd like to compare with the output I'm getting.

AndreMiras:
Oh right actually the recipe that doesn't build for you is `harfbuzz` itself right? So it doesn't really make sense that it's trying to put a `-lharfbuzz` when compiling itself :thinking: 
As for the debug log isn't the one from Travis enough? Otherwise I can try to build one tomorrow

inclement:
@AndreMiras Actually I forgot about travis, but it doesn't print the debug level output which is what I'm most interested in. I've been meaning to add an option to output that to a file (i.e. add a file log output to the default logger).

opacam:
@inclement, Let me do one build and I will send you the log

inclement:
The build log I'm getting is at https://gist.github.com/inclement/484db3a42ec4225b175a3d3e68212707

inclement:
Also posted my config.log at https://gist.github.com/inclement/7f87e55802e76f0c30d9e43d60b726df, could be interesting to compare.

opacam:
The harfbuzz build log: https://gist.github.com/opacam/4e252db3ba4f6d6124bf177833cecdab

inclement:
Thanks @opacam. I have to sleep now, but I'll do some detective work tomorrow. I'm sure it will be some small issue!

opacam:
The config.log at https://gist.github.com/opacam/abcda143a57d8f3c08bcac3816101583

opacam:
No problem @inclement, tomorrow I will be connected almost all day, so let me know if you need anything

opacam:
@inclement, Good morning,

I've find the problem...

I've been checking the logs we post yesterday, they are almost the same but I noticed that your build system has the development libraries for libharfbuzz, mine not, so I installed the `harfbuzz-dev ` package and  :boom:...I faced the same error that you got yesterday...

So **the system's harfbuzz development files are conflicting with our harfbuzz files**...and this shouldn't happen...I will try to find a solution for that...if I remember well we face something similar with libffi recently.

I found this comparing our config.log from harfbuzz:

Your freetype's flags/libs from your harfbuzz's configure.log (line 1122):
```
FREETYPE_CFLAGS='-I/usr/include/freetype2 -I/usr/include/libpng16 -I/usr/include/harfbuzz -I/usr/include/glib-2.0 -I/usr/lib/glib-2.0/include '
FREETYPE_LIBS='-lfreetype '
```

Here are mine before installing `harfbuzz-dev`:
```
FREETYPE_CFLAGS='-I/usr/include/freetype2 -I/usr/include/libpng16'
FREETYPE_LIBS='-lfreetype'
```

So, if you want to do some tests with matplotlib recipe, just removing the `harfbuzz-dev` package should allow you to successfully build harfbuzz's recipe (`harfbuzz-devel` in arch os ...I think)

opacam:
I modified a little both recipes to fix the issue and seems to build fine, still not tested at runtime,I will push the changes when tested... so...**DON'T MERGE THIS**

opacam:
Ok, modified the recipes to fix the conflict between system's harfbuzz development files and our harfbuzz recipe (I also simplified the logic to be more readable and maintainable compared to the first version I posted).

Now we should be able to build the harfbuzz recipe even if we have installed the `harfbuzz-dev` package (debian/ubuntu) or the equivalent in other os.

I tested it in both situations with the Pillow test app (#1826) and seems to work fine for me, travis give us a green tick mark...so...@inclement, when you have time, could you test it again please?

inclement:
Very nice work @opacam, I thought it might be something like this but I got lost inside libtool!

I've tested again, including both with/without harfbuzz in the matplotlib PR, without any issues.

PR #1826, opened 2019-05-27
Add a Pillow test app

opacam:
So we can test Pillow and some of the libraries that Pillow depends on (freetype, harfbuzz, jpeg and png)

**Some notes:**
  -  I used this `testapp` to test #1825
  - The `Pil` operations performed in here are done with *.png images...but it could be enhanced to include some other tests with *.jpeg
  - I included some other tests (`import ctypes` and `test vibration`) but this last one will only work on Android Oreo +. This also applies to all other tests that we trigger the android's vibration and this is happening because `vibrate` was deprecated in API level 26. Since I still haven't the solution for this I put a `todo` as a reminder and wrote in here (maybe we should create a new issue about this...)

Also see: https://developer.android.com/reference/android/os/Vibrator

**As a side note**: I don't know if we should include this testapp in the master branch...whatever the team decide it's fine to me...but I wanted to share it so we could test the changes applied in the mentioned pr...so if anyone feels that we shouldn't merge this let me know or directly close it, no problem :wink:


opacam:
Oh yes, totally agree...but where to start? Implementing all the `testapp`'s tests into `on_device_unit_tests` (#1636)? is this what are you thinking?

AndreMiras:
Yes kinda the idea indeed. There's no urge, but we should think about it

PR #1828, opened 2019-05-28
Update permissions.py

tamano123:
Adding the "ACCESS_FINE_LOCATION" permission


inclement:
Seems good, thanks. Have you any idea why this is excluded in the first place, though?

AndreMiras:
> Have you any idea why this is excluded in the first place, though?

I though it was more like something that got added later on the SDK and we didn't align.

inclement:
@AndreMiras That was my first thought, but I checked and it looks like this has been available since early Android. I feel like there must be a mystery reason :p

PR #1829, opened 2019-05-30
[WIP] Move png to mainline

opacam:
Also modify Pillow recipe because depends on png

This changes are friendly with #1822 which needs shared libraries for png and an specific name for the library `libpng16.so`

**Note:** tested with a Pillow test app and arch armv7a (#1826)

**As a side note:** untouched Pil recipe, because it's removed in #1822, so we avoid solve conflicts in there...but,  we will broke the Pil recipe...so...**be careful with merging this**. That it's why I leave it labeled `WIP` and because it would be great to test with matplotlib recipe before merging this, so we can avoid surprises...


inclement:
Thanks @opacam, I'll pull these commits into #1822 to resolve the pil issue.

opacam:
Oh yes, no problem, feel free to close this pr if you want to include this with matplotlib recipe :wink:

inclement:
Changes merged into #1822, thanks!

PR #1830, opened 2019-05-31
fix for the problem with decorator run_on_ui_thread and the local ref…

oukiar:
This is a workaround for solve the error of "local reference table overflow" that appears when you run multiple times functions with the decorator @run_in_ui_thread.

### DIFF
--- a/pythonforandroid/recipes/android/src/android/runnable.py
+++ b/pythonforandroid/recipes/android/src/android/runnable.py
@@ -10,6 +10,8 @@ from android.config import JAVA_NAMESPACE
 # reference to the activity
 _PythonActivity = autoclass(JAVA_NAMESPACE + '.PythonActivity')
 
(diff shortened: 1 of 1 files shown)


JonasT:
Nice, thanks for helping out with this fix!

I have one suggestion: would you mind adding a comment at some place, e.g. line 53 before `     if f not in __functionstable__: `, that gives a very brief reasoning why this helps? Unless this is obvious to everybody else, but to me it's not 100% clear right now why that code helps (and having a basic understanding of that might be useful next time someone needs to touch it)

inclement:
This looks good to me, thanks @oukiar , I'll add the comment @JonasT suggests and make the change.

inclement:
Hmm, sorry, didn't realise it adds that co-authored message in, my change was trivial. I'll remove that in any future generated messages.

Issue #1782, opened 2019-04-04
Suggestion: getting rid of some things


ghost:
We currently carry around a few things which I think would be worth removing soon:

- [ ] Python 3 CrystaX support: it causes lots of conditionals and special paths, and what does it offer at this point except a worse, outdated python version?
- [x] python2legacy recipe: I understand it allows lowering the API level a little, but only for Python 2 which we want to fade out anyway, so is it really worth keeping around?
- [x] pygame: it's in bad shape, and again adds so many special switches. and there's pygame-sdl2 now I believe, maybe that would even run easily in our SDL2 bootstrap?
- [ ] launcher/`--launcher`: shouldn't this rather be a separate kivy standalone app? this complicates the bootstrapping unnecessarily and there's weird whatever-licensed `renpy` code in the source tree for it. does it even run at this point?

I'm just leaving this list here for discussion

JonasT:
(I personally want to look into a launcher/`--launcher` removal pull request very soon. I might also look into some of the others if you want me to, unless someone else wants to do it of course. For pygame we have a pull request but it seems to have gotten a bit stale)

inclement:
My view is:
- python3crystax: Let's go ahead and remove it, but first at least announce it a bit in advance to try and clearly document the change. I don't want to be too hasty just because this was quite widely used for a couple of years.
- python2legacy: This can go as soon as you like, as far as I'm concerned.
- pygame: This can go too, I'll try to update my old PR or if someone else wants to make one then that's fine too.
- launcher: I'd rather not remove this right now unless it's actively getting in the way, unless the replacement @tito made is fully working. I'm pretty sure it should still run at the moment.

Julian-O:
Rumour is that both launchers are out of action and unlikely to be fixed any time soon, so launcher code could also be removed.

PR #1831, opened 2019-05-31
[R.I.P.] Remove `python2legacy` and `hospython2legacy`

opacam:
**Because:**
  - pygame has been removed (one of the purposes of this recipe was to make it work that bootstrap, alongside to be able to target api levels below 21)
  - the end of life of Python 2 it's near (2020)
  - we still have the python2 recipe (version 2.7.15)

**Caveats:**
  - We loose the ability to target api levels below 21  (Android 5.0 - Lollipop)

**See also:**
  - https://developer.android.com/about/dashboards
  - open discussion about removing things at #1782

**If the code you care about is still on Python 2...the clock is ticking...https://pythonclock.org/**

**As a side note:** I don't think that we can get a green tick mark with this pr in travis...this pr has all the numbers to fail with our `CI` test `rebuild_updated_recipes`...similar than the recently merged pr about the pygame removal


inclement:
I have no objections to this. I'll merge as soon as I have time to review it, or if others are looking at it then go ahead as far as I'm concerned :)

Thanks @opacam

AndreMiras:
Nice to see it goes :wave: 
Funnily my tox `tests/test_graph.py` were failing until I deleted the empty directories left over after rebasing to master.
The error was:
```
AttributeError: module 'pythonforandroid.bootstraps.pygame' has no attribute 'bootstrap'
```
I had to delete the following empty directories:

- pythonforandroid/recipes/pygame*
- pythonforandroid/recipes/*python2legacy

Issue #1833, opened 2019-06-02
Sorting out python-for-android releases


inclement:
For the last few months, we've been mostly encouraging people to use python-for-android master. Buildozer does this by default. This has worked okay, especially for the intended purpose of bringing people up to date due to all the rapid changes, but it really isn't a good long term plan.

Main thoughts:
- Releases should be frequent. p4a releases are actually quite straightforward (just push a new sdist to pypi and increment the version), but I've obviously not done a great job of doing this often.
- It's important for p4a to have its own releases on pypi, but also buildozer needs a good way to use the latest release (and for this to also be recent code, old releases are bad).
- Until recently we did have a `stable` branch that buildozer used, with the idea being that all commits could be merged there once they were tested a bit on master. I think this was not a bad idea, except that we never merged nearly often enough, and once PRs got out of order it was annoying to resolve conflicts (plus I'm not sure if buildozer even behaved right when it came to updating the repo).

I've been wondering more recently if we should try a calver approach, with a release (say) every week. The creation and even pushing of the release package could be automated. This would have the problem that issues merged just before the release time would not be tested for long, but perhaps this could be resolved by e.g. releasing the master branch from 1 week ago, so nothing gets released for at least a week.

This would have the nice feature of a clear release trail on pypi, but also a clear 'most recent' target for buildozer. It would also respect that p4a is a rapidly moving project where it's usually best to be near the top of trunk. I think it might also be an advantage that broken releases could be easily revoked if necessary (not that semantic versions couldn't be, but it somehow seems more natural with scheduled releases), but I'm not sure about it.

Alternatively, we could reinstate the stable branch and automatically schedule merges to it. I think that would be fine, except it doesn't give a clear set of releases for pypi. We could schedule releases on top of that, but then it's the same as the above proposal except with an additional stable branch that only buildozer will really use.

Does anyone have any thoughts about this? All ideas welcome, the above are not strong opinions and I'm sure there could be better ideas.

JonasT:
My first reaction is, a release schedule sounds like a good idea but going more rapid has both advantages and disadvantages. Given how fast we merge or don't merge things sometimes, I think weekly might be too ambitious, although I like the idea of aiming for something regular. But maybe something like every 3 or 4 weeks would make more sense?

[...]

Issue #1788, opened 2019-04-06
Keep track of coverage testing


AndreMiras:
Since we're trying to increase our testing coverage, we need to keep track of it.
This ticket is about investigating which solution to use and implement it.
In a first phase, we don't want to enforce a certain coverage, but at least have an overview of what we currently have and track how it evolves.

opacam:
@AndreMiras, thanks to remind us this `coverage` thing, now I think that we already have it covered thanks to the pull request mentioned above, so I'm closing this issue :wink:, feel free to open it again if you think that we should discuss something

PR #1835, opened 2019-06-03
Add coverage with coveralls and make use of travis stages

opacam:
To achieve the desired result we also modify the travis file, Implementing travis stages which are:
  - **Lint** (our tox tests runs in here and also we collect the coverage data)
  - **Test** (our build tests, which are exactly the same than the ones we have been using for a while, but now those are labelled)

Make use of the stages allow us to sequentially run some tests (without loosing the parallel builds, so the `Test` stage will be runned in parallel mode...the same behaviour than we have right now). Be aware that if the `Lint` stage fails, then the `Test` stage will never be run. This feature has been implemented in here in order to only send the coverage reports in case that we succeed with `Lint` stage.

This is friendly with the idea to implement some kind of automation to auto release/publish p4a versions (see issue/discussion #1833). We could implement it in a new stage (eg: deploy), to build docs, release a new version when master branch is tagged, upload it to pypi....

**As a side note:** notice that we currently perform our tox tests 4 times (one for each test we make)...with this pr we only make the tox tests one time (`Lint` stage), so this should improve the timings with our ci tests

**Notes:**
  - I took myself the liberty to enable `coveralls` for python-for-android (not only for my fork, also enabled for the main project `kivy/python-for-android`, if this pr is not merged, then we should disable it)
  - The added badge in README.md file should be automatically updated to the right value at the time of merging this in master branch
  - The coverage is configured to only track the `pythonforandroid` module 
  - The coverage report is not totally accurate because we skip some tests in travis

**Solves:**  #1788

**Coverage report at:** https://coveralls.io/jobs/49254277

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,11 @@ services:
 
 before_install:
   - travis_retry sudo apt update -qq
-  - travis_retry sudo apt install -qq --no-install-recommends python2.7 python3 python3-venv python3-virtualenv
+  # to successfully send the coveralls reports we need pyOpenSSL
+  - travis_retry sudo apt install -qq --no-install-recommends python2.7 python3 python3-venv python3-virtualenv python3-setuptools python3-pip python3-openssl
   # (venv/virtualenv are both used by tests/test_pythonpackage.py)
   - sudo pip install tox>=2.0
+  - sudo pip3 install coveralls
   # https://github.com/travis-ci/travis-ci/issues/6069#issuecomment-266546552
   - git remote set-branches --add origin master
   - git fetch
@@ -39,6 +41,8 @@ script:
   # case that the travis log doesn't produce any output for more than 10 minutes
   - while sleep 540; do echo "==== Still running (travis, don't kill me) ===="; done &
   - docker build --tag=p4a --file Dockerfile.py3 .
-  - docker run -e CI p4a /bin/sh -c "$COMMAND"
+  - docker run -e CI -e TRAVIS_JOB_ID="$TRAVIS_JOB_ID" -e TRAVIS_BRANCH="$TRAVIS_BRANCH" p4a /bin/sh -c "$COMMAND"
   # kill the background process started before run docker
   - kill %1
+after_success:
+  - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,19 @@ dist: xenial  # needed for more recent python 3 and python3-venv
 
 language: generic
 
+stages:
+  - lint
+  - test
+
 services:
   - docker
 
 before_install:
   - travis_retry sudo apt update -qq
   # to successfully send the coveralls reports we need pyOpenSSL
-  - travis_retry sudo apt install -qq --no-install-recommends python2.7 python3 python3-venv python3-virtualenv python3-setuptools python3-pip python3-openssl
+  - travis_retry sudo apt install -qq --no-install-recommends
+    python2.7 python3 python3-venv python3-virtualenv python3-pip
+    python3-setuptools python3-openssl
   # (venv/virtualenv are both used by tests/test_pythonpackage.py)
   - sudo pip install tox>=2.0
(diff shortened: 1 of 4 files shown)


inclement:
Sounds good to me, but I'll let @AndreMiras check the travis details :)

AndreMiras:
Sounds awesome! I'll try to review tonight, otherwise tomorrow.
Thanks!

.travis.yml
```diff
@@ -20,25 +28,42 @@ env:
...
+      name: "Tox tests and coverage"
+      script:
+        # we want to fail fast on tox errors without having to `docker build` first
+        - tox -- tests/ --ignore tests/test_pythonpackage.py
+        # (we ignore test_pythonpackage.py since these run way too long!!
+        #  test_pythonpackage_basic.py will still be run.)
+      after_success:
+        - coveralls
+
+    - &testing
```
AndreMiras: I'm not aware of this syntax and also the `<<: *testing` below. Could you share me a doc to read or a keyword to look for?

opacam: Oh yes, no problem, I learned this syntax for this pr, they are called `Yaml Anchors & Aliases` and are great to create templates, I discovered this in this article:
  - [Optimizing Mocha's Builds with Travis](https://boneskull.com/mocha-and-travis-ci-build-stages/#buildstage1smoke)

Also mentioned in [Travis docs examples](https://docs.travis-ci.com/user/build-stages/using-yaml-aliases/)
And here some more information: [atlassian - yaml anchors](https://confluence.atlassian.com/bitbucket/yaml-anchors-960154027.html)

opacam: The `yaml Anchors & Aliases` that I wrote works the following way:

First we define a template named testing using the keywords `&testing`
```
    - &testing
      stage: test
      before_script:
        # build docker image
        - docker build --tag=p4a --file Dockerfile.py3 .
        # Run a background process to make sure that travis will not kill our tests in
        # case that the travis log doesn't produce any output for more than 10 minutes
        - while sleep 540; do echo "==== Still running (travis, don't kill me) ===="; done &
      script:
        - docker run -e CI -e TRAVIS_JOB_ID="$TRAVIS_JOB_ID" -e TRAVIS_BRANCH="$TRAVIS_BRANCH" p4a /bin/sh -c "$COMMAND"
      after_script:
        # kill the background process started before run docker
        - kill %1
```
The following code will take the above defined template, actually will also be part of the template but we override later for the other jobs:

```
      name: Python 3 basic
      # overrides requirements to skip `peewee` pure python module, see:
      # https://github.com/kivy/python-for-android/issues/1263#issuecomment-390421054
      env: COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_python3_sqlite_openssl.py apk --sdk-dir $ANDROID_SDK_HOME --ndk-dir $ANDROID_NDK_HOME --requirements libffi,sdl2,pyjnius,kivy,python3,openssl,requests,sqlite3,setuptools'
```
For the following jobs we want to use the same scripts defined in  `&testing` so we use the keywords `<<:*testing` but we want to override the `name` and `env` defined in the template so we do the following:

```
    - <<: *testing
      name: Python 2 basic
      env: COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_python2_sqlite_openssl.py apk --sdk-dir $ANDROID_SDK_HOME --ndk-dir $ANDROID_NDK_HOME --requirements sdl2,pyjnius,kivy,python2,openssl,requests,sqlite3,setuptools'
```

The same for the other jobs...

AndreMiras: Thanks for the explanation makes sense :+1: 
Our Travis file start to get more complex, but that's the way it is I guess.
Dropping python2 at some point may help slightly

opacam:
Yeah, I'm working on that...I'm writing unittests for a couple of uncovered modules (`archs` and `bootstrap`) I almost have it...I will create proper pr when ready :wink:

Issue #1836, opened 2019-06-04
Create a release checklist


inclement:
I'm thinking to create a release checklist and instructions that we can reuse, in preparation for acting on #1833. Once this is complete, I'll move it to proper documentation somewhere.

The following is what I currently would do to make a release. With a git flow workflow we'd want to do the same thing, except with the nice feature that the release gets its own branch to run tests and checks on - I've never been very good about that.

- [ ] Check that the [build is passing](https://travis-ci.org/kivy/python-for-android)
- [ ] Run the tests locally via `tox`: this performs some long-running tests that are skipped on Travis.
- [ ] Build and run the [on_device_unit_tests](https://github.com/kivy/python-for-android/tree/master/testapps/on_device_unit_tests) app using buildozer. Check that they all pass.
- [ ] Build and run the following [testapps](https://github.com/kivy/python-for-android/tree/master/testapps) for arch `armeabi-v7a` and `arm64-v8a`:
  - [ ] `python3 setup_testapp_python3_sqlite_openssl.py apk`
  - [ ] `python3 setup_testapp_python2.py apk`
- [ ] Check that the version number is correct
- [ ] Tag the release
- [ ] Create the release distributions: `python3 setup.py sdist`
- [ ] Upload the release distribution to pypi: `python3 -m twine upload`
  - ^^ this one I'm not certain about, last time we released I found things had changed and twine was now compulsory, I'll check it and fix the instruction on the next release.

The release should also involve an announcement, which I intend to get on top of but don't consider part of this core process.

Questions:
- How does this change for git flow? The release branch procedure is then well defined, but this needs writing down.
- Are there any quick improvements to be made to the tests I've proposed? Obviously in the medium and long term there's lots we'd like to do, but this is the minimal set of things that roughly represents checks I've done in the past (in addition to general use of p4a).
- Is there anything missing from this list?
- How should we actually manage dev branch and release version numbers using calver?

I'm intending to test this procedure as currently written for making a release at the same time as setting up the develop branch, in order to check it works.

JonasT:
`.travis.yml` does currently exclude some `pythonforandroid/pythonpackage.py` tests (intentionally, due ot runtime). `tox` runs them.

Therefore I propose that running `tox` and checking that no tests fail should be added to the release checklist!

inclement:
Good point @JonasT, I've added it to the list.
[...]

PR #1838, opened 2019-06-05
Documented the development and release models

inclement:
Added some documentation about our new git flow structure, and documented the release checklist proposed at #1836.

I propose that once everyone is happy with this, I'll merge the PR, then immediately create the develop branch and the first release branch. We can tidy up any procedural issues in that first release, including making sure the release checklist works okay.

The only real unanswered question is what CalVer format to use. Is there any reason not to go with YYYY.MM.DD? Is some other option more normal (or more [pep-440 compatible](https://www.python.org/dev/peps/pep-0440/))?

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,8 +3,8 @@ python-for-android
 
 [![Build Status](https://travis-ci.org/kivy/python-for-android.svg?branch=master)](https://travis-ci.org/kivy/python-for-android)
 [![Coverage Status](https://coveralls.io/repos/github/kivy/python-for-android/badge.svg?branch=master&kill_cache=1)](https://coveralls.io/github/kivy/python-for-android?branch=master)
-[![Backers on Open Collective](https://opencollective.com/kivy/backers/badge.svg)](#backers) 
-[![Sponsors on Open Collective](https://opencollective.com/kivy/sponsors/badge.svg)](#sponsors) 
+[![Backers on Open Collective](https://opencollective.com/kivy/backers/badge.svg)](#backers)
+[![Sponsors on Open Collective](https://opencollective.com/kivy/sponsors/badge.svg)](#sponsors)
 
 python-for-android is a packager for Python apps on Android. You can
 create your own Python distribution including the modules and
@@ -69,9 +69,15 @@ We also have [#support Discord channel](https://chat.kivy.org/).
 
 ## Contributing
 
-We love pull requests and discussing novel ideas. Check out our
-[contribution guide](http://kivy.org/docs/contribute.html) and feel
-free to improve python-for-android.
+We love pull requests and discussing novel ideas. Check out the Kivy
+project [contribution guide](http://kivy.org/docs/contribute.html) and
+feel free to improve python-for-android.
+
+See [our
+documentation](https://python-for-android.readthedocs.io/en/latest/contribute/)
+for more information about the python-for-android development and
+release model, but don't worry about the details. You just need to
+make a pull request, we'll take care of the rest.
 
 The following mailing list and IRC channel are used exclusively for
 discussions about developing the Kivy framework and its sister projects:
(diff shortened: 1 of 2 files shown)


JonasT:
Actually, git flow doesn't necessarily work with pull request. It usually incorporates release branches and hotfix branches as far as I remember, as well as feature branches with a naming scheme - none of which we really do. 

Therefore I think mentioning git flow is more distracting and confusing than helpful. I would describe what python-for-android does from what I can see as "development on `develop` branch with a pull request flow and releases on the master branch", I would recommend just not mentioning git flow at all since to me it appears to be too different

inclement:
I put the instructions about the pull request because it seemed like a convenient way to centralise release discussion and see all the changes at once. I thought that the proposed branching structure would be exactly what git flow uses, but maybe I've misunderstood something?

For the feature branches, I didn't write about it because we do almost all development in personal branches in our own forks. The naming seemed like a minor point compared to the main development structure, but I'm happy to rethink it. I had imagined that we would be more formal about naming them `feature/whatever` style if pushed to this repo, although I didn't write that down.

JonasT:
@inclement companies who use git flow usually don't use personal repos. that's simply a slightly different approach, and gitflow also uses hotfix branches and a separate forked off branch *for each release* in preparation and other things - it's just way more than we do. so basically it has some similarities but it's just not an exact match, which is why I recommend not mentioning it

inclement:
> and gitflow also uses hotfix branches and a separate forked off branch for each release in preparation and other things - it's just way more than we do

Right, but I'm proposing we *do* start doing this. Do you think it's overkill?

JonasT:
@inclement hmm release branches might be, the reason they're usually done is so that pull request merging and such can continue while the release is prepared. but we don't have a huge pull request volume anyway, and it's normal that we delay them often (not necessarily a bad thing) so I don't see a problem just not merging things for 2 -3 days while we put together a release unless it fixes something in the release

if you want to do it anyway though I don't object, I just think at our current dev speed preparing the release on master could work well enough. of course that'd make the master slightly more unstable, but it would still be more stable than most master branches of most projects on github :smile:

inclement:
Thanks @JonasT. I've pushed a change to de-emphasise git flow except as a reference.

I'd like to leave the release branch stuff in there to see how it goes. We can always relax this if that turns out to be easier, I'm not one for unnecessary procedure. For now I like the idea because it seems like it should be straightforward but nice and clear, something that's been an issue in the past when release preparation has dragged on a bit. It's possible I might be overcompensating though, given that those things were my fault in the first place :p

opacam:
**About packaging and pypi upload**

We could generate the `wheel` alongside the `tar.gz` package like this:
  - Install/upgrade `setuptools` and `wheel`:
    * ```python3 -m pip install --user --upgrade setuptools wheel```
  - Generate distribution archives:
    * ```python3 setup.py sdist bdist_wheel```

And then the twine upload commands would be:
  - Install/upgrade twine:
    * ```python3 -m pip install --user --upgrade twine```
  - Upload the distribution archives with:
    * ```twine upload dist/*```

**Note:** notice that the twine command is triggered without `python`, the latest  `twine` releases comes with a binary

**See also:**
  - [Generating distribution archives](https://packaging.python.org/tutorials/packaging-projects/#generating-distribution-archives)
  - [Uploading distribution archives](https://packaging.python.org/tutorials/packaging-projects/#uploading-the-distribution-archives)

opacam:
About `calver` format, I usuallly prefer the short version of year `YY.MM.DD`, I don't have strong arguments for this, only that it shortens a little the version, that the version is more homogenous and maybe more readable...but after reading a little the link that @inclement posted about pep 440, I have my doubts about using the format I propose or the initial @inclement's proposal, because they mention the following about `date versions` in their examples section:

> Date based releases, using an incrementing serial within each year, skipping zero:
```
2012.1
2012.2
2012.3
...
2012.15
2013.1
2013.2
...
```

So maybe we should take the `pep 440` road...

**See also:** [version scheme examples pep 440 compliant](https://www.python.org/dev/peps/pep-0440/#examples-of-compliant-version-schemes)

JonasT:
I mean realistically python-for-android might not exist in 90 years, but you never know. I'd rather stick with 4 digit year just to be safe

AndreMiras:
Nice proposal overall and thanks for opening the discussion guys :clap: 
I personally like what we have in the proposal. I think sharing git flow for reference is great because it's an improved model and we don't have to maintain a too specific documentation. We also don't have to use the complete model with hotfixes... but I think master, develop, (feature) and release as described is completely fine.
As for the version I'm fine with `YYYY.MM.DD` :+1:

PR #1839, opened 2019-06-06
Release 2019.06.06

inclement:
I've set up the master/develop repository structure, now working to make a first release.

I'm including in this release branch whatever final changes I've found we need to make the develop branch the default. Per procedure, these will of course be merged back into develop as well as master. Let me know if you see anything I've missed.

This is probably last call on the release numbering system. Do we want any suffix, like 20190606.1?

# Release checklist

- [x] Check that the [build is passing](https://travis-ci.org/kivy/python-for-android)
- [x] Run the tests locally via `tox`: this performs some long-running tests that are skipped on Travis.
- [x] Build and run the [on_device_unit_tests](https://github.com/kivy/python-for-android/tree/master/testapps/on_device_unit_tests) app using buildozer. Check that they all pass.
- [x] Build and run the following [testapps](https://github.com/kivy/python-for-android/tree/master/testapps) for arch `armeabi-v7a` and `arm64-v8a`:
  - `python3 setup_testapp_python3_sqlite_openssl.py apk`
    - [x] `armeabi-v7a`
    - [x] `arm64-v8a`
  - `python3 setup_testapp_python2.py apk`
    - [x] `armeabi-v7a`
    - [x] `arm64-v8a`

- [x] Check that the version number is correct


inclement:
I've redirected all existing pull requests from master to the develop branch.

AndreMiras:
As for the suffix that's only needed if we do two releases in one day right? Well if we ever need it then yes we could YYYY.MM.DD.PATCH, but I usually simply YYYY.MM.DD+1 to keep the format consistent

inclement:
Just wondering, does travis provide a way to upload build artifacts somewhere? It would be nice if we could test the APKs built automatically, even if that testing is not itself automatic yet.

AndreMiras:
Yes it does but I've never played with it

AndreMiras:
By the way, now I'm thinking how do we want to handle the `.. versionchanged::` parts? It probably means on `develop` this string will be incomplete because we don't know the version upfront. Then we can review/fill these incomplete strings while preparing the release branch

opacam:
Maybe we should change from `master` to `develop` this line:

```def modified_recipes(branch='origin/master'):```

**At:** https://github.com/kivy/python-for-android/blob/develop/ci/rebuild_updated_recipes.py#L34

inclement:
I'm thinking that on the develop branch we should use version `YYYY.MM.DD.1.dev0`, where the `YYYY.MM.DD` comes from the latest release version. That should mean that pip orders the version correctly. We would expect that the next release would not actually be `YYYY.MM.DD.1`, it's just a convenient way to mark correctly that the develop branch is newer than the master branch. Any objections to taht?

AndreMiras:
That makes total sense :+1:

inclement:
Thanks @opacam, fixed that reference.
As far as I'm concerned this is good to go, I'll merge it as soon as someone approves it again.

PR #1842, opened 2019-06-07
Add unittest for module `pythonforandroid.archs`

opacam:
Because this module isn't covered by our tests and this allow us to improve a little our coverage

### DIFF
--- a/tests/test_archs.py
+++ b/tests/test_archs.py
@@ -1,0 +1,261 @@
+import os
+
+import unittest
+
+try:
+    from unittest import mock
+except ImportError:
+    # `Python 2` or lower than `Python 3.3` does not
+    # have the `unittest.mock` module built-in
+    import mock
+from pythonforandroid.bootstrap import Bootstrap
+from pythonforandroid.distribution import Distribution
+from pythonforandroid.recipe import Recipe
+from pythonforandroid.build import Context
+from pythonforandroid.util import BuildInterruptingException
+from pythonforandroid.archs import (
+    Arch,
+    ArchARM,
+    ArchARMv7_a,
+    ArchAarch_64,
+    Archx86,
+    Archx86_64,
+)
+
+expected_env_gcc_keys = {
+    "CFLAGS",
+    "LDFLAGS",
+    "CXXFLAGS",
+    "TOOLCHAIN_PREFIX",
+    "TOOLCHAIN_VERSION",
+    "CC",
+    "CXX",
+    "AR",
+    "RANLIB",
+    "LD",
+    "LDSHARED",
+    "STRIP",
+    "MAKE",
+    "READELF",
+    "NM",
+    "BUILDLIB_PATH",
+    "PATH",
+    "ARCH",
+    "NDK_API",
+}
+
+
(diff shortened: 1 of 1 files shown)


tests/test_archs.py
```diff
@@ -0,0 +1,247 @@
...
+        self.assertEqual(arch.target, "armv7a-none-linux-androideabi")
+        self.assertEqual(arch.platform_dir, "arch-arm")
+        arch = ArchARM(self.ctx)
+
+        # Check environment flags
+        env = arch.get_env()
+        self.assertIsInstance(env, dict)
+
+        for k in expected_env_gcc_keys:
+            self.assertIn(k, env)
```
AndreMiras: Minor suggestion, checking the intersection and equality should also work I think:
```python
assert expected_env_gcc_keys == env.keys() & expected_env_gcc_keys
```

tests/test_archs.py
```diff
@@ -0,0 +1,247 @@
...
+        self.assertIsInstance(arch.include_dirs, list)
+
+    # Here we mock two functions:
+    # - `ensure_dir` because we don't want to create any directory
+    # - `find_executable` because otherwise we will
+    #   get an error when trying to find the compiler (we are setting some fake
+    #   paths for our android sdk and ndk so probably will not exist)
+    @mock.patch("pythonforandroid.archs.find_executable")
+    @mock.patch("pythonforandroid.build.ensure_dir")
+    def test_arch_arm(self, mock_ensure_dir, mock_find_executable):
```
AndreMiras: Minor: why not making a dedicated `class TestArchARM`. So we create a dedicated test class per class we want to test. That way we don't end up with a too big generic test class. We can still inherit from `TestArch` so we share the `setup()` if needed

tests/test_archs.py
```diff
@@ -0,0 +1,247 @@
...
+        )
+
+    # Here we mock the same functions than the previous tests plus `glob`,
+    # so we make sure that the glob result is the expected even if the folder
+    # doesn't exist, which is probably the case. This has to be done because
+    # here we tests the `get_env` with clang
+    @mock.patch("pythonforandroid.archs.glob")
+    @mock.patch("pythonforandroid.archs.find_executable")
+    @mock.patch("pythonforandroid.build.ensure_dir")
+    def test_arch_armv7a(
```
AndreMiras: Minor suggestion: Same here with dedicated `TestArchARMv7_a`

tests/test_archs.py
opacam: Good point, it works, but I would rather prefer to use this:

`self.assertEqual(expected_env_gcc_keys, env.keys() & expected_env_gcc_keys)`

Mostly because this way we have a more clear message in case of error

AndreMiras: Sure feel free :+1:

tests/test_archs.py
opacam: I Applied your `subclass` suggestion (slightly different from your proposal)

Thanks @AndreMiras!!!  :smile:

AndreMiras: Nice :+1:

PR #1845, opened 2019-06-08
bugfix: unpack for nonzip archives also needs to compare basename(dir)

sfoerster:
This fixes a case when a recipe for a tar.gz archive attempts to mv to a subdirectory of itself.


inclement:
Thanks!

PR #1847, opened 2019-06-08
Add unittest for module `pythonforandroid.distribution`

opacam:
To enhance our code coverage...a little more

**Note:** In order to optimize the tests, I `mocked` any file creation

### DIFF
--- a/tests/test_distribution.py
+++ b/tests/test_distribution.py
@@ -1,0 +1,216 @@
+import os
+import json
+import unittest
+
+try:
+    from unittest import mock
+except ImportError:
+    # `Python 2` or lower than `Python 3.3` does not
+    # have the `unittest.mock` module built-in
+    import mock
+from pythonforandroid.bootstrap import Bootstrap
+from pythonforandroid.distribution import Distribution
+from pythonforandroid.recipe import Recipe
+from pythonforandroid.util import BuildInterruptingException
+from pythonforandroid.build import Context
+
+dist_info_data = {
+    "dist_name": None,
+    "bootstrap": "sdl2",
+    "archs": ["armeabi", "armeabi-v7a", "x86", "x86_64", "arm64-v8a"],
+    "ndk_api": 21,
+    "use_setup_py": False,
+    "recipes": ["hostpython3", "python3", "sdl2", "kivy", "requests"],
+    "hostpython": "/some/fake/hostpython3",
+    "python_version": "3.7",
+}
+
+
+class TestDistribution(unittest.TestCase):
+    def setUp(self):
+        self.ctx = Context()
+        self.ctx.ndk_api = 21
+        self.ctx.android_api = 27
+        self.ctx._sdk_dir = "/opt/android/android-sdk"
+        self.ctx._ndk_dir = "/opt/android/android-ndk"
+        self.ctx.setup_dirs(os.getcwd())
+        self.ctx.recipe_build_order = [
+            "hostpython3",
+            "python3",
+            "sdl2",
+            "kivy",
+        ]
+
+    def setUp_distribution_with_bootstrap(self, bs, **kwargs):
+        # extend the setUp by configuring a distribution, because some test
+        # needs a distribution to be set to be properly tested
+        self.ctx.bootstrap = bs
+        self.ctx.bootstrap.distribution = Distribution.get_distribution(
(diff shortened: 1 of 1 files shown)


tests/test_distribution.py
```diff
@@ -0,0 +1,218 @@
...
+        self.ctx.bootstrap.distribution = Distribution.get_distribution(
+            self.ctx,
+            name=kwargs.pop("name", "test_prj"),
+            recipes=kwargs.pop("recipes", ["python3", "kivy"]),
+            **kwargs
+        )
+
+    def tearDown(self):
+        if self.ctx.bootstrap:
+            self.ctx.bootstrap.distribution = None
```
AndreMiras: Minor: just wondering if this is needed since we loose the reference to `self.ctx.bootstrap.distribution` in the line below when we do ` self.ctx.bootstrap = None`?

opacam: Thanks!!
you are right :+1:

AndreMiras:
Looks sensible thank you. I'll try to spend more time this evening digging into it a bit more. But having more tests and better coverage cannot harm I guess :smile:

tests/test_distribution.py
```diff
@@ -0,0 +1,216 @@
...
+    @mock.patch("pythonforandroid.distribution.exists")
+    def test_folder_exist(self, mock_exists):
+
+        self.setUp_distribution_with_bootstrap(
+            Bootstrap().get_bootstrap("sdl2", self.ctx)
+        )
+        self.ctx.bootstrap.distribution.folder_exists()
+        mock_exists.assert_called_with(
+            self.ctx.bootstrap.distribution.dist_dir
+        )
```
AndreMiras: Just a general note and personal taste.
I usually also verify the number of times it was called. So in this case it would be `assert_called_once_with()`.
Also I actually prefer to use `call_args_list()` and manually assert myself because I feel like it's easier for the human eye to catch where the assertion is.
Again I'm just sharing some personal taste for the sake of sharing it, but I don't feel like one is superior over the other

tests/test_distribution.py
```diff
@@ -0,0 +1,216 @@
...
+        self.ctx.bootstrap.distribution.delete()
+        mock_rmtree.assert_called_with(
+            self.ctx.bootstrap.distribution.dist_dir
+        )
+
+    @mock.patch("pythonforandroid.distribution.exists")
+    def test_get_distribution_no_name(self, mock_exists):
+
+        mock_exists.return_value = False
```
AndreMiras: Minor: not sure we need to mock `exists()` here since `folders` from `get_distributions()` is empty so we won't get into the loop right?

tests/test_distribution.py
```diff
@@ -0,0 +1,216 @@
...
+        )
+        mock_glob.return_value = ["sdl2-python3"]
+        mock_open_dist_info.side_effect = [
+            mock.mock_open(
+                read_data=json.dumps(dist_info_data_no_ndk_api)
+            ).return_value
+        ]
+
+        dists = self.ctx.bootstrap.distribution.get_distributions(self.ctx)
+        self.assertEqual(dists[0].ndk_api, None)
```
AndreMiras: just a random idea, we could also have mocked `warning` and asserted its call

tests/test_distribution.py
opacam: @AndreMiras, 
First of all, **so many thanks to review those tests** you are awesome :+1:, writing and reviewing those `unittest` is not an easy task because enforce us to keep the eyes in more than one file at once and all the `mock.patch` stuff it's sometimes not easy to follow. About `assert_called_once`, you are right, I should check the number of calls we make to those functions, I will address this in another pr.

**Note:** I prefer to use `assert_called_once_with` because we need less code to have the same result

tests/test_distribution.py
opacam: You are right we could do that, but I more concerned to mock the stuff that could slow down our tests (mostly file operations: copy, paste, pack, unpack, read files...)

AndreMiras: yes I just throw that idea, because I may have mocked it, just to show we're throwing a warning in this specific case. But you're covering the case anyway asserting ndk_api is None. By the way if you like self.assert you also have `self.assertIsNone()`.

PR #1849, opened 2019-06-08
Improved release model documentation

inclement:
Some minor improvements to the doc that I made while going through the release.

### DIFF
--- a/doc/source/contribute.rst
+++ b/doc/source/contribute.rst
@@ -14,7 +14,7 @@ and we'll deal with the rest.
 Development model
 -----------------
 
-python-for-android is developed using the following model::
+python-for-android is developed using the following model:
 
 - The ``master`` branch always represents the latest stable release.
 - The ``develop`` branch is the most up to date with new contributions.
@@ -44,9 +44,9 @@ Creating a new release
 
 New releases follow these steps:
 
-- Create a new branch ``release/YYYY.MM.DD`` based on the ``develop`` branch.
-  - ``git checkout -b release/YYYY.MM.DD develop``
-- Create a Github pull request to merge ``release/YYYY.MM.DD`` into ``master``.
+- Create a new branch ``release-YYYY.MM.DD`` based on the ``develop`` branch.
+  - ``git checkout -b release-YYYY.MM.DD develop``
+- Create a Github pull request to merge ``release-YYYY.MM.DD`` into ``master``.
 - Complete all steps in the `release checklist <release_checklist_>`_,
   and document this in the pull request (copy the checklist into the PR text)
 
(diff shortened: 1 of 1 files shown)


doc/source/contribute.rst
```diff
@@ -44,20 +44,24 @@ Creating a new release
 
 New releases follow these steps:
 
-- Create a new branch ``release/YYYY.MM.DD`` based on the ``develop`` branch.
-  - ``git checkout -b release/YYYY.MM.DD develop``
-- Create a Github pull request to merge ``release/YYYY.MM.DD`` into ``master``.
+- Create a new branch ``release-YYYY.MM.DD`` based on the ``develop`` branch.
+  - ``git checkout -b release-YYYY.MM.DD develop``
+- Create a Github pull request to merge ``release-YYYY.MM.DD`` into ``master``.
```
AndreMiras: I also liked the slash `/`, gitflow has this by default

inclement: I have no strong feelings about it, I just happened to use a dash when I made the first branch, I think because there's already a p4a branch with a dash. I'll just revert this part if you like.

doc/source/contribute.rst
```diff
@@ -44,20 +44,24 @@ Creating a new release
...
-- Tag the release commit in ``master``. Include a short summary of the changes.
-- Create the release distributions: ``python3 setup.py sdist``
+- Tag the release commit in ``master``, with tag ``vYYYY.MM.DD``. Include a short summary of the changes.
+- Create the release distributions: ``python3 setup.py sdist bdist_wheel``
 - Upload the release to pypi: ``python3 -m twine upload``.
-- Upload the release ``.tar.gz`` to the Github tag.
+- Add to the Github release page (see e.g. `this example <https://github.com/kivy/python-for-android/releases/tag/v2019.06.06>`__):
+  - The python-for-android README summary
+  - A short list of major changes in this release, if any
+  - A changelog summarising merge commits since the last release
```
AndreMiras: maybe we should start adopting the http://keepachangelog.com approach then.
The idea is, making a release and a changelog takes time and hurts. Then we want to do it more following the "if it hurts do it more" approach. So the we could have a changelog file or directory that we fill more often than just at release time.
I usually use a CHANGELOG.md file, but since p4a has quite some changes each release the file would be pretty big. So we could follow the same approach as Django and have a directory https://github.com/django/django/tree/2.2.2/docs/releases

inclement: I wondered about this, but let's just change this later if it seems like a good idea. I'm not against it if you want to set it up. That said, I think the changelog isn't so bad to construct if the releases are regular, even this time it didn't take long (and the github ui was convenient for it) and it's 3 months since last release rather than 1.

doc/source/contribute.rst
AndreMiras: OK I was wondering what was the reason behind. I think we can use whatever the person doing the release like. The idea behind is the same; having a release branch.
So let's keep it the way you like

inclement:
Merging but happy to revisit the points you mentioned.

doc/source/contribute.rst
AndreMiras: Yes I'm definitely happy to keep it for later if we think it makes sense

AndreMiras:
Let's merge it! I just wanted to have your advice on these points.

PR #1850, opened 2019-06-08
Hotfix 2019.06.06.post0: added long_description_content_type to setup.py

inclement:
This doesn't change p4a's code, just fixes how the long_description is rendered on pypi. Currently pypi displays the markdown as plain text, because it tries and fails to render as rst.

See [2019.06.06 pypi release description](https://pypi.org/project/python-for-android/2019.6.6/) vs [testpypi release where I fixed this](https://test.pypi.org/project/python-for-android/2019.6.6.1.dev0/#description).

I think this may as well be a post release thing, since it's a trivial fix.


inclement:
I've actually gone ahead and pushed this change to pypi, since it's a doc-only change, but we can still merge it this way.

AndreMiras:
We could also have made another release with just that fix e.g. `2019.06.08`

AndreMiras:
That also means we should try to build the doc as part as our automated testing process to be sure

PR #1852, opened 2019-06-08
pythonpackage can't return build requirements for wheels, make it return an error if attempted

JonasT:
This change ensures the pythonpackage functions return an error when someone attempts to get build requirements of a wheel (beyond the regular runtime dependencies) since this is currently not implemented properly.

It might be possible to determine and obtain this info somehow, but since this is not ever actually used by our `setup.py` functionality I have made sure it raises a `NotImplementedError` now if this is ever attempted, instead of returning incorrect data as previously.

This should also fix the test failure of `test_pythonpackage_basic.py` on current latest `develop`.

### DIFF
--- a/tests/test_pythonpackage_basic.py
+++ b/tests/test_pythonpackage_basic.py
@@ -43,6 +43,8 @@ def fake_metadata_extract(dep_name, output_folder, debug=False):
 
             Lorem Ipsum"""
         ))
+    with open(os.path.join(output_folder, "metadata_source"), "w") as f:
+        f.write(u"wheel")  # since we have no pyproject.toml
 
 
 def test__extract_info_from_package():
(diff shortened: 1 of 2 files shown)


JonasT:
This bug mainly manifests itself in that unit test by the way, not any situation I can think of when actually using the `setup.py` in practice. So there shouldn't be any huge reason to rush merging this

tests/test_pythonpackage_basic.py
```diff
@@ -87,9 +89,25 @@ def test_get_dep_names_of_package():
...
     assert "setuptools" not in dep_names
-    dep_names = get_dep_names_of_package("python-for-android",
-                                         include_build_requirements=True)
-    assert "setuptools" in dep_names
+    try:
+        dep_names = get_dep_names_of_package(
+            "python-for-android", include_build_requirements=True,
+            verbose=True,
+        )
+    except NotImplementedError as e:
```
inclement: I'd be tempted to make an `IsWheelError` subclass or something, on the basis that specific exceptions aren't used often enough, but I'm never really sure about this.

JonasT: @inclement I'm somehow afraid somebody will use that to test if a package is a wheel and then end up with broken code once it is actually implemented and this no longer returns an error (or they use it right but we remove the error so it can no longer be imported once this is implemented and therefore no longer ever thrown). So I think I'd rather stick with a `NotImplementedError` here, personally

JonasT: @inclement what are your thoughts on my thought on this? :stuck_out_tongue: other than this I think it's ready for merge

AndreMiras: I'm merging, we can always eventually address that one later if we feel it's critical

inclement: I wouldn't have approved the PR if I thought this needed changing, sorry if that wasn't clear and thanks for explaining the motivation!

opacam:
Thanks @AndreMiras!!, 
I think that we should have all the tests fully working or we could cause a mess without willing, that is why I labeled this as `Critical` (our `CI` tests depends on this), the other ones labeled can wait

AndreMiras:
I agree we should always keep the build green

PR #1855, opened 2019-06-10
Add unittest for `pythonforandroid.util` and...

opacam:
Also:
  - [x] Remove unused functions from `pythonforandroid.util`
  - [x] Make tox use Python 3.7 for the travis's lint stage (so we can use some unittest methods introduced in python 3.6 (see also: [mock docs](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once))
  - [x] Enhance some tests for `test_distribution` verifying the number of calls we made to some functions. @AndreMiras, this is for you, that reviewed the pr where we add those tests and you noticed those missing verifications :+1: (completes the work started at #1847)
  - [x] ~~Temporary skip some  pythonpackage's tests for python > 3.5 in travis (Because it fails and I don't know why it fails. So this needs further investigation and a fix, @JonasT...could you take a look on this?)~~
  - [x] Add docs for the recent added tests: `test_archs`, `test_distribution` and `test_util`

**Notes:**
  - In order to optimize the tests, I mocked any operation that can slow down our tests
  - The last point of the things done may be removed in case that we fix  `the pythonpackage tests` before merging this

**As a side note:** ~~The pythonpackage's tests that fails are not failing for the same reason addressed in #1852 ([see travis error](https://travis-ci.org/kivy/python-for-android/jobs/543675158#L609-L632)). It seems that fails on `_get_system_python_executable`...~~ The pythonpackage tests issues mentioned are addressed at #1856 and #1852

### DIFF
--- a/pythonforandroid/util.py
+++ b/pythonforandroid/util.py
@@ -1,8 +1,6 @@
 import contextlib
 from os.path import exists, join
 from os import getcwd, chdir, makedirs, walk, uname
-import io
-import json
 import sh
 import shutil
 import sys
@@ -60,79 +58,6 @@ def temp_directory():
 def ensure_dir(filename):
     if not exists(filename):
         makedirs(filename)
-
-
-class JsonStore(object):
-    """Replacement of shelve using json, needed for support python 2 and 3.
-    """
-
-    def __init__(self, filename):
-        super(JsonStore, self).__init__()
-        self.filename = filename
-        self.data = {}
-        if exists(filename):
-            try:
-                with io.open(filename, encoding='utf-8') as fd:
-                    self.data = json.load(fd)
-            except ValueError:
-                print("Unable to read the state.db, content will be replaced.")
-
-    def __getitem__(self, key):
-        return self.data[key]
-
-    def __setitem__(self, key, value):
-        self.data[key] = value
-        self.sync()
-
-    def __delitem__(self, key):
-        del self.data[key]
-        self.sync()
-
-    def __contains__(self, item):
-        return item in self.data
-
-    def get(self, item, default=None):
-        return self.data.get(item, default)
-
-    def keys(self):
-        return self.data.keys()
-
-    def remove_all(self, prefix):
-        for key in self.data.keys()[:]:
-            if not key.startswith(prefix):
-                continue
-            del self.data[key]
-        self.sync()
-
-    def sync(self):
-        # http://stackoverflow.com/questions/12309269/write-json-data-to-file-in-python/14870531#14870531
-        if IS_PY3:
-            with open(self.filename, 'w') as fd:
-                json.dump(self.data, fd, ensure_ascii=False)
-        else:
-            with io.open(self.filename, 'w', encoding='utf-8') as fd:
-                fd.write(unicode(json.dumps(self.data, ensure_ascii=False)))  # noqa F821
-
-
-def which(program, path_env):
-    '''Locate an executable in the system.'''
-    import os
-
-    def is_exe(fpath):
-        return os.path.isfile(fpath) and os.access(fpath, os.X_OK)
-
-    fpath, fname = os.path.split(program)
-    if fpath:
-        if is_exe(program):
-            return program
-    else:
-        for path in path_env.split(os.pathsep):
-            path = path.strip('"')
-            exe_file = os.path.join(path, program)
-            if is_exe(exe_file):
-                return exe_file
-
-    return None
 
 
 def get_virtualenv_executable():
--- a/tests/test_distribution.py
+++ b/tests/test_distribution.py
@@ -27,7 +27,14 @@ dist_info_data = {
 
 
 class TestDistribution(unittest.TestCase):
+    """
+    An inherited class of `unittest.TestCase`to test the module
+    :mod:`~pythonforandroid.distribution`.
+    """
+
     def setUp(self):
+        """Configure a :class:`~pythonforandroid.build.Context` so we can
+        perform our unittests"""
         self.ctx = Context()
         self.ctx.ndk_api = 21
         self.ctx.android_api = 27
@@ -42,8 +49,8 @@ class TestDistribution(unittest.TestCase):
         ]
 
     def setUp_distribution_with_bootstrap(self, bs, **kwargs):
-        # extend the setUp by configuring a distribution, because some test
-        # needs a distribution to be set to be properly tested
+        """Extend the setUp by configuring a distribution, because some test
+        needs a distribution to be set to be properly tested"""
         self.ctx.bootstrap = bs
         self.ctx.bootstrap.distribution = Distribution.get_distribution(
             self.ctx,
@@ -53,9 +60,13 @@ class TestDistribution(unittest.TestCase):
         )
 
     def tearDown(self):
+        """Here we make sure that we reset a possible bootstrap created in
+        `setUp_distribution_with_bootstrap`"""
         self.ctx.bootstrap = None
 
     def test_properties(self):
+        """Test that some attributes has the expected result (for now, we check
+        that `__repr__` and `__str__` return the proper values"""
         self.setUp_distribution_with_bootstrap(
             Bootstrap().get_bootstrap("sdl2", self.ctx)
         )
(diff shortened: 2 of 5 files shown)


pythonforandroid/util.py
```diff
@@ -62,79 +60,6 @@ def ensure_dir(filename):
         makedirs(filename)
 
 
-class JsonStore(object):
```
opacam: I didn't find any entry in our code where we use this class, but maybe was introduced for some future feature...@inclement, is that the case? Can we safely remove this class `JsonStore` and the method `which`?

JonasT:
Can you give me the failing tests output? Also it would be really better if you didn't mark them as skip even temporarily, if we ever accidentally merge this that might be a bit of a problem

tests/test_pythonpackage_basic.py
```diff
@@ -80,6 +93,11 @@ def test_get_package_name():
         shutil.rmtree(temp_d)
 
 
+# Todo: Fix CI tests for python > 3.5
+@pytest.mark.skipif(
```
JonasT: please don't :scream: let me look at the output and I'll see what I can do, these shouldn't be skipped

opacam:
@JonasT, thanks for the very quick response!!
See travis job at [this link](https://travis-ci.org/kivy/python-for-android/jobs/543675158)

pythonforandroid/util.py
inclement: I think this is inherited from the original fork of kivy-ios, no objection to deleting it.

tests/test_util.py
```diff
@@ -0,0 +1,134 @@
...
+            ]
+        )
+
+    def test_current_directory_exception(self):
+        # test chdir to non-existing directory, should raise error
+        # for py3 the exception is FileNotFoundError and IOError for py2, to
+        # avoid introduce conditions, we test with a more generic exception
+        with self.assertRaises(OSError):
+            with util.current_directory("/fake/directory"):
+                # the line below will never be executed
```
inclement: If it will never be executed, why not prove it by raising an error? But this is a super minor point, I have no issue with the test.

tests/test_util.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+
+    def test_walk_valid_filens_sample(self):
+        file_ens = util.walk_valid_filens(
+            "/home/opacam/Devel/python-for-android/tests/",
+            ["__pycache__"],
+            ["*.pyc"],
+        )
+        for i in os.walk("/home/opacam/Devel/python-for-android/tests/"):
+            print(i)
```
inclement: Not sure if I'm missing something here, what is the value in printing this? There are no other checks?

pythonforandroid/util.py
AndreMiras: Nice cleanup :clap: 
I love tests for that :smile:

tests/test_distribution.py
```diff
@@ -85,7 +85,7 @@ def test_delete(self, mock_rmtree):
             Bootstrap().get_bootstrap("sdl2", self.ctx)
         )
         self.ctx.bootstrap.distribution.delete()
-        mock_rmtree.assert_called_with(
+        mock_rmtree.assert_called_once_with(
```
AndreMiras: thanks :smile:

tests/test_util.py
```diff
@@ -0,0 +1,134 @@
...
+        util.ensure_dir("fake_directory")
+        mock_makedirs.assert_called_once_with("fake_directory")
+
+    @mock.patch("shutil.rmtree")
+    @mock.patch("pythonforandroid.util.mkdtemp")
+    def test_temp_directory(self, mock_mkdtemp, mock_shutil_rmtree):
+        mock_mkdtemp.return_value = "/temp/any_directory"
+        with util.temp_directory():
+            mock_mkdtemp.assert_called_once()
+        mock_shutil_rmtree.assert_called_once_with("/temp/any_directory")
```
AndreMiras: Minor remark: we could also have `mock_shutil_rmtree.assert_not_called()` while still inside the context manager

tests/test_util.py
AndreMiras: You could use a `pass`, but I think it's also fine as it is. Also do you know you can chain context managers without indenting multiple `with`, sometimes it's readable enough:
```python
with self.assertRaises(OSError), util.current_directory("/fake/directory"):
  pass
```

tests/test_util.py
```diff
@@ -0,0 +1,134 @@
...
+            [
+                mock.call("/temp/any_directory"),
+                mock.call(os.getcwd()),
+            ]
+        )
+
+    def test_current_directory_exception(self):
+        # test chdir to non-existing directory, should raise error
+        # for py3 the exception is FileNotFoundError and IOError for py2, to
+        # avoid introduce conditions, we test with a more generic exception
```
AndreMiras: Minor/question: Just wondering why don't we use docstrings rather than comments like this?
I guess IDE would better understand docstrings for instance and it's usually the way to document a function/method. Why do you think?

tests/test_util.py
AndreMiras: Yes this is definitely some debugging left over. Probably the while test should go

tests/test_util.py
```diff
@@ -0,0 +1,134 @@
...
+        self.assertIsInstance(file_ens, types.GeneratorType)
+        # given the simulated structure we expect:
+        expected_result = {
+            "/fake_dir/README",
+            "/fake_dir/Lib/abc.pyc",
+            "/fake_dir/Lib/ctypes/util.pyc",
+        }
+        result = set()
+        for i in file_ens:
+            result.add(i)
```
AndreMiras: You could probably directly do a:
```python
result = set(file_ens)
```

AndreMiras:
WIP but already looking good well done :muscle: 
Let us know once this is ready for another review/merge

tests/test_util.py
opacam: Yes, I addressed this by documenting the recent modules we wrote,
But be aware that we don't have it linked in our docs source, so it won't be visible until we enable it
:wink:

tests/test_util.py
opacam: Thanks, already removed

opacam:
Aaaa, @JonasT...we still have an error in our tox tests for `pythonpackage_basic`:

https://travis-ci.org/kivy/python-for-android/jobs/544152688#L423-L428

Now it's different than before, we only have an error so we almost have it :smile:

tests/test_util.py
AndreMiras: Awesome thanks!

JonasT:
@opacam you need to make sure the `virtualenv` module is installed in the travis environment *for python 3.7*, it is not present by default and it's not optional

opacam:
Oh yes, I already saw that, the thing is that `virtualenv` it's already in there:
https://github.com/kivy/python-for-android/blob/d2ff49d12f1aef23bac42b148a88d17b132bf5fe/.travis.yml#L18

To demonstrate that it is already installed in the travis's tox job, I created another branch from this pr's branch, with an extra commit with a couple of lines, just before the command to trigger tox:
```
      before_script:
        - pip install virtualenv
      script:
        # we want to fail fast on tox errors without having to `docker build` first
        - tox -- tests/ --ignore tests/test_pythonpackage.py
```
Yo can see that it's already installed in this travis job:

https://travis-ci.org/opacam/python-for-android/jobs/544471193#L263-L264

So  the `pythonpackage` test isn't detecting it...or maybe I miss something....:thinking:

JonasT:
> the thing is that virtualenv it's already in there:

most likely not 3.7. the 3.7 in that travis image is in a really weird directory, the mechanism you used for install most likely installs it *additionally* and not as replacement. so `python3-virtualenv` is probably for 3.6 or whatever the regular system default is, and won't go into the 3.7 site-packages. you could check if there's `pip3.7` or similar, that's most likely the best way to get it installed (via `pip3.7 install -U virtualenv` or similar)

opacam:
@JonasT :+1:, thanks for your help with this issues with `virtualenv`

Now we have a green tick mark in our tox tests :smile:

PR #1856, opened 2019-06-10
Fix locating system python when it's not in $PATH (weird but happens, apparently)

JonasT:
Testing a fix for locating system python on travis. ~~This will most likely not work yet but I need to see the travis output to figure out what is wrong (it's a travis-specific problem) hence I'm already making the pull request~~ This fixes the tests breaking when the system python is not in the `$PATH` env var for some reason (usually that doesn't happen, but with travis & `python3.7` for some reason it does)

### DIFF
--- a/pythonforandroid/pythonpackage.py
+++ b/pythonforandroid/pythonpackage.py
@@ -34,6 +34,7 @@
 
 
 from io import open  # needed for python 2
+import functools
 import os
 from pep517.envbuild import BuildEnvironment
 from pep517.wrappers import Pep517HookCaller
@@ -172,11 +173,23 @@ def _get_system_python_executable():
 
     def python_binary_from_folder(path):
         def binary_is_usable(python_bin):
+            """ Helper function to see if a given binary name refers
+                to a usable python interpreter binary
+            """
+
+            # Abort if path isn't present at all or a directory:
+            if not os.path.exists(
+                os.path.join(path, python_bin)
+            ) or os.path.isdir(os.path.join(path, python_bin)):
+                return
+            # We should check file not found anyway trying to run it,
+            # since it might be a dead symlink:
             try:
                 filenotfounderror = FileNotFoundError
             except NameError:  # Python 2
                 filenotfounderror = OSError
             try:
+                # Run it and see if version output works with no error:
                 subprocess.check_output([
                     os.path.join(path, python_bin), "--version"
                 ], stderr=subprocess.STDOUT)
@@ -202,6 +215,7 @@ def _get_system_python_executable():
     bad_candidates = []
     good_candidates = []
     ever_had_nonvenv_path = False
+    ever_had_path_starting_with_prefix = False
     for p in os.environ.get("PATH", "").split(":"):
         # Skip if not possibly the real system python:
         if not os.path.normpath(p).startswith(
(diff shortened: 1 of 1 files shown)


JonasT:
@opacam this appears to make it work, the main remaining issues being that virtualenv needs to be installed for every python version (including the new python 3.7, which doesn't seem to be the case) and that #1852 also needs to be merged to get rid of all the test errors

pythonforandroid/pythonpackage.py
```diff
@@ -172,11 +173,22 @@ def _get_system_python_executable():
 
     def python_binary_from_folder(path):
         def binary_is_usable(python_bin):
+            """ Helper function to see if a given binary name refers
+                to a usable python interpreter binary
+            """
+
+            # Abort if path isn't present at all or a directory:
+            if not os.path.exists(os.path.join(path, python_bin)) or \
```
opacam: **Minor suggestion:** I think that this is more readable this way and we avoid using backslash:
```
            if not os.path.exists(
                os.path.join(path, python_bin)
            ) or os.path.isdir(os.path.join(path, python_bin)):
                return
```

pythonforandroid/pythonpackage.py
```diff
@@ -202,17 +214,21 @@ def binary_is_usable(python_bin):
...
                 ):
             continue
 
+        ever_had_path_starting_with_prefix = True
+
         # First folders might be virtualenv/venv we want to avoid:
         if not ever_had_nonvenv_path:
             sep = os.path.sep
-            if ("system32" not in p.lower() and "usr" not in p) or \
+            if ("system32" not in p.lower() and "usr" not in p and
```
opacam: **Another suggestion:** as before, I think it's more readable this way:
```
            if (
                (
                    "system32" not in p.lower()
                    and "usr" not in p
                    and not p.startswith("/opt/python")
                )
                or {"home", ".tox"}.intersection(set(p.split(sep)))
                or "users" in p.lower()
            ):
```

pythonforandroid/pythonpackage.py
JonasT: added, thanks :heart:

pythonforandroid/pythonpackage.py
JonasT: changed it to almost this (I think pep8 wants the binary operators like `or` at the end of the line) I hope you like it, thanks for the suggestion :heart:

JonasT:
> fne to me to keep that in your code style

my style is ugly :scream: lol. I just tend to forget, always good to poke me about it

opacam:
Aaa...don't worry about the code style...I have a solution for that

**I will tell you my secret**...I tend to use a little trick before commit/push anything...:laughing:

There is a python module named [`black`](https://github.com/python/black) wich formats the code in a nice way...since I discovered that, it has been arround my head that we should make use of this module in p4a...it would allow us to forget about those code style discussions and focus with the important stuff. It can be integrated with travis (to perform a code style test and uses  a `pyproject.toml` file to do the configuration). It also can be executed targeting a file only or a folder.

I really recommend you that take a look on this project, I think that it's worth it :wink:

inclement:
Looks like travis failed because the last run predates the wheel fix, I've set it running again.

Issue #1854, opened 2019-06-09
apk build error: bzlib.h: No such file or directory


Fak3:
os: openSUSE tumlbeweed linux
p4a version: git master
android ndk version: r17c

I tried to build apk with p4a:
`p4a apk --private /home/z/pproj/paradox_ssh/ --version=1.7 --bootstrap=sdl2 --local-recipes=./recipes --requirements=python3,kivy-async,openssl,sqlite3,pil,requests,sdl2,plyer --whitelist=./whitelist.txt --permission=CALL_PHONE --permission=INTERNET --orientation=portrait --package=org.spbelect.paradox2 --name="paradox2" --release`

The build failed with an error:
> /home/u1/.local/share/python-for-android/build/other_builds/freetype/armeabi-v7a__ndk_target_19/freetype/src/bzip2/ftbzip2.c:52:19: fatal error: bzlib.h: No such file or directory
> #include <bzlib.h>
>                   ^
> compilation terminated.

full p4a stdout: https://gist.github.com/Fak3/5fca7ea090ddcc78be30571c8f4f91d8

I do have bzlib headers installed:
```
> rpm -ql libbz2-devel
/usr/include/bzlib.h
/usr/lib64/libbz2.so
/usr/lib64/pkgconfig/bzip2.pc
/usr/share/doc/packages/libbz2-devel
/usr/share/doc/packages/libbz2-devel/CHANGES
/usr/share/licenses/libbz2-devel
/usr/share/licenses/libbz2-devel/LICENSE
```
But somehow recipe cannot find the headers

opacam:
I think that your problem is that you are trying to build with `ndk-api=19`, you should use at least 21, try adding `--ndk-api=21` to your p4a build command (but first remove any build made with ndk-api 19)

**See also:**
  - https://gist.github.com/Fak3/5fca7ea090ddcc78be30571c8f4f91d8#file-log-txt-L11
  - https://gist.github.com/Fak3/5fca7ea090ddcc78be30571c8f4f91d8#file-log-txt-L282

Fak3:
@opacam Just tried to do that - changed NDKAPI env var to 21, and removed the ~/.local/share/python-for-android/ 
Got the same error.

opacam:
Well then, we may have a bug in `openSUSE`, could you share the freetype's config.log?

**Note:** Given your filesystem and target `Arch` it should be at:
[...]

PR #1857, opened 2019-06-10
Add `--without-bzip2` to freetype's configure args

opacam:
To fix freetype build for `openSUSE` and maybe other oses.

**Note:** It seems that there is a bug in `bzip2-devel` for debian and derivatives which are distributed without a `pkg-config` file, so when we compile freetype with a debian os (or derivative) we build it without bzip2 support, so we enforce to disable bzip2 support, because we know that freetype works fine without bzip2 support

Resolves: #1854


opacam:
The `tox tests` failed, as expected, until we don't merge #1852 we will get the same tox error, we should address that soon...

PR #1862, opened 2019-06-11
Fix various setup.py processing bugs

JonasT:
I made the setup.py processing unintentionally litter a file :grimacing: :grimacing: (in the user project folder of all places) so that's not very good and here's the fix :smile: along with other fixes:

- fix file litter of constraint.txt
- fix --private relative paths during build.py's setup.py install
- fix "" and "." not being recognized as folder by pythonpackage.py

~~Still WIP because I'm still testing it~~ tested with various actual `setup.py`-based builds!

### DIFF
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -1,7 +1,9 @@
 from __future__ import print_function
 
-from os.path import (join, realpath, dirname, expanduser, exists,
-                     split, isdir)
+from os.path import (
+    abspath, join, realpath, dirname, expanduser, exists,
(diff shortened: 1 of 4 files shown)


JonasT:
For what it's worth, the coverage thing: there is no new code except 2 conditionals that were plain missing, is this computed based on functions present? I did split up a big one simply because it got too nested to read well, but there isn't really any new code - and I actually added tests for the broken handling of pythonpackage of `""` and `"."`, so in a way there's more tested than before :woman_shrugging:

PR #1864, opened 2019-06-13
Make the tox jobs run in parallel &...

opacam:
Also move from branch `master` to `develop` our travis file git commands

Make the tox jobs run in parallel will decrease the timings of the lint stage considerably (from ~5' to ~3'). Also, Making the jobs run in parallel will allow us to split the log output for each test, so it will be easier to find the problem, if any

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,8 @@ env:
 
 jobs:
   include:
-    - stage: lint
-      name: "Tox tests and coverage"
+    - &linting
+      stage: lint
       language: python
       python: 3.7
       before_script:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,7 +13,7 @@ services:
 
 before_install:
   # https://github.com/travis-ci/travis-ci/issues/6069#issuecomment-266546552
-  - git remote set-branches --add origin master
+  - git remote set-branches --add origin develop
```
opacam: Since we change our git flow, now our main branch is `develop`, I feel that we must change this...@AndreMiras, is this right?

.travis.yml
```diff
@@ -23,8 +23,8 @@ env:
 
 jobs:
   include:
-    - stage: lint
-      name: "Tox tests and coverage"
+    - &linting
```
opacam: Here we apply the same technique that we use in the `test` stage, explained [here](https://github.com/kivy/python-for-android/pull/1835#discussion_r291112884)

.travis.yml
```diff
@@ -45,6 +45,14 @@ jobs:
         - tox -- tests/ --ignore tests/test_pythonpackage.py
         # (we ignore test_pythonpackage.py since these run way too long!!
         #  test_pythonpackage_basic.py will still be run.)
+      name: "Tox Pep8"
+      env: TOXENV=pep8
```
opacam: Here we use a tox feature, explained [here](https://tox.readthedocs.io/en/latest/example/general.html#selecting-one-or-more-environments-to-run-tests-against)

.travis.yml
AndreMiras: Yes completely :+1:

PR #1867, opened 2019-06-16
Fix debug build not resulting in gdb-debuggable build

etc0de:
This pull request does the following things:

- fixes that a debug (non-`--release`) `.apk` build won't actually enable gdb debugging in the manifest
- renames `build.py`'s `parse_args` to `parse_args_and_make_package` because that is what it appears to be actually doing smile
- makes SDL2 and any standard `NDKRecipe` build with `NDK_DEBUG=1` when `--release` is not specified to add debugging symbols
- fixes debugging symbols being stripped even when not using `--release`
  which makes using gdb a hassle


pythonforandroid/recipe.py
```diff
@@ -695,6 +701,7 @@ def build_arch(self, arch, *extra_args):
             shprint(
                 sh.ndk_build,
                 'V=1',
+                'NDK_DEBUG=' + ("1" if self.build_as_debuggable else "0"),
```
AndreMiras: I like the way you did because it's clear/explicit, but in theory you could directly cast to int.
Hence:
```python
'NDK_DEBUG=' + int(self.build_as_debuggable),
```
See:
```python
>>> int(True)
1
>>> int(False)
0
```

JonasT: hmm good point. I'll go with whatever you guys prefer, I personally like the explicit way though. unlike in other languages I think the python ternary operator is very neat and readable, while casting bool to int seems a bit arcane to me

JonasT:
@AndreMiras it actually just leverages the `--release` (or lack of that being specified) option of `toolchain.py`. I haven't added any new options, I just added code to pass that into the `build.py` which so far didn't actually retain any knowledge of that option. So it shouldn't clash with anything, and doesn't actually change any options either. I've just updated the `--release` option's text because I found it completely impossible to understand :smile: :smile: :smile:

pythonforandroid/recipe.py
AndreMiras: Yes let's keep it the explicit way :+1:

JonasT:
For what it's worth this is incredibly useful even if just because now `--debug` will also enable SDL2's asserts which can help a lot dealing with crashes or anything will potentially crash. I've run multiple builds with this by now and can only recommend it for inclusion based on my own experience!

JonasT:
As far as I'm concerned this is ready for some more testing & merge,  I manually extracted the `.apk` and checked with `file`it has debugging symbols so I'm fairly confident the basic functionality works!

opacam:
@JonasT, good work :smile:!!

I've tested this with `--release` enabled and without it, with our testapp_sqlite_openssl for python3. I did not checked symbols but it definitively produces bigger apk in non `--release` mode and it works at runtime in both cases. Aaa, to be able to install the apk in `--release` mode, it first must be signed, I did that with a debug key:

```
jarsigner -verbose -sigalg SHA1withRSA -digestalg SHA1 -keystore ~/.android/debug.keystore bdisttest_python3_sqlite_openssl_googlendk-1.1-release-unsigned.apk androiddebugkey -storepass android
```
I also reviewed the code and it looks good to me, so you have my OK, but I will leave the final word to @AndreMiras or @inclement

**Note:** I see that you don't save the build mode into `dist_info.json`  maybe we should or not...:thinking:...better to discuss this point at #1926 (I could add it in there...if we take that road... :wink: )

**As a side note:** I usually don't publish my apks and with this I will end up with bigger apks...unless I build it with `--release` and sign my apk...so I'm wondering...should we have an build mode which strips debug symbols but not enforces to sign apk?...let's say something like `--test-release` or something like that?

JonasT:
Thanks for the feedback :heart: 

You should always sign any public release with your own developer key so only you can publish updates. So I don't think a test release no symbols but with debug signing makes too much sense. However, I think it might make sense to have a debug release with symbols that isn't signed with some random debug key - but that wasn't possible before this pull request either, so I'd say it is a separate feature

AndreMiras:
Thanks again for the effort and discussion guys! I'll review/test/merge tonight.
I think you both have a valid point regarding the debug and apk size.
On one hand yes public release should always be signed with dev key indeed. On the other hands I see one use cases where it's not the case by default.
For instance I publish APK through [F-Droid](https://f-droid.org) via `buildozer android debug`. The F-Droid server doesn't have access to my signing keys. However there's probably a way to sign with their keys, but well p4a/buildozer support in F-Droid is pretty inexistant currently :stuck_out_tongue: 
Nevertheless I think this PR is already a great improvement and will help a lot debugging runtime libraries crashes for instance. So if I don't see regressions I think we're good for a merge

AndreMiras:
Not a good start, this breaks in buildozer it seems.
```
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/tmp/trash/.buildozer/android/platform/python-for-android/pythonforandroid/toolchain.py", line 1201, in <module>
    main()
  File "/tmp/trash/.buildozer/android/platform/python-for-android/pythonforandroid/toolchain.py", line 1195, in main
    ToolchainCL()
  File "/tmp/trash/.buildozer/android/platform/python-for-android/pythonforandroid/toolchain.py", line 580, in __init__
    if args.build_mode == "release":
AttributeError: 'Namespace' object has no attribute 'build_mode'
```
Complete debug log:
```
andre@linux:/tmp/trash$ buildozer android debug
# Check configuration tokens
# Ensure build layout
# Create directory /tmp/trash/.buildozer
# Create directory /tmp/trash/.buildozer/applibs
# Create directory /tmp/trash/.buildozer/android/platform
# Create directory /tmp/trash/.buildozer/android/app
# Check configuration tokens
# Read available permissions from api-versions.xml
# Preparing build
# Check requirements for android
# Run 'dpkg --version'
# Cwd None
Debian 'dpkg' package management program version 1.19.0.5 (amd64).
This is free software; see the GNU General Public License version 2 or
later for copying conditions. There is NO warranty.
# Search for Git (git)
#  -> found at /usr/bin/git
# Search for Cython (cython)
#  -> found at /home/andre/.local/bin/cython
# Search for Java compiler (javac)
#  -> found at /usr/lib/jvm/java-8-openjdk-amd64/bin/javac
# Search for Java keytool (keytool)
#  -> found at /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/keytool
# Install platform
# Run 'git clone -b make_debug_release_debuggable --single-branch https://github.com/JonasT/python-for-android.git python-for-android'
# Cwd /tmp/trash/.buildozer/android/platform
Cloning into 'python-for-android'...
# Run '/usr/bin/python3 -m pip install -q --user \'appdirs\' \'colorama>=0.3.3\' \'jinja2\' \'six\' \'enum34; python_version<"3.4"\' \'sh>=1.10; sys_platform!="nt"\' \'pep517\' \'pytoml\' \'virtualenv\''
# Cwd None
You are using pip version 19.0.3, however version 19.1.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
# Apache ANT found at /home/andre/.buildozer/android/platform/apache-ant-1.9.4
# Android SDK found at /home/andre/.buildozer/android/platform/android-sdk
# Android NDK found at /home/andre/.buildozer/android/platform/android-ndk-r17c
# Installing/updating SDK platform tools if necessary
# Run '/home/andre/.buildozer/android/platform/android-sdk/tools/bin/sdkmanager tools platform-tools'
# Cwd /home/andre/.buildozer/android/platform/android-sdk
[=======================================] 100% Computing updates...             
# Run '/home/andre/.buildozer/android/platform/android-sdk/tools/bin/sdkmanager --update'
# Cwd /home/andre/.buildozer/android/platform/android-sdk
[=======================================] 100% Computing updates...             
# Updating SDK build tools if necessary
# Run '/home/andre/.buildozer/android/platform/android-sdk/tools/bin/sdkmanager --list'
# Cwd /home/andre/.buildozer/android/platform/android-sdk
Installed packages:=====================] 100% Computing updates...             
  Path                   | Version    | Description                    | Location               
  -------                | -------    | -------                        | -------                
  build-tools;29.0.0     | 29.0.0     | Android SDK Build-Tools 29     | build-tools/29.0.0/    
  build-tools;29.0.0-rc3 | 29.0.0 rc3 | Android SDK Build-Tools 29-rc3 | build-tools/29.0.0-rc3/
  build-tools;29.0.1     | 29.0.1     | Android SDK Build-Tools 29.0.1 | build-tools/29.0.1/    
  platform-tools         | 29.0.1     | Android SDK Platform-Tools     | platform-tools/        
  platforms;android-27   | 3          | Android SDK Platform 27        | platforms/android-27/  
  tools                  | 26.1.1     | Android SDK Tools 26.1.1       | tools/                 

Available Packages:
  Path                                                                                     | Version      | Description                                                         
  -------                                                                                  | -------      | -------                                                             
  add-ons;addon-google_apis-google-15                                                      | 3            | Google APIs                                                         
  add-ons;addon-google_apis-google-16                                                      | 4            | Google APIs                                                         
  add-ons;addon-google_apis-google-17                                                      | 4            | Google APIs                                                         
  add-ons;addon-google_apis-google-18                                                      | 4            | Google APIs                                                         
  add-ons;addon-google_apis-google-19                                                      | 20           | Google APIs                                                         
  add-ons;addon-google_apis-google-21                                                      | 1            | Google APIs                                                         
  add-ons;addon-google_apis-google-22                                                      | 1            | Google APIs                                                         
  add-ons;addon-google_apis-google-23                                                      | 1            | Google APIs                                                         
  add-ons;addon-google_apis-google-24                                                      | 1            | Google APIs                                                         
  add-ons;addon-google_gdk-google-19                                                       | 11           | Glass Development Kit Preview                                       
  build-tools;19.1.0                                                                       | 19.1.0       | Android SDK Build-Tools 19.1                                        
  build-tools;20.0.0                                                                       | 20.0.0       | Android SDK Build-Tools 20                                          
  build-tools;21.1.2                                                                       | 21.1.2       | Android SDK Build-Tools 21.1.2                                      
  build-tools;22.0.1                                                                       | 22.0.1       | Android SDK Build-Tools 22.0.1                                      
  build-tools;23.0.1                                                                       | 23.0.1       | Android SDK Build-Tools 23.0.1                                      
  build-tools;23.0.2                                                                       | 23.0.2       | Android SDK Build-Tools 23.0.2                                      
  build-tools;23.0.3                                                                       | 23.0.3       | Android SDK Build-Tools 23.0.3                                      
  build-tools;24.0.0                                                                       | 24.0.0       | Android SDK Build-Tools 24                                          
  build-tools;24.0.1                                                                       | 24.0.1       | Android SDK Build-Tools 24.0.1                                      
  build-tools;24.0.2                                                                       | 24.0.2       | Android SDK Build-Tools 24.0.2                                      
  build-tools;24.0.3                                                                       | 24.0.3       | Android SDK Build-Tools 24.0.3                                      
  build-tools;25.0.0                                                                       | 25.0.0       | Android SDK Build-Tools 25                                          
  build-tools;25.0.1                                                                       | 25.0.1       | Android SDK Build-Tools 25.0.1                                      
  build-tools;25.0.2                                                                       | 25.0.2       | Android SDK Build-Tools 25.0.2                                      
  build-tools;25.0.3                                                                       | 25.0.3       | Android SDK Build-Tools 25.0.3                                      
  build-tools;26.0.0                                                                       | 26.0.0       | Android SDK Build-Tools 26                                          
  build-tools;26.0.1                                                                       | 26.0.1       | Android SDK Build-Tools 26.0.1                                      
  build-tools;26.0.2                                                                       | 26.0.2       | Android SDK Build-Tools 26.0.2                                      
  build-tools;26.0.3                                                                       | 26.0.3       | Android SDK Build-Tools 26.0.3                                      
  build-tools;27.0.0                                                                       | 27.0.0       | Android SDK Build-Tools 27                                          
  build-tools;27.0.1                                                                       | 27.0.1       | Android SDK Build-Tools 27.0.1                                      
  build-tools;27.0.2                                                                       | 27.0.2       | Android SDK Build-Tools 27.0.2                                      
  build-tools;27.0.3                                                                       | 27.0.3       | Android SDK Build-Tools 27.0.3                                      
  build-tools;28.0.0                                                                       | 28.0.0       | Android SDK Build-Tools 28                                          
  build-tools;28.0.1                                                                       | 28.0.1       | Android SDK Build-Tools 28.0.1                                      
  build-tools;28.0.2                                                                       | 28.0.2       | Android SDK Build-Tools 28.0.2                                      
  build-tools;28.0.3                                                                       | 28.0.3       | Android SDK Build-Tools 28.0.3                                      
  build-tools;29.0.0                                                                       | 29.0.0       | Android SDK Build-Tools 29                                          
  build-tools;29.0.1                                                                       | 29.0.1       | Android SDK Build-Tools 29.0.1                                      
  cmake;3.10.2.4988404                                                                     | 3.10.2       | CMake 3.10.2.4988404                                                
  cmake;3.6.4111459                                                                        | 3.6.4111459  | CMake 3.6.4111459                                                   
  docs                                                                                     | 1            | Documentation for Android SDK                                       
  emulator                                                                                 | 29.0.11      | Android Emulator                                                    
  extras;android;gapid;1                                                                   | 1.0.3        | GPU Debugging tools                                                 
  extras;android;gapid;3                                                                   | 3.1.0        | GPU Debugging tools                                                 
  extras;android;m2repository                                                              | 47.0.0       | Android Support Repository                                          
  extras;google;auto                                                                       | 1.1          | Android Auto Desktop Head Unit emulator                             
  extras;google;google_play_services                                                       | 49           | Google Play services                                                
  extras;google;instantapps                                                                | 1.7.0        | Google Play Instant Development SDK                                 
  extras;google;m2repository                                                               | 58           | Google Repository                                                   
  extras;google;market_apk_expansion                                                       | 1            | Google Play APK Expansion library                                   
  extras;google;market_licensing                                                           | 1            | Google Play Licensing Library                                       
  extras;google;simulators                                                                 | 1            | Android Auto API Simulators                                         
  extras;google;webdriver                                                                  | 2            | Google Web Driver                                                   
  extras;m2repository;com;android;support;constraint;constraint-layout-solver;1.0.0        | 1            | Solver for ConstraintLayout 1.0.0                                   
  extras;m2repository;com;android;support;constraint;constraint-layout-solver;1.0.0-alpha4 | 1            | com.android.support.constraint:constraint-layout-solver:1.0.0-alpha4
  extras;m2repository;com;android;support;constraint;constraint-layout-solver;1.0.0-alpha8 | 1            | Solver for ConstraintLayout 1.0.0-alpha8                            
  extras;m2repository;com;android;support;constraint;constraint-layout-solver;1.0.0-beta1  | 1            | Solver for ConstraintLayout 1.0.0-beta1                             
  extras;m2repository;com;android;support;constraint;constraint-layout-solver;1.0.0-beta2  | 1            | Solver for ConstraintLayout 1.0.0-beta2                             
  extras;m2repository;com;android;support;constraint;constraint-layout-solver;1.0.0-beta3  | 1            | Solver for ConstraintLayout 1.0.0-beta3                             
  extras;m2repository;com;android;support;constraint;constraint-layout-solver;1.0.0-beta4  | 1            | Solver for ConstraintLayout 1.0.0-beta4                             
  extras;m2repository;com;android;support;constraint;constraint-layout-solver;1.0.0-beta5  | 1            | Solver for ConstraintLayout 1.0.0-beta5                             
  extras;m2repository;com;android;support;constraint;constraint-layout-solver;1.0.1        | 1            | Solver for ConstraintLayout 1.0.1                                   
  extras;m2repository;com;android;support;constraint;constraint-layout-solver;1.0.2        | 1            | Solver for ConstraintLayout 1.0.2                                   
  extras;m2repository;com;android;support;constraint;constraint-layout;1.0.0               | 1            | ConstraintLayout for Android 1.0.0                                  
  extras;m2repository;com;android;support;constraint;constraint-layout;1.0.0-alpha4        | 1            | com.android.support.constraint:constraint-layout:1.0.0-alpha4       
  extras;m2repository;com;android;support;constraint;constraint-layout;1.0.0-alpha8        | 1            | ConstraintLayout for Android 1.0.0-alpha8                           
  extras;m2repository;com;android;support;constraint;constraint-layout;1.0.0-beta1         | 1            | ConstraintLayout for Android 1.0.0-beta1                            
  extras;m2repository;com;android;support;constraint;constraint-layout;1.0.0-beta2         | 1            | ConstraintLayout for Android 1.0.0-beta2                            
  extras;m2repository;com;android;support;constraint;constraint-layout;1.0.0-beta3         | 1            | ConstraintLayout for Android 1.0.0-beta3                            
  extras;m2repository;com;android;support;constraint;constraint-layout;1.0.0-beta4         | 1            | ConstraintLayout for Android 1.0.0-beta4                            
  extras;m2repository;com;android;support;constraint;constraint-layout;1.0.0-beta5         | 1            | ConstraintLayout for Android 1.0.0-beta5                            
  extras;m2repository;com;android;support;constraint;constraint-layout;1.0.1               | 1            | ConstraintLayout for Android 1.0.1                                  
  extras;m2repository;com;android;support;constraint;constraint-layout;1.0.2               | 1            | ConstraintLayout for Android 1.0.2                                  
  lldb;2.0                                                                                 | 2.0.2558144  | LLDB 2.0                                                            
  lldb;2.1                                                                                 | 2.1.2852477  | LLDB 2.1                                                            
  lldb;2.2                                                                                 | 2.2.3271982  | LLDB 2.2                                                            
  lldb;2.3                                                                                 | 2.3.3614996  | LLDB 2.3                                                            
  lldb;3.0                                                                                 | 3.0.4213617  | LLDB 3.0                                                            
  lldb;3.1                                                                                 | 3.1.4508709  | LLDB 3.1                                                            
  ndk-bundle                                                                               | 20.0.5594570 | NDK                                                                 
  ndk;20.0.5594570                                                                         | 20.0.5594570 | NDK (Side by side) 20.0.5594570                                     
  patcher;v4                                                                               | 1            | SDK Patch Applier v4                                                
  platform-tools                                                                           | 29.0.1       | Android SDK Platform-Tools                                          
  platforms;android-10                                                                     | 2            | Android SDK Platform 10                                             
  platforms;android-11                                                                     | 2            | Android SDK Platform 11                                             
  platforms;android-12                                                                     | 3            | Android SDK Platform 12                                             
  platforms;android-13                                                                     | 1            | Android SDK Platform 13                                             
  platforms;android-14                                                                     | 4            | Android SDK Platform 14                                             
  platforms;android-15                                                                     | 5            | Android SDK Platform 15                                             
  platforms;android-16                                                                     | 5            | Android SDK Platform 16                                             
  platforms;android-17                                                                     | 3            | Android SDK Platform 17                                             
  platforms;android-18                                                                     | 3            | Android SDK Platform 18                                             
  platforms;android-19                                                                     | 4            | Android SDK Platform 19                                             
  platforms;android-20                                                                     | 2            | Android SDK Platform 20                                             
  platforms;android-21                                                                     | 2            | Android SDK Platform 21                                             
  platforms;android-22                                                                     | 2            | Android SDK Platform 22                                             
  platforms;android-23                                                                     | 3            | Android SDK Platform 23                                             
  platforms;android-24                                                                     | 2            | Android SDK Platform 24                                             
  platforms;android-25                                                                     | 3            | Android SDK Platform 25                                             
  platforms;android-26                                                                     | 2            | Android SDK Platform 26                                             
  platforms;android-27                                                                     | 3            | Android SDK Platform 27                                             
  platforms;android-28                                                                     | 6            | Android SDK Platform 28                                             
  platforms;android-29                                                                     | 1            | Android SDK Platform 29                                             
  platforms;android-7                                                                      | 3            | Android SDK Platform 7                                              
  platforms;android-8                                                                      | 3            | Android SDK Platform 8                                              
  platforms;android-9                                                                      | 2            | Android SDK Platform 9                                              
  sources;android-15                                                                       | 2            | Sources for Android 15                                              
  sources;android-16                                                                       | 2            | Sources for Android 16                                              
  sources;android-17                                                                       | 1            | Sources for Android 17                                              
  sources;android-18                                                                       | 1            | Sources for Android 18                                              
  sources;android-19                                                                       | 2            | Sources for Android 19                                              
  sources;android-20                                                                       | 1            | Sources for Android 20                                              
  sources;android-21                                                                       | 1            | Sources for Android 21                                              
  sources;android-22                                                                       | 1            | Sources for Android 22                                              
  sources;android-23                                                                       | 1            | Sources for Android 23                                              
  sources;android-24                                                                       | 1            | Sources for Android 24                                              
  sources;android-25                                                                       | 1            | Sources for Android 25                                              
  sources;android-26                                                                       | 1            | Sources for Android 26                                              
  sources;android-27                                                                       | 1            | Sources for Android 27                                              
  sources;android-28                                                                       | 1            | Sources for Android 28                                              
  system-images;android-10;default;armeabi-v7a                                             | 5            | ARM EABI v7a System Image                                           
  system-images;android-10;default;x86                                                     | 5            | Intel x86 Atom System Image                                         
  system-images;android-10;google_apis;armeabi-v7a                                         | 6            | Google APIs ARM EABI v7a System Image                               
  system-images;android-10;google_apis;x86                                                 | 6            | Google APIs Intel x86 Atom System Image                             
  system-images;android-14;default;armeabi-v7a                                             | 2            | ARM EABI v7a System Image                                           
  system-images;android-15;default;armeabi-v7a                                             | 5            | ARM EABI v7a System Image                                           
  system-images;android-15;default;x86                                                     | 5            | Intel x86 Atom System Image                                         
  system-images;android-15;google_apis;armeabi-v7a                                         | 6            | Google APIs ARM EABI v7a System Image                               
  system-images;android-15;google_apis;x86                                                 | 6            | Google APIs Intel x86 Atom System Image                             
  system-images;android-16;default;armeabi-v7a                                             | 6            | ARM EABI v7a System Image                                           
  system-images;android-16;default;mips                                                    | 1            | MIPS System Image                                                   
  system-images;android-16;default;x86                                                     | 6            | Intel x86 Atom System Image                                         
  system-images;android-16;google_apis;armeabi-v7a                                         | 6            | Google APIs ARM EABI v7a System Image                               
  system-images;android-16;google_apis;x86                                                 | 6            | Google APIs Intel x86 Atom System Image                             
  system-images;android-17;default;armeabi-v7a                                             | 6            | ARM EABI v7a System Image                                           
  system-images;android-17;default;mips                                                    | 1            | MIPS System Image                                                   
  system-images;android-17;default;x86                                                     | 4            | Intel x86 Atom System Image                                         
  system-images;android-17;google_apis;armeabi-v7a                                         | 6            | Google APIs ARM EABI v7a System Image                               
  system-images;android-17;google_apis;x86                                                 | 6            | Google APIs Intel x86 Atom System Image                             
  system-images;android-18;default;armeabi-v7a                                             | 5            | ARM EABI v7a System Image                                           
  system-images;android-18;default;x86                                                     | 4            | Intel x86 Atom System Image                                         
  system-images;android-18;google_apis;armeabi-v7a                                         | 6            | Google APIs ARM EABI v7a System Image                               
  system-images;android-18;google_apis;x86                                                 | 6            | Google APIs Intel x86 Atom System Image                             
  system-images;android-19;default;armeabi-v7a                                             | 5            | ARM EABI v7a System Image                                           
  system-images;android-19;default;x86                                                     | 6            | Intel x86 Atom System Image                                         
  system-images;android-19;google_apis;armeabi-v7a                                         | 38           | Google APIs ARM EABI v7a System Image                               
  system-images;android-19;google_apis;x86                                                 | 38           | Google APIs Intel x86 Atom System Image                             
  system-images;android-21;android-tv;armeabi-v7a                                          | 3            | Android TV ARM EABI v7a System Image                                
  system-images;android-21;android-tv;x86                                                  | 3            | Android TV Intel x86 Atom System Image                              
  system-images;android-21;default;armeabi-v7a                                             | 4            | ARM EABI v7a System Image                                           
  system-images;android-21;default;x86                                                     | 5            | Intel x86 Atom System Image                                         
  system-images;android-21;default;x86_64                                                  | 5            | Intel x86 Atom_64 System Image                                      
  system-images;android-21;google_apis;armeabi-v7a                                         | 30           | Google APIs ARM EABI v7a System Image                               
  system-images;android-21;google_apis;x86                                                 | 30           | Google APIs Intel x86 Atom System Image                             
  system-images;android-21;google_apis;x86_64                                              | 30           | Google APIs Intel x86 Atom_64 System Image                          
  system-images;android-22;android-tv;x86                                                  | 3            | Android TV Intel x86 Atom System Image                              
  system-images;android-22;default;armeabi-v7a                                             | 2            | ARM EABI v7a System Image                                           
  system-images;android-22;default;x86                                                     | 6            | Intel x86 Atom System Image                                         
  system-images;android-22;default;x86_64                                                  | 6            | Intel x86 Atom_64 System Image                                      
  system-images;android-22;google_apis;armeabi-v7a                                         | 24           | Google APIs ARM EABI v7a System Image                               
  system-images;android-22;google_apis;x86                                                 | 24           | Google APIs Intel x86 Atom System Image                             
  system-images;android-22;google_apis;x86_64                                              | 24           | Google APIs Intel x86 Atom_64 System Image                          
  system-images;android-23;android-tv;armeabi-v7a                                          | 12           | Android TV ARM EABI v7a System Image                                
  system-images;android-23;android-tv;x86                                                  | 19           | Android TV Intel x86 Atom System Image                              
  system-images;android-23;android-wear;armeabi-v7a                                        | 6            | Android Wear ARM EABI v7a System Image                              
  system-images;android-23;android-wear;x86                                                | 6            | Android Wear Intel x86 Atom System Image                            
  system-images;android-23;default;armeabi-v7a                                             | 6            | ARM EABI v7a System Image                                           
  system-images;android-23;default;x86                                                     | 10           | Intel x86 Atom System Image                                         
  system-images;android-23;default;x86_64                                                  | 10           | Intel x86 Atom_64 System Image                                      
  system-images;android-23;google_apis;armeabi-v7a                                         | 31           | Google APIs ARM EABI v7a System Image                               
  system-images;android-23;google_apis;x86                                                 | 31           | Google APIs Intel x86 Atom System Image                             
  system-images;android-23;google_apis;x86_64                                              | 31           | Google APIs Intel x86 Atom_64 System Image                          
  system-images;android-24;android-tv;x86                                                  | 20           | Android TV Intel x86 Atom System Image                              
  system-images;android-24;default;arm64-v8a                                               | 7            | ARM 64 v8a System Image                                             
  system-images;android-24;default;armeabi-v7a                                             | 7            | ARM EABI v7a System Image                                           
  system-images;android-24;default;x86                                                     | 8            | Intel x86 Atom System Image                                         
  system-images;android-24;default;x86_64                                                  | 8            | Intel x86 Atom_64 System Image                                      
  system-images;android-24;google_apis;arm64-v8a                                           | 25           | Google APIs ARM 64 v8a System Image                                 
  system-images;android-24;google_apis;armeabi-v7a                                         | 25           | Google APIs ARM EABI v7a System Image                               
  system-images;android-24;google_apis;x86                                                 | 25           | Google APIs Intel x86 Atom System Image                             
  system-images;android-24;google_apis;x86_64                                              | 25           | Google APIs Intel x86 Atom_64 System Image                          
  system-images;android-24;google_apis_playstore;x86                                       | 19           | Google Play Intel x86 Atom System Image                             
  system-images;android-25;android-tv;x86                                                  | 14           | Android TV Intel x86 Atom System Image                              
  system-images;android-25;android-wear-cn;armeabi-v7a                                     | 4            | China version of Android Wear ARM EABI v7a System Image             
  system-images;android-25;android-wear-cn;x86                                             | 4            | China version of Android Wear Intel x86 Atom System Image           
  system-images;android-25;android-wear;armeabi-v7a                                        | 3            | Android Wear ARM EABI v7a System Image                              
  system-images;android-25;android-wear;x86                                                | 3            | Android Wear Intel x86 Atom System Image                            
  system-images;android-25;default;x86                                                     | 1            | Intel x86 Atom System Image                                         
  system-images;android-25;default;x86_64                                                  | 1            | Intel x86 Atom_64 System Image                                      
  system-images;android-25;google_apis;arm64-v8a                                           | 16           | Google APIs ARM 64 v8a System Image                                 
  system-images;android-25;google_apis;armeabi-v7a                                         | 16           | Google APIs ARM EABI v7a System Image                               
  system-images;android-25;google_apis;x86                                                 | 16           | Google APIs Intel x86 Atom System Image                             
  system-images;android-25;google_apis;x86_64                                              | 16           | Google APIs Intel x86 Atom_64 System Image                          
  system-images;android-25;google_apis_playstore;x86                                       | 9            | Google Play Intel x86 Atom System Image                             
  system-images;android-26;android-tv;x86                                                  | 12           | Android TV Intel x86 Atom System Image                              
  system-images;android-26;android-wear-cn;x86                                             | 4            | China version of Android Wear Intel x86 Atom System Image           
  system-images;android-26;android-wear;x86                                                | 4            | Android Wear Intel x86 Atom System Image                            
  system-images;android-26;default;x86                                                     | 1            | Intel x86 Atom System Image                                         
  system-images;android-26;default;x86_64                                                  | 1            | Intel x86 Atom_64 System Image                                      
  system-images;android-26;google_apis;x86                                                 | 14           | Google APIs Intel x86 Atom System Image                             
  system-images;android-26;google_apis;x86_64                                              | 14           | Google APIs Intel x86 Atom_64 System Image                          
  system-images;android-26;google_apis_playstore;x86                                       | 7            | Google Play Intel x86 Atom System Image                             
  system-images;android-27;android-tv;x86                                                  | 7            | Android TV Intel x86 Atom System Image                              
  system-images;android-27;default;x86                                                     | 1            | Intel x86 Atom System Image                                         
  system-images;android-27;default;x86_64                                                  | 1            | Intel x86 Atom_64 System Image                                      
  system-images;android-27;google_apis;x86                                                 | 9            | Google APIs Intel x86 Atom System Image                             
  system-images;android-27;google_apis_playstore;x86                                       | 3            | Google Play Intel x86 Atom System Image                             
  system-images;android-28;android-tv;x86                                                  | 8            | Android TV Intel x86 Atom System Image                              
  system-images;android-28;android-wear-cn;x86                                             | 3            | China version of Wear OS Intel x86 Atom System Image                
  system-images;android-28;android-wear;x86                                                | 3            | Wear OS Intel x86 Atom System Image                                 
  system-images;android-28;default;x86                                                     | 4            | Intel x86 Atom System Image                                         
  system-images;android-28;default;x86_64                                                  | 4            | Intel x86 Atom_64 System Image                                      
  system-images;android-28;google_apis;x86                                                 | 9            | Google APIs Intel x86 Atom System Image                             
  system-images;android-28;google_apis;x86_64                                              | 9            | Google APIs Intel x86 Atom_64 System Image                          
  system-images;android-28;google_apis_playstore;x86                                       | 8            | Google Play Intel x86 Atom System Image                             
  system-images;android-28;google_apis_playstore;x86_64                                    | 8            | Google Play Intel x86 Atom_64 System Image                          
  system-images;android-29;default;x86                                                     | 6            | Intel x86 Atom System Image                                         
  system-images;android-29;default;x86_64                                                  | 6            | Intel x86 Atom_64 System Image                                      
  system-images;android-29;google_apis;x86                                                 | 6            | Google APIs Intel x86 Atom System Image                             
  system-images;android-29;google_apis;x86_64                                              | 6            | Google APIs Intel x86 Atom_64 System Image                          
  system-images;android-29;google_apis_playstore;x86                                       | 6            | Google Play Intel x86 Atom System Image                             
  system-images;android-29;google_apis_playstore;x86_64                                    | 6            | Google Play Intel x86 Atom_64 System Image                          
  system-images;android-Q;android-tv;x86                                                   | 1            | Android TV Intel x86 Atom System Image                              
  tools                                                                                    | 26.1.1       | Android SDK Tools                                                   

# Check that aidl can be executed
# Search for Aidl
# Run '/home/andre/.buildozer/android/platform/android-sdk/build-tools/29.0.1/aidl'
# Cwd None
# Downloading platform api target if necessary
# Android packages installation done.
# Check application requirements
# Check garden requirements
# Compile platform
# Run '/usr/bin/python3 -m pythonforandroid.toolchain create --dist_name=myapp --bootstrap=sdl2 --requirements=python3,kivy --arch armeabi-v7a --copy-libs --color=always --storage-dir="/tmp/trash/.buildozer/android/platform/build" --ndk-api=21'
# Cwd /tmp/trash/.buildozer/android/platform/python-for-android
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/tmp/trash/.buildozer/android/platform/python-for-android/pythonforandroid/toolchain.py", line 1201, in <module>
    main()
  File "/tmp/trash/.buildozer/android/platform/python-for-android/pythonforandroid/toolchain.py", line 1195, in main
    ToolchainCL()
  File "/tmp/trash/.buildozer/android/platform/python-for-android/pythonforandroid/toolchain.py", line 580, in __init__
    if args.build_mode == "release":
AttributeError: 'Namespace' object has no attribute 'build_mode'
# Command failed: /usr/bin/python3 -m pythonforandroid.toolchain create --dist_name=myapp --bootstrap=sdl2 --requirements=python3,kivy --arch armeabi-v7a --copy-libs --color=always --storage-dir="/tmp/trash/.buildozer/android/platform/build" --ndk-api=21
# ENVIRONMENT:
#     CLUTTER_IM_MODULE = 'xim'
#     LC_ALL = 'en_US.UTF-8'
#     NVM_DIR = '/home/andre/bin/nvm-0.33.11'
#     LS_COLORS = 'rs=0:di=01;34:ln=01;36:mh=00:pi=40;33:so=01;35:do=01;35:bd=40;33;01:cd=40;33;01:or=40;31;01:mi=00:su=37;41:sg=30;43:ca=30;41:tw=30;42:ow=34;42:st=37;44:ex=01;32:*.tar=01;31:*.tgz=01;31:*.arc=01;31:*.arj=01;31:*.taz=01;31:*.lha=01;31:*.lz4=01;31:*.lzh=01;31:*.lzma=01;31:*.tlz=01;31:*.txz=01;31:*.tzo=01;31:*.t7z=01;31:*.zip=01;31:*.z=01;31:*.Z=01;31:*.dz=01;31:*.gz=01;31:*.lrz=01;31:*.lz=01;31:*.lzo=01;31:*.xz=01;31:*.zst=01;31:*.tzst=01;31:*.bz2=01;31:*.bz=01;31:*.tbz=01;31:*.tbz2=01;31:*.tz=01;31:*.deb=01;31:*.rpm=01;31:*.jar=01;31:*.war=01;31:*.ear=01;31:*.sar=01;31:*.rar=01;31:*.alz=01;31:*.ace=01;31:*.zoo=01;31:*.cpio=01;31:*.7z=01;31:*.rz=01;31:*.cab=01;31:*.wim=01;31:*.swm=01;31:*.dwm=01;31:*.esd=01;31:*.jpg=01;35:*.jpeg=01;35:*.mjpg=01;35:*.mjpeg=01;35:*.gif=01;35:*.bmp=01;35:*.pbm=01;35:*.pgm=01;35:*.ppm=01;35:*.tga=01;35:*.xbm=01;35:*.xpm=01;35:*.tif=01;35:*.tiff=01;35:*.png=01;35:*.svg=01;35:*.svgz=01;35:*.mng=01;35:*.pcx=01;35:*.mov=01;35:*.mpg=01;35:*.mpeg=01;35:*.m2v=01;35:*.mkv=01;35:*.webm=01;35:*.ogm=01;35:*.mp4=01;35:*.m4v=01;35:*.mp4v=01;35:*.vob=01;35:*.qt=01;35:*.nuv=01;35:*.wmv=01;35:*.asf=01;35:*.rm=01;35:*.rmvb=01;35:*.flc=01;35:*.avi=01;35:*.fli=01;35:*.flv=01;35:*.gl=01;35:*.dl=01;35:*.xcf=01;35:*.xwd=01;35:*.yuv=01;35:*.cgm=01;35:*.emf=01;35:*.ogv=01;35:*.ogx=01;35:*.aac=00;36:*.au=00;36:*.flac=00;36:*.m4a=00;36:*.mid=00;36:*.midi=00;36:*.mka=00;36:*.mp3=00;36:*.mpc=00;36:*.ogg=00;36:*.ra=00;36:*.wav=00;36:*.oga=00;36:*.opus=00;36:*.spx=00;36:*.xspf=00;36:'
#     LESSCLOSE = '/usr/bin/lesspipe %s %s'
#     XDG_MENU_PREFIX = 'gnome-'
#     LANG = 'en_US.UTF-8'
#     DISPLAY = ':0'
#     GNOME_SHELL_SESSION_MODE = 'ubuntu'
#     COLORTERM = 'truecolor'
#     NVM_CD_FLAGS = ''
#     USERNAME = 'andre'
#     XDG_VTNR = '2'
#     SSH_AUTH_SOCK = '/run/user/1000/keyring/ssh'
#     MANDATORY_PATH = '/usr/share/gconf/ubuntu.mandatory.path'
#     XDG_SESSION_ID = '2'
#     USER = 'andre'
#     DESKTOP_SESSION = 'ubuntu'
#     QT4_IM_MODULE = 'xim'
#     TEXTDOMAINDIR = '/usr/share/locale/'
#     GNOME_TERMINAL_SCREEN = '/org/gnome/Terminal/screen/8b3bb1a0_6e26_47e9_bf09_cba324fef17b'
#     DEFAULTS_PATH = '/usr/share/gconf/ubuntu.default.path'
#     PWD = '/tmp/trash'
#     HOME = '/home/andre'
#     TEXTDOMAIN = 'im-config'
#     SSH_AGENT_PID = '2204'
#     QT_ACCESSIBILITY = '1'
#     XDG_SESSION_TYPE = 'x11'
#     XDG_DATA_DIRS = '/usr/share/ubuntu:/usr/local/share:/usr/share:/var/lib/snapd/desktop'
#     XDG_SESSION_DESKTOP = 'ubuntu'
#     GJS_DEBUG_OUTPUT = 'stderr'
#     GTK_MODULES = 'gail:atk-bridge'
#     WINDOWPATH = '2'
#     TERM = 'xterm-256color'
#     SHELL = '/bin/bash'
#     VTE_VERSION = '5202'
#     QT_IM_MODULE = 'ibus'
#     XMODIFIERS = '@im=ibus'
#     IM_CONFIG_PHASE = '2'
#     NVM_BIN = '/home/andre/bin/nvm-0.33.11/versions/node/v10.13.0/bin'
#     XDG_CURRENT_DESKTOP = 'ubuntu:GNOME'
#     GPG_AGENT_INFO = '/run/user/1000/gnupg/S.gpg-agent:0:1'
#     GNOME_TERMINAL_SERVICE = ':1.82'
#     XDG_SEAT = 'seat0'
#     SHLVL = '1'
#     LANGUAGE = 'en_US.UTF-8'
#     GDMSESSION = 'ubuntu'
#     GNOME_DESKTOP_SESSION_ID = 'this-is-deprecated'
#     LOGNAME = 'andre'
#     DBUS_SESSION_BUS_ADDRESS = 'unix:path=/run/user/1000/bus'
#     XDG_RUNTIME_DIR = '/run/user/1000'
#     XAUTHORITY = '/run/user/1000/gdm/Xauthority'
#     XDG_CONFIG_DIRS = '/etc/xdg/xdg-ubuntu:/etc/xdg'
#     PATH = '/home/andre/.buildozer/android/platform/apache-ant-1.9.4/bin:/home/andre/bin/nvm-0.33.11/versions/node/v10.13.0/bin:/home/andre/bin/node_modules/.bin:/home/andre/.local/bin:/home/andre/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin'
#     GJS_DEBUG_TOPICS = 'JS ERROR;JS LOG'
#     SESSION_MANAGER = 'local/linux:@/tmp/.ICE-unix/1937,unix/linux:/tmp/.ICE-unix/1937'
#     LESSOPEN = '| /usr/bin/lesspipe %s'
#     GTK_IM_MODULE = 'ibus'
#     LC_TIME = 'en_GB.UTF-8'
#     _ = '/home/andre/.local/bin/buildozer'
#     PACKAGES_PATH = '/home/andre/.buildozer/android/packages'
#     ANDROIDSDK = '/home/andre/.buildozer/android/platform/android-sdk'
#     ANDROIDNDK = '/home/andre/.buildozer/android/platform/android-ndk-r17c'
#     ANDROIDAPI = '27'
#     ANDROIDMINAPI = '21'
# 
# Buildozer failed to execute the last command
# The error might be hidden in the log above this error
# Please read the full log, and search for it before
# raising an issue with buildozer itself.
# In case of a bug report, please add a full log with log_level = 2
andre@linux:/tmp/trash$
```
If we had better test coverage we would have caught that earlier I guess. I hope I can make some time to have basic `toolchain.py` coverage soon

JonasT:
@AndreMiras I added one first very basic test making sure that calling the argument parsers still works (which would have caught this mistake) :smile:

AndreMiras:
Thanks for the update. Could you review https://github.com/kivy/python-for-android/pull/1933 so we can rebase your PR on it once it's merged?

pythonforandroid/recipe.py
```diff
@@ -577,12 +581,15 @@ def list_recipes(cls, ctx):
...
-    def get_recipe(cls, name, ctx):
+    def get_recipe(cls, name, ctx, build_as_debuggable=False):
         '''Returns the Recipe with the given name, if it exists.'''
         name = name.lower()
         if not hasattr(cls, "recipes"):
             cls.recipes = {}
         if name in cls.recipes:
+            if ctx is not None:
+                cls.recipes[name].ctx = ctx
+            cls.recipes[name].build_as_debuggable = build_as_debuggable
```
AndreMiras: This is really really confusing me. Why do we even have to do this since it's been done here below via:
```python
recipe.build_as_debuggable = build_as_debuggable
recipe.ctx = ctx
cls.recipes[name.lower()] = recipe
```
Plus the whole thing seems unnatural to me, we're passing Recipe attributes from here while it's usually set a recipe level, feels like we're missing design patterns/vision somehow, no?
Couldn't the context be used to pass such things around?

pythonforandroid/toolchain.py
```diff
@@ -204,9 +208,10 @@ def build_dist_from_args(ctx, dist, args):
                   ignore_project_setup_py=getattr(
                       args, "ignore_setup_py", False
                   ),
+                  debug_build=(args.build_mode == "debug"),
                  )
 
-    ctx.bootstrap.run_distribute()
+    ctx.bootstrap.run_distribute(debug_build=(args.build_mode == "debug"))
```
AndreMiras: Minor: if it's the second time we use such `debug_build` as a boolean, maybe we could create a temp variable.
```python
debug_build = args.build_mode == "debug"
```

tests/test_distribution.py
```diff
@@ -84,11 +84,12 @@ def test_folder_exist(self, mock_exists):
         :meth:`~pythonforandroid.distribution.Distribution.folder_exist` is
         called once with the proper arguments."""
 
+        mock_exists.return_value = False
         self.setUp_distribution_with_bootstrap(
             Bootstrap().get_bootstrap("sdl2", self.ctx)
         )
         self.ctx.bootstrap.distribution.folder_exists()
-        mock_exists.assert_called_once_with(
+        mock_exists.assert_called_with(
```
AndreMiras: So is it called more than once now, how is that?
Is it worth documenting it with `call_args_list` then?

pythonforandroid/recipe.py
JonasT: @AndreMiras for what it's worth the additional `cls.recipes[name].ctx = ctx` here is vital because the unit tests will call this function and obtain the same recipe via *multiple* contexts (for the different tests) and break otherwise in weird ways

Regarding `build_as_debuggable`, you are completely right. I will look into moving it :+1: :clap: makes more sense on the context

tests/test_distribution.py
JonasT: It is called more than once, I have actually been thinking I should probably check why :joy: I changed it because it didn't seem obvious to me why it'd be important that it is only called once, it just didn't seem like a condition important to guarantee

JonasT:
Thanks for all the feedback! I'll rebase and put `build_as_debuggable` on the build context, so I'll make this `WIP` for now since it doesn't make sense to merge this yet in its current form

pythonforandroid/recipe.py
AndreMiras: Thanks for looking into this :hearts:

tests/test_distribution.py
AndreMiras: Yes I agree the condition may indeed not be important to guarantee. But if the behavior changed and it's not so obvious why, sometimes it's worth to dig slightly more. We may see we introduced something we didn't want to.
I can help digging later when you're done introducing the context update

pythonforandroid/build.py
```diff
@@ -91,6 +91,9 @@ class Context(object):
     '''A build context. If anything will be built, an instance this class
     will be instantiated and used to hold all the build state.'''
 
+    build_as_debuggable = False
+    # Whether to build with debugging symbols
```
AndreMiras: super-nitpick: I think we usually put comments the line before or the line in front. Docstrings however goes after

pythonforandroid/build.py
```diff
@@ -549,7 +552,10 @@ def build_recipes(build_order, python_modules, ctx, project_dir,
             ('The requirements ({}) were not found as recipes, they will be '
              'installed with pip.').format(', '.join(python_modules)))
 
-    recipes = [Recipe.get_recipe(name, ctx) for name in build_order]
+    recipes = [
+        Recipe.get_recipe(name, ctx)
+        for name in build_order
+    ]
```
AndreMiras: Nitpick: At some point we may introduce black and this will move again which introduces noise.
Basically since that line wasn't already 79 chars, black would just make it one line again I suppose.

pythonforandroid/recipe.py
```diff
@@ -583,6 +583,8 @@ def get_recipe(cls, name, ctx):
         if not hasattr(cls, "recipes"):
             cls.recipes = {}
         if name in cls.recipes:
+            if ctx is not None:
+                cls.recipes[name].ctx = ctx
```
AndreMiras: That really gets me curious :thinking:
Why would we need to override the context of the cached recipe? It should already be there via this https://github.com/kivy/python-for-android/blob/v2019.07.08/pythonforandroid/recipe.py#L610 right?

pythonforandroid/recipe.py
```diff
@@ -998,7 +1001,8 @@ def build_cython_components(self, arch):
                 info('First build appeared to complete correctly, skipping manual'
                      'cythonising.')
 
-            self.strip_object_files(arch, env)
+            if not self.ctx.build_as_debuggable:
+                self.strip_object_files(arch, env)
```
AndreMiras: Minor suggestion: just wondering since we're calling this snippet multiple times, maybe we want to make a short helper function/method out of it, e.g.
```python
strip_if_release(self, ctx, env):
    if ctx.build_as_debuggable():
        return
    self.strip_object_files(ctx.archs[0], env, build_dir=ctx.build_dir)
```
Do you think that would make sense?

pythonforandroid/toolchain.py
```diff
@@ -587,6 +594,9 @@ def add_parser(subparsers, *args, **kwargs):
 
         self.ctx = Context()
         self.ctx.use_setup_py = getattr(args, "use_setup_py", True)
+        self.ctx.build_as_debuggable = getattr(
+            args, "build_mode", "debug"
```
AndreMiras: Minor: seems like we're repeating ourselves in a similar fashion few times, enough to make an helper function maybe.
Maybe something like:
```python
def build_as_debuggable(args):
    return getattr(args, "build_mode", "debug") == "release"
```

pythonforandroid/recipe.py
JonasT: Well the fishiness is that the code is written under the assumption just one context will ever be relevant which obviously isn't true at least for the tests, so the whole static caching may not be an ideal design. But for regular program execution of p4a it probably won't matter, but sure it is pretty weird in its inherent assumptions - not that this is really caused by this pull request

pythonforandroid/build.py
JonasT: oh oops, I actually just misread the later lines thinking they had the comments after and was slightly confused by that :joy: that makes so much more sense

pythonforandroid/recipe.py
AndreMiras: OK then if it's the test that are misbehaving, should the tests be updated?
Anyway that looked properly weird to me so I checked out the branch and tried removing that lines. Tests are passing just fine

JonasT:
> Something that disappoints me is it has absolutely no unit tests.

oh no :scream: and yeah, you're right


I'm on it! :+1: I should manage to produce at least one decent test for this on the `Context` level I'm sure

pythonforandroid/recipe.py
JonasT: @AndreMiras it depends on the test execution order I think, and if you just look at the tests it is obvious it is inherently, necessarily going to blow up. Look in `tests/test_recipe.py`, multiple tests construct a `Context()` more than once and these contexts are used for obtaining a recipe, so obviously if any ever obtain the same recipe you already got the disaster spelled out (because now you can possibly get the recipe attached with the wrong context of a previous test, and that can introduce all sorts of subtle wrong behavior)

Fixing the tests from my first conclusions doesn't make too much sense, because why even have the `ctx` parameter for getting a recipe if it's ignored? I think the bandaid most reasonable fix is what I added here, and the long-term fix might be to rethink this cached recipe approach in different ways (e.g. cache it with a dictionary mapping to context with per-context-and-recipe-pair singleton rather than global per-recipe singleton)

AndreMiras: OK I think I understand what you mean. But then in this case I still feel we're increasing the technical debt (if any) by adding this line that just feels unnatural to me :confused: 
Unit tests should be definition be isolated and not leave the context modified from one to another unless we specifically want to. And we should not put a workaround in the code to fix this misbehaviour. I don't think static attributes is anything new in Python or in other languages, same things like with singleton and stuff. When we have one we need to carefully write tests to not leave it in a state we don't want to. I know I'm completely guilty of writing some of such tests, thinking "it's fine I'll address when it impacts us"

JonasT: Just to emphasize this again, I think it is wrong to fix this in the tests. `get_recipe` has a context/`ctx` parameter, it is completely counter-intuitive and somewhat nonsensical to me that it is ever ignored. So if there is any fix I think it has to be inside `get_recipe`

I'm open to dropping this from the pull request (I put it on because I saw spurious test failures) but I would expect us to see weird errors down the line unless we do something about it. I agree this bandaid is not the nicest way to fix it, but breaking up the singleton design entirely also could have tons of implications and break a lot of things when done thoughtlessly, so I don't think a good solution will come quickly.

I'm against "fixing" anything in the tests themselves since again, I think this is just hiding the design problem of `get_recipe()` here and a worse technical debt than this workaround

AndreMiras: Very sad to ear :cry: 
In the order of preference I personally have (the first being my preferred solution):
1) of course a proper fix by design if you think it's a design flow (but as you said that would involve a lot of changes)
2) introduce "work-around" in the tests (I've implemented an example in https://github.com/kivy/python-for-android/pull/1948)
3) introduce "work-around" in the code (what you're currently implementing)

Anyway let's assume we keep your implementation of `3)`. Then could you please write a dedicated and comprehensive test (and docstring) showing this is needed and avoiding later regressions. I reckon the test should be fairly easy to write I think.
Then two thing come to my mind.
1) with your workaround, do we still need this line then https://github.com/kivy/python-for-android/blob/v2019.07.08/pythonforandroid/recipe.py#L610 ?
2) why in your workaround `None` is treated differently? Let's cover that case as well

JonasT: I think a work-around in a test is just masking it, so I really don't like the idea at all. However what you proposed as workaround actually seems like a useful long-term stability measure so it doesn't seem like the worst idea - just not good for dealing with this actual issue

I think we need to fix it, I'm just not sure how yet. I'll sleep about it and see if I can think of what might be the best way to address this :+1:

JonasT: (oh and I'll also back this "fix" out of this pull request here unless you disagree, since this is growing way too big/unrelated and needs more thought. I suggest we should handle it separately)

AndreMiras: Yes I would prefer to not have any of theses workaround or fixes in the PR. But if we do, then I would like them properly unit tested

JonasT:
@AndreMiras I'm struggling a bit to get some of the mocking to work, you wouldn't happen to know why travis/tox is showing this error?

`TypeError: Need a valid target to patch. You supplied: 'project_has_setup_py'`

For mocking `def project_has_setup_py(...)` in `pythonforandroid/build.py` like this in `tests/test_build.py`:

https://github.com/kivy/python-for-android/blob/5f30077961608fb802f3eb3fafea66ec8afe878c/tests/test_build.py#L37

Also, the `pythonforandroid.util.current_directory` `mock.patch` doesn't error out, but I've noticed it doesn't actually work - when the test runs that far it doesn't call the mocked routine but the original one, and again I don't know why :thinking: 

I'll probably find out myself soon enough, but if you have a minute to have a look & a good idea that could be super useful :heart: :heart_eyes:

AndreMiras:
Yes the mock path is wrong. Most of the time you need to mock the path where it's imported, not where it's defined.
Hence it should be `pythonforandroid.build.current_directory` not `pythonforandroid.util.current_directory`

tests/test_build.py
```diff
@@ -23,3 +23,33 @@ def test_run_pymodules_install_optional_project_dir(self):
...
         assert m_info.call_args_list[-1] == mock.call(
             'No Python modules and no setup.py to process, skipping')
+
+    def test_strip_if_debuggable(self):
+        ctx = mock.Mock()
+        ctx.python_recipe.major_minor_version_string = "python3.6"
+        ctx.get_site_packages_dir.return_value = "test-doesntexist"
+        ctx.build_dir = "nonexistant_directory"
+        ctx.archs = ["arm64"]
+        ctx.build_as_debuggable = False
```
AndreMiras: Minor: This one is not needed right, since you're doing `ctx.build_as_debuggable = True` later on just before calling `run_pymodules_install`

tests/test_build.py
```diff
@@ -23,3 +23,33 @@ def test_run_pymodules_install_optional_project_dir(self):
...
+
+        modules = ["mymodule"]
+        project_dir = None
+        with mock.patch('pythonforandroid.build.info'), \
+                mock.patch('sh.Command'),\
+                mock.patch('pythonforandroid.build.open'),\
+                mock.patch('pythonforandroid.build.shprint'),\
+                mock.patch('pythonforandroid.build.current_directory'),\
+                mock.patch('pythonforandroid.build.CythonRecipe') as m_CythonRecipe, \
```
AndreMiras: Minor: I usually try to patch the actual method to have the least impact, hence something like:
```python
mock.patch('pythonforandroid.build.CythonRecipe.strip_object_files') as m_strip_object_files
```

tests/test_build.py
```diff
@@ -23,3 +23,33 @@ def test_run_pymodules_install_optional_project_dir(self):
             assert run_pymodules_install(ctx, modules, project_dir) is None
         assert m_info.call_args_list[-1] == mock.call(
             'No Python modules and no setup.py to process, skipping')
+
+    def test_strip_if_debuggable(self):
+        ctx = mock.Mock()
```
AndreMiras: I'm surprised we're not using an actual `Context` object, but I guess why not :smile:

tests/test_build.py
JonasT: it felt safer that way, to reduce any side effects I'm not interested in and I don't really need any functionality of the recipe anyway in this test, so it was trivial to mock a bit more far-reaching

pythonforandroid/recipe.py
JonasT: I just looked at this, and since it would need different parameters and essentially just saves one line in two places, it seems to me like this would actually complicate code flow more than worth in saving. But that's just my opinion and as usual I'm happy to add it in if you  really want me to :+1:

pythonforandroid/toolchain.py
JonasT: I just tried this, and I have the feeling the actual problem is we want something like #1812 instead so we can do simple `(args.build_mode == "debug")` everywhere and it'll look ok. The problem is that sometimes "release" and "debug" are checked separately and it looks like someone may want to add a mode in these places so dropping this in everywhere would kinda look weird I think, but in only half the places it also looks weird. I think the real problem really is `hasattr`/`getattr` making this look ugly

pythonforandroid/build.py
JonasT: good point, removed it!

pythonforandroid/recipe.py
AndreMiras: Yes I'm also fine you keep it this way :+1:

pythonforandroid/toolchain.py
AndreMiras: Yes `hasattr`/`getattr` definitely makes it look hugly, but testing for debug or release for me is just like testing for True or False. It's a two state, so you can always negate it. Yes it can also be three state, but I don't see it as a problem.
Anyway the PR looks fine overall and it's unit tested, so I'm also happy to keep it this way

tests/test_build.py
```diff
@@ -23,3 +23,32 @@ def test_run_pymodules_install_optional_project_dir(self):
...
+                mock.patch('pythonforandroid.build.current_directory'),\
+                mock.patch('pythonforandroid.build.CythonRecipe') as m_CythonRecipe, \
+                mock.patch('pythonforandroid.build.project_has_setup_py') as m_project_has_setup_py, \
+                mock.patch('pythonforandroid.build.run_setuppy_install'):
+            m_project_has_setup_py.return_value = False
+
+            # Make sure it is NOT called when debug build:
+            ctx.build_as_debuggable = True
+            assert run_pymodules_install(ctx, modules, project_dir) is None
+            self.assertFalse(m_CythonRecipe().strip_object_files.called)
```
AndreMiras: Minor: why don't we make it consistent via the `assert` keyword?
```python
assert m_CythonRecipe().strip_object_files.called is False
```
Note how I purposely didn't write:
```python
assert not m_CythonRecipe().strip_object_files.called
```
Because I also want to enforce type checking, specially for mocks and unit testing

AndreMiras:
Hi @jtc0de, I got side tracked and never merged. But this is looking good I think.
Can you rebase so I can test on an actual app and merge?

tmarplatt:
Hi @etc0de I have a few questions: does this fix enable using `ndk-gdb` for native debugging? I see that the manifest file enables `android:debuggable`, which is a requirement.

Have you tested debugging built apps on an Android device? If so, what is the process like?

Another: Say a recipe (which is not inheriting `NDKRecipe`) builds a library and installs it under site-packages. What requirement is there to use the library's debug symbols for debugging a crash or segfault?

etc0de:
> Have you tested debugging built apps on an Android device? If so, what is the process like?

Yes, you open the resulting APK with the debugging symbols directly in Android Studio (without using a regular project file) and then just press the regular debug launch button and the IDE does everything like launching it, connecting gdb, etc. This also already works *without* debugging symbols and without this pull request, but obviously then all the traces you get in gdb are useless & cryptic which this pull request addresses. It just fixes the gdb output, basically.

> Another: Say a recipe (which is not inheriting NDKRecipe) builds a library and installs it under site-packages. What requirement is there to use the library's debug symbols for debugging a crash or segfault?

The recipe needs to be changed to add in debugging symbols when the build context's `build_as_debuggable` is `True` as I did for the `NDKRecipe`. There *might* be some way for p4a to force this for some via some `CFLAGS` environment hacks added in, but I haven't tried that yet. Without any such changes, these libraries just won't have proper symbols and as before the traces going into it in gdb just won't show useful info.

inclement:
Just fixed the only merge conflict, assuming the tests pass then let's go ahead and merge.

etc0de:
@inclement what sort of conflict did you get? I just attempted a rebase as a test and it went without a conflict, maybe merging develop into it isn't that optimal

AndreMiras:
Looking good to me, thanks for fixing the conflicts.
I would say let's not wait for further conflicts. I'll try to run an app on device today and merge if nothing breaks

PR #1869, opened 2019-06-16
customizability

zworkb:
# customizability options for p4a

This PR introduces a set of options which deal with integration of java/kotlin libraries

- android entrypoint: this option already existed in buildozer but was not implemented in p4a.
    this allows to specify a custom class inherited class from org.kivy.android.PythonActivity as entrypoint
- android apptheme: specifies a custom app-theme for the main activity, will be generated into the manifest
- compile-options
- gradle-repository: with this option it is possible to specify additional repositories for the gradle build
- packaging-options: results in packagingOptions in the gradle script, can be necessary to resolve conflicts in java/kotlin packages

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -663,6 +663,24 @@ tools directory of the Android SDK.
                               'https://developer.android.com/guide/'
                               'topics/manifest/'
                               'activity-element.html'))
+
+    ap.add_argument('--android-entrypoint', dest='android_entrypoint',
+                    default='org.kivy.android.PythonActivity',
+                    help='Defines which java class will be used for startup, usually a subclass of PythonActivity')
+    ap.add_argument('--android-apptheme', dest='android_apptheme',
+                    default='@android:style/Theme.NoTitleBar',
+                    help='Defines which app theme should be selected for the main activity')
+    ap.add_argument('--add-compile-option', dest='compile_options', default=[],
+                    action='append', help='add compile options to gradle.build')
+    ap.add_argument('--add-gradle-repository', dest='gradle_repositories',
+                    default=[],
+                    action='append',
+                    help='Ddd a repository for gradle')
+    ap.add_argument('--add-packaging-option', dest='packaging_options',
+                    default=[],
+                    action='append',
+                    help='Dndroid packaging options')
+
     ap.add_argument('--wakelock', dest='wakelock', action='store_true',
                     help=('Indicate if the application needs the device '
                           'to stay on'))
--- a/pythonforandroid/bootstraps/common/build/templates/build.tmpl.gradle
+++ b/pythonforandroid/bootstraps/common/build/templates/build.tmpl.gradle
@@ -13,68 +13,82 @@ allprojects {
     repositories {
         google()
         jcenter()
-	    flatDir {
-	    	dirs 'libs'
-	    }
+        {%- for repo in args.gradle_repositories %}
+        {{repo}}
+        {%- endfor %}
+        flatDir {
+            dirs 'libs'
+        }
     }
 }
 
 apply plugin: 'com.android.application'
 
 android {
-	compileSdkVersion {{ android_api }}
-	buildToolsVersion '{{ build_tools_version }}'
-	defaultConfig {
-		minSdkVersion {{ args.min_sdk_version }}
-		targetSdkVersion {{ android_api }}
-		versionCode {{ args.numeric_version }}
-		versionName '{{ args.version }}'
-	}
+    compileSdkVersion {{ android_api }}
+    buildToolsVersion '{{ build_tools_version }}'
+    defaultConfig {
+        minSdkVersion {{ args.min_sdk_version }}
+        targetSdkVersion {{ android_api }}
+        versionCode {{ args.numeric_version }}
+        versionName '{{ args.version }}'
+    }
 
-	{% if args.sign -%}
-	signingConfigs {
-		release {
-			storeFile file(System.getenv("P4A_RELEASE_KEYSTORE"))
-			keyAlias System.getenv("P4A_RELEASE_KEYALIAS")
-			storePassword System.getenv("P4A_RELEASE_KEYSTORE_PASSWD")
-			keyPassword System.getenv("P4A_RELEASE_KEYALIAS_PASSWD")
-		}
-	}
+    {% if args.sign -%}
+    signingConfigs {
+        release {
+            storeFile file(System.getenv("P4A_RELEASE_KEYSTORE"))
+            keyAlias System.getenv("P4A_RELEASE_KEYALIAS")
+            storePassword System.getenv("P4A_RELEASE_KEYSTORE_PASSWD")
+            keyPassword System.getenv("P4A_RELEASE_KEYALIAS_PASSWD")
+        }
+    }
     {%- endif %}
 
-	buildTypes {
-		debug {
-		}
-		release {
-			{% if args.sign -%}
-			signingConfig signingConfigs.release
-			{%- endif %}
-		}
-	}
+    {% if args.packaging_options -%}
+    packagingOptions {
+        {%- for option in args.packaging_options %}
+        {{option}}
+        {%- endfor %}
+    }
+    {%- endif %}
+
+    buildTypes {
+        debug {
+        }
+        release {
+            {% if args.sign -%}
+            signingConfig signingConfigs.release
+            {%- endif %}
+        }
+    }
 
     compileOptions {
         sourceCompatibility JavaVersion.VERSION_1_7
         targetCompatibility JavaVersion.VERSION_1_7
+        {%- for option in args.compile_options %}
+        {{option}}
+        {%- endfor %}
     }
 
     sourceSets {
         main {
             jniLibs.srcDir 'libs'
-            }
+        }
     }
 
 }
 
 dependencies {
-	{%- for aar in aars %}
-	compile(name: '{{ aar }}', ext: 'aar')
-	{%- endfor -%}
-        {%- for jar in jars %}
-        compile files('src/main/libs/{{ jar }}')
-        {%- endfor -%}
-	{%- if args.depends -%}
-	{%- for depend in args.depends %}
-	compile '{{ depend }}'
-	{%- endfor %}
-	{%- endif %}
+    {%- for aar in aars %}
+    compile(name: '{{ aar }}', ext: 'aar')
+    {%- endfor -%}
+    {%- for jar in jars %}
+    compile files('src/main/libs/{{ jar }}')
+    {%- endfor -%}
+    {%- if args.depends -%}
+    {%- for depend in args.depends %}
+    compile '{{ depend }}'
+    {%- endfor %}
+    {%- endif %}
 }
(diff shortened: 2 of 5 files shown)


JonasT:
Thanks, it does look very interesting and the options seem reasonably named!

Please don't be discouraged but looking at it I still have to ask: I'm wondering, how common is it to use additional Java libraries? For me Cython, C or pyjnius always seemed like the obvious options. I'm asking because it does appear to add a medium amount of complexity to the gradle file and some options that aren't immediately obvious to someone not familiar with the ecosystem like me (and I imagine other people helping out with p4a) and I think there is some benefit in keeping things simple

Also related to that, what does the app theme do other than change the title bar? Or is that only relevant if using Java libraries for UI? (If yes, then again I'm wondering if that is actually a thing many p4a users actually do and if that is worth adding)

zworkb:
Thanks for your answer!

as you mentioned pyjnius is the obvious way to call into java, that is right. These options help to integrate java/kotlin libraries that can be called via jnius.
Perhaps the best is that I explain the general usecase from a customer project that we are currently working on:
It involves a Python based backend running on the device which deals with object detection using OpenCV, database storage and server communication. It is a customer requirement that the whole frontend is implemented in native Android with Kotlin to provide a native Android user experience.
This app makes heavy use of the camera and the Kivy camera has on some device a startup time for the live view of over 15 secs, which was inacceptable for the customer needs (that was only one reason for the wish for a native UI)

We tried different ways to do that, one was a service_only bootstrap which worked out fine but involves installation of two different apk's and twice acquiring permissions (for frontend and backend).
This lead to the solution where we pack the frontend into an .aar and ship that with the kivy project, which works fine now using these additions to p4a and buildozer.

These requirements involved the integration of a set of java dependencies (that is already provided by the p4a-option --depend and others).
The integration of these external dependencies lead to the necessity of adding additional maven repositories, setting java compatibility levels, packaging options.

The integration of the Kotlin-based Activity made it necessary to set a different AppTheme (the one from the native activity) in the AndroidManifest, otherwise the gradle build fails with a clash in merging the manifest of the .aar and the one of the p4a toolchain. 

The bottom line is that these additions help out in projects where p4a is used as a backend for a native UI.

I hope I could point out the use of these additional options fitting to the options already implemented for integration of java/kotlin dependencies.
  - 
I you are interested I can provide an example that showcases the native UI with a p4a backend.

As I have seen this could also be seen as an answer to [this old question on reddit](https://www.reddit.com/r/Kotlin/comments/6dst4u/python_backend_kotling_frontend_possible/)

rnixx:
@JonasT 

> I'm wondering if that is actually a thing many p4a users actually do and if that is worth adding

At least it gives the p4a user the option to do it, thus yes, it's worth it. If it's then used by many p4a users is not really predictable. The Approach of including native UI parts as aar (or any other lib) has some going for it and the changes are not immersive, the "keep it simple" argument feels a little misplaced.

JonasT:
>  At least it gives the p4a user the option to do it, thus yes, it's worth it

I feel like we spend a lot of time about removing options, like pygame bootstrap, at some point soon we might deprecate python 2, and maybe the launcher will at some point become an external thing (at least I suggested that.) This doesn't seem like a huge addition so forgive me that I go on so much about this, if the others feel like this is worth merging I won't stand in the way for sure, but it is a road to hell paved with good intentions to just put in all options just because 1-2 users would use it

>  If it's then used by many p4a users is not really predictable

Well I hoped there were some known other users asking for it. I have never seen any, hence my question. And it doesn't seem that obvious to me to do, since if you want to get that deep into java APIs, you're bound to get into a painful clashing area at some point - wouldn't most just write a Java app then? (I don't know the answer, I'm just posing the question)

>  the "keep it simple" argument feels a little misplaced.

Maybe. It is a simple addition after all. But I asked it (and would do so again) because this did strike me as a particular one that seems quite niche and possibly outside of what most p4a users would do. Maybe I am wrong, but that was just my impression so hence the question. It wasn't my intention to block this, but just to bring in something worth considering if anyone else possibly felt the same way

rnixx:
> I feel like we spend a lot of time on removing options, like pygame bootstrap, at some point soon we might deprecate python 2, and maybe the launcher will at some point become an external thing

Pygame bootstrap is legacy, python2 is EOL, both have full featured replacements. These can safely be deprecated/removed without regression. Removing options would be something different.

> that seems quite niche and possibly outside of what most p4a users would do

Your're right. But i don't get the mood about keeping a simple change out just because it might be used only by a minority of users.

zworkb:
>  ... so please don't be angry ...

no worries, nobody is angry !

I think p4a is more than a compatibility layer for Kivy. It brings the strength of Python which is not just the language per se, but also parts of it's ecosystem through recipes.

As I have already stated above there are situations where I need a native UI and can now have the best of two worlds.

I have put together a simple [deonstrator](https://github.com/zworkb/kivy-native-ui-demo) that shows how to integrate a kotlin based activity into a kivy/p4a app and let it communicate via websockets with the python service.

JonasT:
>  But i don't get the mood about keeping a simple change out

For me particular the reason is in part because the gradle part has given me some headaches in particular, so I don't think the cost of making it more complex for the future is zero. So I'm not a fan of increasing the complexity of it "just to have more options", personally. But maybe let's hear the opinions of someone else like @AndreMiras or @inclement ?

AndreMiras:
First of all thanks for spending time investigating and contributing with that PR @zworkb.
Also thank you guys for having a discussion about it :+1: 

I'm usually very cautious/fearful when touching the core specially when dealing with a niche use-case indeed. So yes I'm a bit like @JonasT here, but of course it doesn't mean that I don't want to review/merge.
Also I know @inclement is usually more easy going and relaxed regarding adding new features to the core, plus he knows p4a way more than the most of us.
Thing is now and inside the context of this PR, it might be obvious (or not) this feature doesn't break anything. But it's true that it may introduce some headaches or bugs later while touching let's say `build.tmpl.gradle` and not realising some part are being used in this PR. Specially since this like the other 70% of our code base is not properly cover tested yet.

Thanks again, I'll try to make some time reviewing it this week

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -665,6 +665,30 @@ def parse_args(args=None):
                               'https://developer.android.com/guide/'
                               'topics/manifest/'
                               'activity-element.html'))
+
+    ap.add_argument('--android-entrypoint', dest='android_entrypoint',
+                    default='org.kivy.android.PythonActivity',
+                    help='defines which java class sill be used for startup, usually a subclass of PythonActivity')
```
inclement: Minor: typo "sill" for "will"

inclement: Also the help strings should start with a capital letter for consistency.

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -665,6 +665,30 @@ def parse_args(args=None):
                               'https://developer.android.com/guide/'
                               'topics/manifest/'
                               'activity-element.html'))
+
+    ap.add_argument('--android-entrypoint', dest='android_entrypoint',
```
inclement: This one seems fine, I don't think we could support the same thing any other way with such a simple api.

Minor: I'd probably pull out the 'org.kivy.android.PythonActivity' to a constant declaration somewhere.

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -665,6 +665,30 @@ def parse_args(args=None):
...
+                    default='org.kivy.android.PythonActivity',
+                    help='defines which java class sill be used for startup, usually a subclass of PythonActivity')
+    ap.add_argument('--android-apptheme', dest='android_apptheme',
```
inclement: As with the android entrypoint, this seems okay (albeit even more niche), it's a very simple option and there's no simple alternative.

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -665,6 +665,30 @@ def parse_args(args=None):
...
+                    default='@android:style/Theme.NoTitleBar',
+                    help='defines which app theme should be selected for the main activity')
+    ap.add_argument('--add-java-source-compat', dest='java_source_compat',
```
inclement: I don't like source compat and target compat options so much, see main comment.

pythonforandroid/bootstraps/sdl2/build/templates/AndroidManifest.tmpl.xml
```diff
@@ -54,7 +54,7 @@
     <application android:label="@string/app_name"
                  android:icon="@drawable/icon"
                  android:allowBackup="{{ args.allow_backup }}"
-                 android:theme="@android:style/Theme.NoTitleBar{% if not args.window %}.Fullscreen{% endif %}"
+                 android:theme="{{args.android_apptheme}}{% if not args.window %}.Fullscreen{% endif %}"
```
inclement: Mildly awkward here that only the sdl2 AndroidManifest is changed - I guess this should be implemented in every bootstrap?

pythonforandroid/bootstraps/common/build/templates/build.tmpl.gradle
```diff
@@ -13,68 +13,85 @@ allprojects {
     repositories {
```
inclement: The changes here are much more minimal than I thought at first glance, due to the indentation changes.

inclement:
And further to that, thanks very much @zworkb for raising these thoughtful additions, I'm glad p4a was flexible enough to let you get this working.

zworkb:
thanks @inclement for your comment and suggestions. 
I like the idea with the more common option '--compile-options' and will change that as well the mentioned typos and minor changes, probably this weekend.

pythonforandroid/bootstraps/common/build/build.py
zworkb: done

pythonforandroid/bootstraps/common/build/build.py
zworkb: as suggested I introduced --add-compile-option

pythonforandroid/bootstraps/sdl2/build/templates/AndroidManifest.tmpl.xml
zworkb: i did that now also for the other bootstraps

zworkb:
> And further to that, thanks very much @zworkb for raising these thoughtful additions, I'm glad p4a was flexible enough to let you get this working.

@inclement I have now implemented all your suggestions, so the PR should be ready for merging.
When you merge it, please also merge kivy/buildozer#919 which implements for corresponding options for buildozer. 

Again thanks for your support!

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -663,6 +663,24 @@ def parse_args(args=None):
...
                               'topics/manifest/'
                               'activity-element.html'))
+
+    ap.add_argument('--android-entrypoint', dest='android_entrypoint',
+                    default='org.kivy.android.PythonActivity',
+                    help='Defines which java class will be used for startup, usually a subclass of PythonActivity')
+    ap.add_argument('--android-apptheme', dest='android_apptheme',
+                    default='@android:style/Theme.NoTitleBar',
+                    help='Defines which app theme should be selected for the main activity')
+    ap.add_argument('--add-compile-option', dest='compile_options', default=[],
```
rnixx: --add-compile-options <- typo?

zworkb: no, its no typo, its a parameter thats adds a single option to a list of options (action='append')

PR #1871, opened 2019-06-17
Fix wrong env variable for `hostpython build path` in `archs.py`

opacam:
The hostpython path that we have doesn't point to the right location (probably was the path for the old python2/python2legacy recipe). As per the current python versions that we maintain, the right path is the one of this pr (inside `native-build` folder)


inclement:
Thanks @opacam

PR #1872, opened 2019-06-18
Add unittest for module `pythonforandroid.bootstrap`

opacam:
In this `unittest` we will also cover all the bootstraps we have at this time, so this would increase the coverage a little more than the other tests I recently introduced.

Also should be mentioned that, in order that to not increase the time of our tox tests, I mocked a lot of functions, so this way the test is performed with less than a second and it covers above a 90% of the lines of  the tested files (which I think that should be the final goal for our unittests...above an **90%**)

**As a side note:** I introduced the bootstraps tests in here because our bootstraps (sdl2, webview...)  inherits from `Bootstrap`

### DIFF
--- a/tests/test_bootstrap.py
+++ b/tests/test_bootstrap.py
@@ -1,0 +1,545 @@
+import os
+import sh
+
+import unittest
+
+try:
+    from unittest import mock
+except ImportError:
+    # `Python 2` or lower than `Python 3.3` does not
+    # have the `unittest.mock` module built-in
+    import mock
+from pythonforandroid.bootstrap import Bootstrap
+from pythonforandroid.distribution import Distribution
+from pythonforandroid.recipe import Recipe
+from pythonforandroid.archs import ArchARMv7_a
+from pythonforandroid.build import Context
+
+
+class BaseClassSetupBootstrap(object):
+    """
+    An class object which is intended to be used as a base class to configure
+    an inherited class of `unittest.TestCase`. This class will override the
+    `setUp` and `tearDown` methods.
+    """
+
+    def setUp(self):
+        self.ctx = Context()
+        self.ctx.ndk_api = 21
+        self.ctx.android_api = 27
+        self.ctx._sdk_dir = "/opt/android/android-sdk"
+        self.ctx._ndk_dir = "/opt/android/android-ndk"
+        self.ctx.setup_dirs(os.getcwd())
+        self.ctx.recipe_build_order = [
+            "hostpython3",
+            "python3",
+            "sdl2",
+            "kivy",
+        ]
+
+    def setUp_distribution_with_bootstrap(self, bs):
+        """
+        Extend the setUp by configuring a distribution, because some test
+        needs a distribution to be set to be properly tested
+        """
+        self.ctx.bootstrap = bs
+        self.ctx.bootstrap.distribution = Distribution.get_distribution(
+            self.ctx, name="test_prj", recipes=["python3", "kivy"]
+        )
+
+    def tearDown(self):
+        """
+        Extend the `tearDown` by configuring a distribution, because some test
+        needs a distribution to be set to be properly tested
+        """
+        self.ctx.bootstrap = None
+
+
+class TestBootstrapBasic(BaseClassSetupBootstrap, unittest.TestCase):
+    """
+    An inherited class of `BaseClassSetupBootstrap` and `unittest.TestCase`
+    which will be used to perform tests for the methods/attributes shared
+    between all bootstraps which inherits from class
+    :class:`~pythonforandroid.bootstrap.Bootstrap`.
+    """
+
+    def test_attributes(self):
+        """A test which will initialize a bootstrap and will check if the
+        values are the expected.
+        """
+        bs = Bootstrap().get_bootstrap("sdl2", self.ctx)
+        self.assertEqual(bs.name, "sdl2")
+        self.assertEqual(bs.jni_dir, "sdl2/jni")
+        self.assertEqual(bs.get_build_dir_name(), "sdl2-python3")
+
+        # test dist_dir error
+        bs.distribution = None
+        with self.assertRaises(SystemExit) as e:
+            bs.dist_dir
+        self.assertEqual(e.exception.args[0], 1)
+
+        # test dist_dir success
+        self.setUp_distribution_with_bootstrap(bs)
+        self.assertTrue(bs.dist_dir.endswith("dists/test_prj"))
+
+    def test_build_dist_dirs(self):
+        """A test which will initialize a bootstrap and will check if the
+        directories we set has the values that we expect. Here we test methods:
+
+            - :meth:`~pythonforandroid.bootstrap.Bootstrap.get_build_dir`
+            - :meth:`~pythonforandroid.bootstrap.Bootstrap.get_dist_dir`
+            - :meth:`~pythonforandroid.bootstrap.Bootstrap.get_common_dir`
+        """
+        bs = Bootstrap().get_bootstrap("sdl2", self.ctx)
+
+        self.assertTrue(
+            bs.get_build_dir().endswith("build/bootstrap_builds/sdl2-python3")
+        )
+        self.assertTrue(bs.get_dist_dir("test_prj").endswith("dists/test_prj"))
+        self.assertTrue(
+            bs.get_common_dir().endswith("pythonforandroid/bootstraps/common")
+        )
+
+    def test_list_bootstraps(self):
+        """A test which will initialize a bootstrap and will check if the
+        method :meth:`~pythonforandroid.bootstrap.Bootstrap.list_bootstraps`
+        returns the expected values, which should be: `empty", `service_only`,
+        `webview` and `sdl2`
+        """
+        expected_bootstraps = {"empty", "service_only", "webview", "sdl2"}
+        set_of_bootstraps = set(Bootstrap().list_bootstraps())
+        self.assertEqual(
+            expected_bootstraps, expected_bootstraps & set_of_bootstraps
+        )
+        self.assertEqual(len(expected_bootstraps), len(set_of_bootstraps))
+
+    def test_get_bootstraps_from_recipes(self):
+        """A test which will initialize a bootstrap and will check if the
+        method :meth:`~pythonforandroid.bootstrap.Bootstrap.
+        get_bootstraps_from_recipes` returns the expected values
+        """
+        recipes_sdl2 = {"sdl2", "python3", "kivy"}
+        bs = Bootstrap().get_bootstrap_from_recipes(recipes_sdl2, self.ctx)
+
+        self.assertEqual(bs.name, "sdl2")
+
+        # test wrong recipes
+        wrong_recipes = {"python2", "python3", "pyjnius"}
+        bs = Bootstrap().get_bootstrap_from_recipes(wrong_recipes, self.ctx)
+        self.assertIsNone(bs)
+
+    @mock.patch("pythonforandroid.bootstrap.ensure_dir")
+    def test_prepare_dist_dir(self, mock_ensure_dir):
+        """A test which will initialize a bootstrap and will check if the
+        method :meth:`~pythonforandroid.bootstrap.Bootstrap.prepare_dist_dir`
+        successfully calls once the method `endure_dir`
+        """
+        bs = Bootstrap().get_bootstrap("sdl2", self.ctx)
+
+        bs.prepare_dist_dir("fake_name")
+        mock_ensure_dir.assert_called_once_with(bs.dist_dir)
+
+    @mock.patch("pythonforandroid.bootstrap.open", create=True)
+    @mock.patch("pythonforandroid.util.chdir")
+    @mock.patch("pythonforandroid.bootstrap.shutil.copy")
+    @mock.patch("pythonforandroid.bootstrap.os.makedirs")
+    def test_bootstrap_prepare_build_dir(
+        self, mock_os_makedirs, mock_shutil_copy, mock_chdir, mock_open
+    ):
+        """A test which will initialize a bootstrap and will check if the
+        method :meth:`~pythonforandroid.bootstrap.Bootstrap.prepare_build_dir`
+        successfully calls the methods that we need to prepare a build dir.
+        """
+
+        # prepare bootstrap
+        bs = Bootstrap().get_bootstrap("service_only", self.ctx)
+        self.ctx.bootstrap = bs
+
+        # test that prepare_build_dir runs (notice that we mock
+        # any file/dir creation so we can speed up the tests)
+        bs.prepare_build_dir()
+
+        # make sure that the open command has been called only once
+        mock_open.assert_called_once_with("project.properties", "w")
+
+        # check that the other mocks we made are actually called
+        mock_os_makedirs.assert_called()
+        mock_shutil_copy.assert_called()
+        mock_chdir.assert_called()
+
+    @mock.patch("pythonforandroid.bootstrap.os.path.isfile")
+    @mock.patch("pythonforandroid.bootstrap.os.path.exists")
+    @mock.patch("pythonforandroid.bootstrap.os.unlink")
+    @mock.patch("pythonforandroid.bootstrap.open", create=True)
+    @mock.patch("pythonforandroid.util.chdir")
+    @mock.patch("pythonforandroid.bootstrap.sh.ln")
+    @mock.patch("pythonforandroid.bootstrap.listdir")
+    @mock.patch("pythonforandroid.bootstrap.sh.mkdir")
+    @mock.patch("pythonforandroid.bootstrap.sh.rm")
+    def test_bootstrap_prepare_build_dir_with_java_src(
+        self,
+        mock_sh_rm,
+        mock_sh_mkdir,
+        mock_listdir,
+        mock_sh_ln,
+        mock_chdir,
+        mock_open,
+        mock_os_unlink,
+        mock_os_path_exists,
+        mock_os_path_isfile,
+    ):
+        """A test which will initialize a bootstrap and will check perform
+        another test for method
+        :meth:`~pythonforandroid.bootstrap.Bootstrap.prepare_build_dir`. In
+        here we will simulate that we have `with_java_src` set to some value.
+        """
+        self.ctx.symlink_java_src = ["some_java_src"]
+        mock_listdir.return_value = [
+            "jnius",
+            "kivy",
+            "Kivy-1.11.0.dev0-py3.7.egg-info",
+            "pyjnius-1.2.1.dev0-py3.7.egg",
+        ]
+
+        # prepare bootstrap
+        bs = Bootstrap().get_bootstrap("sdl2", self.ctx)
+        self.ctx.bootstrap = bs
+
+        # test that prepare_build_dir runs (notice that we mock
+        # any file/dir creation so we can speed up the tests)
+        bs.prepare_build_dir()
+        # make sure that the open command has been called only once
+        mock_open.assert_called_with("project.properties", "w")
+
+        # check that the symlink was made 4 times and that
+        self.assertEqual(
+            len(mock_sh_ln.call_args_list), len(mock_listdir.return_value)
+        )
+        for i, directory in enumerate(mock_listdir.return_value):
+            self.assertTrue(
+                mock_sh_ln.call_args_list[i][0][1].endswith(directory)
+            )
+
+        # check that the other mocks we made are actually called
+        mock_sh_rm.assert_called()
+        mock_sh_mkdir.assert_called()
+        mock_chdir.assert_called()
+        mock_os_unlink.assert_called()
+        mock_os_path_exists.assert_called()
+        mock_os_path_isfile.assert_called()
+
+
+class GenericBootstrapTest(BaseClassSetupBootstrap):
+    """
+    An inherited class of `BaseClassSetupBootstrap` which will extends his
+    functionality by adding some generic bootstrap tests, so this way we can
+    test all our sub modules of :mod:`~pythonforandroid.bootstraps` from within
+    this module.
+
+    .. warning:: This is supposed to be used as a base class, so please, don't
+                 use this directly.
+    """
+
+    @property
+    def bootstrap_name(self):
+        """Subclasses must have property 'bootstrap_name'. It should be the
+        name of the bootstrap to test"""
+        raise NotImplementedError("Not implemented in GenericBootstrapTest")
+
+    @mock.patch("pythonforandroid.bootstraps.service_only.open", create=True)
+    @mock.patch("pythonforandroid.bootstraps.webview.open", create=True)
+    @mock.patch("pythonforandroid.bootstraps.sdl2.open", create=True)
+    @mock.patch("pythonforandroid.distribution.open", create=True)
+    @mock.patch(
+        "pythonforandroid.python.GuestPythonRecipe.create_python_bundle"
+    )
+    @mock.patch("pythonforandroid.bootstrap.Bootstrap.strip_libraries")
+    @mock.patch("pythonforandroid.util.exists")
+    @mock.patch("pythonforandroid.util.chdir")
+    @mock.patch("pythonforandroid.bootstrap.listdir")
+    @mock.patch("pythonforandroid.bootstrap.sh.rm")
+    @mock.patch("pythonforandroid.bootstrap.sh.cp")
+    def test_run_distribute(
+        self,
+        mock_sh_cp,
+        mock_sh_rm,
+        mock_listdir,
+        mock_chdir,
+        mock_ensure_dir,
+        mock_strip_libraries,
+        mock_create_python_bundle,
+        mock_open_dist_files,
+        mock_open_sdl2_files,
+        mock_open_webview_files,
+        mock_open_service_only_files,
+    ):
+        """
+        A test for any overwritten method of
+        `~pythonforandroid.bootstrap.Bootstrap.run_distribute`. Here we mock
+        any file/dir operation that it could slow down our tests, and there is
+        a lot to mock, because the `run_distribute` method it should take care
+        of prepare all compiled files to generate the final `apk`. The targets
+        of this test will be:
+
+            - :meth:`~pythonforandroid.bootstraps.sdl2.BootstrapSdl2
+              .run_distribute`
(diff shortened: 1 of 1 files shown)


AndreMiras:
Awesome, I'll review it pretty soon.
I saw there's codacy being introduced, is it not a bit early? We're not even pep8 compliant

tests/test_bootstrap.py
```diff
@@ -0,0 +1,551 @@
...
+    def test_get_bootstraps_from_recipes(self):
+        """A test which will initialize a bootstrap and will check if the
+        method :meth:`~pythonforandroid.bootstrap.Bootstrap.
+        get_bootstraps_from_recipes` returns the expected values
+        """
+        recipes_service = {"pyjnius", "python3"}
+        bs = Bootstrap().get_bootstrap_from_recipes(recipes_service, self.ctx)
+        self.setUp_distribution_with_bootstrap(bs)
+
+        self.assertEqual(bs.name, "service_only")
```
AndreMiras: This guy seems to be returning `webview` on my setup for some reason :thinking: 
```
[andre:~/workspace/python-for-android] develop+* 1 ± venv/bin/pytest tests/test_bootstrap.py 
================================================================================================================================== test session starts ==================================================================================================================================
platform linux -- Python 3.6.7, pytest-4.2.0, py-1.7.0, pluggy-0.8.1
rootdir: /home/andre/workspace/python-for-android, inifile:
collected 23 items                                                                                                                                                                                                                                                                      

tests/test_bootstrap.py ...F...................                                                                                                                                                                                                                                   [100%]

======================================================================================================================================= FAILURES ========================================================================================================================================
__________________________________________________________________________________________________________________ TestBootstrapBasic.test_get_bootstraps_from_recipes __________________________________________________________________________________________________________________

self = <test_bootstrap.TestBootstrapBasic testMethod=test_get_bootstraps_from_recipes>

    def test_get_bootstraps_from_recipes(self):
        """A test which will initialize a bootstrap and will check if the
        method :meth:`~pythonforandroid.bootstrap.Bootstrap.
        get_bootstraps_from_recipes` returns the expected values
        """
        recipes_service = {"pyjnius", "python3"}
        bs = Bootstrap().get_bootstrap_from_recipes(recipes_service, self.ctx)
        self.setUp_distribution_with_bootstrap(bs)
    
>       self.assertEqual(bs.name, "service_only")
E       AssertionError: 'webview' != 'service_only'
E       - webview
E       + service_only

tests/test_bootstrap.py:134: AssertionError
--------------------------------------------------------------------------------------------------------------------------------- Captured stderr call ----------------------------------------------------------------------------------------------------------------------------------
[INFO]:    Trying to find a bootstrap that matches the given recipes.
[INFO]:    Found 3 acceptable bootstraps: ['webview', 'sdl2', 'service_only']
[INFO]:    No existing dists meet the given requirements!
----------------------------------------------------------------------------------------------------------------------------------- Captured log call -----------------------------------------------------------------------------------------------------------------------------------
bootstrap.py               156 INFO     [INFO]:    Trying to find a bootstrap that matches the given recipes.
bootstrap.py               186 INFO     [INFO]:    Found 3 acceptable bootstraps: ['webview', 'sdl2', 'service_only']
bootstrap.py               189 INFO     [INFO]:    Using the first of these: webview
distribution.py            111 INFO     [INFO]:    No existing dists meet the given requirements!
=================================================================================================================================== warnings summary ====================================================================================================================================
pythonforandroid/util.py:23
  /home/andre/workspace/python-for-android/pythonforandroid/util.py:23: DeprecationWarning: WgetDownloader style of invoking requests is deprecated. Use newer urlopen functions/methods
    urlretrieve = WgetDownloader().retrieve

-- Docs: https://docs.pytest.org/en/latest/warnings.html
==================================================================================================================== 1 failed, 22 passed, 1 warnings in 0.31 seconds ====================================================================================================================
[andre:~/workspace/python-for-android] develop+* 1 ±
```

tests/test_bootstrap.py
```diff
@@ -0,0 +1,551 @@
...
+        values are the expected.
+        """
+        bs = Bootstrap().get_bootstrap("sdl2", self.ctx)
+        self.assertEqual(bs.name, "sdl2")
+        self.assertEqual(bs.jni_dir, "sdl2/jni")
+        self.assertEqual(bs.get_build_dir_name(), "sdl2-python3")
+
+        # test dist_dir error
+        with self.assertRaises(SystemExit) as e:
+            getattr(bs, "dist_dir")
```
AndreMiras: Minor: just wondering why don't we simply `bs.dist_dir`?

tests/test_bootstrap.py
```diff
@@ -0,0 +1,551 @@
...
+        """
+        Extend the `tearDown` by configuring a distribution, because some test
+        needs a distribution to be set to be properly tested
+
+        .. note:: Here we need to reset `self.ctx.bootstrap.distribution`
+                  because, otherwise we will be unable to successfully pass the
+                  test :meth:`TestBootstrapBasic.test_attributes`
+        """
+        if self.ctx.bootstrap:
+            self.ctx.bootstrap.distribution = None
```
AndreMiras: Just wondering:
If this is just needed for `TestBootstrapBasic.test_attributes` then why don't we explicitly make that change in this test?
Also I can't grasp how this work, if just doing `self.ctx.bootstrap = None` is not enough so it means we have some leftover `distribution` references somewhere, right? Do you know why/where?

tests/test_bootstrap.py
```diff
@@ -0,0 +1,551 @@
...
+            getattr(bs, "dist_dir")
+        self.assertEqual(e.exception.args[0], 1)
+
+        # test dist_dir success
+        self.setUp_distribution_with_bootstrap(bs)
+        self.assertTrue(bs.dist_dir.endswith("dists/test_prj"))
+
+    def test_build_dist_dirs(self):
+        """A test which will initialize a bootstrap and will check if the
+        directories we set has the values that we expect. Her we test methods:
```
AndreMiras: Minor typo: `s/Her /Here /`

tests/test_bootstrap.py
```diff
@@ -0,0 +1,551 @@
...
+
+            - :meth:`~pythonforandroid.bootstrap.Bootstrap.get_build_dir`
+            - :meth:`~pythonforandroid.bootstrap.Bootstrap.get_dist_dir`
+            - :meth:`~pythonforandroid.bootstrap.Bootstrap.get_common_dir`
+        """
+        bs = Bootstrap().get_bootstrap("sdl2", self.ctx)
+        self.setUp_distribution_with_bootstrap(bs)
```
AndreMiras: Minor: may not be needed

tests/test_bootstrap.py
```diff
@@ -0,0 +1,551 @@
...
+
+    def test_list_bootstraps(self):
+        """A test which will initialize a bootstrap and will check if the
+        method :meth:`~pythonforandroid.bootstrap.Bootstrap.list_bootstraps`
+        returns the expected values, which should be: `empty", `service_only`,
+        `webview` and `sdl2`
+        """
+        # prepare bootstrap and distribution
+        bs = Bootstrap().get_bootstrap("sdl2", self.ctx)
+        self.setUp_distribution_with_bootstrap(bs)
```
AndreMiras: Minor: may not be needed

tests/test_bootstrap.py
```diff
@@ -0,0 +1,551 @@
...
+
+        expected_bootstraps = {"empty", "service_only", "webview", "sdl2"}
+        set_of_bootstraps = set(Bootstrap().list_bootstraps())
+        self.assertEqual(
+            expected_bootstraps, expected_bootstraps & set_of_bootstraps
+        )
```
AndreMiras: Why are we intersecting expected_bootstraps & set_of_bootstraps?
Is it because it's OK to have `set_of_bootstraps = expected_bootstraps + {'whatever_other_bootstrap'}` while not OK to have less than `expected_bootstraps` in `set_of_bootstraps`? But then the the assertEqual in the next line defeats that purpose no?

tests/test_bootstrap.py
```diff
@@ -0,0 +1,551 @@
...
+        mock_open.side_effect = [
+            mock.mock_open(read_data="target=android-21").return_value
+        ]
+
+        # prepare bootstrap
+        bs = Bootstrap().get_bootstrap("service_only", self.ctx)
+        self.ctx.bootstrap = bs
+
+        # test that prepare_build_dir runs (notice that we mock
+        # any file/dir creation so we can speed up the tests)
```
AndreMiras: :+1:  not only we (may) speed things over, but more importantly (IMO) we keep things clean as we don't mess up with the file system

tests/test_bootstrap.py
```diff
@@ -0,0 +1,551 @@
...
+        mock_chdir,
+        mock_open,
+    ):
+        """A test which will initialize a bootstrap and will check if the
+        method :meth:`~pythonforandroid.bootstrap.Bootstrap.prepare_build_dir`
+        successfully calls the methods that we need to prepare a build dir.
+        """
+        mock_isdir.return_value = True
+        mock_sh_mkdir.return_value = "sh.mkdir successfully mocked"
+        mock_sh_rm.return_value = "sh.rm successfully mocked"
```
AndreMiras: Minor: just curious why would we need to set the return values for some of these like `mock_sh_rm` and `mock_sh_mkdir`?

tests/test_bootstrap.py
```diff
@@ -0,0 +1,551 @@
...
+        mock_isdir,
+        mock_sh_rm,
+        mock_sh_mkdir,
+        mock_listdir,
+        mock_sh_ln,
+        mock_os_makedirs,
+        mock_shutil_copy,
+        mock_chdir,
+        mock_open,
+    ):
```
AndreMiras: Actually most of the mocks are not being hit here because `symlink_java_src` is `False`, right?

tests/test_bootstrap.py
```diff
@@ -0,0 +1,551 @@
...
+        # prepare bootstrap
+        bs = Bootstrap().get_bootstrap("sdl2", self.ctx)
+        self.ctx.bootstrap = bs
+
+        # test that prepare_build_dir runs (notice that we mock
+        # any file/dir creation so we can speed up the tests)
+        bs.prepare_build_dir()
+        # make sure that the open command has been called only once
+        mock_open.assert_called_with("project.properties", "w")
+        mock_open.reset_mock()
```
AndreMiras: Minor: Maybe we can also quick check that the symlink was made 4 times since we listdir returned 4 dirs?
E.g. something like:
```python
self.assertEqual(len(mock_sh_ln.call_args_list), len(mock_listdir.return_value)
```
Or if we want to go a bit further, we could check the symlink source, e.g.
```python
for i, directory in enumerate(mock_listdir.return_value):
    self.assertTrue(mock_sh_ln.call_args_list[i][0][1].endswith(directory))
```

tests/test_bootstrap.py
```diff
@@ -0,0 +1,551 @@
...
+            "Kivy-1.11.0.dev0-py3.7.egg-info",
+            "pyjnius-1.2.1.dev0-py3.7.egg",
+        ]
+        mock_sh_ln.return_value = True
+        mock_os_makedirs.return_value = True
+        mock_shutil_copy.return_value = True
+        mock_chdir.return_value = True
+        mock_open.side_effect = [
+            mock.mock_open(read_data="target=android-21").return_value
+        ]
```
AndreMiras: This is specifically confusing me :thinking:

tests/test_bootstrap.py
```diff
@@ -0,0 +1,551 @@
...
+        .. note:: Here we need to reset `self.ctx.bootstrap.distribution`
+                  because, otherwise we will be unable to successfully pass the
+                  test :meth:`TestBootstrapBasic.test_attributes`
+        """
+        if self.ctx.bootstrap:
+            self.ctx.bootstrap.distribution = None
+        self.ctx.bootstrap = None
+
+
+class TestBootstrapBasic(BaseClassSetupBootstrap, unittest.TestCase):
```
AndreMiras: I'm already not a big fan of multiple inheritance, even for mixins. And that's my personal opinion.
But here anyway `BaseClassSetupBootstrap` was already inheriting `unittest.TestCase` what do you think?

AndreMiras:
This is some impressive testing on here, well done @opacam :+1: 
I was about to review `GenericBootstrapTest`, but I'd like your feedback on some of my comments first. Just to see if we're on the same page or if I'm completely off :smile: 
Thanks again for your efforts, the result is great!

tests/test_bootstrap.py
opacam: I changed the recipes of this tests in such a way that we could only get only one result so we can avoid this issue (I suspect that we got different results because we use `sets` in some calculations and probably we convert it to a list at some point...anyway...the result that you got I consider it valid...so I think it's better to avoid these ambiguities).
**Thanks!!**

tests/test_bootstrap.py
opacam: aaa, this is because I have a linting running in background inside my `ide` and it marks your solution as an operation that does nothing...it seems that is happy with `getattr`...the final result is the same...but...since it seems that you have seen something strange in here, I have changed it as you proposed :wink:

tests/test_bootstrap.py
opacam: I addressed this issue as you proposed but changing it a little those lines of code. I reset it in the `test_attributes`, but now we do it directly from the created bootstrap not from `ctx.bootstrap`, and it seems that it's working fine :smile:

tests/test_bootstrap.py
opacam: Yea, the main idea is that, in case that we add a new bootstrap and we don't update this test, the alarm should be triggered. Notice that the first test will only contemplate a bootstrap removal or rename, but not an addition...so the second test fills this gap.

tests/test_bootstrap.py
opacam: You are right, We don't need to do this, Thanks @AndreMiras, I forgot to remove it :smile:

tests/test_bootstrap.py
opacam: Yes, but I think that We actually need this mocking because we need to copy all the files of the bootstrap...so `symlink_java_src` it's an extra...unless I miss something...

tests/test_bootstrap.py
opacam: Thanks @AndreMiras :smile:
I introduced your suggestion, if you think that we should add more tests let me know (actually I'm pretty sure that we can test a lot more, but for starters it's fine...we got almost all code executed and we improve our coverage...)

tests/test_bootstrap.py
opacam: Aaa...yea...we need to do this because we test the `open` inside a Context Manager and the default Mock object lacks of  the attributes `__enter__` and `__exit__` wich will lead into an error, adding into the final `.return_value` solves the problem...
maybe do you have a better approach?

tests/test_bootstrap.py
opacam: Why you don't like `multiple-inheritance`?
In some situations it makes a lot of sense to use it...it's because of the `MRO` which could be tricky to follow?

tests/test_bootstrap.py
AndreMiras: oh OK I was also wondering if it was maybe for the linting, thanks for clarifying

tests/test_bootstrap.py
AndreMiras: Yes I definitely think what you had was more than good for a start :+1:

tests/test_bootstrap.py
AndreMiras: Yes mainly because it can get things harder to follow sometimes. I've learn programming the OOP way and I think I tend to abuse inheritance in general.
Anyway I think your code is pretty straight forward and readable

tests/test_bootstrap.py
AndreMiras: Thanks for the update, then it means you probably discovered a bug in this component :clap: 
Basically https://github.com/kivy/python-for-android/blob/v2019.06.06/pythonforandroid/bootstrap.py#L190
So at some point we assume something is ordered but it's not. I think the problem may actually come from us assuming the [listdir()](https://github.com/kivy/python-for-android/blob/v2019.06.06/pythonforandroid/bootstrap.py#L145) of [list_bootstraps()](https://github.com/kivy/python-for-android/blob/v2019.06.06/pythonforandroid/bootstrap.py#L141) is in a known/consistent order while it's not.
I've created https://github.com/kivy/python-for-android/issues/1875 to investigate, let me know what you think

tests/test_bootstrap.py
AndreMiras: What I mean is for instance `mock_sh_rm` is never called because it's only being called if `symlink_java_src` is `True`.
You can verify this with `mock_sh_rm.called` after calling `bs.prepare_build_dir()`.
Same thing with most of the mocks in this test, for instance `mock_listdir`, `mock_sh_mkdir`, `mock_isdir` and `mock_sh_ln`. So since it's not being called, we could just not mock it and the code would look easier. That's only when `symlink_java_src` is `False` of course.

tests/test_bootstrap.py
AndreMiras: Which error? Because it's a magic mock to me and so is your return_value here.
Personally I would drop all the `mock_open.side_effect....` part. The tests should still pass without it and I can't see why it's needed :thinking:

tests/test_bootstrap.py
```diff
@@ -0,0 +1,523 @@
...
+    @mock.patch("pythonforandroid.bootstrap.shutil.copy")
+    @mock.patch("pythonforandroid.bootstrap.sh.ln")
+    @mock.patch("pythonforandroid.bootstrap.listdir")
+    @mock.patch("pythonforandroid.bootstrap.sh.Command")
+    @mock.patch("pythonforandroid.bootstrap.sh.zip")
+    @mock.patch("pythonforandroid.bootstrap.sh.mkdir")
+    @mock.patch("pythonforandroid.bootstrap.sh.rm")
+    @mock.patch("pythonforandroid.bootstrap.sh.cp")
+    @mock.patch("pythonforandroid.bootstrap.isdir")
+    def test_run_distribute(
```
AndreMiras: I think the following mocks are not necessary as no code path would reach them in the given tests.

- mock_isdir
- mock_sh_mkdir
- mock_sh_zip
- mock_sh_command
- mock_sh_ln
- mock_shutil_copy
- mock_shutil_copy2
- mock_os_unlink
- mock_os_path_isfile
- mock_get_python_install_dir

Removing them would make the code more readable.
Maybe as a rule of thumb we should try to prove `assert mock.called is True` or even `assert mock.mock_calls != []` to show the mock is actually being called? What do you think?

tests/test_bootstrap.py
opacam: You are right, We could have a problem with this because we don't have any control of what bootstrap will be returned in case that there are multiple possibilities...also I think that this bootstraps list should not been sorted alphabetically instead should be done based on requested recipes...but better to discuss it in the issue you created

**Note:** I cannot take the credits  of discover this bug, actually we both found it...you pointed exactly what is the problem and your local testing triggered the the alarm  :clap:

tests/test_bootstrap.py
opacam: @AndreMiras,You are right, I was trying to reinvent the wheel, we don't need that line **thanks!!** :smile:

tests/test_bootstrap.py
opacam: I think that you are right, we should check that all the mocks are being called, so I revised all this pull request and I think that now all the mocks are being tested in some way or other so we should be free of unnecessary mocks...so please, let me know if you find one so I can fix it before merge this

tests/test_bootstrap.py
```diff
@@ -0,0 +1,556 @@
...
+        """
+        self.ctx.symlink_java_src = ["some_java_src"]
+        mock_isdir.return_value = False
+        mock_listdir.return_value = [
+            "jnius",
+            "kivy",
+            "Kivy-1.11.0.dev0-py3.7.egg-info",
+            "pyjnius-1.2.1.dev0-py3.7.egg",
+        ]
+        mock_sh_ln.return_value.return_value = mock.Mock()
```
AndreMiras: Why is this for? Basically it would make the `ln()()` call return a `Mock`, but it almost does that by default since it's a `MagicMock` hence `ln()()` returns a `MagicMock` even if you remove that line. So I'm not sure why this is needed :thinking:

tests/test_bootstrap.py
```diff
@@ -0,0 +1,556 @@
...
+        bs = Bootstrap().get_bootstrap("service_only", self.ctx)
+        self.ctx.bootstrap = bs
+
+        # test that prepare_build_dir runs (notice that we mock
+        # any file/dir creation so we can speed up the tests)
+        bs.prepare_build_dir()
+
+        # make sure that the open command has been called only once
+        mock_open.assert_called_once_with("project.properties", "w")
+        mock_open.reset_mock()
```
AndreMiras: Minor: why?

tests/test_bootstrap.py
```diff
@@ -0,0 +1,556 @@
...
+    @mock.patch("pythonforandroid.bootstrap.open", create=True)
+    @mock.patch("pythonforandroid.util.chdir")
+    @mock.patch("pythonforandroid.bootstrap.sh.ln")
+    @mock.patch("pythonforandroid.bootstrap.listdir")
+    @mock.patch("pythonforandroid.bootstrap.sh.mkdir")
+    @mock.patch("pythonforandroid.bootstrap.sh.rm")
+    @mock.patch("pythonforandroid.bootstrap.isdir")
+    def test_bootstrap_prepare_build_dir_with_java_src(
+        self,
+        mock_isdir,
```
AndreMiras: Minor: I don't think this mock is called

tests/test_bootstrap.py
```diff
@@ -0,0 +1,556 @@
...
+        # prepare bootstrap
+        bs = Bootstrap().get_bootstrap("sdl2", self.ctx)
+        self.ctx.bootstrap = bs
+
+        # test that prepare_build_dir runs (notice that we mock
+        # any file/dir creation so we can speed up the tests)
+        bs.prepare_build_dir()
+        # make sure that the open command has been called only once
+        mock_open.assert_called_with("project.properties", "w")
+        mock_open.reset_mock()
```
AndreMiras: Minor: why?

tests/test_bootstrap.py
```diff
@@ -0,0 +1,556 @@
...
+        self.setUp_distribution_with_bootstrap(bs)
+
+        self.ctx.hostpython = "/some/fake/hostpython3"
+        self.ctx.python_recipe = Recipe.get_recipe("python3", self.ctx)
+        self.ctx.python_modules = ["requests"]
+        self.ctx.archs = [ArchARMv7_a(self.ctx)]
+
+        mock_open_dist_files.side_effect = [
+            mock.mock_open(read_data="target=android-21").return_value
+        ]
```
AndreMiras: Sorry this is still confusing me. What are these for again? :grimacing:

tests/test_bootstrap.py
```diff
@@ -0,0 +1,556 @@
...
+        self.assertEqual(cp_java_call[0], sh.cp)
+        self.assertTrue(cp_java_call[2].endswith("classes.jar"))
+        self.assertEqual(cp_java_call[3], "libs/file.jar")
+        cp_libs_call, kw = mock_shprint.call_args_list[2]
+        self.assertEqual(cp_libs_call[0], sh.cp)
+        self.assertEqual(cp_libs_call[2], "/fakedir/file.aar")
+        self.assertEqual(cp_libs_call[3], libs_dir)
+        mock_bs_dir.assert_has_calls([mock.call("libs"), mock.call(libs_dir)])
+        mock_glob.assert_called()
+        reset_mocks()
```
AndreMiras: Mock would stop living once outside the decorated method, hence this last one is probably not necessary

tests/test_bootstrap.py
```diff
@@ -0,0 +1,556 @@
...
+        mock_ensure_dir,
+        mock_sh_command,
+        mock_sh_print,
+    ):
+        mock_find_executable.return_value = "arm-linux-androideabi-gcc"
+        mock_sh_command.return_value = mock.Mock()
+        mock_sh_print.return_value.stdout = (
+            "/fake_dir/libsqlite3.so"
+            "\n/fake_dir/libpng16.so\n".encode("ascii")
+        )
```
AndreMiras: Minor: We don't have to fake this one unless it really serves a purpose in the test.
I mean why would we want to `mock_sh_print.return_value.stdout =` here?

tests/test_bootstrap.py
```diff
@@ -0,0 +1,556 @@
...
+    @mock.patch("pythonforandroid.archs.find_executable")
+    def test_bootstrap_strip(
+        self,
+        mock_find_executable,
+        mock_ensure_dir,
+        mock_sh_command,
+        mock_sh_print,
+    ):
+        mock_find_executable.return_value = "arm-linux-androideabi-gcc"
+        mock_sh_command.return_value = mock.Mock()
```
AndreMiras: Minor: this is (already) a `MagicMock` by default, why changing it to a `Mock`? Let's be lazy unless required not to

tests/test_bootstrap.py
opacam: Yes, this is wrong  :sweat_smile:, I wanted to check that we actually `wrote` the expected values and this is not the way to go, I will change that

**Thanks!!**

tests/test_bootstrap.py
```diff
@@ -0,0 +1,545 @@
...
+        }
+        mock_open_bs = mock_open_bootstraps[self.bootstrap_name]
+        # test that the expected calls has been called
+        for expected_call in expected_open_calls[self.bootstrap_name]:
+            self.assertIn(expected_call, mock_open_bs.call_args_list)
+        # test that the write function has been called with the expected args
+        self.assertIn(
+            mock.call().__enter__().write("sdk.dir=/opt/android/android-sdk"),
+            mock_open_bs.mock_calls,
+        )
```
opacam: @AndreMiras,
now we test that we actually we write the expected values...this (and the lines below) replaces the `read_data` thing that it was wrong :wink:

And **thanks for reviewing this**, your reviews allowed us to improve this pr a lot :heart:, now it looks much cleaner and we do more testing than the first posted version.

AndreMiras: That definitely makes more sense to me thanks! :clap:

PR #1877, opened 2019-06-21
fix ctypes-util-find-library issue for python2

surbhicis:
Adding patch for fixing ctypes-util-find-library issue for python2 support


AndreMiras:
Thank you for the PR. I know it's a relatively simple change, but you know we're slowly decommissioning Python2 right?
Why don't you try to migrate your app to Python3?

surbhicis:
@AndreMiras I am planning to port app for python3 but somehow until it takes we need to have support as kivy/p4a have recipe py2 support so it should also have fixes for ctypes find_library :)

PR #1722, opened 2019-02-20
Initial migration to NDK r19 (Part I - The core)

opacam:
The time has come to move into ndk r19, there is big changes in android's ndk r19 which will make live easier at compile time because (extracted from android docs: https://developer.android.com/ndk/guides/other_build_systems):

`
As of NDK r19, the toolchains installed by default with the NDK may be used in-place. The make_standalone_toolchain.py script is no longer needed for interfacing with arbitrary build systems.
`

So, this implies that we can remove some cflags that we were using (`isystem`, `sysroot`, `gcc-toolchain`...). I think that this is good for python-for-android maintainability....but we will loose the `gcc` compiler because has been deprecated in android's ndk (partially removed in ndk r18 as far as I know), so this will mean the beginning of the end of python2legacy recipe (openssl will not work anymore for python2legacy, pygame will be removed soon, python2legacy needs the gcc compiler...)

**Things done here:**

  - [x] Rework of the arch environment to work with ndk r19+
      - Modifies target so we can use the toolchains `in-place` (as if we were using a standalone toolchain)
      - Removes some cflags that we don't need anymore
      - Move `macros` and `includes` from `CFLAGS` to `CPPFLAGS` (because this is the way it's supposed to be)
      - Move libraries from `LDFLAGS` to `LDLIBS` (again, the same as above)
      - Remove hardcoded cpu count in `MAKE`
      - Remove `gcc` of environment (because it has been partially removed in ndk r18)
      - Group environment variables and add some comments (to put some order and make easier for us to maintain it)
  - [x] Adapt **python2** and **python3** recipes to new clang triplet (**requires an android ndk version >= r19**)
  - [x] ~~Add libz recipe~~ Read android's ndk zlib version and add flags to python's recipe `env`, to solve build problems with python's zlib module (I think that this could also solve the `zip/unzip` issues reported in issue #1388)
  - [x] Update docker files to use ndk r19b
  - [x] Fix numpy for ndk r19
  - [x] Update documentation
  - [x] Update `test_archs` and `test_bootstrap`

**Those points it will be done later in another prs:**
  - ~~Remove pygame bootstrap (already started by @inclement in #1670)~~
  - ~~Remove python2legacy (we introduced python2legacy recipe in order to made it work the pygame's bootstrap...without pygame...there is no point to maintain it..plus it won't get built with ndk r19)~~
[...]

PR #1883, opened 2019-06-22
Make it raise an error if an old ndk is used

opacam:
In order to avoid issues with old android's NDKs, we force to raise an error in case that an NDK version lower than the specified is detected. We also add the ability to extract the android's NDK letter version, which for now, will only be used to inform the user of the version which is using.

~~**Important note:** **Don't merge this before the ndk-r19 core part I** (we can easily adapt this for ndk-r17 but I leave it to `ndk-19` so we make sure that travis fails as expected), once the Part I gets merged, we can rebase this into develop, push it and the tests should pass **or we could adapt it to ndk-r17 and merge it before**~~

**Note:** The idea to create this pr comes from [this comment](https://github.com/kivy/python-for-android/pull/1722#issuecomment-502844382) from the pull request `ndk-r19 initial migration`

**See also:** The reported error in [travis test](https://travis-ci.org/kivy/python-for-android/jobs/549176086#L2332)

### DIFF
--- a/pythonforandroid/recommendations.py
+++ b/pythonforandroid/recommendations.py
@@ -10,28 +10,80 @@ MIN_NDK_VERSION = 17
 MAX_NDK_VERSION = 17
 
 RECOMMENDED_NDK_VERSION = '17c'
-OLD_NDK_MESSAGE = 'Older NDKs may not be compatible with all p4a features.'
 NEW_NDK_MESSAGE = 'Newer NDKs may not be fully supported by p4a.'
+NDK_DOWNLOAD_URL = 'https://developer.android.com/ndk/downloads/'
 
 
 def check_ndk_version(ndk_dir):
-    # Check the NDK version against what is currently recommended
+    """
+    Check the NDK version against what is currently recommended and raise an
+    exception of :class:`~pythonforandroid.util.BuildInterruptingException` in
+    case that the user tries to use an NDK lower than minimum supported,
+    specified via attribute `MIN_NDK_VERSION`.
+
+    .. versionchanged:: 2019.06.06.1.dev0
+        Added the ability to get android's ndk `letter version` and also
+        rewrote to raise an exception in case that an NDK version lower than
+        the minimum supported is detected.
+    """
     version = read_ndk_version(ndk_dir)
 
     if version is None:
-        return  # if we failed to read the version, just don't worry about it
+        warning(
+            'Unable to read the ndk version, assuming that you are using an'
+            ' NDK greater than {min_supported} (the minimum ndk required to'
+            ' use p4a successfully).\n'
+            'Note: If you got build errors, consider to download the'
+            ' recommended ndk version which is {rec_version} and try'
+            ' it again (after removing all the files generated with this'
+            ' build). To download the android NDK visit the following page: '
+            '{ndk_url}'.format(
+                min_supported=MIN_NDK_VERSION,
+                rec_version=RECOMMENDED_NDK_VERSION,
+                ndk_url=NDK_DOWNLOAD_URL,
+            )
+        )
+        return
+
+    # create a dictionary which will describe the relationship of the android's
+    # ndk minor version with the `human readable` letter version, egs:
+    # Pkg.Revision = 17.1.4828580 => ndk-17b
+    # Pkg.Revision = 17.2.4988734 => ndk-17c
+    # Pkg.Revision = 19.0.5232133 => ndk-19 (No letter)
+    minor_to_letter = {0: ''}
+    minor_to_letter.update(
+        {n + 1: chr(i) for n, i in enumerate(range(ord('b'), ord('b') + 25))}
+    )
 
     major_version = version.version[0]
+    letter_version = minor_to_letter[version.version[1]]
+    string_version = '{major_version}{letter_version}'.format(
+        major_version=major_version, letter_version=letter_version
+    )
 
-    info('Found NDK revision {}'.format(version))
+    info('Found NDK version {}'.format(string_version))
 
     if major_version < MIN_NDK_VERSION:
-        warning('Minimum recommended NDK version is {}'.format(
-            RECOMMENDED_NDK_VERSION))
-        warning(OLD_NDK_MESSAGE)
+        raise BuildInterruptingException(
+            'Unsupported NDK version detected {user_version}\n'
+            '* Note: Minimum supported NDK version is {min_supported}'.format(
+                user_version=string_version, min_supported=MIN_NDK_VERSION
+            ),
+            instructions=(
+                'Please, go to the android ndk page ({ndk_url}) and download a'
+                ' supported version.\n*** The currently recommended NDK'
+                ' version is {rec_version} ***'.format(
+                    ndk_url=NDK_DOWNLOAD_URL,
+                    rec_version=RECOMMENDED_NDK_VERSION,
+                )
+            ),
+        )
     elif major_version > MAX_NDK_VERSION:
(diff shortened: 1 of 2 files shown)


AndreMiras:
Nice PR!
Yes I would be tempted to adapt it for `ndk-r17` so we can make an early merge with easy adaptations later.
Also we may want to unit test this feature

opacam:
Ok, adapted to the current situation `ndk-r17` and also added `unittests` for `recommendations` module.

**As a side note:** I skipped some of the tests for py2, because I make use of  `assertLog` which is introduced in Python 3.4 and it fails for our tox py2 tests. There are solutions for this, but it would add an extra complexity and I don't think that it's worth it...@AndreMiras, let me know if you think different in this matter :wink:

inclement:
Please feel free to make this Python 3 only if you prefer, I'll fill out https://github.com/kivy/python-for-android/pull/1918 soon and I think it shouldn't be an issue to merge that PR first.

I didn't raise the question there yet, but 3.4+ seems like a fine support base for now.

pythonforandroid/recommendations.py
```diff
@@ -10,28 +10,80 @@
...
     if version is None:
-        return  # if we failed to read the version, just don't worry about it
+        warning(
+            'Unable to read the ndk version, assuming that you are using an'
+            ' NDK greater than {min_supported} (the minimum ndk required to'
+            ' use p4a successfully).\n'
+            'Note: If you got build errors, consider to download the'
+            ' recommended ndk version which is {rec_version} and try'
+            ' it again (after removing all the files generated with this'
+            ' build). To download the android NDK visit the following page: '
```
inclement: I think this should be worded more directly, something like:
```
warning("Unable to read the NDK version from the given directory {}".format(ndk_dir))
warning("Make sure your NDK version is greater than {min_supported}. If you get build errors, download the recommended NDK {rec_version} from {ndk_url}".format(...))
```

pythonforandroid/recommendations.py
```diff
@@ -10,28 +10,80 @@
...
-    info('Found NDK revision {}'.format(version))
+    info('Found NDK version {}'.format(string_version))
 
     if major_version < MIN_NDK_VERSION:
-        warning('Minimum recommended NDK version is {}'.format(
-            RECOMMENDED_NDK_VERSION))
-        warning(OLD_NDK_MESSAGE)
+        raise BuildInterruptingException(
+            'Unsupported NDK version detected {user_version}\n'
+            '* Note: Minimum supported NDK version is {min_supported}'.format(
```
inclement: Again, I'd write something shorter like "The minimum supported NDK version is {min_supported}. You can download it from {ndk_url}".

tests/test_recommendations.py
```diff
@@ -0,0 +1,191 @@
...
+        Test that a fake ndk dir give us two messages:
+            - first should be an `INFO` log
+            - second should be an `WARNING` log
+        """
+        with self.assertLogs(level="INFO") as cm:
+            check_ndk_version(self.ndk_dir)
+        self.assertEqual(
+            cm.output,
+            [
+                "INFO:p4a:[INFO]:    Could not determine NDK version, "
```
inclement: I think all these exact output checks may be annoying to keep updated. How about storing the unformatted strings in recommendations.py, and importing them here when checking the printed output matches them? I wouldn't necessarily suggest that for most log output, but I think it makes sense for this kind of important message which is a meaningful string in itself and not a throwaway log.

More generally, it might be nice to have helper functions that can pattern match against the expected log level (info, warning, ...) and some of the string content, when the whole string isn't important. That's just a thought, not a request here.

opacam:
> I didn't raise the question there yet, but 3.4+ seems like a fine support base for now.

Aaahhh, I would go with Python 3.6+ so we could make use of `f-strings` without external dependencies.

AndreMiras:
Yes @opacam of course we can skip python2 now :+1: 
Yes @inclement good question for python3, honestly I'd rather just support 3.6 and 3.7. It's purely arbitrary. Just because I like f-strings and also Ubuntu LTS doesn't have 3.5 :stuck_out_tongue:

inclement:
@opacam I'm not certain about 3.6+ but I'm open for discussion, feel free to raise it on the "drop python2" PR. I suggested assuming 3.4+ here only because that's the version you mentioned needing.

opacam:
Guys, If you don't mind, please don't merge this, I would like to address the @inclement comments before :wink:

opacam:
Ok, I think I addressed all of the @inclement comments, going to dinner now...I will check later or tomorrow, let me know if you find anything to be changed :smile:

pythonforandroid/recommendations.py
```diff
@@ -92,14 +164,19 @@ def check_target_api(api, arch):
 MIN_NDK_API = 21
 RECOMMENDED_NDK_API = 21
 OLD_NDK_API_MESSAGE = ('NDK API less than {} is not supported'.format(MIN_NDK_API))
+TARGET_NDK_API_GREATER_THAN_TARGET_API_MESSAGE = (
+    'Target NDK API is {ndk_api}, '
+    'higher than the target Android API {android_api}.'
+)
```
AndreMiras: Just wondering, why in p4a we don't have all the const on the top of the file? It's probably fine, it's just that I'm more used to do it this way, or even sometimes in a dedicated file

tests/test_recommendations.py
```diff
@@ -0,0 +1,204 @@
...
+        mock_open_src_prop.side_effect = [
+            mock.mock_open(
+                read_data="Pkg.Revision = 17.2.4988734"
+            ).return_value
+        ]
+        version = read_ndk_version(self.ndk_dir)
+        mock_open_src_prop.assert_called_once_with(
+            join(self.ndk_dir, "source.properties")
+        )
+        assert version == "17.2.4988734"
```
AndreMiras: minor: inconsistent assert. I personally prefer this version, but it's not what you seem to be using in the rest of the file.
Same thing in `test_read_ndk_version_error` below.
But anyway we're far from being consistent in this with p4a currently. Don't know if we want to address or not

tests/test_recommendations.py
```diff
@@ -0,0 +1,204 @@
...
+            cm.output,
+            ["INFO:p4a:[INFO]:    {}".format(PARSE_ERROR_NDK_MESSAGE)],
+        )
+        mock_open_src_prop.assert_called_once_with(
+            join(self.ndk_dir, "source.properties")
+        )
+        assert version is None
+
+    def test_check_target_api_error_arch_armeabi(self):
+
```
AndreMiras: Nitpick: why are we sometimes leaving a blank space after method definition and sometimes not?

tests/test_recommendations.py
```diff
@@ -0,0 +1,204 @@
...
+            TARGET_NDK_API_GREATER_THAN_TARGET_API_MESSAGE.format(
+                ndk_api=ndk_api, android_api=android_api
+            ),
+        )
+
+    @unittest.skipIf(running_in_py2, "`assertLogs` requires Python 3.4+")
+    def test_check_ndk_api_warning_old_ndk(self):
+        """
+        Given an `android api` lower than the supported by p4a, we should
+        get an `BuildInterruptingException`.
```
AndreMiras: This comment seem to be a lie, right?

AndreMiras:
Very neat! I'm going to give it a try and if nothing seems to break, I'm in for a merge :+1:

inclement:
I'm assuming this is fine to merge. Thanks @opacam :)

PR #1886, opened 2019-06-25
[WIP] Fix foreground notification being mandatory and more

JonasT:
Fix foreground notification being mandatory and more. Details:

- Fixes that right now the service always slaps up a notification
  without this being configurable at runtime. It only is configurable
  right now if setting the foreground property via --service, which
  cannot be done when using the simpler service/main.py entrypoint

- Fixes service_only service code template not rendering .foreground
  correctly since it added an outdated, useless function name override

- Fixes notification channel name which is visible to end user
  hardcoding "python" in its name which is not ideal for end user
  naming (since the average user might not even know what python is)

- Fixes override of doStartForeground() leading to quite some
  error-prone code duplication


JonasT:
Closing in favor of #1888

PR #1887, opened 2019-06-26
Fix Bootstrap.get_bootstrap_from_recipes() so it's smarter and deterministic, fixes #1875

### LINKED ISSUE #1875: get_bootstrap_from_recipes() result consistency
In https://github.com/kivy/python-for-android/pull/1872#discussion_r294949284 we realised get_bootstrap_from_recipes() was not always returning consistent result across setups.
This is an investigation task to determine if it's a bug or a feature and if it's a bug to fix it.

JonasT:
This should fix `Bootstrap.get_bootstrap_from_recipes()` not being deterministic, which fixes #1875

### DIFF
--- a/pythonforandroid/bootstrap.py
+++ b/pythonforandroid/bootstrap.py
@@ -1,10 +1,11 @@
+import functools
+import glob
+import importlib
+import os
 from os.path import (join, dirname, isdir, normpath, splitext, basename)
 from os import listdir, walk, sep
 import sh
 import shlex
-import glob
-import importlib
-import os
 import shutil
 
 from pythonforandroid.logger import (warning, shprint, info, logger,
@@ -32,6 +33,35 @@ def copy_files(src_root, dest_root, override=True):
                     shutil.copy(src_file, dest_file)
             else:
                 os.makedirs(dest_file)
+
+
+default_recipe_priorities = [
+    "webview", "sdl2", "service_only"  # last is highest
+]
+# ^^ NOTE: these are just the default priorities if no special rules
+# apply (which you can find in the code below), so basically if no
+# known graphical lib or web lib is used - in which case service_only
+# is the most reasonable guess.
+
+
+def _cmp_bootstraps_by_priority(a, b):
+    def rank_bootstrap(bootstrap):
+        """ Returns a ranking index for each bootstrap,
+            with higher priority ranked with higher number. """
+        if bootstrap.name in default_recipe_priorities:
+            return default_recipe_priorities.index(bootstrap.name) + 1
+        return 0
+
+    # Rank bootstraps in order:
+    rank_a = rank_bootstrap(a)
+    rank_b = rank_bootstrap(b)
+    if rank_a != rank_b:
+        return (rank_b - rank_a)
+    else:
+        if a.name < b.name:  # alphabetic sort for determinism
+            return -1
+        else:
+            return 1
 
 
 class Bootstrap(object):
@@ -138,36 +168,43 @@ class Bootstrap(object):
         self.distribution.save_info(self.dist_dir)
 
     @classmethod
-    def list_bootstraps(cls):
+    def all_bootstraps(cls):
         '''Find all the available bootstraps and return them.'''
         forbidden_dirs = ('__pycache__', 'common')
         bootstraps_dir = join(dirname(__file__), 'bootstraps')
+        result = set()
         for name in listdir(bootstraps_dir):
             if name in forbidden_dirs:
                 continue
             filen = join(bootstraps_dir, name)
             if isdir(filen):
-                yield name
+                result.add(name)
+        return result
 
     @classmethod
-    def get_bootstrap_from_recipes(cls, recipes, ctx):
-        '''Returns a bootstrap whose recipe requirements do not conflict with
-        the given recipes.'''
+    def get_usable_bootstraps_for_recipes(cls, recipes, ctx):
+        '''Returns all bootstrap whose recipe requirements do not conflict
+        with the given recipes, in no particular order.'''
         info('Trying to find a bootstrap that matches the given recipes.')
         bootstraps = [cls.get_bootstrap(name, ctx)
-                      for name in cls.list_bootstraps()]
-        acceptable_bootstraps = []
+                      for name in cls.all_bootstraps()]
+        acceptable_bootstraps = set()
+
+        # Find out which bootstraps are acceptable:
         for bs in bootstraps:
             if not bs.can_be_chosen_automatically:
                 continue
-            possible_dependency_lists = expand_dependencies(bs.recipe_depends)
+            possible_dependency_lists = expand_dependencies(bs.recipe_depends, ctx)
             for possible_dependencies in possible_dependency_lists:
                 ok = True
+                # Check if the bootstap's dependencies have an internal conflict:
                 for recipe in possible_dependencies:
                     recipe = Recipe.get_recipe(recipe, ctx)
                     if any([conflict in recipes for conflict in recipe.conflicts]):
                         ok = False
                         break
+                # Check if bootstrap's dependencies conflict with chosen
+                # packages:
                 for recipe in recipes:
                     try:
                         recipe = Recipe.get_recipe(recipe, ctx)
@@ -180,14 +217,58 @@ class Bootstrap(object):
                         ok = False
                         break
                 if ok and bs not in acceptable_bootstraps:
-                    acceptable_bootstraps.append(bs)
+                    acceptable_bootstraps.add(bs)
+
         info('Found {} acceptable bootstraps: {}'.format(
             len(acceptable_bootstraps),
             [bs.name for bs in acceptable_bootstraps]))
-        if acceptable_bootstraps:
-            info('Using the first of these: {}'
-                 .format(acceptable_bootstraps[0].name))
-            return acceptable_bootstraps[0]
+        return acceptable_bootstraps
+
+    @classmethod
+    def get_bootstrap_from_recipes(cls, recipes, ctx):
+        '''Picks a single recommended default bootstrap out of
+           all_usable_bootstraps_from_recipes() for the given reicpes,
+           and returns it.'''
+
+        known_web_packages = {"flask"}  # to pick webview over service_only
+        recipes_with_deps_lists = expand_dependencies(recipes, ctx)
+        acceptable_bootstraps = cls.get_usable_bootstraps_for_recipes(
+            recipes, ctx
+        )
+
+        def have_dependency_in_recipes(dep):
+            for dep_list in recipes_with_deps_lists:
+                if dep in dep_list:
+                    return True
+            return False
+
+        # Special rule: return SDL2 bootstrap if there's an sdl2 dep:
+        if (have_dependency_in_recipes("sdl2") and
+                "sdl2" in [b.name for b in acceptable_bootstraps]
+                ):
+            info('Using sdl2 bootstrap since it is in dependencies')
+            return cls.get_bootstrap("sdl2", ctx)
+
+        # Special rule: return "webview" if we depend on common web recipe:
+        for possible_web_dep in known_web_packages:
+            if have_dependency_in_recipes(possible_web_dep):
+                # We have a web package dep!
+                if "webview" in [b.name for b in acceptable_bootstraps]:
+                    info('Using webview bootstrap since common web packages '
+                         'were found {}'.format(
+                             known_web_packages.intersection(recipes)
+                         ))
+                    return cls.get_bootstrap("webview", ctx)
+
+        prioritized_acceptable_bootstraps = sorted(
+            list(acceptable_bootstraps),
+            key=functools.cmp_to_key(_cmp_bootstraps_by_priority)
+        )
+
+        if prioritized_acceptable_bootstraps:
+            info('Using the highest ranked/first of these: {}'
+                 .format(prioritized_acceptable_bootstraps[0].name))
+            return prioritized_acceptable_bootstraps[0]
         return None
 
     @classmethod
@@ -299,9 +380,26 @@ class Bootstrap(object):
                 shprint(sh.rm, '-rf', d)
 
 
-def expand_dependencies(recipes):
+def expand_dependencies(recipes, ctx):
+    """ This function expands to lists of all different available
+        alternative recipe combinations, with the dependencies added in
+        ONLY for all the not-with-alternative recipes.
+        (So this is like the deps graph very simplified and incomplete, but
+         hopefully good enough for most basic bootstrap compatibility checks)
+    """
+
+    # Add in all the deps of recipes where there is no alternative:
+    recipes_with_deps = list(recipes)
+    for entry in recipes:
+        if not isinstance(entry, (tuple, list)) or len(entry) == 1:
+            if isinstance(entry, (tuple, list)):
+                entry = entry[0]
+            recipe = Recipe.get_recipe(entry, ctx)
+            recipes_with_deps += recipe.depends
+
+    # Split up lists by available alternatives:
     recipe_lists = [[]]
-    for recipe in recipes:
+    for recipe in recipes_with_deps:
         if isinstance(recipe, (tuple, list)):
             new_recipe_lists = []
             for alternative in recipe:
@@ -311,6 +409,6 @@ def expand_dependencies(recipes):
                     new_recipe_lists.append(new_list)
             recipe_lists = new_recipe_lists
         else:
-            for old_list in recipe_lists:
-                old_list.append(recipe)
+            for existing_list in recipe_lists:
+                existing_list.append(recipe)
     return recipe_lists
--- a/tests/test_bootstrap.py
+++ b/tests/test_bootstrap.py
@@ -1,6 +1,6 @@
+
 import os
 import sh
-
 import unittest
 
 try:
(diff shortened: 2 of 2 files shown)


AndreMiras:
Thanks @JonasT for the pull request :pray: 
I think @opacam was suggesting to maybe go a bit deeper than just sorting it, see his comment there https://github.com/kivy/python-for-android/pull/1872#discussion_r296253683
Also we may want to unit test this change

JonasT:
@AndreMiras added it to the test! If I understand @opacam 's suggestion basically is "don't return the first alphabetical one but something better", but this is strictly speaking not related to the determinism issue and also something that doesn't concern `Bootstrap.list_bootstraps` but rather should be additional logic in `Bootstrap.get_bootstrap_from_recipes`. So IMHO it makes sense to add this change anyway

tests/test_bootstrap.py
```diff
@@ -100,18 +100,29 @@ def test_build_dist_dirs(self):
...
         `webview` and `sdl2`
         """
-        expected_bootstraps = {"empty", "service_only", "webview", "sdl2"}
-        set_of_bootstraps = set(Bootstrap().list_bootstraps())
+
+        mock_listdir.return_value = \
+            ["empty", "sdl2", "service_only", "webview"]
+        expected_bootstraps = sorted(
+            ["empty", "service_only", "webview", "sdl2"]
+        )
```
AndreMiras: minor suggestion to improve the readability of this block:
```python
listdir_ret = ["empty", "service_only", "webview", "sdl2"]
mock_listdir.return_value = listdir_ret
expected_bootstraps = sorted(listdir_ret)
```
That way we make it clear that we're talking about the same list once sorted once not. Because here you somehow shuffled your lists and at first sight we don't see they're the same or why they're shuffled.

tests/test_bootstrap.py
```diff
@@ -100,18 +100,29 @@ def test_build_dist_dirs(self):
...
+        self.assertEqual(
+            expected_bootstraps, list(Bootstrap().list_bootstraps())
+        )
+
+        # Make sure order remains stable even if listdir() order varies:
+        mock_listdir.return_value = \
+            ["empty", "webview", "sdl2", "service_only"]
```
AndreMiras: Minor: then you could use `listdir_ret.reverse()` base don the suggestion above

AndreMiras:
Yes @JonasT I think you have a fair point, thanks

opacam:
**IMHO**, It does not make sense to me to return the results sorted in this case. I think that we will only add useless code...so I would rather do it once and well done (with some logic I mean), based on recipes, something like:

  - if `flask` in recipes I would expect an `webview` bootstrap
  - if `kivy` in recipes I would expect an `sdl2` bootstrap
  - all the other combinations I would make it go to `service_only` bootstrap...unless someone thinks about that we should do some other special match

If we apply the `sorted way` proposed in here, it would only lead to inconsistencies (like we have now), but if all of you want to go this `sorted way`, then let's do it...we can apply some kind of logic later.

JonasT:
Yeah @opacam I can see your point, maybe `list_bootstraps` should actually return a set... I'll try something else hang on

JonasT:
@opacam ok I made something new which fixes `Bootstrap.get_bootstrap_from_recipes`, is that more like what you had in mind?

AndreMiras:
Yes agree guys, I also thought this functionally should be a set. But then I started imagining all the refactor that would require and I thought "OK maybe not for now" :stuck_out_tongue: 
Anyway if you feel like giving it a try let's do it :smile: 
I'll take a quick first look now and continue reviewing later

pythonforandroid/bootstrap.py
```diff
@@ -138,36 +139,52 @@ def run_distribute(self):
         self.distribution.save_info(self.dist_dir)
 
     @classmethod
-    def list_bootstraps(cls):
+    def all_bootstraps(cls):
         '''Find all the available bootstraps and return them.'''
         forbidden_dirs = ('__pycache__', 'common')
         bootstraps_dir = join(dirname(__file__), 'bootstraps')
+        result = set()
```
AndreMiras: Yes also thought "this should be a set". But the fact that it was an iterator before made me more cautious. But seeing the code I'm not seeing why not being an iterator would be an issue. The `result` set would be anything enormous, right, so no risk to load it on the stack, correct?

tests/test_bootstrap.py
```diff
@@ -100,33 +102,92 @@ def test_build_dist_dirs(self):
...
+    def test_expand_dependencies(self):
+        # Test depenency expansion of a recipe with no alternatives:
+        expanded_result_1 = expand_dependencies(["pysdl2"], self.ctx)
+        assert({"sdl2", "pysdl2", "python3"} in
+               [set(s) for s in expanded_result_1])
+
+        # Test all alternatives are listed (they won't have dependencies
+        # expanded since expand_dependencies() is too simplistic):
+        expanded_result_2 = expand_dependencies([("pysdl2", "kivy")], self.ctx)
+        assert([["pysdl2"], ["kivy"]] == expanded_result_2)
```
AndreMiras: Minor: `assert` is still a keyword, not a function :smile: 
If you like the function stile you can also use the `self.assertEqual()` or `self.assertTrue()`. That would also make it more consistent with the style used in this file

tests/test_bootstrap.py
```diff
@@ -100,33 +102,92 @@ def test_build_dist_dirs(self):
...
+
+        @mock.patch("pythonforandroid.recipe.Recipe.get_recipe")
+        def do_test_get_bootstraps_from_recipes(mock_get_recipe):
+            """A test which will initialize a bootstrap and will check if the
+            method :meth:`~pythonforandroid.bootstrap.Bootstrap.
+            get_bootstraps_from_recipes` returns the expected values
+            """
+
+            # Use original get_recipe() until we need something else later:
+            mock_get_recipe.side_effect = original_get_recipe
```
AndreMiras: then if you don't need it mocked now, use the mock.patch context manager rather than the function decorator

tests/test_bootstrap.py
```diff
@@ -100,33 +102,92 @@ def test_build_dist_dirs(self):
...
+
+            # Test that SDL2 works with kivy:
+            recipes_sdl2 = {"sdl2", "python3", "kivy"}
+            bs = Bootstrap().get_bootstrap_from_recipes(recipes_sdl2, self.ctx)
```
AndreMiras: Minor: I saw that we did it in other places prior your change, but if `get_bootstrap_from_recipes()` is a static method, then let's not make an instance of the `Bootstrap` class. So:
```python
bs = Bootstrap.get_bootstrap_from_recipes(recipes_sdl2, self.ctx)
```

pythonforandroid/bootstrap.py
```diff
@@ -180,14 +197,61 @@ def get_bootstrap_from_recipes(cls, recipes, ctx):
...
+        # Special rule: return SDL2 bootstrap if there's an sdl2 dep:
+        if (have_dependency_in_recipes("sdl2") and
+                "sdl2" in [b.name for b in acceptable_bootstraps]
+                ):
+            info('Using sdl2 bootstrap since it is in dependencies')
+            return cls.get_bootstrap("sdl2", ctx)
+
+        # Special rule: return "webview" if we depend on common web recipe:
+        for possible_web_dep in known_web_packages:
+            if have_dependency_in_recipes(possible_web_dep):
```
AndreMiras: Just some thoughts without deep diving into it. So I'm not sure if we could get any better, but the time complexity of this one is concerning me.
Let me know if I'm seeing this correctly or not.
`have_dependency_in_recipes()` itself seems `O(n²)` because we have a `in` in a `for` loop.
Then we call `have_dependency_in_recipes()` itself in another for loop so that gives us a `O(n³)`.
If by any chance `dep_list` is a `set` then the `if dep in dep_list:` is not `O(n)` anymore, but `O(1)`. So then the overall complexity would be `O(n²)`.

pythonforandroid/bootstrap.py
```diff
@@ -138,36 +139,52 @@ def run_distribute(self):
...
         info('Trying to find a bootstrap that matches the given recipes.')
         bootstraps = [cls.get_bootstrap(name, ctx)
-                      for name in cls.list_bootstraps()]
-        acceptable_bootstraps = []
+                      for name in cls.all_bootstraps()]
+        acceptable_bootstraps = set()
+        known_web_packages = {"flask"}  # to pick webview over service_only
+        default_recipe_priorities = [
+            "webview", "sdl2", "service_only"  # last is highest
+        ]
```
AndreMiras: Just some thoughts, what if someone uses `flask` as a dep because they need a sub component of it, but still want to be `sdl2` or `service_only`. Probably in this case they can just force the boostrap passing it as param? Just curious/wondering

pythonforandroid/bootstrap.py
```diff
@@ -138,36 +139,52 @@ def run_distribute(self):
...
             if name in forbidden_dirs:
                 continue
             filen = join(bootstraps_dir, name)
             if isdir(filen):
-                yield name
+                result.add(name)
+        return result
 
     @classmethod
     def get_bootstrap_from_recipes(cls, recipes, ctx):
```
AndreMiras: feel like this method is big now and could be split furthermore by concerns, what do you think?

pythonforandroid/bootstrap.py
```diff
@@ -180,14 +197,61 @@ def get_bootstrap_from_recipes(cls, recipes, ctx):
...
+                return (a.name - b.name)  # alphabetic sort for determinism
+
+        prioritized_acceptable_bootstraps = sorted(
+            list(acceptable_bootstraps),
+            key=functools.cmp_to_key(bootstrap_priority)
+        )
+
+        if prioritized_acceptable_bootstraps:
+            info('Using the highest ranked/first of these: {}'
+                 .format(prioritized_acceptable_bootstraps[0].name))
```
AndreMiras: Minor: The message say `of these` so I was expecting to see the list. But we actually show the one being picked up, which is fine, but then the log wording should change a bit

pythonforandroid/bootstrap.py
```diff
@@ -180,14 +197,61 @@ def get_bootstrap_from_recipes(cls, recipes, ctx):
...
+            rank_b = rank_bootstrap(b)
+            if rank_a != rank_b:
+                return (rank_b - rank_a)
+            else:
+                return (a.name - b.name)  # alphabetic sort for determinism
+
+        prioritized_acceptable_bootstraps = sorted(
+            list(acceptable_bootstraps),
+            key=functools.cmp_to_key(bootstrap_priority)
+        )
```
AndreMiras: very interesting helper function. I understand you didn't want to expose it because it's a util function only being used in `get_bootstrap_from_recipes()`, but then:
1) it makes the overall `get_bootstrap_from_recipes()` method larger
2) the `bootstrap_priority()` helper is not unit testable

Maybe we should define it as a private top function inside the file `bootstrap.py` file? I have the feeling it would make the overall thing more readable and more testable

pythonforandroid/bootstrap.py
```diff
@@ -299,9 +363,26 @@ def fry_eggs(self, sitepackages):
...
+        (So this is like the deps graph very simplified and incomplete, but
+         hopefully good enough for most basic bootstrap compatibility checks)
+    """
+
+    # Add in all the deps of recipes where there is no alternative:
+    recipes_with_deps = list(recipes)
+    for entry in recipes:
+        if not isinstance(entry, (tuple, list)) or len(entry) == 1:
+            if isinstance(entry, (tuple, list)):
+                entry = entry[0]
```
AndreMiras: There's something fishy here :fish: :smile: 
First it doesn't seem covered, second it means `expand_dependencies()` would be a bit polymorphic in a way which can lead to unexpected behavior or code not so easy to follow.

AndreMiras:
I made a first review round, your approach looks promising, thank you for giving it a try :clap: 
I hope I can dig deeper and also test it later this week

opacam:
>  @opacam ok I made something new which fixes Bootstrap.get_bootstrap_from_recipes, is that more like what you had in mind?

@JonasT, oh yeah, this looks far better, thanks for doing this :smile:

**Note:** I have took a superficial look and it looks good, I only miss one thing, I think that maybe we should add some documentation about this new method to find the right bootstrap via user recipes...maybe a note in `doc/source/bootstraps.rst` explaining this new behaviour?

pythonforandroid/bootstrap.py
JonasT: SDL2 overrides webview in the special behavior, so that is no prob. The other way around is though, e.g. my website backend actually uses SDL2 :joy: but yeah in that case you just need to specify the bootstrap manually

pythonforandroid/bootstrap.py
JonasT: sorry I don't see it :joy: the previous one I finally split was due to nesting not to length, I think "it's big" is still a somewhat poor criteria (on its own) for a split. UNLESS you want to be able to test the sub parts separately of course, although I'm not sure that is necessary...? I would personally leave it as is :woman_shrugging:

pythonforandroid/bootstrap.py
JonasT: What do you mean with polymorphic? This is the usual behavior all graph lists work, in fact I did not introduce this at all but just stuck to it as it already behaved (see the low loop.) So no, this is as it should be. (Dependencies can be a string or a tuple of alternatives, that is how things are used everywhere)

The coverage is a good point though, let me look into that

pythonforandroid/bootstrap.py
AndreMiras: OK thanks for the info :+1:

pythonforandroid/bootstrap.py
AndreMiras: Yes definitely big is a poor and arbitrary criteria, just like the 80 char lines of the pep. It probably started from some feelings and adjusted over time/researches. That "it's big" feeling would vary from one to another of course.
I wanted to share the feeling before looking for solution to see maybe luckily something super clever would have come to your or other people mind :smile:

pythonforandroid/bootstrap.py
AndreMiras: Thanks for the reply. Great that you will look at the coverage, who knows maybe we will come across another old bug or dead code :smile:

pythonforandroid/bootstrap.py
AndreMiras: Then see taking a quick look I already see some possible split since the comments and the logs say it all, for instance:
`# Find out which bootstraps are acceptable:`
Then a big block and then:
`info('Found {} acceptable bootstraps: {}`
For me that definitely feels like a split. Not only it will make things more readable:
`acceptable_bootstraps = find_acceptable_bootstraps(...)`
But also make it more unit testable :wink:

JonasT: Yeah and it also obscures the control flow and introduces interdependencies and more interfaces making changes later harder. I definitely see your point but I don't think chasing that far after tests is necessarily good, so if you ask me I think this is better off as it is UNTIL we actually want to use it separately, or it reaches a length where this is more warranted. It is also not such a deeply complex and core functionality that I'd think this hurts us a lot not being tested more granularly.

But as usual, I'm happy to leave the final decision to you guys

tests/test_bootstrap.py
JonasT: ah yea dunno why I didn't :smile: looks much better with context manager

pythonforandroid/bootstrap.py
AndreMiras: That's seemed like an easy refactor to me, but OK :cry: 
As usual I'm nitpicking, but won't block a PR for something like this. So if you don't feel like it let's leave it for now. Maybe at some point it bugs so much that I'll do it myself.
Thanks for discussing it :+1:

pythonforandroid/bootstrap.py
```diff
@@ -34,6 +35,33 @@ def copy_files(src_root, dest_root, override=True):
...
+    "webview", "sdl2", "service_only"  # last is highest
+]
+# ^^ NOTE: these are just the default priorities if no special rules
+# apply (which you can find in the code below), so basically if no
+# known graphical lib or web lib is used - in which case service_only
+# is the most reasonable guess.
+
+
+def _cmp_bootstraps_by_priority(a, b):
+    global default_recipe_priorities
```
AndreMiras: This is confusing me, why do you need the `global` keyword if you're not updating the reference of the object?

tests/test_bootstrap.py
```diff
@@ -100,32 +104,106 @@ def test_build_dist_dirs(self):
             bs.get_common_dir().endswith("pythonforandroid/bootstraps/common")
         )
 
-    def test_list_bootstraps(self):
+    def test__cmp_bootstraps_by_priority(self):
+        # Test service_only has higher priority than sdl2:
+        self.assertTrue(_cmp_bootstraps_by_priority(
+            Bootstrap.get_bootstrap("service_only", self.ctx),
+            Bootstrap.get_bootstrap("sdl2", self.ctx)
+        ) < 0)
```
AndreMiras: Awesome thanks!
Then we can write a simple test for covering the alphabetic sorting `return (a.name - b.name)  # alphabetic sort for determinism`
So if basically if I understood correctly, we just need to comp two bootstraps that're not in `default_recipe_priorities` list. And if we don't have any, we can just patch the list to make it empty.
Would that make sense?

JonasT: Ah, neat idea! I'll add that, thanks :heart_eyes:

pythonforandroid/bootstrap.py
JonasT: Ohhh then I need to remove it!! I somehow thought it would be less confusing to make sure people realize it's not a local var, but I guess this is more confusing isn't it? I'll remove it :grinning:

pythonforandroid/bootstrap.py
JonasT: oh no I haven't convinced you even in the slightest have I :scream: oh dear. maybe I'm wrong then, I'll split it up :+1: I mean I can see your side a little, and in the long run I feel like you had so far a slightly better grasp at writing nice code than I have

AndreMiras: That's so nice of you :blush: 
Thank you @JonasT !

JonasT:
Okay so I think I worked pretty much everything in (if not I probably forgot, poke me). In my tests it worked, so maybe it would be a good time to check out if this is worth integrating, after all it is quite self-contained and does some nice little local code improvements

PR #1888, opened 2019-06-26
Fix foreground notification being mandatory and more (attempt 2)

JonasT:
- Adds start_service_not_as_foreground() to avoid always displaying
  an associated notification without this being configurable at runtime.
  It only is configurable right now if setting the foreground property
  via --service, which cannot be done when using the simpler
  service/main.py entrypoint

- Fixes service_only service code template not rendering .foreground
  correctly since it added an outdated, useless function name override

- Fixes notification channel name which is visible to end user
  hardcoding "python" in its name which is not ideal for end user
  naming (since the average user might not even know what python is)

- Fixes override of doStartForeground() leading to quite some
  error-prone code duplication

### DIFF
--- a/pythonforandroid/bootstraps/common/build/templates/Service.tmpl.java
+++ b/pythonforandroid/bootstraps/common/build/templates/Service.tmpl.java
@@ -1,15 +1,8 @@
 package {{ args.package }};
 
-import android.os.Build;
-import java.lang.reflect.Method;
-import java.lang.reflect.InvocationTargetException;
 import android.content.Intent;
 import android.content.Context;
-import android.app.Notification;
-import android.app.PendingIntent;
-import android.os.Bundle;
 import org.kivy.android.PythonService;
-import org.kivy.android.PythonActivity;
 
 
 public class Service{{ name|capitalize }} extends PythonService {
@@ -20,41 +13,9 @@ public class Service{{ name|capitalize }} extends PythonService {
     }
     {% endif %}
 
-    {% if not foreground %}
     @Override
-    public boolean canDisplayNotification() {
-        return false;
-    }
-    {% endif %}
-
-    @Override
-    protected void doStartForeground(Bundle extras) {
-        Notification notification;
-        Context context = getApplicationContext();
-        Intent contextIntent = new Intent(context, PythonActivity.class);
-        PendingIntent pIntent = PendingIntent.getActivity(context, 0, contextIntent,
-            PendingIntent.FLAG_UPDATE_CURRENT);
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.HONEYCOMB) {
-            notification = new Notification(
-                context.getApplicationInfo().icon, "{{ args.name }}", System.currentTimeMillis());
-            try {
-                // prevent using NotificationCompat, this saves 100kb on apk
-                Method func = notification.getClass().getMethod(
-                    "setLatestEventInfo", Context.class, CharSequence.class,
-                    CharSequence.class, PendingIntent.class);
-                func.invoke(notification, context, "{{ args.name }}", "{{ name| capitalize }}", pIntent);
-            } catch (NoSuchMethodException | IllegalAccessException |
-                     IllegalArgumentException | InvocationTargetException e) {
-            }
-        } else {
-            Notification.Builder builder = new Notification.Builder(context);
-            builder.setContentTitle("{{ args.name }}");
-            builder.setContentText("{{ name| capitalize }}");
-            builder.setContentIntent(pIntent);
-            builder.setSmallIcon(context.getApplicationInfo().icon);
-            notification = builder.build();
-        }
-        startForeground({{ service_id }}, notification);
+    protected int getServiceId() {
+        return {{ service_id }};
     }
 
     static public void start(Context ctx, String pythonServiceArgument) {
@@ -62,8 +23,11 @@ public class Service{{ name|capitalize }} extends PythonService {
         String argument = ctx.getFilesDir().getAbsolutePath() + "/app";
         intent.putExtra("androidPrivate", ctx.getFilesDir().getAbsolutePath());
         intent.putExtra("androidArgument", argument);
+        intent.putExtra("serviceTitle", "{{ args.name }}");
+        intent.putExtra("serviceDescription", "{{ name|capitalize }}");
         intent.putExtra("serviceEntrypoint", "{{ entrypoint }}");
         intent.putExtra("pythonName", "{{ name }}");
+        intent.putExtra("serviceStartAsForeground", "{{ foreground|lower }}");
         intent.putExtra("pythonHome", argument);
         intent.putExtra("pythonPath", argument + ":" + argument + "/lib");
         intent.putExtra("pythonServiceArgument", pythonServiceArgument);
--- a/pythonforandroid/bootstraps/service_only/build/templates/Service.tmpl.java
+++ b/pythonforandroid/bootstraps/service_only/build/templates/Service.tmpl.java
@@ -22,15 +22,10 @@ public class Service{{ name|capitalize }} extends PythonService {
     }
     {% endif %}
 
-    {% if foreground %}
-    /**
-     * {@inheritDoc}
-     */
     @Override
-    public boolean getStartForeground() {
-        return true;
+    protected int getServiceId() {
+        return {{ service_id }};
     }
-    {% endif %}
 
     public static void start(Context ctx, String pythonServiceArgument) {
         String argument = ctx.getFilesDir().getAbsolutePath() + "/app";
@@ -41,6 +36,7 @@ public class Service{{ name|capitalize }} extends PythonService {
         intent.putExtra("serviceTitle", "{{ name|capitalize }}");
         intent.putExtra("serviceDescription", "");
         intent.putExtra("pythonName", "{{ name }}");
+        intent.putExtra("serviceStartAsForeground", "{{ foreground|lower }}");
         intent.putExtra("pythonHome", argument);
         intent.putExtra("androidUnpack", argument);
         intent.putExtra("pythonPath", argument + ":" + argument + "/lib");
--- a/pythonforandroid/recipes/android/src/android/_android.pyx
+++ b/pythonforandroid/recipes/android/src/android/_android.pyx
@@ -280,17 +280,29 @@ class AndroidBrowser(object):
 import webbrowser
 webbrowser.register('android', AndroidBrowser)
 
-cdef extern void android_start_service(char *, char *, char *)
-def start_service(title=None, description=None, arg=None):
-    cdef char *j_title = NULL
-    cdef char *j_description = NULL
-    if title is not None:
-        j_title = <bytes>title
-    if description is not None:
-        j_description = <bytes>description
-    if arg is not None:
-        j_arg = <bytes>arg
-    android_start_service(j_title, j_description, j_arg)
+
+def start_service(title="Background Service",
+                  description="", arg="",
+                  as_foreground=True):
+    # Legacy None value support (for old function signature style):
+    if title is None:
+        title = "Background Service"
+    if description is None:
+        description = ""
+    if arg is None:
+        arg = ""
+
+    # Start service:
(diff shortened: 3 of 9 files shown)


JonasT:
@AndreMiras you seem to be using to be services a lot, could you possibly test if this pull request breaks any of your service things or works fine?

AndreMiras:
Thanks @inclement for reviewing and thank you @JonasT for pinging me. I'll give it a try hopefully over the weekend so you can merge it.
I'll keep you posted

JonasT:
@inclement good point! for what it's worth I'm pretty sure I didn't touch any of the SDK/NDK/Android-facing APIs but only all the p4a handling before, but one can never be safe enough so I wouldn't mind more testing. I tested with Android 8.1 / API level 27 on a real device (the app being built with target api 27 and min-api 21)

AndreMiras:
I'm getting build errors.
```
[INFO]:    Selecting java build tool:
[INFO]:    Detected highest available build tools version to be 29.0.1
[INFO]:        Building with gradle, as gradle executable is present
[DEBUG]:   -> running gradlew assembleDebug
[DEBUG]:
[DEBUG]:        > Task :processDebugManifest 
[DEBUG]:        /home/andre/workspace/EtherollApp/.buildozer/android/platform/build/dists/etheroll/src/main/AndroidManifest.xml:35:5-81 Warning:
[DEBUG]:                Element uses-permission#android.permission.WRITE_EXTERNAL_STORAGE at AndroidManifest.xml:35:5-81 duplicated with element declared at AndroidManifest.xml:28:5-81
[DEBUG]:
[DEBUG]:        > Task :compileDebugJavaWithJavac FAILED
[DEBUG]:        /home/andre/workspace/EtherollApp/.buildozer/android/platform/build/dists/etheroll/src/main/java/com/github/andremiras/etheroll/ServiceService.java:11: error: cannot find symbol
[DEBUG]:            @override
[DEBUG]:             ^
[DEBUG]:          symbol:   class override
[DEBUG]:          location: class ServiceService
[DEBUG]:        Note: Some input files use or override a deprecated API.
[DEBUG]:        Note: Recompile with -Xlint:deprecation for details.
[DEBUG]:        Note: Some input files use unchecked or unsafe operations.
[DEBUG]:        Note: Recompile with -Xlint:unchecked for details.
[DEBUG]:        1 error
[DEBUG]:
[DEBUG]:
[DEBUG]:        FAILURE: Build failed with an exception.
[DEBUG]:
[DEBUG]:        * What went wrong:
[DEBUG]:        Execution failed for task ':compileDebugJavaWithJavac'.
[DEBUG]:        > Compilation failed; see the compiler error output for details.
[DEBUG]:
[DEBUG]:        * Try:
[DEBUG]:        Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
[DEBUG]:
[DEBUG]:        * Get more help at https://help.gradle.org
[DEBUG]:
[DEBUG]:        BUILD FAILED in 1s
[DEBUG]:        15 actionable tasks: 15 executed
```
See full [build.log](https://github.com/kivy/python-for-android/files/3438245/build.log) and [buildozer spec](https://github.com/kivy/python-for-android/files/3438246/buildozer.txt) attached

pythonforandroid/bootstraps/service_only/build/templates/Service.tmpl.java
```diff
@@ -22,15 +22,10 @@ public int startType() {
...
     {% endif %}
 
-    {% if foreground %}
-    /**
-     * {@inheritDoc}
-     */
-    @Override
-    public boolean getStartForeground() {
-        return true;
+    @override
```
AndreMiras: Should be `@Override` with uppercase `O` it seems

pythonforandroid/bootstraps/common/build/templates/Service.tmpl.java
```diff
@@ -20,50 +13,25 @@ public int startType() {
...
-        } else {
-            Notification.Builder builder = new Notification.Builder(context);
-            builder.setContentTitle("{{ args.name }}");
-            builder.setContentText("{{ name| capitalize }}");
-            builder.setContentIntent(pIntent);
-            builder.setSmallIcon(context.getApplicationInfo().icon);
-            notification = builder.build();
-        }
-        startForeground({{ service_id }}, notification);
+    @override
```
AndreMiras: Should be `@Override` with uppercase `O` it seems

pythonforandroid/bootstraps/common/build/templates/Service.tmpl.java
```diff
@@ -20,50 +13,25 @@ public int startType() {
...
         intent.putExtra("androidArgument", argument);
+        intent.putExtra("serviceTitle", "{{ args.name }}");
+        intent.putExtra("serviceDescription", "{{ name|capitalize }}");
         intent.putExtra("serviceEntrypoint", "{{ entrypoint }}");
         intent.putExtra("pythonName", "{{ name }}");
+        {% if foreground %}
+        intent.putExtra("serviceStartAsForeground", "true");
+        {% else %}
+        intent.putExtra("serviceStartAsForeground", "false");
+        {% endif %}
```
AndreMiras: how about simply:
```
intent.putExtra("serviceStartAsForeground", "{{ foreground|lower }}");
```

pythonforandroid/bootstraps/service_only/build/templates/Service.tmpl.java
```diff
@@ -41,6 +36,11 @@ public static void start(Context ctx, String pythonServiceArgument) {
         intent.putExtra("serviceTitle", "{{ name|capitalize }}");
         intent.putExtra("serviceDescription", "");
         intent.putExtra("pythonName", "{{ name }}");
+        {% if foreground %}
+        intent.putExtra("serviceStartAsForeground", "true");
+        {% else %}
+        intent.putExtra("serviceStartAsForeground", "false");
+        {% endif %}
```
AndreMiras: Same here:
```
intent.putExtra("serviceStartAsForeground", "{{ foreground|lower }}");
``

File: pythonforandroid/recipes/android/src/android/_android.pyx [hunk 5]
```diff
@@ -280,17 +280,20 @@ class AndroidBrowser(object):
...
-        j_title = <bytes>title
-    if description is not None:
-        j_description = <bytes>description
-    if arg is not None:
-        j_arg = <bytes>arg
-    android_start_service(j_title, j_description, j_arg)
+
+def start_service(title=None, description=None, arg=None,
+                  as_foreground=True):
+    from jnius import autoclass
```
AndreMiras: This is already imported at module level

pythonforandroid/recipes/android/src/android/_android.pyx
```diff
@@ -280,17 +280,20 @@ class AndroidBrowser(object):
...
+    mActivity = autoclass('org.kivy.android.PythonActivity').mActivity
```
AndreMiras: Also already available at module level.
https://github.com/kivy/python-for-android/blob/v2019.07.08/pythonforandroid/recipes/android/src/android/_android.pyx#L164
That probably requires some refactor though :smile:

pythonforandroid/recipes/android/src/android/_android.pyx
```diff
@@ -280,17 +280,20 @@ class AndroidBrowser(object):
...
+    if as_foreground:
+        mActivity.start_service(
+            title or "Background Service", description or "", arg or ""
+        )
+    else:
+        mActivity.start_service_not_as_foreground(
+            title or "Background Service", description or "", arg or ""
+        )
```
AndreMiras: why don't you make these default function param then?
```python
def start_service(
    title="Background Service",
    description="",
    arg="",
    as_foreground=True):
```

inclement:
@AndreMiras When making many comments, could you use 'start a review' to submit them all at once? When they're made as individual reviews they cause an email notification for every one!

AndreMiras:
Will try to next time, thanks for the heads up :smile:

pythonforandroid/bootstraps/common/build/templates/Service.tmpl.java
JonasT: oh no that's no good :open_mouth: that is why I want to get some of these merged :scream: I must have done something weird rebasing, I'll do a manual comparison to my rebased combined strange branch I actually used for build and make sure I didn't miss out anything important since apparently at least one rebase conflict ended up weird :grimacing:

AndreMiras: Well that probably mean we should introduce service compilation to our CI. But at the same time it's time/resources every time we add something new to it

JonasT: Wouldn't it be enough to add `service/main.py` to an existing test app? I think this particular bug should also have been triggered if there had been a service used in addition rather than a standalone service `service_only` test app

JonasT: Oh, fascinating! This is not actually a rebase error, it seems when using `service/main.py` (which is what I tested) this indeed isn't rendered out. So that's the reason, it needed a `service_only` test to show up

JonasT: Ah, it's probably the named extra services with the `--service` parameter. We have three ways to do services, I forgot (`service/main.py` with no name which is what I tested, `--service` with named services, and `service_only`)

pythonforandroid/recipes/android/src/android/_android.pyx
JonasT: yeah I've kept the `mActivity` for now, seems a bit ugly to have that outside the function and rely on that being available

pythonforandroid/recipes/android/src/android/_android.pyx
JonasT: good idea! I did so now, but added some extra code to translate `None` to the new defaults - someone might have read the old signature and passed `None` explicitly, and I don't want to break someone's old code

pythonforandroid/bootstraps/service_only/build/templates/Service.tmpl.java
JonasT: good catch :+1:

pythonforandroid/bootstraps/common/build/templates/Service.tmpl.java
AndreMiras: Yes indeed, I'm using the `--service service:service/main.py`, see from the the logs.
Complete command was:
```
/usr/bin/python3 -m pythonforandroid.toolchain apk --debug --bootstrap=sdl2 --dist_name etheroll --name Etheroll --version 2019.0624 --package com.github.andremiras.etheroll --android_api 27 --minsdk 21 --ndk-api 21 --private /home/andre/workspace/EtherollApp/.buildozer/android/app --permission INTERNET --permission WRITE_EXTERNAL_STORAGE --presplash /home/andre/workspace/EtherollApp/docs/images/etheroll-logo.png --icon /home/andre/workspace/EtherollApp/docs/images/icon.png --orientation portrait --window --service service:service/main.py --copy-libs --local-recipes /home/andre/workspace/EtherollApp/src/python-for-android/recipes --blacklist /home/andre/workspace/EtherollApp/blacklist.txt --arch armeabi-v7a --color=always --storage-dir="/home/andre/workspace/EtherollApp/.buildozer/android/platform/build" --ndk-api=21
```

pythonforandroid/recipes/android/src/android/_android.pyx
AndreMiras: I didn't think about this case, who cares about retro compatibility in the Python world :sweat_smile:

JonasT:
@AndreMiras I just rebased, and I actually ran into the same problem as with the SDL2 pull request: `test_folder_exist` in `tests/test_distribution.py` incorrectly doesn't set `mock_exists.return_value = False` which leads to test failure since the code tries to open a non-existing file, and if I fix it then I also need to change `mock_exists.assert_called_once_with` to `mock_exists.assert_called_with` or it'll fail complaining it was called twice :woman_shrugging: so this problem seems to be actually unrelated to any of my changes in any of these two pull requests

Issue #1873, opened 2019-06-18
numpy recipe bug arm64-v8a


musarra:
### Versions

* Python: 3
* OS: linux
* Kivy: 1.11
* Cython: not sure

### Description

// REPLACE ME: What are you trying to get done, what has happened, what went wrong, and what did you expect?

Google Play does not support 32 bit libraries and the armeabi-v7a cannot be used. This means the apk I created and worked for months could not be uploaded. I tried to fix that issue changing the architecture to 64 bit arm64-v8a. Problem is that I have numpy in requirements and it has an issue with the recipe. The apk get's built but crashes immediately, it never opens.

I am not expert programmer but my suggestions to solve/check the issue looking to numpy recipe are:
1) this url = 'https://pypi.python.org/packages/source.....etc' does not exist. I would replace for https://pypi.org/simple/numpy/numpy-1.16.4.zip instead
2) in recipe, version = '1.15.1'. In numpy website, version = 1.16.4

I built a very simple HelloWorld app in kivy to demonstrate this issue below.

Command:
```sh
buildozer android debug
```

### Logs
06/18 23:34:21: Launching hwtest-0.1-debug
$ adb shell am start -n "org.test.hwtest/org.kivy.android.PythonActivity" -a android.intent.action.MAIN -c android.intent.category.LAUNCHER
Client not ready yet..Waiting for process to come online
Waiting for process to come online
Capturing and displaying logcat messages from application. This behavior can be disabled in the "Logcat output" section of the "Debugger" settings page.
I/python: [INFO   ] [Factory     ] 184 symbols loaded
I/python: [INFO   ] [Image       ] Providers: img_tex, img_dds, img_sdl2, img_gif (img_pil, img_ffpyplayer ignored)
I/python:  Traceback (most recent call last):
       File "/home/musarra/Documents/pyprojects/helloworld/.buildozer/android/platform/build/build/python-installs/hwtest/numpy/core/__init__.py", line 16, in <module>
I/python:  ImportError: dlopen failed: cannot locate symbol "__fixunstfsi" referenced by "/data/data/org.test.hwtest/files/app/_python_bundle/site-packages/numpy/core/multiarray.so"...
     
     During handling of the above exception, another exception occurred:
     
     Traceback (most recent call last):
       File "/home/musarra/Documents/pyprojects/helloworld/.buildozer/android/app/main.py", line 3, in <module>
[...]

PR #1892, opened 2019-06-26
Updated numpy recipe to version 1.16.4

inclement:
This seems to fix the recipe, it works for me on both armeabi-v7a and arm64-v8a. Resolves https://github.com/kivy/python-for-android/issues/1873.


inclement:
This seems good to me now, commits tidied up and working fine for me.

opacam:
:thinking:...why is failing the matplotlib recipe in `CI`?

...maybe numpy recipe has troubles in finding our compiled graphic libraries (freetype and png), see: https://travis-ci.org/kivy/python-for-android/jobs/551991431#L5947-L5958

I just made a build with the current numpy recipe and matplotlib to check and I got this for the same lines I pointed above:
```
[DEBUG]:   	REQUIRED DEPENDENCIES AND EXTENSIONS
[DEBUG]:   	                 numpy: yes [not found. pip may install it below.]
[DEBUG]:   	      install_requires: yes [handled by setuptools]
[DEBUG]:   	                libagg: yes [pkg-config information for 'libagg' could not
[DEBUG]:   	                        be found. Using local copy.]
[DEBUG]:   	              freetype: yes [version unknown]
[DEBUG]:   	                   png: yes [version 1.6.36]
[DEBUG]:   	                 qhull: yes [pkg-config information for 'libqhull' could not
[DEBUG]:   	                        be found. Using local copy.]
```

inclement:
There's a fair chance that the matplotlib recipe never passed the CI test, it's only coming up now because the changed recipes test catches it. It looks like the issues raised relate to the local environment where it isn't sufficiently separated from the matplotlib recipe build, I'll try to check this out in a VM.

opacam:
Yeah, you are right, I just tested it, I made a silly change in `matplotlib` recipe with current develop branch, to force CI tests and it also fails with the same error message, so ...I think that we should add `matplotlib` to `BROKEN_RECIPES_PYTHON3` in `ci/constants.py` with some message explaining this (we can do it later)

inclement:
@opacam I updated the broken recipes in this PR, could you approve again if it looks good to you?

pythonforandroid/recipes/matplotlib/mpl_android_fixes.patch
```diff
@@ -4,26 +4,26 @@ index fc82d5d..2067db0 100644
...
-+        return ['numpy==1.15.1']  # to match p4a's target version
- 
++        return ['numpy==1.16.4']  # to match p4a's target version
+
      def get_install_requires(self):
 -        return ['numpy>=1.10.0']
-+        return ['numpy==1.15.1']  # to match p4a's target version
- 
- 
++        return ['numpy==1.16.4']  # to match p4a's target version
```
AndreMiras: Just wondering if this patch is really required? Prior the patch it's with `numpy>=1.10.0` which doesn't conflict with `numpy==1.16.4` correct?

ci/constants.py
```diff
@@ -77,6 +78,7 @@ class TargetPython(Enum):
     # mpmath package with a version >= 0.19 required
     'sympy',
     'vlc',
+    'matplotlib',  # https://github.com/kivy/python-for-android/issues/1900
```
AndreMiras: Thanks for keeping our CI green and thanks for tracking the issue :clap:

pythonforandroid/recipes/matplotlib/mpl_android_fixes.patch
inclement: Yes, it's required, we need matplotlib to install and build against the same version of numpy that p4a builds or there can be incompatibilities. It would be better to get matplotlib to actually use the numpy headers from the p4a build, but wasn't necessary to get the recipe working so I didn't bother.

AndreMiras: OK thanks for the info :+1:

PR #1898, opened 2019-06-28
Fix `run_pymodules_install` when `project_dir` isn't supplied

opacam:


pythonforandroid/build.py
```diff
@@ -721,9 +721,9 @@ def run_pymodules_install(ctx, modules, project_dir=None,
     info('*** PYTHON PACKAGE / PROJECT INSTALL STAGE ***')
     modules = list(filter(ctx.not_has_package, modules))
 
-    # We change current working directory later, so this
-    # has to be an absolute path:
-    project_dir = abspath(project_dir)
+    # We change current working directory later, so this has to be an absolute
+    # path or `None` in case that we didn't supply the `project_dir` via kwargs
+    project_dir = abspath(project_dir) if project_dir else None
```
opacam: Since we set `None` as default kwarg value for `project_dir` in method `run_pymodules_install`, this line can make it crash the build if we call the mentioned method without providing the kwarg, so we must make sure that  we call `abspath` without `None` or it will make fail the build

AndreMiras:
Here is the regression test https://github.com/kivy/python-for-android/pull/1899

PR #1899, opened 2019-06-28
Simple run_pymodules_install test, refs #1898

AndreMiras:
Regression test for the fix introduced in #1898.

### DIFF
--- a/tests/test_build.py
+++ b/tests/test_build.py
@@ -1,0 +1,25 @@
+import unittest
+
+try:
+    from unittest import mock
+except ImportError:
+    # `Python 2` or lower than `Python 3.3` does not
+    # have the `unittest.mock` module built-in
+    import mock
+from pythonforandroid.build import run_pymodules_install
+
+
+class TestBuildBasic(unittest.TestCase):
+
+    def test_run_pymodules_install_optional_project_dir(self):
+        """
+        Makes sure the `run_pymodules_install()` doesn't crash when the
+        `project_dir` optional parameter is None, refs #1898
+        """
+        ctx = mock.Mock()
+        modules = []
+        project_dir = None
+        with mock.patch('pythonforandroid.build.info') as m_info:
+            assert run_pymodules_install(ctx, modules, project_dir) is None
+        assert m_info.call_args_list[-1] == mock.call(
+            'No Python modules and no setup.py to process, skipping')


AndreMiras:
The aim is definitely not to deeply test anything, but to at least check that the bug address in #1898 is now covered by a simple test

tests/test_build.py
```diff
@@ -0,0 +1,23 @@
+try:
+    from unittest import mock
+except ImportError:
+    # `Python 2` or lower than `Python 3.3` does not
```
inclement: This reminds me we still need to decide when p4a drops python2 support. Soon would probably be good, it's especially annoying nowadays to do python2 compatibility stuff when it will be necessary for only a short time.

tests/test_build.py
```diff
@@ -0,0 +1,23 @@
...
+    # have the `unittest.mock` module built-in
+    import mock
+from pythonforandroid.build import run_pymodules_install
+
+
+class TestBuildBasic:
```
opacam: Is there any reason to non inherit from `unittest.TestCase`?

I think that inheriting from `unittest.TestCase` has several advantages:
  -  Allows some ides to detect the `unittest class` and you can run it from the ide with a couple of mouse clicks (`pycharm` case...probably `eclipse` too...but has been years since I used it...so I'm not sure of this last one)
  -  Allows us to use the `assert methods of unit.TestCase`, which increases the readability

Also I may add that the class `unittest.TestCase` has been made specially to deal with tests, I don't see the reason of not using it...am I missing something?

tests/test_build.py
AndreMiras: Completely agree. Also I can't wait to use f-strings and annotations :smile

opacam: Yes, I'm also waiting for python 3's `f-strings`, but this will require `python >= 3.6`...**I have no problem with that**

**As a side note:** we could already use python 3 `f-strings` but now it would require the install of some extra python package like [fmt](https://github.com/damnever/fmt) (I've been tempted of doing this in some prs :laughing:)

tests/test_build.py
AndreMiras: Yes the reason is, don't use something if you don't need it, as simple as that :smile:
Also known as YAGNI https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it
I mean we use `pytest` from `tox` which advocates for light code with no unnecessary boiler plate. https://docs.pytest.org/en/latest/assert.html
> pytest has support for showing the values of the most common subexpressions including calls, attributes, comparisons, and binary and unary operators. (...). This allows you to use the idiomatic python constructs without boilerplate code while not losing introspection information.

I hope nowadays IDE can work with pytest, but I don't know because I've never used any.

As for the readability I actually think `assert` is far more readable than `self.assertSomething`.
The human eye just has to scan for `assert` it's always this keyword with always a space straight after. So no silly camelCase (shame on you Python!), no `self.` and no variations (assertEqual, assertTrue, assertIsNone, assertIsNotNone...). So I honestly thing the core unittest module makes test less readable than going straight with pytest and plan `assert`.

Nevertheless, I've added the inheritance boilerplate for you, I know it's gonna take time to convince people, so let's try one step at a time :smile:

PR #1901, opened 2019-06-30
Add documentation: testing a pull request

opacam:
Due to the **lack of documentation on how to test a pull request**, sometimes we are enforced to fill pull requests with support comments about how to do the testing or we directly don't answer (to not fill the pr with unrelated information), yeah...I'm guilty of that in #1722...sorry :peace_symbol: . So to remediate this issue, I created this pull request to solve this once and for all, but guys, I need your help in here, because each one of us can have his own method, so it would be great to share our testing methods and write them in the docs, for future reference and to redirect users in case that we need to do so with a minimal interference possible for the affected pull request.

I put three methods in the initial documentation, but I need that all of you review this very carefully because I could miss something (actually I think that I'm missing something...) or it may exist a better ways to accomplish this...probably...which I'll be happy to learn :smile:...so I labeled this as **WIP**, **until** we make sure that **this documentation is well enough to be published and contemplates the best methods** to do the `pull requests` testing

### DIFF
--- a/doc/source/testing_pull_requests.rst
+++ b/doc/source/testing_pull_requests.rst
@@ -1,0 +1,226 @@
+Testing an python-for-android pull request
+==========================================
+
+In order to test a pull request, we recommend to consider the following points:
+
+  #. of course, check if the overall thing makes sense
+  #. is the CI passing? if not what specifically fails
+  #. is it working locally at compile time?
+  #. is it working on device at runtime?
+
+This document will focus on the third point:
+`is it working locally at compile time?` so we will give some hints about how
+to proceed in order to create a local copy of the pull requests and build an
+apk. We expect that the contributors has enough criteria/knowledge to perform
+the other steps mentioned, so let's begin...
+
+To create an apk from a python-for-android pull request we contemplate three
+possible scenarios:
+
+  - using python-for-android commands directly from the pull request files
+    that we want to test, without installing it (the recommended way for most
+    of the test cases)
+  - installing python-for-android using the github's branch of the pull request
+  - using buildozer and a custom app
+
+We will explain the first two methods using one of the distributed
+python-for-android test apps and we assume that you already have the
+python-for-android dependencies installed. For the `buildozer` method we also
+expect that you already have a a properly working app to test and a working
+installation/configuration of `buildozer`. There is one step that it's shared
+with all the testing methods that we propose in here...we named it
+`Common steps`.
+
+
+Common steps
+^^^^^^^^^^^^
+The first step to do it's to get a copy of the pull request, we can do it of
+several ways, and that it will depend of the circumstances but all the methods
+presented here will do the job, so...
+
+Fetch the pull request by number
+--------------------------------
+For the example, we will use `1901` for the example) and the pull request
+branch that we will use is `feature-fix-numpy`, then you will use a variation
+of the following git command:
+`git fetch origin pull/<#>/head:<local_branch_name>`, eg.::
+
+    .. codeblock:: bash
+
+        git fetch upstream pull/1901/head:feature-fix-numpy
+
+.. note:: Notice that we fetch from `upstream`, since that is the original
+          project, where the pull request is supposed to be
+
+.. tip:: The amount of work of some users maybe worth it to add his remote
+       to your fork's git configuration, to do so with the imaginary
+       github user `Obi-Wan Kenobi` which nickname is `obiwankenobi`, you
+       will do::
+
+          .. codeblock:: bash
+
+              git remote add obiwankenobi https://github.com/obiwankenobi/python-for-android.git
+
+       And to fetch the pull request branch that we put as example, you
+       would do::
+
+          .. codeblock:: bash
+
+              git fetch obiwankenobi
+              git checkout obiwankenobi/feature-fix-numpy
+
+
+Clone the pull request branch from the user's fork
+--------------------------------------------------
+Sometimes you may prefer to use directly the fork of the user, so you will get
+the nickname of the user who created the pull request, let's take the same
+imaginary user than before `obiwankenobi`::
+
+    .. codeblock:: bash
+
+        git clone -b feature-fix-numpy \
+            --single-branch \
+            https://github.com/obiwankenobi/python-for-android.git \
+            p4a-feature-fix-numpy
+
+Here's the above command explained line by line:
+
+- `git clone -b feature-fix-numpy`: we tell git that we want to clone the
+  branch named `feature-fix-numpy`
+- `--single-branch`: we tell git that we only want that branch
+- `https://github.com/obiwankenobi/python-for-android.git`: noticed the
+  nickname of the user that created the pull request: `obiwankenobi` in the
+  middle of the line? that should be changed as needed for each pull
+  request that you want to test
+- `p4a-feature-fix-numpy`: the name of the cloned repository, so we can
+  have multiple clones of different prs in the same folder
+
+.. note:: You can view the author/branch information looking at the
+          subtitle of the pull request, near the pull request status (expected
+          an `open` status)
+
+Using python-for-android commands directly from the pull request files
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Enter inside the directory of the cloned repository in the above
+  step and run p4a command with proper args, eg::
+
+    .. codeblock:: bash
+
+        cd p4a-feature-fix-numpy
+        PYTHONPATH=. python3 -m pythonforandroid.toolchain apk \
+            --private=testapps/testapp_sqlite_openssl \
+            --dist-name=dist_test_app_python3_libs \
+            --package=org.kivy \
+            --name=test_app_python3_sqlite_openssl \
+            --version=0.1 \
+            --requirements=requests,peewee,sdl2,pyjnius,kivy,python3 \
+            --ndk-dir=/media/DEVEL/Android/android-ndk-r20 \
+            --sdk-dir=/media/DEVEL/Android/android-sdk-linux \
+            --android-api=27 \
+            --arch=arm64-v8a \
+            --permission=INTERNET \
+            --debug
+
+Things that you should know:
+
+
+    - The example above will build an testapp we will make use of the files of
+      the testapp named `testapp_sqlite_openssl.py` but we don't use the setup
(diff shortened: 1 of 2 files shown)


doc/source/testing_pull_requests.rst
```diff
@@ -0,0 +1,178 @@
...
+    .. codeblock:: bash
+
+        cd testapps
+
+- Run the build of the apk via the freshly installed copy of python-for-android
+  by running a similar command than below::
+
+    .. code-block:: bash
+
+        python3 setup_testapp_python3_sqlite_openssl.py apk \
```
opacam: @inclement, I need your help in here, I learned this second method from you and I have my IDE configured to do the tests without installing python-for-android as described in this method...so this should work without any p4a install right? Well...when I try to run the command from a terminal and without p4a installed, it fails with: `error: invalid command 'apk'` so I have the feeling that I miss something...what I'm missing?

inclement: You do need to install python-for-android for this to work - if you look in the p4a setup.py, the `apk` argument to setup.py is one of the things that gets installed as part of p4a.

doc/source/testing_pull_requests.rst
```diff
@@ -0,0 +1,178 @@
...
+^^^^^^^^^^^^
+
+- open a terminal, create a test directory and enter inside it, eg.::
+
+    .. codeblock:: bash
+
+        mkdir p4a_pull_requests
+        cd p4a_pull_requests
+
+- clone the pull request you wanna to test, let's imagine that you wanna to
```
AndreMiras: s/wanna to/want to/

doc/source/testing_pull_requests.rst
```diff
@@ -0,0 +1,178 @@
...
+
+Common steps
+^^^^^^^^^^^^
+
+- open a terminal, create a test directory and enter inside it, eg.::
+
+    .. codeblock:: bash
+
+        mkdir p4a_pull_requests
+        cd p4a_pull_requests
```
AndreMiras: Minor: trivial, I would skip this part

doc/source/testing_pull_requests.rst
```diff
@@ -0,0 +1,178 @@
...
+
+- clone the pull request you wanna to test, let's imagine that you wanna to
+  test a pull request whose branch is named `feature-fix-numpy` and the github
+  nickname of the user who created the pull request is named `obiwankenoby`::
```
AndreMiras: Minor: Obi-Wan Kenobi is actually spelled with a `i` not a `y` :smile:

doc/source/testing_pull_requests.rst
```diff
@@ -0,0 +1,178 @@
...
+
+    - `git clone -b feature-fix-numpy`: we tell git that we want to clone the
+      branch named `feature-fix-numpy`
+    - `--single-branch`: we tell git that we only want that branch
+    - `https://github.com/obiwankenoby/python-for-android.git`: noticed the
+      nickname of the user that created the pull request: `obiwankenoby` in the
+      middle of the line? that should be changed as needed for each pull
+      request that you wanna to test
+    - `p4a-feature-fix-numpy`: the name of the cloned repository, so we can
+      have multiple clones of different prs in the same folder
```
AndreMiras: I actually assumed this part is known by contributors, so I would also have skipped that part. But it's good that you have it there then.
There's also a "shortcut" to actually fetch the PR given its number:
```sh
git fetch origin pull/<#>/head:<local_branch_name
```
For instance for yours:
```
git fetch upstream pull/1901/head:feature-docs-test-pr
remote: Enumerating objects: 6, done.
remote: Counting objects: 100% (6/6), done.
remote: Compressing objects: 100% (2/2), done.
remote: Total 6 (delta 4), reused 6 (delta 4), pack-reused 0
Unpacking objects: 100% (6/6), done.
From https://github.com/kivy/python-for-android
 * [new ref]           refs/pull/1901/head -> feature-docs-test-pr
[git checkout feature-docs-test-pr
Switched to branch 'feature-docs-test-pr'
```
The workflow I personally use most of the time is to add the remote and checkout the branch. So for yours:
```sh
git remote add opacam https://github.com/opacam/python-for-android.git
git fetch opacam
git checkout opacam/feature-docs-test-pr
```

doc/source/testing_pull_requests.rst
```diff
@@ -0,0 +1,178 @@
...
+Installing python-for-android using the github's branch of the pull request
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+- Enter inside the directory of the cloned repository mentioned in
+  `Common steps` and install it via pip, eg.::
+
+    .. codeblock:: bash
+
+        cd p4a-feature-fix-numpy
+        pip3 install . --upgrade --user
```
AndreMiras: Minor: we can also suggest to use a virtualenv for not messing up with user ~/.local

doc/source/testing_pull_requests.rst
opacam: Aaa, thanks for the clarification...with your info I just found why I can do this without install p4a from my ide

Seeing my forked python-for-android files, I have an directory that we don't have committed `python_for_android.egg-info` (because at some point I installed p4a from my fork and the `egg folder` was generated... I never took care to delete that folder)...that folder seems that is the key, because it contains the `entrypoints.txt` ...the mentioned `apk` is in there...If I remove this file I got the exact error I mentioned before...

So i think that we just found another test method without install...I will try to write this one...

thanks @inclement :smile:

doc/source/testing_pull_requests.rst
```diff
@@ -0,0 +1,178 @@
...
+
+        - `--ndk-dir`: An absolute path to your android's NDK dir
+        - `--sdk-dir`: An absolute path to your android's SDK dir
+
+.. warning:: Once you finish the pull request tests remember to go back to the
+             master or develop versions of python-for-android, since you just
+             installed the python-for-android files of the `pull request`
+
+Using buildozer with a custom app
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
AndreMiras: I actually often use a method pretty close to that one. The method is terribly wrong so I won't share it in the docs.
I edit: `~/.local/lib/python3.6/site-packages/buildozer/targets/android.py` and replace the clone command by the fork.
BTW I think at some point buildozer should make it possible from the spec file, to not only chose the branch, but also the fork.

doc/source/testing_pull_requests.rst
opacam: Oh :scream:, How the hell I spelled wrong the **Obi-Wan**'s last name?
forgive me for this, this is major not minor :laughing:
I will definitively change that...

**¡¡¡Thanks!!!**

and

**May the Force be with you** :wink:

doc/source/testing_pull_requests.rst
opacam: Oh yeah, I will write that :wink:

doc/source/testing_pull_requests.rst
opacam: Ok, I've been investigating a little this thing, and I can say that I realized that, some time ago, I hacked my `site-packages` for the python version that I usually use to test p4a  (**I didn't remember that**)...so it's really necessary to install the package. So better that we stick  to the words I wrote for this method, but I will explain what the hell I did to make it work, which I don't think that we should put in the documentation, but still maybe someone want to do this:

  - I installed the python-for-android package from my development fork of p4a
  - I removed the `pythonforandroid` folder from the installed package in site-packages
  - I created a symlink who points to my `pythonforandroid` folder of my fork of p4a

Doing it this way I can use this method without doing any install (**of course...it's always installed**)...I just switch to the branch I want to test and execute the command for the testapp. So this method it's interesting for development (the best way to apply this is to have an python's virtualenv for only p4a, so you don't mess up with the system's python)

doc/source/testing_pull_requests.rst
```diff
@@ -0,0 +1,226 @@
...
+      by editing this file
+    - be sure to at least edit the following arguments when running the above
+      command, since the default set in there it's unlikely that match your
+      installation:
+
+        - `--ndk-dir`: An absolute path to your android's NDK dir
+        - `--sdk-dir`: An absolute path to your android's SDK dir
+
+.. tip:: if you don't want to mess up with the system's python, you could do
+          the same steps but inside a virtualenv
```
opacam: @AndreMiras, here you have the virtualenv suggestion :+1: 

**Thanks**

doc/source/testing_pull_requests.rst
opacam: > BTW I think at some point buildozer should make it possible from the spec file, to not only chose the branch, but also the fork.

Yeah...you are right...this should be implemented in buildozer :+1:

doc/source/testing_pull_requests.rst
```diff
@@ -0,0 +1,226 @@
+Testing an python-for-android pull request
+==========================================
+
+In order to test a pull request, we recommend to consider the following points:
+
+  #. of course, check if the overall thing makes sense
+  #. is the CI passing? if not what specifically fails
+  #. is it working locally at compile time?
+  #. is it working on device at runtime?
```
opacam: @AndreMiras, I also adapted your approval comment in here, as the introduction of the whole thing :wink:

let me know if it's ok

**Thanks!!** :smile:

doc/source/testing_pull_requests.rst
AndreMiras: Oh yes I've used the symlink approach quite a lot also. Pretty useful, but indeed better leaving this hack undocumented :smile:

doc/source/testing_pull_requests.rst
```diff
@@ -0,0 +1,226 @@
...
+imaginary user than before `obiwankenobi`::
+
+    .. codeblock:: bash
+
+        git clone -b feature-fix-numpy \
+            --single-branch \
+            https://github.com/obiwankenobi/python-for-android.git \
+            p4a-feature-fix-numpy
+
+For what it's worth it, about the command above explained line by line:
```
AndreMiras: Minor: something looks off in this sentence. Maybe we could try something like:
`Here's the above command explained line by line:`

opacam: **Thanks**, this is already fixed :wink:

AndreMiras:
Is this still WIP? For me it's OK for a merge :+1:

PR #1911, opened 2019-07-10
Feature gitignore additions

opacam:
We can generate some files/folders, under some circumstances, which we don't want that git tracks them, so here we add some entries to `.gitignore`, to make sure that git ignores them:
  - files/folders generated by our unit tests
  - files/folders generated by our coverage
  - folders generated by our testapps (in case that we build some testapp directly from inside the source code folder)
  - file `.directory` auto generated by `Dolphin`, the KDE's file manager, when browsing a folder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ coverage.xml
 
 # testapp's build folder
 testapps/build/
+
+# Dolphin (the KDE file manager autogenerates the file `.directory`)
+.directory
--- a/.gitignore
+++ b/.gitignore
@@ -21,4 +21,14 @@ __pycache__/
 
 #idea/pycharm
 .idea/
-.tox
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+*.cover
+.pytest_cache/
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -21,4 +21,20 @@ __pycache__/
...
+nosetests.xml
+coverage.xml
+*.cover
+.pytest_cache/
+
+# testapp's build folder
+testapps/build/
+
+# Dolphin (the KDE file manager autogenerates the file `.directory`)
+.directory
```
AndreMiras: are you trying to sneak your file manager leftover to this PR :trollface:

.gitignore
```diff
@@ -21,4 +21,20 @@ __pycache__/
...
-.tox
\ No newline at end of file
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.coverage
+.coverage.*
+.cache
+nosetests.xml
```
AndreMiras: we're not using this guy, are we?

.gitignore
opacam: You caught me :laughing:, anyway...probably will be useful for someone else don't you think :wink:?

.gitignore
opacam: You are right, we are not using this, so, since this is already merged,  I made the correction in (#1915)

Thanks @AndreMiras!!

Issue #1905, opened 2019-07-01
Proposal: drop CrystaX support and code base


AndreMiras:
Since python3 works on pure Android NDK and p4a CrystaX is not really tested/maintained, maybe it's time for us to completely drop it from the code base.
What do you guys think?

opacam:
Yes, totally agree :+1:
We should get rid of  **python3crystax**...
**IMHO**, right now, I think that it only adds an unnecessary complexity

PR #1913, opened 2019-07-10
Drop CrystaX support and code base

opacam:
There is an open  issue, a proposal to drop CrystaX support (#1905), so this PR takes care of this...in case that the team wants to take this road, we will see in here I guess..:wink:

~~**Note:** labeled `WIP` because the tests must be updated~~


doc/source/buildoptions.rst
```diff
@@ -33,25 +33,20 @@ e.g. ``--requirements=python3``.
...
+
+python-for-android originally supported Python 3 using the CrystaX NDK.  Since
+we have a working python3 recipe, we don't support CrystaX NDK anymore. If you
+were using `python3crystax`, we recommend to give it a try to the new `python3`
+recipe.
+
+.. note:: Since we don't support `python3crystax` anymore, the old instructions
+          has been removed from here. If you, still have the need to make use
+          of this old recipe, you should do it with an old `python-for-android`
+          release. Probably, a got starting point would be `version 0.7.0
```
AndreMiras: Do you meen "a good starting point? `s/a got/a good/

doc/source/docker.rst
```diff
@@ -17,12 +17,11 @@ already have Docker preinstalled and set up.
...
-   dependencies. Within the custom diff made by building the distribution
-   there is another big chunk of space eaten. The very basic stuff such as
-   a distribution with: CPython 3, setuptools, Python for Android ``android``
-   module, SDL2 (+ deps), PyJNIus and Kivy takes almost 13 GB. Check your free
-   space first!
+   Within the Docker image there is Android SDK and NDK + various dependencies.
+   Within the custom diff made by building the distribution there is another
+   big chunk of space eaten. The very basic stuff such as a distribution with:
+   CPython 3, setuptools, Python for Android ``android`` module, SDL2 (+ deps),
+   PyJNIus and Kivy takes almost 13 GB. Check your free space first!
```
AndreMiras: Since CrystaX is gone from the Dockerfile I think the image size is way less.
DockerHub shows 2 GB.
https://cloud.docker.com/u/kivy/repository/docker/kivy/python-for-android/tags

pythonforandroid/recipes/genericndkbuild/__init__.py
```diff
@@ -7,7 +7,7 @@ class GenericNDKBuildRecipe(BootstrapNDKRecipe):
     version = None
     url = None
 
-    depends = [('python2', 'python3', 'python3crystax')]
+    depends = [('python2', 'python3')]
```
AndreMiras: As I understood python2 and 3 are implicitly in the `depends` list aslso if we remove then correct?

AndreMiras:
Awesome start :clap: can't wait to see it goes :smile:

pythonforandroid/recipes/genericndkbuild/__init__.py
opacam: mmm...I don't think so, those are only implicit in any recipe that inherits of `PythonRecipe`, but not for `BootstrapNDKRecipe`...or maybe I miss something?

pythonforandroid/recipe.py
```diff
@@ -725,14 +725,44 @@ class PythonRecipe(Recipe):
     This is almost always what you want to do.'''
 
     setup_extra_args = []
-    '''List of extra arugments to pass to setup.py'''
+    '''List of extra arguments to pass to setup.py'''
+
+    depends = [('python2', 'python3')]
+    '''
+    .. note:: it's important to keep this depends as a class attribute, outside
```
JonasT: The note is helpful, but it seems a bit overly long. I am not trying to be nitpicky, but it is actually a problem that when all explanatory comments get too long one tends to not read them. So I recommend shortening it down to:

```
    .. note:: it's important to keep this depends as a class attribute outside
              `__init__` because sometimes we only initialize the class, so
              the `__init__` call won't be called and the deps would be missing
              (which breaks the dependency graph computation)
```

pythonforandroid/recipe.py
```diff
@@ -725,14 +725,44 @@ class PythonRecipe(Recipe):
...
+        ):
+            # we overwrote `depends` in inherited recipe, so we must add it
+            # again the python versions as dependencies, but we only do this in
+            # case that the sub classes recipe does not contain any python
+            # version as dependency because it may be some recipes only
+            # compatible with a single version of python
+            depends = self.depends
+            depends.append(('python2', 'python3'))
+            depends = list(set(depends))
+            self.depends = depends
```
JonasT: This looks like the wrong place to do it to me. I think this should be removed and handled in `pythonforandroid/graph.py` instead, because this would also ensure that not calling `super().__init__()` is never a problem. I it is "obvious" that any recipe is meant to use python so semantically I don't think it is important or relevant to enforce this here, and it's completely fine to fix this in graph.py since it maintains the spirit of the class if it's not in here (if that explanation makes any sense to you)

pythonforandroid/recipe.py
```diff
@@ -725,14 +725,44 @@ class PythonRecipe(Recipe):
...
+    '''
+    .. note:: it's important to keep this depends as a class attribute, outside
+              `__init__` because, sometimes, we only initialize the object, so
+              the `__init__` call it won't be called, which will lead to not
+              have the python versions as a dependencies and it will cause a
+              tremendous `test_graph` error (difficult to track) and also, the
+              build order for dependencies will not be computed as expected (if
+              computed...). So be very careful with this line!!
+
+    .. warning:: this `depends` may be overwrote in inherited classes of
```
JonasT: I would shorten this to:

```
 .. warning:: don't forget to call `super().__init__()` in any recipe's `__init__`, or
              otherwise it may not be ensured that it depends on python2 or python3
              which can break the dependency graph
```

pythonforandroid/recipe.py
```diff
@@ -725,14 +725,44 @@ class PythonRecipe(Recipe):
...
-        depends = list(set(depends))
-        self.depends = depends
+        if not any(
+            [
+                d
+                for d in {'python2', 'python3', ('python2', 'python3')}
+                if d in self.depends
+            ]
+        ):
+            # we overwrote `depends` in inherited recipe, so we must add it
```
JonasT: I would repharse this comment to:

```
# We ensure here that the recipe depends on python even it overrode `depends`.
# We only do this if it doesn't already depend on any python,
# since some recipes intentionally don't depend on/work with all python variants
```

JonasT:
Looks good to me in overall, I suggested a few minor things but don't consider them blocking on the merge.

I suggest we should merge this soon, because it looks like this may conflict with things and crystax really is annoying to keep around

doc/source/buildoptions.rst
```diff
@@ -33,25 +33,20 @@ e.g. ``--requirements=python3``.
...
-This uses the prebuilt Python from the `CrystaX NDK
-<https://www.crystax.net/android/ndk>`__, a drop-in replacement for
-Google's official NDK which includes many improvements. You
-*must* use the CrystaX NDK 10.3.0 or higher when building with
-python3. You can get it `here
-<https://www.crystax.net/en/download>`__.
+~~~~~~~~~~~~~~~
+
+python-for-android originally supported Python 3 using the CrystaX NDK.  Since
+we have a working python3 recipe, we don't support CrystaX NDK anymore. If you
```
inclement: I would be more direct with this kind of doc, you can just write 

> python-for-android no longer supports building for Python 3 using the CrystaX NDK. Instead, use the `python3` recipe, which can be built using the normal Google NDK.

> The last python-for-android version supporting CrystaX was 0.7.0.

No need for this to hold up merging the PR, we can improve the doc later (there are plenty of parts that I wrote that also need improvement), just noting as something I've been thinking about.

inclement:
Not merging because I didn't check the status of the other comments, but it's fine to merge for me.

opacam:
Thanks @JonasT, @inclement and @AndreMiras,

 for your reviews and text corrections :+1:

All your suggestions as been applied as you wanted, so I think that we have this :smile:

inclement:
Thanks @opacam!

PR #1914, opened 2019-07-10
Fix crash when guessing Bootstrap (expand_dependencies)

opacam:
When a pure python package it's supplied inside the dependency list (for the method `expand_dependencies`), we will get a crash because we don't have a recipe for it, unless we contemplate that situation

**Note:** We missed that because we haven't a pure python package in our CI` builds for the [testapps](https://github.com/kivy/python-for-android/blob/develop/.travis.yml#L73-L81)

**As a side note:** The `expand_dependencies` method was recently introduced in PR #1887


JonasT:
Oopsie, did I break things again :grimacing: :grimacing: :grimacing: the fix looks exactly like I would have done it, so good work :heart: and sorry :joy:

PR #1922, opened 2019-07-15
Add optional android fileprovider.

Fak3:
https://developer.android.com/reference/android/support/v4/content/FileProvider

If the option `--fileprovider-paths=./mypaths.xml` is given, then this xml file is
copied to apk `res/xml/file_paths.xml` and <provider> is added to the AndroidManifest.xml


sanderland:
Has this been ignored and abandoned, or has anything similar made its way in? Struggling to have my kivy app open an external app.

kollivier:
Flagging that we are using this support as well and it would be nice if we could get it merged! :)

Ham-Merhead:
I'm trying to share a file on Android 10, which seems pretty basic functionality.

But permissions means I need to create a 'content' Uri rather than a 'file' Uri.

Seems the only way to do this is via a ContentProvider such as FileProvider, but this requires androidx.

So
1) This is really important, file 'share' is basic functionality. Thank you.
2) Consider defaults for the xml files, a Python user is here because they don't want to think about that stuff.
3)  The Android docs say ContentProvider must be sub classed, but anything I try to do in pyjnius wont be seen in Java. And if I try to instance ContentProvider, I get:
JNI DETECTED ERROR IN APPLICATION: can't make objects of type android.content.ContentProvider: 0x712ee630
So a minimal Python/pyjnius example would be really helpful.

Thanks again.

AndreMiras:
@Fak3  thank you for the pull request and sorry that it wasn't reviewed earlier.
Thanks also to the other to show interest regarding the feature.
Before we can merge it could you address the following?
- rebase to recent upstream/develop changes
- leverage `android.gradle_dependencies` refs https://github.com/kivy/python-for-android/pull/2200#issuecomment-637435022
- write unit tests

Ideally squash the commits after addressing so the PR is still one commit

wajdyz:
Seems to be a must API 30 onwards.
More restrictions by Google now on using storage

antocuni:
Unfortunately the original branch has conflicts with the `develop` branch of p4a. I rebased it on top of the latest version (6ed70232) and published the rebased branch here:
https://github.com/antocuni/python-for-android/tree/fileprovider-rebased

See also this example to see it in action:
https://github.com/antocuni/plyer_camera_example

Zen-CODE:
@Fak3 Any word on resolving the conflicts? Would be nice to get this in.....

Android-for-Python:
There is another way to generate a uri : `uri = context.getContentResolver().insert()`

Issue #1920, opened 2019-07-14
Allow to reuse existing dist without rebuild


Fak3:
Currently there is no way to tell `p4a apk` to use existing dist without rebuilding it. The `--no-allow-replace-dist` flag results in failure:

> [INFO]:    No existing dists meet the given requirements!
> [ERROR]:   Build failed: Asked for dist with name paradox with recipes (python3, kivy_myasync, openssl, sqlite3, pillow, pytz, sdl2, ansimarkup, app-state, better-exceptions-fork, certifi, chardet, colorama, docutils, getinstance, h11, h2, hpack, http3, hyperframe, idna, lockorator, loguru, minidjango, plyer, Pygments, python-dateutil, requests, requests-async, rfc3986, six, sniffio, sqlparse, urllib3) and NDK API None, but a dist with this name already exists and has either incompatible recipes (freetype, hostpython3, jpeg, libffi, openssl, png, sdl2_image, sdl2_mixer, sdl2_ttf, sqlite3, python3, sdl2, pytz, setuptools, six, idna, pillow, pyjnius, requests, android, kivy_myasync, minidjango, docutils, http3, app-state, chardet, requests-async, sqlparse, hyperframe, certifi, plyer, ansimarkup, sniffio, pygments, colorama, rfc3986, urllib3, python-dateutil, h2, getinstance, loguru, better-exceptions-fork, lockorator, h11, hpack) or NDK API 21

The failure is due to the p4a wrongly assumes the dist requirements aren't matching: #1882

It would be nice to have a way to tell p4a to reuse the existing dist even if it thinks that existing dist does not meet dependencies.

inclement:
@Fak3 Thanks for these reports and sorry for not responding sooner, I'm trying to catch up with it now.

Issue #1917, opened 2019-07-12
Changing android.arch or android.api does not trigger the appropriate re-build.


brentpicasso:
If android.arch or android.api is changed and build is run again, it does not pick up the changes. 

a clean must be performed before the changes are reflected. 

Expected: the changes should be detected and appropriate artifacts should be re-built.

inclement:
Thanks for the report. For general reference if anyone is looking at this:

- android.arch should go in the dist_info.json and used to separate dists. In fact, we should automatically create multiple dists with the arch in the name
- I think the android.api thing is just a bug, because the android API is hardcoded in one of the config files within the dist. There's an old issue about this. I don't think this is actually necessary, so it should be possible to build for multiple android apis from a single dist.

PR #1926, opened 2019-07-20
[WIP] Enhance distribution check when reusing distributions and...

opacam:
**...Skip some unneeded rebuilds**

The main reason to do this, it's because we usually have to deal with issues related with old cached builds as well as the reuse of a distributions it seems to not work as expected, so this PR is intended to deal with this.

**Things made in here:**
  - Save into dist_info.json file more information:
      + android api
      + archs
      + p4a version
      + recipes version
  - Perform more checks when we find a dist with the same name:  
      + check if android api is the same as the one we request
      + check if distribution's arch is the same that we are requesting
      + check if p4a version we are using is the same as the one we used for the built distribution (and if not, force a cleanup of the build directory)
       + check recipes version and remove old build in the case that we detect an outdated recipe version
  - Fix Recipes.clean_build, so we can remove old builds
  - Implement BootstrapNDKRecipe.should_build , so we can remove old bootstrap builds (Here I make use of `ndk-build clean` which it seems to remove all bootstrap builds and libraries regarding that which recipe we are targeting...:thinking:...I suppose it's fine for now (it simplifies a little the thing)
  - ~~Implement `GuestPythonRecipe.should_build` so we can skip python rebuild~~ (removed commit since it's already done at #1951)
  - change bootstrap build directory (to ensure that when we change arch and we already have build with another arch, we don't get conflicts)
  - detect when a PythonRecipe needs a rebuild, so we remove our `python-installs` directory (yeah, it's a little drastic this point but it would do the job and simplifies the code)
  - remove recipe's generated libraries from libs_collection when we detected a changed library recipe

**Things to consider...I may forgot something in here:**
  - The build/dist directories will grow up considerably
  - ~~we never remove recipe's generates libraries from `libs_collection` when we detected a changed recipe (we only remove the build directory...so this may cause some troubles in case that `Recipe.should_build` checks for the library into that folder)...mmm~~
  - buildozer need to be updated to work with this (since the dist name it will be modified with a suffix)

**Closes:** #1920 and #1917

**I leave this `WIP` because:**
  - we may want to discuss the whole idea...we could simply force a full clean in each build (sure that it would be simpler)
  - There are some parts of the code that may be improved  or even moved from one place to another...I have to review it...but if anyone sees something, I'll be glad to discuss/fix it
  - I tested different situations and this thing seems to work, but I may forgot something...so better other eyes take look on this
  -  we may want to split this thing into multiple PRs...sometimes things make sense in one PR, but when PRs go big it complicates things for all of us...should we have a workflow to deal with this? Maybe the [stacked pull requests](https://graysonkoonce.com/stacked-pull-requests-keeping-github-diffs-small/) method could be a solution?
     + **As a side note:** I think this method could only be used for those ones with write permissions to repo, since needs to create a new upstream branch plus it requires to merge things in `reverse` order)

### DIFF
--- a/pythonforandroid/distribution.py
+++ b/pythonforandroid/distribution.py
@@ -99,7 +99,9 @@ class Distribution(object):
 
         # 0) Check if a dist with that name already exists
         if name is not None and name:
-            possible_dists = [d for d in possible_dists if d.name == name]
+            possible_dists = [
+                d for d in possible_dists if d.name.startswith(name)
+            ]
             if possible_dists:
                 name_match_dist = possible_dists[0]
 
@@ -224,7 +226,10 @@ class Distribution(object):
                 i += 1
             name = filen.format(i)
 
-        dist.name = name
+        # we add a suffix to our dist name, so we avoid to
+        # overwrite them when building for different arch
+        suffix_arch = f"_{{{'·'.join([arch.arch for arch in ctx.archs])}}}"
+        dist.name = name + suffix_arch
         dist.dist_dir = join(ctx.dist_dir, dist.name)
         dist.recipes = recipes_with_version
         dist.archs = req_archs
--- a/pythonforandroid/distribution.py
+++ b/pythonforandroid/distribution.py
@@ -20,6 +20,7 @@ class Distribution(object):
 
     name = None  # A name identifying the dist. May not be None.
     needs_build = False  # Whether the dist needs compiling
+    needs_clean_build = False  # Whether the build needs a full clean
     url = None
     dist_dir = None  # Where the dist dir ultimately is. Should not be None.
     ndk_api = None
@@ -129,6 +130,11 @@ class Distribution(object):
                  not require_perfect_match)):
                 info_notify('{} has compatible recipes, using this one'
                             .format(dist.name))
+                if dist.p4a_version != __version__:
+                    # We build this dist with a different version of p4a, so
+                    # we mark it as a dist that requires a clean build environ
+                    # (to avoid old cached builds issues)
+                    dist.needs_clean_build = True
                 return dist
 
         assert len(possible_dists) < 2
@@ -209,7 +215,9 @@ class Distribution(object):
                 dist.dist_dir = folder
                 dist.needs_build = False
                 dist.recipes = dist_info['recipes']
-                for entry in {'archs', 'ndk_api', 'android_api'}:
+                for entry in {
+                    'archs', 'ndk_api', 'android_api', 'p4a_version'
+                }:
                     setattr(dist, entry, dist_info.get(entry, None))
                     if entry not in dist_info:
                         warning(
--- a/pythonforandroid/distribution.py
+++ b/pythonforandroid/distribution.py
@@ -3,8 +3,11 @@ import glob
 import json
 
 from pythonforandroid import __version__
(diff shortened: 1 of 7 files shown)


inclement:
Thanks @opacam, I have no immediate objection but I won't have time to look at this properly until at least next week.

JonasT:
I actually support the idea of forcing clean builds, and adding mechanisms to speed up build in general instead. E.g. separate distribution-independent download caching, promoting the use of ccache more which works fine, and other such stuff.

The build just seems too complex to be worth caching if you throw in external packages that may have new releases, change in deps, and other things of the world changing.

So my vote would be to drop all the ways to make a distribution "unique" and cacheable, and just make a fresh one for each build (and throw the old ones away unless some `--no-cleanup-distribution` is used for debugging)

pythonforandroid/distribution.py
```diff
@@ -19,14 +23,21 @@ class Distribution(object):
...
 
     archs = []
     '''The arch targets that the dist is built for.'''
 
-    recipes = []
+    recipes = {}
+    '''A dictionary which holds recipes information. Each key is the name of
+    the recipe, and the value is recipe's version'''
+
+    recipes_to_rebuild = set()
```
AndreMiras: Minor: why not being consistent and use the literal `{}`?

pythonforandroid/distribution.py
```diff
@@ -83,23 +93,62 @@ def get_distribution(cls, ctx, name=None, recipes=[],
...
+                    dist.ndk_api != ctx.ndk_api,
+                    dist.android_api != ctx.android_api,
+                    set(dist.archs) != set(req_archs),
+                    dist.ndk_api is None,
+                ]
+            ):
                 continue
-            for recipe in recipes:
-                if recipe not in dist.recipes:
+            checked_dist_recipes = set()
```
AndreMiras: same here for string literals `{}` which by the way is slightly faster in Python (because `set` would require alias resolution), but who cares :smile:

pythonforandroid/distribution.py
opacam: I don't think that python have a literal for an empty set...`{}` this will be interpreted as  a dict, unless we put some values in the set creation...I'm right doesn't?

AndreMiras: You're completely right! And I should get more sleep :sweat_smile:

pythonforandroid/distribution.py
AndreMiras: Nevermind ref https://github.com/kivy/python-for-android/pull/1926#discussion_r308382952
Thanks @opacam

inclement:
Just noting here since I mentioned it on #1716 - I think I'd like to propose some simplifications to this PR before we merge it. In the meantime I'm working on a simpler change to support multiple arch builds for the same dist name, just to make sure this annoyance is resolved in the next release (which I'll make as soon as this change is added).

opacam:
I close this for several reasons:
  - there are merge conflicts which are hard to solve
  - when reviewing the code I find some parts that I don't like
  - hard to review this PR
  - how I deal with `android_api` is not right

But, I will create a new set of PRs to deal with the problems we have when reusing distributions, which...I hope...it will be easier/faster to review :wink:

**¡¡Thanks all for your feedback in here!!** :smile:

PR #1928, opened 2019-07-23
Basic toolchain.py unit tests

AndreMiras:
First simple set of tests for toolchain.py
Also refactors `Context.prepare_build_environment()` slightly. Splits
concerns to improve readability and ease unit testing.

### DIFF
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -27,6 +27,66 @@ from pythonforandroid.recommendations import (
     RECOMMENDED_NDK_API, RECOMMENDED_TARGET_API)
 
 
+def get_cython_path():
+    for cython_fn in ("cython", "cython3", "cython2", "cython-2.7"):
+        cython = sh.which(cython_fn)
+        if cython:
+            return cython
+    raise BuildInterruptingException('No cython binary found.')
+
+
+def get_ndk_platform_dir(ndk_dir, ndk_api, arch):
+    ndk_platform_dir_exists = True
+    platform_dir = arch.platform_dir
+    ndk_platform = join(
+        ndk_dir,
+        'platforms',
+        'android-{}'.format(ndk_api),
+        platform_dir)
+    if not exists(ndk_platform):
+        warning("ndk_platform doesn't exist: {}".format(ndk_platform))
+        ndk_platform_dir_exists = False
+    return ndk_platform, ndk_platform_dir_exists
+
+
+def get_toolchain_versions(ndk_dir, arch):
+    toolchain_versions = []
+    toolchain_path_exists = True
+    toolchain_prefix = arch.toolchain_prefix
+    toolchain_path = join(ndk_dir, 'toolchains')
+    if isdir(toolchain_path):
+        toolchain_contents = glob.glob('{}/{}-*'.format(toolchain_path,
+                                                        toolchain_prefix))
+        toolchain_versions = [split(path)[-1][len(toolchain_prefix) + 1:]
+                              for path in toolchain_contents]
+    else:
+        warning('Could not find toolchain subdirectory!')
+        toolchain_path_exists = False
+    return toolchain_versions, toolchain_path_exists
+
+
+def get_targets(sdk_dir):
+    if exists(join(sdk_dir, 'tools', 'bin', 'avdmanager')):
+        avdmanager = sh.Command(join(sdk_dir, 'tools', 'bin', 'avdmanager'))
+        targets = avdmanager('list', 'target').stdout.decode('utf-8').split('\n')
+    elif exists(join(sdk_dir, 'tools', 'android')):
+        android = sh.Command(join(sdk_dir, 'tools', 'android'))
+        targets = android('list').stdout.decode('utf-8').split('\n')
+    else:
+        raise BuildInterruptingException(
+            'Could not find `android` or `sdkmanager` binaries in Android SDK',
+            instructions='Make sure the path to the Android SDK is correct')
+    return targets
+
+
+def get_available_apis(sdk_dir):
+    targets = get_targets(sdk_dir)
+    apis = [s for s in targets if re.match(r'^ *API level: ', s)]
+    apis = [re.findall(r'[0-9]+', s) for s in apis]
+    apis = [int(s[0]) for s in apis if s]
+    return apis
+
+
 class Context(object):
     '''A build context. If anything will be built, an instance this class
     will be instantiated and used to hold all the build state.'''
@@ -238,20 +298,7 @@ class Context(object):
         self.android_api = android_api
 
         check_target_api(android_api, self.archs[0].arch)
-
-        if exists(join(sdk_dir, 'tools', 'bin', 'avdmanager')):
-            avdmanager = sh.Command(join(sdk_dir, 'tools', 'bin', 'avdmanager'))
-            targets = avdmanager('list', 'target').stdout.decode('utf-8').split('\n')
-        elif exists(join(sdk_dir, 'tools', 'android')):
-            android = sh.Command(join(sdk_dir, 'tools', 'android'))
-            targets = android('list').stdout.decode('utf-8').split('\n')
-        else:
-            raise BuildInterruptingException(
-                'Could not find `android` or `sdkmanager` binaries in Android SDK',
-                instructions='Make sure the path to the Android SDK is correct')
-        apis = [s for s in targets if re.match(r'^ *API level: ', s)]
-        apis = [re.findall(r'[0-9]+', s) for s in apis]
-        apis = [int(s[0]) for s in apis if s]
+        apis = get_available_apis(self.sdk_dir)
         info('Available Android APIs are ({})'.format(
             ', '.join(map(str, apis))))
         if android_api in apis:
@@ -327,46 +374,21 @@ class Context(object):
         if not self.ccache:
             info('ccache is missing, the build will not be optimized in the '
                  'future.')
-        for cython_fn in ("cython", "cython3", "cython2", "cython-2.7"):
-            cython = sh.which(cython_fn)
-            if cython:
-                self.cython = cython
-                break
-        else:
-            raise BuildInterruptingException('No cython binary found.')
-        if not self.cython:
-            ok = False
-            warning("Missing requirement: cython is not installed")
+        self.cython = get_cython_path()
 
         # This would need to be changed if supporting multiarch APKs
         arch = self.archs[0]
-        platform_dir = arch.platform_dir
         toolchain_prefix = arch.toolchain_prefix
-        toolchain_version = None
-        self.ndk_platform = join(
-            self.ndk_dir,
-            'platforms',
-            'android-{}'.format(self.ndk_api),
-            platform_dir)
-        if not exists(self.ndk_platform):
-            warning('ndk_platform doesn\'t exist: {}'.format(
-                self.ndk_platform))
-            ok = False
+        self.ndk_platform, ndk_platform_dir_exists = get_ndk_platform_dir(
+            self.ndk_dir, self.ndk_api, arch)
+        ok = ok and ndk_platform_dir_exists
 
         py_platform = sys.platform
         if py_platform in ['linux2', 'linux3']:
             py_platform = 'linux'
-
-        toolchain_versions = []
-        toolchain_path = join(self.ndk_dir, 'toolchains')
-        if isdir(toolchain_path):
-            toolchain_contents = glob.glob('{}/{}-*'.format(toolchain_path,
-                                                            toolchain_prefix))
-            toolchain_versions = [split(path)[-1][len(toolchain_prefix) + 1:]
-                                  for path in toolchain_contents]
-        else:
-            warning('Could not find toolchain subdirectory!')
-            ok = False
+        toolchain_versions, toolchain_path_exists = get_toolchain_versions(
+            self.ndk_dir, arch)
+        ok = ok and toolchain_path_exists
         toolchain_versions.sort()
 
         toolchain_versions_gcc = []
--- a/tests/test_toolchain.py
+++ b/tests/test_toolchain.py
@@ -1,0 +1,89 @@
+import sys
+import pytest
+import mock
+from pythonforandroid.toolchain import ToolchainCL
+from pythonforandroid.util import BuildInterruptingException
+
+
+def patch_sys_argv(argv):
+    return mock.patch('sys.argv', argv)
+
+
+def patch_argparse_print_help():
+    return mock.patch('argparse.ArgumentParser.print_help')
+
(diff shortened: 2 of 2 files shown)


pythonforandroid/build.py
```diff
@@ -327,46 +374,21 @@ def prepare_build_environment(self,
...
-            cython = sh.which(cython_fn)
-            if cython:
-                self.cython = cython
-                break
-        else:
-            raise BuildInterruptingException('No cython binary found.')
-        if not self.cython:
-            ok = False
-            warning("Missing requirement: cython is not installed")
+        self.cython = get_cython_path()
```
AndreMiras: Not only we moved this block in a `get_cython_path()` helper above, but also removed the `warning()` that seemed dead code since we're raising exception

AndreMiras:
The `toolchain.py` looks painful to unit test. Still we should slowly, but surely get there.
This already increased the coverage slightly `COVERAGE INCREASED (+4.4%) TO 36.775%`, refs https://coveralls.io/jobs/51258249
BTW I saw @KeyWeeUsr was requested for review, that was a misclick, sorry for the spam

opacam:
Good, some more coverage is welcome, thanks to take care of this :smile:

Mmm...it seems that my local tests fails:

```
opacam@DEPOBOX:~/Devel/python-for-android$ tox -r -e py37 -- tests/test_toolchain.py
GLOB sdist-make: /home/opacam/Devel/python-for-android/setup.py
py37 recreate: /home/opacam/Devel/python-for-android/.tox/py37
py37 installdeps: mock, pytest, virtualenv
py37 inst: /home/opacam/Devel/python-for-android/.tox/.tmp/package/3/python-for-android-2019.7.8.1.dev0.zip
py37 installed: appdirs==1.4.3,atomicwrites==1.3.0,attrs==19.1.0,colorama==0.4.1,importlib-metadata==0.18,Jinja2==2.10.1,MarkupSafe==1.1.1,mock==3.0.5,more-itertools==7.2.0,packaging==19.0,pep517==0.5.0,pluggy==0.12.0,py==1.8.0,pyparsing==2.4.1.1,pytest==5.0.1,python-for-android==2019.7.8.1.dev0,pytoml==0.1.21,sh==1.12.14,six==1.12.0,virtualenv==16.7.0,wcwidth==0.1.7,zipp==0.5.2
py37 run-test-pre: PYTHONHASHSEED='1110765035'
py37 run-test: commands[0] | pytest tests/test_toolchain.py
============================================================================================================================= test session starts ==============================================================================================================================
platform linux -- Python 3.7.4, pytest-5.0.1, py-1.8.0, pluggy-0.12.0
cachedir: .tox/py37/.pytest_cache
rootdir: /home/opacam/Devel/python-for-android
collected 4 items                                                                                                                                                                                                                                                              

tests/test_toolchain.py ..F.                                                                                                                                                                                                                                             [100%]

=================================================================================================================================== FAILURES ===================================================================================================================================
_________________________________________________________________________________________________________________________ TestToolchainCL.test_create __________________________________________________________________________________________________________________________

self = <test_toolchain.TestToolchainCL object at 0x7f28b7837c10>

    def test_create(self):
        """
        Basic `create` distribution test.
        """
        argv = ['toolchain.py', 'create', '--sdk-dir=/tmp/android-sdk', '--ndk-dir=/tmp/android-ndk']
        with mock.patch('sys.argv', argv), \
                mock.patch('pythonforandroid.build.get_available_apis') as m_get_available_apis, \
                mock.patch('pythonforandroid.build.get_toolchain_versions') as m_get_toolchain_versions, \
                mock.patch('pythonforandroid.build.get_ndk_platform_dir') as m_get_ndk_platform_dir, \
                mock.patch('pythonforandroid.build.get_cython_path') as m_get_cython_path, \
                mock.patch('pythonforandroid.toolchain.build_dist_from_args') as m_build_dist_from_args:
            m_get_available_apis.return_value = [27]
            m_get_toolchain_versions.return_value = (['4.9'], True)
            m_get_ndk_platform_dir.return_value = (
                '/tmp/android-ndk/platforms/android-21/arch-arm', True)
            ToolchainCL()
        assert m_get_available_apis.call_args_list == [mock.call('/tmp/android-sdk')]
        assert m_get_toolchain_versions.call_args_list == [
            mock.call('/tmp/android-ndk', mock.ANY)]
        assert m_get_cython_path.call_args_list == [mock.call()]
>       assert m_build_dist_from_args.call_count == 1
E       AssertionError: assert 0 == 1
E        +  where 0 = <MagicMock name='build_dist_from_args' id='139812854396752'>.call_count

tests/test_toolchain.py:53: AssertionError
----------------------------------------------------------------------------------------------------------------------------- Captured stderr call ---------------------------------------------------------------------------------------------------------------------------
```

The same error when running for python2...it seems that `build_dist_from_args` is never called in my local tests, but it clearly works for CI tests...:thinking:

tests/test_toolchain.py
```diff
@@ -0,0 +1,62 @@
...
+                mock.patch('argparse.ArgumentParser.print_help') as m_print_help:
+            ToolchainCL()
+        assert ex_info.value.code == 1
+        assert m_print_help.call_args_list == [mock.call()]
+
+    def test_create(self):
+        """
+        Basic `create` distribution test.
+        """
+        argv = ['toolchain.py', 'create', '--sdk-dir=/tmp/android-sdk', '--ndk-dir=/tmp/android-ndk']
```
opacam: About the tox error I got, mentioned in [this comment](https://github.com/kivy/python-for-android/pull/1928#issuecomment-514932867), the addition of `dist_name` in argv solves the problem:
```
        argv = [
            'toolchain.py',
            'create',
            '--sdk-dir=/tmp/android-sdk',
            '--ndk-dir=/tmp/android-ndk',
            '--dist-name=test_toolchain',
        ]
```

tests/test_toolchain.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class TestToolchainCL:
+
+    def test_help(self):
+        """
+        Calling with `--help` should print help and exit 0.
+        """
+        argv = ['toolchain.py', '--help', '--storage-dir=/tmp']
+        with mock.patch('sys.argv', argv), pytest.raises(SystemExit) as ex_info, \
```
opacam: could we avoid this backslash?
I suggest something like:
```
        with mock.patch('sys.argv', argv), pytest.raises(
            SystemExit
        ) as ex_info, mock.patch(
            'argparse.ArgumentParser.print_help'
        ) as m_print_help:
            ToolchainCL()
```
Which also will be below 80 characters, so it will be pep8 friendly

tests/test_toolchain.py
```diff
@@ -0,0 +1,62 @@
...
+            ToolchainCL()
+        assert ex_info.value.code == 1
+        assert m_print_help.call_args_list == [mock.call()]
+
+    def test_create(self):
+        """
+        Basic `create` distribution test.
+        """
+        argv = ['toolchain.py', 'create', '--sdk-dir=/tmp/android-sdk', '--ndk-dir=/tmp/android-ndk']
+        with mock.patch('sys.argv', argv), \
```
opacam: Again here, the backslashes in this context manager:
```
        with mock.patch('sys.argv', argv), mock.patch(
            'pythonforandroid.build.get_available_apis'
        ) as m_get_available_apis, mock.patch(
            'pythonforandroid.build.get_toolchain_versions'
        ) as m_get_toolchain_versions, mock.patch(
            'pythonforandroid.build.get_ndk_platform_dir'
        ) as m_get_ndk_platform_dir, mock.patch(
            'pythonforandroid.build.get_cython_path'
        ) as m_get_cython_path, mock.patch(
            'pythonforandroid.toolchain.build_dist_from_args'
        ) as m_build_dist_from_args:
            m_get_available_apis.return_value = [27]
            m_get_toolchain_versions.return_value = (['4.9'], True)
            m_get_ndk_platform_dir.return_value = (
                '/tmp/android-ndk/platforms/android-21/arch-arm',
                True,
            )
```
This will also be pep8 friendly

tests/test_toolchain.py
```diff
@@ -0,0 +1,62 @@
...
+                mock.patch('pythonforandroid.build.get_toolchain_versions') as m_get_toolchain_versions, \
+                mock.patch('pythonforandroid.build.get_ndk_platform_dir') as m_get_ndk_platform_dir, \
+                mock.patch('pythonforandroid.build.get_cython_path') as m_get_cython_path, \
+                mock.patch('pythonforandroid.toolchain.build_dist_from_args') as m_build_dist_from_args:
+            m_get_available_apis.return_value = [27]
+            m_get_toolchain_versions.return_value = (['4.9'], True)
+            m_get_ndk_platform_dir.return_value = (
+                '/tmp/android-ndk/platforms/android-21/arch-arm', True)
+            ToolchainCL()
+        assert m_get_available_apis.call_args_list == [mock.call('/tmp/android-sdk')]
```
opacam: I also would shorten those lines to be below 80 characters:
```
        assert m_get_available_apis.call_args_list == [
            mock.call('/tmp/android-sdk')
        ]
        assert m_get_toolchain_versions.call_args_list == [
            mock.call('/tmp/android-ndk', mock.ANY)
        ]
```

tests/test_toolchain.py
```diff
@@ -0,0 +1,62 @@
...
+            mock.call('/tmp/android-ndk', mock.ANY)]
+        assert m_get_cython_path.call_args_list == [mock.call()]
+        assert m_build_dist_from_args.call_count == 1
+
+    def test_create_no_sdk_dir(self):
+        """
+        The `--sdk-dir` is mandatory to `create` a distribution.
+        """
+        argv = ['toolchain.py', 'create']
+        with mock.patch('sys.argv', argv), pytest.raises(BuildInterruptingException) as ex_info:
```
opacam: And this context manager (below 80 characters):

```
        with mock.patch('sys.argv', argv), pytest.raises(
            BuildInterruptingException
        ) as ex_info:
```

tests/test_toolchain.py
```diff
@@ -0,0 +1,62 @@
...
+            ToolchainCL()
+        assert ex_info.value.message == 'Android SDK dir was not specified, exiting.'
```
opacam: And also would shorten this assert, something like:
```
        assert (
            ex_info.value.message
            == 'Android SDK dir was not specified, exiting.'
        )
```

tests/test_toolchain.py
```diff
@@ -0,0 +1,62 @@
...
+class TestToolchainCL:
+
+    def test_help(self):
+        """
+        Calling with `--help` should print help and exit 0.
+        """
+        argv = ['toolchain.py', '--help', '--storage-dir=/tmp']
+        with mock.patch('sys.argv', argv), pytest.raises(SystemExit) as ex_info, \
+                mock.patch('argparse.ArgumentParser.print_help') as m_print_help:
+            ToolchainCL()
```
JonasT: Wouldn't something like this be a little easier to understand:

```
subprocess.check_output([
    sys.executable, "-c", "from pythonforandroid.toolchain import main(); main", "--help"
], cwd=os.path.join(os.path.dirname(__file__), ".."))
```

Or is that only me? Mock is fine but if it's not needed I would still rather use standard methods
---
There is also a lot of repetition of this block in particular:
```
        with mock.patch('sys.argv', argv), pytest.raises(SystemExit) as ex_info, \
                mock.patch('argparse.ArgumentParser.print_help') as m_print_help:
            ToolchainCL()
```
No matter if we end up using mock or not, maybe a helper function for that would be useful?

tests/test_toolchain.py
JonasT: Strangely enough I just tested it and it runs through fine for me, locally. I tested with python 3.7.3 and python 2.7.16 on Fedora 30. Or was this already fixed?

AndreMiras: No I haven't applied the fix yet. That's a very strange behaviour indeed.
Thank you @opacam for sharing a fix. I'll look into it

tests/test_toolchain.py
AndreMiras: The subprocess is overkill to me, it's really too much top level integration test.
Regarding the patching helper, yes I agree, I actually do this in my projects and at work, but since it's not common in p4a I didn't dare to introduce it.
I'll then

JonasT: > it's really too much top level integration test.

I'll let you decide but just as a note: that seems like an irrelevant point to me, because isn't this what the entire purpose is? Check if the overall tool responds fine to `--help`? It seems a little to me *with this particular argument* you're trying to isolate a component here with no benefit other than possibly complicating the test, and also removing it further from the actual usage later. But of course there are other completely valid reasons to stick with your current solution, e.g. if you find it more readable (which I don't but I can see others maybe would)

But keep in mind my proposal would also possibly greatly reduce the maintenance burden if `toolchain.py` is ever heavily refactored to also adjust the test here, because it nicely stays out of how the functionality needs to be performed as long as it works right without a crash, which I think is an additional benefit!

But that's just my personal opinion of course and I don't think this should hold off a merge

tests/test_toolchain.py
AndreMiras: Addressed

tests/test_toolchain.py
AndreMiras: Addressed

tests/test_toolchain.py
AndreMiras: Addressed

tests/test_toolchain.py
AndreMiras: I also hate backslashes, but for context managers, I haven't found a more readable way to write it so far.
I'm not yet convinced this way is more readable, but I gave it a try nonetheless :smile:

tests/test_toolchain.py
AndreMiras: Addressed

tests/test_toolchain.py
AndreMiras: I've added it, build is still passing for me, let's see for Travis and for you then :smile:

opacam: Yeah, now works, for CI and for my local tox tests too :smile:

Thanks @AndreMiras!!

tests/test_toolchain.py
AndreMiras: Yes I definitely don't share this point of view regarding readability, I find that `subprocess.check_output` absolutely ugly. Not to say about parsing eventual errors that it could output or the fact that's unclear which python executable it would run, the fact that you need to play around with the cwd. For me it's just a big nope. I've hardly ever seen people writing unit tests calling `subprocess` :confused:

JonasT: it's `sys.executable` do people not usually know that? :smile: but it's fine, leave it with mocking I am clearly making you confused :joy: (I'm weird that's normal)

PR #1931, opened 2019-07-25
Add a document describing how p4a interacts with pip & python packages

JonasT:
This pull request adds an internal documentation (not build as part of user docs, I just put it in there so we can find it again - it's mainly meant to just sit around in a place where we can find it) describing how p4a interacts with pip & python packages. [This is mainly meant for pip developers and others outside of our team to be able to get an idea what we're doing](https://github.com/kivy/python-for-android/issues/1932), although of course it could also be interesting to newcomers to our own team!

**Please have fun reading and tell me if anything doesn't make sense!**

### DIFF
--- a/doc/source/contribute.rst
+++ b/doc/source/contribute.rst
@@ -79,3 +79,154 @@ Release checklist
     - [ ] `armeabi-v7a`
     - [ ] `arm64-v8a`
 - [ ] Check that the version number is correct
+
+
+
+How python-for-android uses `pip`
+---------------------------------
+
+*Last update: July 2019*
+
+This section is meant to provide a quick summary how
+p4a (=python-for-android) uses pip and python packages in
+its build process.
+**It is written for a python
+packagers point of view, not for regular end users or
+contributors,** to assist with making pip developers and
+other packaging experts aware of p4a's packaging needs.
+
+Please note this section just attempts to neutrally list the
+current mechanisms, so some of this isn't necessarily meant
+to stay but just how things work inside p4a in
+this very moment.
+
+
+Basic concepts
+~~~~~~~~~~~~~~
+
+*(This part repeats other parts of the docs, for the sake of
+making this a more independent read)*
+
+p4a builds & packages a python application for use on Android.
+It does this by providing a Java wrapper, and for graphical applications
+an SDL2-based wrapper which can be used with the kivy UI toolkit if
+desired (or alternatively just plain PySDL2). Any such python application
+will of course have further library dependencies to do its work.
+
+p4a supports two types of package dependencies for a project:
+
+**Recipe:** install script in custom p4a format. Can either install
+C/C++ or other things that cannot be pulled in via pip, or things
+that can be installed via pip but break on android by default.
+These are maintained primarily inside the p4a source tree by p4a
+contributors and interested folks.
+
+**Python package:** any random pip python package can be directly
+installed if it doesn't need adjustments to work for Android.
+
+p4a will map any dependency to an internal recipe if present, and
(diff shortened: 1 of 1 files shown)


doc/internal-dev-info/how-pi4a-uses-pip.md
```diff
@@ -0,0 +1,157 @@
...
+desired (or alternatively just plain PySDL2).
+
+Any such python application will have dependencies, obviously.
+
+p4a supports two types of package dependencies for a project:
+
+**Recipe:** install script in custom p4a format. Can either install
+C/C++ or other things that cannot be pulled in via pip, or things
+that can be installed via pip but break on android by default.
+These are maintained inside the p4a source tree by p4a contributors
```
inclement: Strictly they're only primarily maintained inside the p4a source tree, local recipes do work fine.

inclement: Also missing full stop :p

inclement:
Oh, and as for where this goes, I don't think it would be inappropriate to put it in the documentation. It could be e.g. a subpage of "Development and Contributing". I don't think that would cause issues, and it would be nice to have it be discoverable and easily referenceable in the compiled documentation.

JonasT:
@inclement good point! I added into the "Development and Contributing" section now. will also make it easier to link once it's out on read the docs

doc/source/contribute.rst
```diff
@@ -79,3 +79,154 @@ Release checklist
...
+   pep517 library (if there is a `pyproject.toml` or `setup.py`)
+
+3. The collected dependencies are mapped to p4a's recipes if any are
+   available for them,
+   otherwise they're kept around as external regular package references
+
+4. All the dependencies mapped to recipes are built via p4a's internal
+   mechanisms to build these recipes. (This
+   may or may not indirectly use pip, depending on whether the recipe
+   wraps a python package or not and uses pip to install ro not)
```
AndreMiras: minor: typo `ro not` -> `or not`
Also minor, I'm curious about why that `(This` is alone on the lines above :stuck_out_tongue:

doc/source/contribute.rst
```diff
@@ -79,3 +79,154 @@ Release checklist
...
+dealing with python packages:
+
+- Packages will work fine without a recipe if they would also build
+  on Linux ARM, don't use any API not available in the NDK if they
+  use native code, and don't use any weird compiler flags the toolchain
+  doesn't like if they use native code. The package also needs to
+  work with cross compilation
+
+- There is currently no easy way for a package to know it is being
+  cross-compiled (at least that we know of) other than examining the
```
AndreMiras: I'm not sure what we try to say here. But command like `file` on a shared object would give info on the target arch it's compiled to right?

JonasT: Yes but usually you'd set that via a standardized env var and have some way e.g. in setuptools to differentiate the targets or something, and we're not really enabling or aware of such a way and not recommending this to anyone. Also, most install scripts probably just don't expect to be cross compiled, and maybe they should - it was just an acknowledgement that right now most install scripts are left mostly unaware by how we set things up

JonasT: Like, this is not our own problem alone. setuptools and such probably should provide better ways too. But it seems to me that working together with setuptools/pip people this could be possibly handled better than it is right now

AndreMiras: > it was just an acknowledgement that right now most install scripts are left mostly unaware by how we set things up

OK I understand the purpose of the paragraph now, thanks :+1:

doc/source/contribute.rst
JonasT: fixed it!

PR #1933, opened 2019-07-27
Increases toolchain.py test coverage

AndreMiras:
Increases `test_create()` coverage demonstrating crash referenced in:
<https://github.com/kivy/python-for-android/pull/1867#issuecomment-514352035>
Adds `test_recipes()` checking if it prints out without crashing.

### DIFF
--- a/pythonforandroid/toolchain.py
+++ b/pythonforandroid/toolchain.py
@@ -739,6 +739,15 @@ class ToolchainCL(object):
                 sys.argv.append(arg)
 
     def recipes(self, args):
+        """
+        Prints recipes basic info, e.g.
+        ```
+        python3      3.7.1
+            depends: ['hostpython3', 'sqlite3', 'openssl', 'libffi']
+            conflicts: ['python2']
+            optional depends: ['sqlite3', 'libffi', 'openssl']
+        ```
+        """
         ctx = self.ctx
         if args.compact:
             print(" ".join(set(Recipe.list_recipes(ctx))))
--- a/tests/test_toolchain.py
+++ b/tests/test_toolchain.py
@@ -1,6 +1,8 @@
+import io
 import sys
 import pytest
 import mock
+from pythonforandroid.recipe import Recipe
 from pythonforandroid.toolchain import ToolchainCL
 from pythonforandroid.util import BuildInterruptingException
 
@@ -11,6 +13,10 @@ def patch_sys_argv(argv):
 
 def patch_argparse_print_help():
     return mock.patch('argparse.ArgumentParser.print_help')
+
+
+def patch_sys_stdout():
+    return mock.patch('sys.stdout', new_callable=io.StringIO)
 
 
 def raises_system_exit():
@@ -51,6 +57,8 @@ class TestToolchainCL:
             'create',
             '--sdk-dir=/tmp/android-sdk',
             '--ndk-dir=/tmp/android-ndk',
+            '--bootstrap=service_only',
+            '--requirements=python3',
             '--dist-name=test_toolchain',
         ]
         with patch_sys_argv(argv), mock.patch(
@@ -62,8 +70,11 @@ class TestToolchainCL:
         ) as m_get_ndk_platform_dir, mock.patch(
             'pythonforandroid.build.get_cython_path'
         ) as m_get_cython_path, mock.patch(
-            'pythonforandroid.toolchain.build_dist_from_args'
-        ) as m_build_dist_from_args:
+            'pythonforandroid.toolchain.build_recipes'
+        ) as m_build_recipes, mock.patch(
+            'pythonforandroid.bootstraps.service_only.'
+            'ServiceOnlyBootstrap.run_distribute'
+        ) as m_run_distribute:
             m_get_available_apis.return_value = [27]
             m_get_toolchain_versions.return_value = (['4.9'], True)
             m_get_ndk_platform_dir.return_value = (
@@ -74,16 +85,54 @@ class TestToolchainCL:
         assert m_get_toolchain_versions.call_args_list == [
             mock.call('/tmp/android-ndk', mock.ANY)]
         assert m_get_cython_path.call_args_list == [mock.call()]
-        assert m_build_dist_from_args.call_count == 1
+        build_order = [
+            'hostpython3', 'libffi', 'openssl', 'sqlite3', 'python3',
+            'genericndkbuild', 'setuptools', 'six', 'pyjnius', 'android',
+        ]
+        python_modules = []
+        context = mock.ANY
+        project_dir = None
+        assert m_build_recipes.call_args_list == [
+            mock.call(
+                build_order,
+                python_modules,
+                context,
+                project_dir,
+                ignore_project_setup_py=False
+            )
+        ]
+        assert m_run_distribute.call_args_list == [mock.call()]
 
     def test_create_no_sdk_dir(self):
         """
         The `--sdk-dir` is mandatory to `create` a distribution.
         """
         argv = ['toolchain.py', 'create']
-        with mock.patch('sys.argv', argv), pytest.raises(
+        with patch_sys_argv(argv), pytest.raises(
             BuildInterruptingException
         ) as ex_info:
             ToolchainCL()
         assert ex_info.value.message == (
             'Android SDK dir was not specified, exiting.')
+
+    @pytest.mark.skipif(sys.version_info < (3, 0), reason="requires python3")
+    def test_recipes(self):
+        """
+        Checks the `recipes` command prints out recipes information without crashing.
+        """
+        argv = ['toolchain.py', 'recipes']
+        with patch_sys_argv(argv), patch_sys_stdout() as m_stdout:
+            ToolchainCL()
+        # check if we have common patterns in the output
+        expected_strings = (
+            'conflicts:',
+            'depends:',
+            'kivy',
+            'optional depends:',
+            'python3',
+            'sdl2',
+        )
+        for expected_string in expected_strings:
+            assert expected_string in m_stdout.getvalue()
+        # deletes static attribute to not mess with other tests
+        del Recipe.recipes


AndreMiras:
This is still in WIP, because for some reason the `test_recipes()` later make the `tests/recipes/test_reportlab.py` fail. I need to investigate, but this probably has something to do with the way we dynamically load recipes.
For records, the error is:
```
__________________________________________________ TestReportLabRecipe.test_prebuild_arch __________________________________________________

self = <test_reportlab.TestReportLabRecipe testMethod=test_prebuild_arch>

    def test_prebuild_arch(self):
        """
        Makes sure `prebuild_arch()` runs without error and patches `setup.py`
        as expected.
        """
        # `prebuild_arch()` dynamically replaces strings in the `setup.py` file
        setup_path = os.path.join(self.recipe_dir, 'setup.py')
        with open(setup_path, 'w') as setup_file:
            setup_file.write('_FT_LIB_\n')
            setup_file.write('_FT_INC_\n')
    
        # these sh commands are not relevant for the test and need to be mocked
        with \
                patch('sh.patch'), \
                patch('sh.touch'), \
                patch('sh.unzip'), \
                patch('os.path.isfile'):
>           self.recipe.prebuild_arch(self.arch)

tests/recipes/test_reportlab.py:52:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pythonforandroid/recipes/reportlab/__init__.py:33: in prebuild_arch
    ft_dir = ft.get_build_dir(arch.arch)
pythonforandroid/recipe.py:275: in get_build_dir
    return join(self.get_build_container_dir(arch), self.name)
pythonforandroid/recipe.py:264: in get_build_container_dir
    dir_name, '{}__ndk_target_{}'.format(arch, self.ctx.ndk_api))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <pythonforandroid.build.Context object at 0x7f86ad48e710>

    @property
    def ndk_api(self):
        '''The API number compile against'''
        if self._ndk_api is None:
>           raise ValueError('Tried to access ndk_api but it has not '
                             'been set - this should not happen, something '
                             'went wrong!')
E           ValueError: Tried to access ndk_api but it has not been set - this should not happen, something went wrong!

pythonforandroid/build.py:203: ValueError
----------------------------------------------------------- Captured stderr call -----------------------------------------------------------
```

AndreMiras:
I've fixed the test and we now have a +15.8% of coverage (to 53.148%), awesome.
The problem cames from the static `Recipe.recipes` attribute. We may encounter similar issues in the future as we write more tests. I think the fix I've have is OK enough for now as I don't want to bring tearDown, fixtures or context manager yet.

tests/test_toolchain.py
```diff
@@ -62,8 +70,11 @@ def test_create(self):
         ) as m_get_ndk_platform_dir, mock.patch(
             'pythonforandroid.build.get_cython_path'
         ) as m_get_cython_path, mock.patch(
-            'pythonforandroid.toolchain.build_dist_from_args'
-        ) as m_build_dist_from_args:
+            'pythonforandroid.toolchain.build_recipes'
+        ) as m_build_recipes, mock.patch(
+            'pythonforandroid.bootstraps.service_only.'
+            'ServiceOnlyBootstrap.run_distribute'
+        ) as m_run_distribute:
```
JonasT: This is just a generic gut feeling that might be wrong, but:

My intuitive reaction here is: this seems like this is mocked too much and will be a huge pain once we change some things. (I also find at some point it gets quite challenging to read)

Is there really no easier way to call this, maybe e.g. actually more on a real directory with test data and just check the output? I know this goes a bit back to me suggesting `subprocess` instead of mocking but here it seems to actually grow to such a size that I'm wondering if here it might be justified, it just looks a little like we're testing a too big component here in a mock way that is too large to be reasonably mock-tested in isolation

tests/test_toolchain.py
```diff
@@ -74,16 +85,54 @@ def test_create(self):
         assert m_get_toolchain_versions.call_args_list == [
             mock.call('/tmp/android-ndk', mock.ANY)]
         assert m_get_cython_path.call_args_list == [mock.call()]
-        assert m_build_dist_from_args.call_count == 1
+        build_order = [
+            'hostpython3', 'libffi', 'openssl', 'sqlite3', 'python3',
+            'genericndkbuild', 'setuptools', 'six', 'pyjnius', 'android',
+        ]
+        python_modules = []
```
JonasT: maybe put the `[]` as a direct parameter into the `mock.call` and add ` #python_modules parameter` comment instead? I assume you put this separately for readability, but it did throw me off at first that this is so weirdly put separate

tests/test_toolchain.py
```diff
@@ -74,16 +85,54 @@ def test_create(self):
...
+    @pytest.mark.skipif(sys.version_info < (3, 0), reason="requires python3")
+    def test_recipes(self):
+        """
+        Checks the `recipes` command prints out recipes information without crashing.
+        """
+        argv = ['toolchain.py', 'recipes']
+        with patch_sys_argv(argv), patch_sys_stdout() as m_stdout:
+            ToolchainCL()
+        # check if we have common patterns in the output
+        expected_strings = (
```
JonasT: I like this, very readable in my opinion!

tests/test_toolchain.py
JonasT: Okay I thought about this more, and what I said is nonsense :joy: creating a distribution will compile all the recipes, right? So there is no quick way to test that in isolation and what you're doing makes perfect sense. I guess I just hoped there was an easier, less depending-exactly-on-so-many-specific-function-calls kind of way, but oh well :woman_shrugging:

tests/test_toolchain.py
AndreMiras: Yes this is exactly for readability, I specifically prefer explicit code rather than a comment.
I mean what you say is also fine and I know people do it and then if I were to be consistent I should do the same for at least `context` and `project_dir` as well right? Then the code would look something like:
```python
assert m_build_recipes.call_args_list == [
    mock.call(
        build_order,
        # python_modules parameter
        [],
        # context parameter
        mock.ANY,
        # project_dir parameter
        None,
        ignore_project_setup_py=False
    )
]
```
I honestly don't like it at all :confused: but if you definitely prefer comment than code in this case, I can still had it. And what about `build_order` would you also prefer literal with a comment? Or perhaps I misunderstood what you're suggesting :thinking: ?

tests/test_toolchain.py
AndreMiras: Thanks, I almost missed it as it was "resolved" actually :stuck_out_tongue:

tests/test_toolchain.py
JonasT: yes you got my gist! what about put just a little differently:

```
assert m_build_recipes.call_args_list == [
    mock.call(
        build_order,
        [],                # python_modules
        mock.ANY,          # context
        None,              # project_dir
        ignore_project_setup_py=False
    )
]
```
Something like that?

I am only suggesting something different because at first glance the separately written block made me think there was something else going on in addition, so it did throw me off for a second. I think putting it as a comment like that does the same job while avoiding that

But as usual whatever you prefer, was just what I personally thought with my occasionally weird style preferences :smile:

AndreMiras:
Thanks for the quick review @JonasT. Yes I do agree with your strike-through comment regarding the integration test. It's indeed far from being a strickly unit test indeed, but it was easy enough to write it that way while also helping raising issues, increasing coverage and not introducing any kind of lag/flakkness in our testing, so it feel good enough for now

tests/test_toolchain.py
AndreMiras: OK I understand what you say I think and thanks for suggesting, but yes I definitely prefer to use code rather than comments in this case.
I hope you're fine that I keep it this way then

AndreMiras:
@opacam I'm merging it prior your review, but I'll keep an eye on your post merge comments if you have any so I can address on subsequent PRs

pythonforandroid/toolchain.py
```diff
@@ -739,6 +739,15 @@ def _read_configuration():
...
     def recipes(self, args):
+        """
+        Prints recipes basic info, e.g.
+        ```
+        python3      3.7.1
+            depends: ['hostpython3', 'sqlite3', 'openssl', 'libffi']
+            conflicts: ['python2']
+            optional depends: ['sqlite3', 'libffi', 'openssl']
+        ```
+        """
```
opacam: **minor:** I'm not sure how it will be rendered this docstring in sphinx...maybe I would use a `code-block` directive, something like:

```
        Prints recipes basic info, e.g.
        .. code-block:: bash

            python3      3.7.1
            depends: ['hostpython3', 'sqlite3', 'openssl', 'libffi']
            conflicts: ['python2']
            optional depends: ['sqlite3', 'libffi', 'openssl']
 ```

But, @AndreMiras, don't need to change this, if I remember well, I think that we don't have enabled the build of technical documentation, so it wont be visible for now.

Apart from this, I would like to mention that overall it looks good, clean and very readable.Also, I made tox test on my machine, with current develop branch , since it's already merged, without any issue.

@AndreMiras, thanks for this PR, good job, as usual :+1:

and @JonasT, thanks for your review, I was out for the weekend and I was unable to take care of the review, so yours sped up the merging :smile:

AndreMiras: I don't know how I missed that comment. I'd be happy to address it still

PR #1936, opened 2019-07-28
Also copy the service/main.py when building with setup.py

jtc0de:
When using `--use-setup-py` and a `setup.py` instead of `--requirements`, only the `main.py` will currently get copied into the final `.apk` as well as the site packages. That makes using an unnamed service (without `--service`) impossible since `service/main.py` is no longer copied into the `.apk`, and an installed entrypoint via `setup.py` would be in another location.

This commit fixes this by also copying `service/main.py` into the `.apk` when present in addition to `main.py` (while still not copying any other files)

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -321,16 +321,36 @@ main.py that loads it.''')
                       'full private data into .apk.')
                 tar_dirs.append(args.private)
             else:
-                print('Copying main.py ONLY, since other app data is '
-                      'expected in site-packages.')
+                print("Copying main.py's ONLY, since other app data is "
+                      "expected in site-packages.")
                 main_py_only_dir = tempfile.mkdtemp()
                 _temp_dirs_to_clean.append(main_py_only_dir)
-                if exists(join(args.private, "main.pyo")):
-                    shutil.copyfile(join(args.private, "main.pyo"),
-                                    join(main_py_only_dir, "main.pyo"))
-                elif exists(join(args.private, "main.py")):
-                    shutil.copyfile(join(args.private, "main.py"),
-                                    join(main_py_only_dir, "main.py"))
+
+                # Check all main.py files we need to copy:
+                copy_paths = ["main.py", join("service", "main.py")]
+                for copy_path in copy_paths:
+                    variants = [
+                        copy_path,
+                        copy_path.partition(".")[0] + ".pyc",
+                        copy_path.partition(".")[0] + ".pyo",
+                    ]
+                    # Check in all variants with all possible endings:
+                    for variant in variants:
+                        if exists(join(args.private, variant)):
+                            # Make sure surrounding directly exists:
+                            dir_path = os.path.dirname(variant)
+                            if (len(dir_path) > 0 and
+                                    not exists(
+                                        join(main_py_only_dir, dir_path)
+                                    )):
+                                os.mkdir(join(main_py_only_dir, dir_path))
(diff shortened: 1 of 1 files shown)


pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -331,6 +331,20 @@ def make_package(args):
                 elif exists(join(args.private, "main.py")):
                     shutil.copyfile(join(args.private, "main.py"),
                                     join(main_py_only_dir, "main.py"))
+                if exists(join(args.private, "service", "main.pyo")):
```
inclement: I would pull the repeated calls to `join` into a variable, since the arguments are the same each time.

JonasT:
> I haven't tried to service/main.py method for years, and I had the impression that it's broken or very buggy. Am I wrong about that?

Actually yes, I *only* use that one and it works perfectly fine! :smile: I otherwise would have made a pull request, because I don't like `--service` since I tend to avoid anything that needs additional command line flags. I prefer going with ways that kinda work just from the files and `setup.py` without too many special settings

pythonforandroid/bootstraps/common/build/build.py
opacam: Ei...shouldn't be the file `main.pyc`?

`.pyo` was for python 2, but, as far as I know, for python 3 should be `.pyc`...since we are moving to python3...maybe we should change all this `.pyo` extensions to `.pyc`...or maybe set the file extension based on the python version used, to be compatible with both versions of python...

JonasT:
Tested it now, appears to work just fine in the latest version! :+1:

pythonforandroid/bootstraps/common/build/build.py
JonasT: Both `.pyc` and `.pyo` are handled now, thanks for pointing this out!

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -321,16 +321,36 @@ def make_package(args):
...
+                    # Check in all variants with all possible endings:
+                    for variant in variants:
+                        if exists(join(args.private, variant)):
+                            # Make sure surrounding directly exists:
+                            dir_path = os.path.dirname(variant)
+                            if (len(dir_path) > 0 and
+                                    not exists(
+                                        join(main_py_only_dir, dir_path)
+                                    )):
+                                os.mkdir(join(main_py_only_dir, dir_path))
```
AndreMiras: Minor: What about creating the dir yes or yes? You know like `mkdir -p` so we can skip the if?
e.g.
```python
os.mkdir(join(main_py_only_dir, dir_path), exist_ok=True)
```

JonasT: I am trained to avoid this because of Python 2. Are we sure this code doesn't run in Python 2 when deploying for Python 2 android targets even with Python 3 host? It *might* be ok but I haven't tested just yet, and that's why I usually just avoid Python 3-only things

AndreMiras: Yes I see, let's play it safe then :+1:
Even though I don't really mind Python2 being broken :trollface:

inclement: There's also `ensure_dir` in util.py, which I used to use for this.

JonasT: Hm my intuition is to stick more to standard functions for readability, especially since this will solve itself once we drop Python 2 entirely also as a target (which I assume we will do *eventually*?)

PR #1937, opened 2019-07-28
Call Cython via `python -m Cython` rather than system-wide binary

### LINKED ISSUE #1885: TypeError: must be str, not bytes
### Versions

* Python: 3.5.3 (running in virtualenv and in root)
  (seems to be running Python 3.7.1 on Android based on the logs)
* OS: Ubuntu 17:04
* Kivy: 1.9.1
* Cython: 0.28.6
* Andoid 7.0 (on Samsung Galaxy S6)

### Description

I have compiled a very basic helloWorld app as below. The app works fine on Ubuntu and also compiles fine into apk using buildozer. 

````
__version__ = '1.0'
from kivy.app import App
from kivy.uix.button import Button
 
class Hello(App):
    def build(self):
        btn = Button(text='Hello World')
        return  btn
 
Hello().run()
````

When I run on Android the app immediately closes before displaying the hello world screen. After working through logs produced with adb logcat I see the following...
````
06-25 19:48:52.272  2100  2118 I python  :  Traceback (most recent call last):
06-25 19:48:52.272  2100  2118 I python  :    File "/home/lee/KivyProjects/helloWorld/.buildozer/android/app/main.py", line 3, in <module>
06-25 19:48:52.273  2100  2118 I python  :    File "/home/lee/KivyProjects/helloWorld/.buildozer/android/platform/build/build/python-installs/helloWorld/kivy/app.py", line 319, in <module>
06-25 19:48:52.273  2100  2118 I python  :    File "/home/lee/KivyProjects/helloWorld/.buildozer/android/platform/build/build/python-installs/helloWorld/kivy/base.py", line 26, in <module>
06-25 19:48:52.274  2100  2118 I python  :    File "/home/lee/KivyProjects/helloWorld/.buildozer/android/platform/build/build/python-installs/helloWorld/kivy/clock.py", line 410, in <module>
06-25 19:48:52.274  2100  2118 I python  :    File "<frozen importlib._bootstrap>", line 983, in _find_and_load
06-25 19:48:52.274  2100  2118 I python  :    File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
06-25 19:48:52.275  2100  2118 I python  :    File "<frozen importlib._bootstrap>", line 668, in _load_unlocked
06-25 19:48:52.275  2100  2118 I python  :    File "<frozen importlib._bootstrap>", line 638, in _load_backward_compatible
06-25 19:48:52.276  2100  2118 I python  :    File "/home/lee/KivyProjects/helloWorld/.buildozer/android/platform/build/build/other_builds/python3-libffi-openssl-sqlite3/armeabi-v7a__ndk_target_21/python3/Lib/ctypes/util.py", line 73, in <module>
06-25 19:48:52.276  2100  2118 I python  :    File "/home/lee/KivyProjects/helloWorld/.buildozer/android/platform/build/build/python-installs/helloWorld/android/__init__.py", line 8, in <module>
06-25 19:48:52.277  2100  2118 I python  :    File "android/_android.pyx", line 162, in init android._android
06-25 19:48:52.277  2100  2118 I python  :  TypeError: must be str, not bytes
06-25 19:48:52.277  2100  2118 I python  : Python for android ended.
````

This appears to be the same issue raised as 1691 under the title "unicode error during startup (python3, numpy, opencv) - patch included #1691". If I am reading this patch correctly the current file _android.pyx should have the following changes

removed this line
    python_act = autoclass(JAVA_NAMESPACE + u'.PythonActivity')

added these lines
    java_namespace = JAVA_NAMESPACE.decode() if isinstance(JAVA_NAMESPACE, bytes) else JAVA_NAMESPACE
    python_act = autoclass(java_namespace  + u'.PythonActivity')

Yet when I look at the current code on Github for _android.pyx I do not see these changes applied. Has this fix somehow been removed from the current version? How can i apply this fix to my system?

### PR DESCRIPTION
**Pull request by JonasT**
This changes calls to Cython to `python -m Cython` to avoid picking one not matching the current python version (which can happen if just calling `Cython`). This should fix #1885

### DIFF
--- a/Dockerfile.py3
+++ b/Dockerfile.py3
@@ -94,7 +94,7 @@ ENV WORK_DIR="${HOME_DIR}" \
 # install system dependencies
 RUN ${RETRY} apt -y install -qq --no-install-recommends \
         python3 virtualenv python3-pip python3-venv \
-        wget lbzip2 patch sudo \
+        wget lbzip2 patch sudo python python-pip \
     && apt -y autoremove
 
 # build dependencies
@@ -122,8 +122,8 @@ RUN useradd --create-home --shell /bin/bash ${USER}
 RUN usermod -append --groups sudo ${USER}
 RUN echo "%sudo ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
 
-
-RUN pip3 install --upgrade cython==0.28.6
+# install cython for python 2 (for python 3 it's inside the venv)
+RUN pip2 install --upgrade Cython==0.28.6
 
 WORKDIR ${WORK_DIR}
 COPY --chown=user:user . ${WORK_DIR}
@@ -133,4 +133,5 @@ USER ${USER}
 # install python-for-android from current branch
 RUN virtualenv --python=python3 venv \
     && . venv/bin/activate \
+    && pip3 install --upgrade Cython==0.28.6 \
     && pip3 install -e .
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -27,14 +27,6 @@ from pythonforandroid.recommendations import (
     RECOMMENDED_NDK_API, RECOMMENDED_TARGET_API)
 
 
-def get_cython_path():
-    for cython_fn in ("cython", "cython3", "cython2", "cython-2.7"):
-        cython = sh.which(cython_fn)
-        if cython:
-            return cython
-    raise BuildInterruptingException('No cython binary found.')
-
-
 def get_ndk_platform_dir(ndk_dir, ndk_api, arch):
     ndk_platform_dir_exists = True
     platform_dir = arch.platform_dir
@@ -111,7 +103,6 @@ class Context(object):
     use_setup_py = False
 
     ccache = None  # whether to use ccache
-    cython = None  # the cython interpreter name
 
     ndk_platform = None  # the ndk platform directory
 
@@ -374,7 +365,14 @@ class Context(object):
         if not self.ccache:
             info('ccache is missing, the build will not be optimized in the '
                  'future.')
-        self.cython = get_cython_path()
+        try:
+            subprocess.check_output([
+                "python3", "-m", "cython", "--help",
+            ])
+        except subprocess.CalledProcessError:
+            warning('Cython for python3 missing. If you are building for '
+                    ' a python 3 target (which is the default)'
+                    ' then THINGS WILL BREAK.')
 
         # This would need to be changed if supporting multiarch APKs
         arch = self.archs[0]
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -1024,8 +1024,11 @@ class CythonRecipe(PythonRecipe):
             del cyenv['PYTHONPATH']
         if 'PYTHONNOUSERSITE' in cyenv:
             cyenv.pop('PYTHONNOUSERSITE')
-        cython_command = sh.Command(self.ctx.cython)
-        shprint(cython_command, filename, *self.cython_args, _env=cyenv)
+        python_command = sh.Command("python{}".format(
+            self.ctx.python_recipe.major_minor_version_string.split(".")[0]
+        ))
+        shprint(python_command, "-m", "Cython.Build.Cythonize",
+                filename, *self.cython_args, _env=cyenv)
 
     def cythonize_build(self, env, build_dir="."):
         if not self.cythonize:
--- a/Dockerfile.py2
+++ b/Dockerfile.py2
@@ -122,8 +122,6 @@ RUN usermod -append --groups sudo ${USER}
 RUN echo "%sudo ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
 
 
-RUN pip install --upgrade cython==0.28.6
-
 WORKDIR ${WORK_DIR}
 COPY --chown=user:user . ${WORK_DIR}
 RUN chown --recursive ${USER} ${ANDROID_SDK_HOME}
@@ -132,4 +130,5 @@ USER ${USER}
 # install python-for-android from current branch
 RUN virtualenv --python=python venv \
     && . venv/bin/activate \
+    && pip install --upgrade cython==0.28.6 \
     && pip install -e .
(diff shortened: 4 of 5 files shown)

### CODE REVIEW

File: pythonforandroid/build.py [hunk 1]
```diff
@@ -374,7 +365,6 @@ def prepare_build_environment(self,
         if not self.ccache:
             info('ccache is missing, the build will not be optimized in the '
                  'future.')
-        self.cython = get_cython_path()
```
inclement: Could we still add a check here for Cython being available, or a BuildInterruptingException if not?

JonasT: Oh, good point! Sure, that would be possible, let me have a look at it

Dockerfile.py3
```diff
@@ -107,6 +107,7 @@ RUN dpkg --add-architecture i386 \
         libpangox-1.0-0:i386 libpangoxft-1.0-0:i386 libidn11:i386 \
         zip zlib1g-dev zlib1g:i386 \
     && apt -y autoremove
+RUN ${RETRY} pip3 install -U Cython
```
AndreMiras: Wait what? It's already there https://github.com/kivy/python-for-android/blob/v2019.07.08/Dockerfile.py3#L126

pythonforandroid/build.py
```diff
@@ -374,7 +365,14 @@ def prepare_build_environment(self,
...
+        try:
+            subprocess.check_output([
+                "python3", "-m", "cython", "--help",
+            ])
+        except subprocess.CalledProcessError:
+            warning('Cython for python3 missing. If you are building for '
+                    ' a python 3 target (which is the default)'
+                    ' then THINGS WILL BREAK.')
```
AndreMiras: If things would definitely break. Then let's not be gentle and let's fail fast with `BuildInterruptingException`, what do you guys think?

pythonforandroid/recipe.py
```diff
@@ -1024,8 +1024,12 @@ def cythonize_file(self, env, build_dir, filename):
             del cyenv['PYTHONPATH']
         if 'PYTHONNOUSERSITE' in cyenv:
             cyenv.pop('PYTHONNOUSERSITE')
-        cython_command = sh.Command(self.ctx.cython)
-        shprint(cython_command, filename, *self.cython_args, _env=cyenv)
+        cython_command = sh.Command(
+            "python" +
+            self.ctx.python_recipe.major_minor_version_string.split(".")[0]
```
AndreMiras: nitpick: I prefer `format()`

pythonforandroid/recipe.py
```diff
@@ -1024,8 +1024,12 @@ def cythonize_file(self, env, build_dir, filename):
             del cyenv['PYTHONPATH']
         if 'PYTHONNOUSERSITE' in cyenv:
             cyenv.pop('PYTHONNOUSERSITE')
-        cython_command = sh.Command(self.ctx.cython)
-        shprint(cython_command, filename, *self.cython_args, _env=cyenv)
+        cython_command = sh.Command(
```
AndreMiras: technically this isn't a cython command, but a python command, correct?
The cython however comes later once you adds the `-m cython` to it few lines below

pythonforandroid/build.py
JonasT: Only if it's a python 3 target. We still support python 2, and at this stage the recipes aren't known :woman_shrugging: so as the text says, it would break but only for a python 3 build. I did some poking around the code and I found no way to tell this in this function

AndreMiras: Oh Python2, right... Thanks I almost forgot about this folk :joy:

Dockerfile.py3
```diff
@@ -133,4 +135,5 @@ USER ${USER}
 # install python-for-android from current branch
 RUN virtualenv --python=python3 venv \
     && . venv/bin/activate \
+    && pip3 install -U Cython \
```
AndreMiras: Isn't the one installed system wide enough?
https://github.com/kivy/python-for-android/blob/v2019.07.08/Dockerfile.py3#L126

JonasT: Hm true I may have interpreted that one error situation wrong, let me try it might work without that :+1:

JonasT: Yup system-wide is enough, apparently :joy: well-spotted!

AndreMiras: I understand, thanks for investigating!
Then maybe we should ditch the system one?

AndreMiras: Ouch, I just saw the build is failing for Python2.
https://travis-ci.org/kivy/python-for-android/jobs/567481453
```
2943  RAN: /usr/bin/python2 -m Cython.Build.Cythonize ./jnius/jnius.pyx
2944
2945  STDOUT:
2946/usr/bin/python2: No module named Cython.Build
```
So that's properly weird and annoying. Basically it means if we want to support the "Python2 basic" apk build in the CI, we do need the system cython on python2. While only having the venv cython3. That would look kinda inconsistent in the Dockerfile, but with some comments that should make it.
So basically we want to have:
```
RUN pip2 install --upgrade Cython==0.28.6
```
And also in the venv:
```
&& pip3 install --upgrade Cython==0.28.6
```
By the way, let's take the opportunity to make them consistent.

- uppercase `Cython` because that's it's naming on pypi
- long argument `--upgrade` because we love explicit
- explicit "officially p4a tested" version `0.28.6` in all places (eventually via variable)

JonasT:
I just realized I never tested if this *actually* fixes #1885, even though I believe this is the "more correct" way to call it in any case. Does anyone have a test environment for that around? Otherwise I suppose I should set myself one up real quick and actually give that a try

Dockerfile.py3
```diff
@@ -122,8 +122,8 @@ RUN useradd --create-home --shell /bin/bash ${USER}
 RUN usermod -append --groups sudo ${USER}
 RUN echo "%sudo ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
 
-
-RUN pip3 install --upgrade cython==0.28.6
+# install cython for python 2 (for python 3 it's inside the venv)
+pip2 install --upgrade Cython==0.28.6
```
AndreMiras: You're missing the `RUN` here:
```
Error response from daemon: Dockerfile parse error line 126: unknown instruction: PIP2
```
ahah don't give up :sweat_smile:

AndreMiras:
> even though I believe this is the "more correct" way to call it in any case

Yes same feeling here, so we should make it into the tree in any case (after the build passes of course)

Dockerfile.py3
JonasT: > ahah don't give up :sweat_smile:

I won't I'm stubborn :joy:

AndreMiras:
Good job! Feel free to merge once the build turns :heavy_check_mark:

PR #1938, opened 2019-07-28
Added setuptools to Kivy recipe requirements

inclement:
Kivy master (and future Kivy releases) depends on setuptools now, whereas it didn't before. I think we should just add it to the requirements now, to avoid issues when people try to install from master. It doesn't seem worth messing around with making setuptools a requirement only if Kivy > 1.11 is being targeted.


inclement:
Thanks @opacam. I saw that comment, but I didn't realise that setuptools is effectively a default dependency now. I've seen a user have issues using p4a master, but in this case it looks like they were probably using an older version of p4a where this dependency chain wasn't the same.

PR #1944, opened 2019-07-30
[LIBS - PART I] Initial refactor of library recipes

opacam:
This PR will introduce a new cls attribute: `built_libraries`.
This `built_libraries` attribute must be a dictionary. The keys are the full library name, e.g.: `libffi.so`. And the values must be the relative path of the library, e.g: `.libs`. So the this cls attribute for libffi recipe it would look like:
    ```built_libraries = {'libffi.so': '.libs'}```

This new cls attribute will allow us to detect library recipes and to refactor common operations that we do for those kind of recipes:
  - copy library into right location
  - implement `should_build` for library recipes

Those two things will make that our rebuilds to be more consistent and I hope that it will be easier to maintain. Also I was thinking that it would be great to refactor the build methods that we use in those recipes, but this is another chapter of p4a story...and maybe better to wait for the NDK r19 migration thing

This could be safely merged without affecting the library recipes untouched here, which will behave as usual. This is because the `build_arch` method, for library recipes only, it will be split into two jobs:
 - do_build_libs
 - do_install_libs

**So, once explained a little this PR, in here we do:**
  - [x] Add attribute `built_libraries`
  - [x] Add methods to refactor common library operations:
    - [x] do_build_libs
    - [x] do_install_libs
  - [x] implement `should_build` for library recipes
  - [x] implement basic tests for the newly introduced `Recipe` methods
  - [x] Make use of those Library attributes/methods for some basic set of recipes:
    - [x] libffi
    - [x] openssl
    - [x] png
     - [x] jpeg
     - [x] freetype
     - [x] harfbuzz
     - [x] libcurl
     - [x] libzbar
     - [x] libiconv
     - [x] libexpat
     - [x] libogg
     - [x] libxml2
     - [x] libxslt
     - [x] libshine
     - [x] libx264
     - [x] libglob
     - [x] libsodium
     - [x] libsecp256k1

**Things to do in seperate PRs**, to be easy to review and because those points will start to conflict with the `NDK r19 migration` thing:
  - Make use of Library attributes/methods introduced in this PR for un covered library recipes in here. Here we have two situations:
     - library recipes that depends on android's STL library (almost all the work done in here will depend of the `NDK r19 migration` thing)
     - all remaining library recipes, wich are not STL lib dependant and that are not implemented in here (because I was unable to perform a successful build with them using arch `arm64-v8a`...so I think it would  be better to deal with those recipes in seperate PRs and later...with the `NDK r19 migration` thing merged)

**Notes about changed recipes:** all the recipes touched in here almost have the same changes:
  - remove `should_build` method (to make use of the one implemented in base class for library recipes)
  - rename `build_arch` to `do_build_libs`
  - remove the copy of the library (because that will be done automatically by the method `do_install_libs` implemented in base class)
  - fixed the imports due to refactoring

### DIFF
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -516,6 +516,19 @@ class Recipe(with_metaclass(RecipeMeta)):
         if hasattr(self, build):
             getattr(self, build)()
 
+    def install_libraries(self, arch):
+        '''This method is always called after `build_arch`. In case that we
+        detect a library recipe, defined by the class attribute
+        `built_libraries`, we will copy all defined libraries into the
+         right location.
+        '''
+        if not self.built_libraries:
+            return
+        shared_libs = [
+            lib for lib in self.get_libraries(arch) if lib.endswith(".so")
+        ]
+        self.install_libs(arch, *shared_libs)
+
     def postbuild_arch(self, arch):
         '''Run any post-build tasks for the Recipe. By default, this checks if
         any postbuild_archname methods exist for the archname of the
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -498,9 +498,14 @@ class Recipe(with_metaclass(RecipeMeta)):
 
     def should_build(self, arch):
         '''Should perform any necessary test and return True only if it needs
-        building again.
+        building again. Per default we implement a library test, in case that
+        we detect so.
 
         '''
+        if self.built_libraries:
+            return not all(
+                exists(lib) for lib in self.get_libraries(arch.arch)
+            )
         return True
 
     def build_arch(self, arch):
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -573,6 +573,27 @@ class Recipe(with_metaclass(RecipeMeta)):
     def has_libs(self, arch, *libs):
         return all(map(lambda l: self.ctx.has_lib(arch.arch, l), libs))
 
+    def get_libraries(self, arch_name, in_context=False):
+        """Return the full path of the library depending on the architecture.
+        Per default, the build library path it will be returned, unless
+        `get_libraries` has been called with kwarg `in_context` set to
+        True.
+
+        .. note:: this method should be used for library recipes only
+        """
+        recipe_libs = set()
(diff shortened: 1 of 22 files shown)


AndreMiras:
From a very quick check, great idea and awesome work! :clap: 
But why do you go nuts on it from the first PR? :smile: I mean you could tacle `pythonforandroid/recipe.py` and maybe a little specific recipe as an example, that's fine. And then when it gets to the tree later on, you can address some more recipes, and again some more later.
Now you did most of the hard work so that's fine I guess. But keep it for later, I don't think a PR has to cover all the possible cases with 1000 lines of diff for hitting the tree.

pythonforandroid/recipes/libshine/__init__.py
```diff
@@ -23,7 +24,7 @@ def build_arch(self, arch):
                     '--enable-static',
                     '--prefix={}'.format(realpath('.')),
                     _env=env)
-            shprint(sh.make, '-j4', _env=env)
+            shprint(sh.make, '-j', str(cpu_count()), _env=env)
```
AndreMiras: ahah thanks for sneaking technical debt fixes :clap:

pythonforandroid/recipe.py
```diff
@@ -479,18 +498,42 @@ def apply_patches(self, arch, build_dir=None):
...
         any build_archname methods exist for the archname of the current
         architecture, and runs them if so.'''
         build = "build_{}".format(arch.arch)
         if hasattr(self, build):
             getattr(self, build)()
+        # in case that we detect a library recipe, we split the build in two
+        # steps, so we can control the copy and check of the built libraries
+        if self.built_libraries:
+            self.do_build_libs(arch)
+            self.do_install_libs(arch)
```
inclement: There's nothing really wrong with this, but I mildly dislike replacing `build_arch` with `do_build_libs` in a subset of recipes.

I think it could be better to add an `install_libs` method that called after build_arch in `build.build_recipes`, and which does the job currently covered by `do_install_libs`. That way, all the recipes could continue to use `build_arch` the same way they currently do, which is nice for consistency.

inclement:
This has picked up some conflicts, but it's good to merge as far as I'm concerned.

PR #1947, opened 2019-07-31
[LIBS - PART II] Part II of NDK r19 migration - Initial STL lib migration

opacam:
This is the second part of the libraries refactoring, but also it is the second part of the NDK r19 migration. In here we implement some attributes/methods to cls `Recipe` that will allow us to refactor some recipes that depends on android's STL library which now will default to `c++_shared` because, as per android's NDK version r19, is the only STL library available.

Things done here:
  - Add android's STL lib support to `Recipe`
  - Make` CppCompiledComponentsPythonRecipe` use Recipe's STL support
  - Make icu a library recipe with STL support (reworked recipe, since contains further changes)
  - Rework pyicu recipe to match latest icu changes
  - Add tests for recipe with STL support
  - Add tests for icu/pyicu recipes

**This PR shouldn't be merged before** the `LIBS - PART I` (#1944)


tests/recipes/test_icu.py
```diff
@@ -0,0 +1,167 @@
+import os
```
AndreMiras: Awesome that you're unit testing it :+1:

inclement:
Even more great work @opacam, I've merged all the PRs whose builds were currently passing. You are truly the recipe expert!

opacam:
@inclement and @AndreMiras

Thanks for your positive comments :blush:, just trying to follow you guys, you are the real experts en here :1st_place_medal: not me ;)

**Notes**:
  - I though that this would take us more time (specially because I had to leave for some days...still out...so **thank you guys to take care of this, for the reviews and the mergings**)
  - Unfortunately, I didn't had time to add tests for all the recipes touched in the other parts (I may have some of them in my main computer...I don't remember cause now working on a crappy laptop :joy_cat: ...I will check next days)

PR #1948, opened 2019-07-31
Adds new `backup_attributes` context manager for tests

AndreMiras:
This context manager will makes it possible to rollback an object state
after leaving the context manager.
This is a follow up of
<https://github.com/kivy/python-for-android/pull/1867#discussion_r309345405>

Also address docstring format as suggested by @opacam in
<https://github.com/kivy/python-for-android/pull/1933#discussion_r308116515>

### DIFF
--- a/pythonforandroid/toolchain.py
+++ b/pythonforandroid/toolchain.py
@@ -741,12 +741,12 @@ class ToolchainCL(object):
     def recipes(self, args):
         """
         Prints recipes basic info, e.g.
-        ```
-        python3      3.7.1
-            depends: ['hostpython3', 'sqlite3', 'openssl', 'libffi']
-            conflicts: ['python2']
-            optional depends: ['sqlite3', 'libffi', 'openssl']
-        ```
+        .. code-block:: bash
+
+            python3      3.7.1
+                depends: ['hostpython3', 'sqlite3', 'openssl', 'libffi']
+                conflicts: ['python2']
+                optional depends: ['sqlite3', 'libffi', 'openssl']
         """
         ctx = self.ctx
         if args.compact:
--- a/tests/test_toolchain.py
+++ b/tests/test_toolchain.py
@@ -2,6 +2,7 @@ import io
 import sys
 import pytest
 import mock
+from util import backup_attributes
 from pythonforandroid.recipe import Recipe
 from pythonforandroid.toolchain import ToolchainCL
 from pythonforandroid.util import BuildInterruptingException
@@ -121,7 +122,10 @@ class TestToolchainCL:
         Checks the `recipes` command prints out recipes information without crashing.
         """
         argv = ['toolchain.py', 'recipes']
-        with patch_sys_argv(argv), patch_sys_stdout() as m_stdout:
+        with (
+                patch_sys_argv(argv)), (
+                patch_sys_stdout()) as m_stdout, (
+                backup_attributes(Recipe, {'recipes'})):
             ToolchainCL()
         # check if we have common patterns in the output
         expected_strings = (
@@ -134,5 +138,3 @@ class TestToolchainCL:
         )
         for expected_string in expected_strings:
             assert expected_string in m_stdout.getvalue()
-        # deletes static attribute to not mess with other tests
-        del Recipe.recipes
--- a/tests/test_util.py
+++ b/tests/test_util.py
@@ -8,6 +8,7 @@ except ImportError:
     # `Python 2` or lower than `Python 3.3` does not
     # have the `unittest.mock` module built-in
     import mock
+import util as test_util
 from pythonforandroid import util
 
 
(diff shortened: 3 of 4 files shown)


tests/test_toolchain.py
```diff
@@ -121,7 +122,10 @@ def test_recipes(self):
         Checks the `recipes` command prints out recipes information without crashing.
         """
         argv = ['toolchain.py', 'recipes']
-        with patch_sys_argv(argv), patch_sys_stdout() as m_stdout:
+        with (
+                patch_sys_argv(argv)), (
+                patch_sys_stdout()) as m_stdout, (
+                backup_attributes(Recipe, {'recipes'})):
```
AndreMiras: in Python3 that could also be used as a decorator, e.g.
```
@backup_attributes(Recipe, {'recipes'})
def test_recipes(self):
    pass
```

JonasT:
As I already stated, I think this is just waiting for other code someone writing in the future to blow up that uses `get_recipe` with different `ctx`/`Context()` parameters (which the function interfaces invites one to do). Instead, I would suggest writing a test that actually supplies different contexts when getting the same recipe and checks that the context is set correctly. I think this is like introducing a really mean and hidden source of bugs intentionally and keeping that working, it makes no sense to me

JonasT:
On further inspection it also seems a bit over the top, but I like predictable tests. So I'm not necessarily opposed. But I think as long as it masks the actual issue I don't think it's the right point in time to merge this

tests/test_toolchain.py
opacam: I'm wondering, will this work when used alongside with `mock.patch` decorator?

pythonforandroid/toolchain.py
```diff
@@ -741,12 +741,12 @@ def _read_configuration():
     def recipes(self, args):
         """
         Prints recipes basic info, e.g.
-        ```
-        python3      3.7.1
-            depends: ['hostpython3', 'sqlite3', 'openssl', 'libffi']
-            conflicts: ['python2']
-            optional depends: ['sqlite3', 'libffi', 'openssl']
-        ```
+        .. code-block:: bash
```
opacam: @AndreMiras, thanks to take care of this docstring :+1:

tests/test_toolchain.py
```diff
@@ -2,6 +2,7 @@
 import sys
 import pytest
 import mock
+from util import backup_attributes
```
opacam: I'm not sure if the name you give to the file can cause troubles in some import situations...since we already have a file with the same name in `pythonforandroid.util` :thinking:...I would give it another filename to avoid strange situations...but maybe there is no need...what do you think about this?

tests/test_toolchain.py
AndreMiras: Depends of what you patch and if you patch before the `backup_attributes` or after I guess

tests/test_toolchain.py
AndreMiras: Yes it also bugged me. Actually I usually call it `utils.py`, then I changed it to be consistent. But still yes it also bugs me that it looks the same and one is for tests only. I don't have preferences, tell me how you like it to be called and I can call it that way :smile:

tests/test_toolchain.py
opacam: Ok, we'll see then, thanks!

tests/test_toolchain.py
opacam: - `tools`
  - `utilities`
  - `assets`
  ...

JonasT:
Ok I thought about this some more after sleeping, and now I'm not so sure this pull request is a good idea:

Why are we trying to arbitrarily reset a small part of p4a's entire global state? Wouldn't it be a much cleaner idea to just use a different interpreter per test if we really wanted to guarantee that? Otherwise, why are we just caring about this specific small part now and not resetting everything else, what's the meaning behind that? Because just for this bug it doesn't really make sense to me, that bug should rather be test-covered & fixed instead of avoided through a state reset anyway

So in overall I'm really not sure where this pull request is getting us and I find it hard to see it as anything but a complication with little benefit and a weird partial focus :no_mouth: would be happy to be convinced of something else though

tests/test_toolchain.py
JonasT: What about `test_utils.py`?

opacam: @JonasT, I would avoid `test_utils`...I think that unittests looks into the first part of the filename to determine if a file is a test or not...am I right?

Plus the mentioned file is not a test...it's a set of utilities to help us testing...I think that it's confusing...

JonasT: `utils_for_testing.py`? :smile:

opacam: That looks better :wink:

Issue #1817, opened 2019-05-07
Cannot build APK with buildozer


jairoerazobotero:
### Versions
Python: 3.7.1
OS: MacOS Mojave 10.14.4

### Description
I'm trying to build a simple kivy app with buildozer. I already tried rebuilding (deleting .buildozer folder then running buildozer -v android debug, reinstall buildozer via pip install buildozer, but I still encounter the same issue. I need help. The code of my app is [here](https://github.com/jairoerazobotero/testchess) in branch Tests.

### buildozer.spec
Command:

`buidozer -v android debug`

Spec file:
```[app]

# (str) Title of your application
title = TestChess

# (str) Package name
package.name = testchess

# (str) Package domain (needed for android/ios packaging)
package.domain = org.test

# (str) Source code where the main.py live
source.dir = .

# (list) Source files to include (let empty to include all the files)
source.include_exts = py,png,jpg,kv,atlas

# (list) List of inclusions using pattern matching
#source.include_patterns = assets/*,images/*.png

# (list) Source files to exclude (let empty to not exclude anything)
#source.exclude_exts = spec

# (list) List of directory to exclude (let empty to not exclude anything)
#source.exclude_dirs = tests, bin

# (list) List of exclusions using pattern matching
#source.exclude_patterns = license,images/*/*.jpg

# (str) Application versioning (method 1)
version = 0.1

# (str) Application versioning (method 2)
# version.regex = __version__ = ['"](.*)['"]
# version.filename = %(source.dir)s/main.py

# (list) Application requirements
# comma separated e.g. requirements = sqlite3,kivy
requirements = python3,kivy

# (str) Custom source folders for requirements
# Sets custom source for any requirements with recipes
# requirements.source.kivy = ../../kivy

# (list) Garden requirements
#garden_requirements =

# (str) Presplash of the application
#presplash.filename = %(source.dir)s/data/presplash.png

# (str) Icon of the application
#icon.filename = %(source.dir)s/data/icon.png

# (str) Supported orientation (one of landscape, sensorLandscape, portrait or all)
orientation = portrait

# (list) List of service to declare
#services = NAME:ENTRYPOINT_TO_PY,NAME2:ENTRYPOINT2_TO_PY

#
# OSX Specific
#

#
# author = © Copyright Info

# change the major version of python used by the app
osx.python_version = 3

# Kivy version to use
osx.kivy_version = 1.9.1

#
# Android specific
#

# (bool) Indicate if the application should be fullscreen or not
[...]

Issue #1800, opened 2019-04-20
buildozer failed clang++: error: linker command failed with exit code 1


hoangvuche:
### Versions

* Python: 3.7.1
* OS: Mac Mojave 10.14.2
* Kivy: 1.10.1
* Cython: 0.28.1

### Description
I can't build my kivy app for an apk targeting Android with the log as follows.
I have few experience in debuging this, could someone pls help me with the root cause and solution?

### buildozer.spec

Command:
```sh
buildozer android debug
```

Spec file:
```
[app]

# (str) Title of your application
title = My news

# (str) Package name
package.name = myapp

# (str) Package domain (needed for android/ios packaging)
package.domain = ruche.news

# (str) Source code where the main.py live
source.dir = .

# (list) Source files to include (let empty to include all the files)
source.include_exts = py,png,jpg,kv,atlas

# (list) List of inclusions using pattern matching
#source.include_patterns = assets/*,images/*.png

# (list) Source files to exclude (let empty to not exclude anything)
#source.exclude_exts = spec

# (list) List of directory to exclude (let empty to not exclude anything)
source.exclude_dirs = tests, bin, venv

# (list) List of exclusions using pattern matching
#source.exclude_patterns = license,images/*/*.jpg

# (str) Application versioning (method 1)
version = 0.1

# (str) Application versioning (method 2)
# version.regex = __version__ = ['"](.*)['"]
# version.filename = %(source.dir)s/main.py

# (list) Application requirements
# comma separated e.g. requirements = sqlite3,kivy
requirements = python3,kivy

# (str) Custom source folders for requirements
# Sets custom source for any requirements with recipes
# requirements.source.kivy = ../../kivy

# (list) Garden requirements
#garden_requirements =

# (str) Presplash of the application
#presplash.filename = %(source.dir)s/data/presplash.png

# (str) Icon of the application
#icon.filename = %(source.dir)s/data/icon.png

# (str) Supported orientation (one of landscape, sensorLandscape, portrait or all)
orientation = portrait

# (list) List of service to declare
#services = NAME:ENTRYPOINT_TO_PY,NAME2:ENTRYPOINT2_TO_PY

#
# OSX Specific
#

#
# author = © Copyright Info

# change the major version of python used by the app
osx.python_version = 3

# Kivy version to use
osx.kivy_version = 1.9.1

#
# Android specific
#

# (bool) Indicate if the application should be fullscreen or not
fullscreen = 1

[...]

PR #1951, opened 2019-08-02
 Fix build for case-insensitive FS and add CI test for OSX

opacam:
It turns out that the generated binary for `MacOS` and `Cygwin` is not `python`...it's `python.exe`...this leads to not set correctly the `ctx.hostpython` variable for `Mac OS`. This rename of the python binary for  `MacOS` is to, probably,  avoid issues with the  `Mac OS` filesystem which could be Case Insensitive and it would conflict with the build folder named `Python`).

I hope that this solve the `Mac OS` build issue :crossed_fingers:

**Note:** this should be tested by some mac user, unfortunately I could not test it because I don't have a mac...so I recommend to merge after someone test this, but all seems indicate that this is the problem of p4a with Mac OS...thanks to [this comment](https://github.com/kivy/python-for-android/issues/1817#issuecomment-517338012) we got the solution :smile:,**Thanks** @TheSin-

**See also:** https://github.com/python/cpython/blob/3.7/README.rst#build-instructions

Closes: #1817
Closes: #1800
Closes: #1682
Closes: #1647

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,11 +16,6 @@ before_install:
   - git remote set-branches --add origin develop
   - git fetch
 
-env:
-  global:
-    - ANDROID_SDK_HOME=/opt/android/android-sdk
-    - ANDROID_NDK_HOME=/opt/android/android-ndk
-
 jobs:
   include:
     - &linting
@@ -65,19 +60,41 @@ jobs:
         # case that the travis log doesn't produce any output for more than 10 minutes
         - while sleep 540; do echo "==== Still running (travis, don't kill me) ===="; done &
       script:
-        - docker run -e CI -e TRAVIS_JOB_ID="$TRAVIS_JOB_ID" -e TRAVIS_BRANCH="$TRAVIS_BRANCH" p4a /bin/sh -c "$COMMAND"
+        - >
+          docker run
+          -e CI
+          -e TRAVIS_JOB_ID
+          -e TRAVIS_BRANCH
+          -e ANDROID_SDK_HOME="/home/user/.android/android-sdk"
+          -e ANDROID_NDK_HOME="/home/user/.android/android-ndk"
+          p4a /bin/sh -c "$COMMAND"
       after_script:
         # kill the background process started before run docker
         - kill %1
-      name: Python 3 armeabi-v7a
+      name: Python 3 arm64-v8a
       # overrides requirements to skip `peewee` pure python module, see:
       # https://github.com/kivy/python-for-android/issues/1263#issuecomment-390421054
-      env: COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_python3_sqlite_openssl.py apk --sdk-dir $ANDROID_SDK_HOME --ndk-dir $ANDROID_NDK_HOME --requirements libffi,sdl2,pyjnius,kivy,python3,openssl,requests,sqlite3,setuptools' --arch=armeabi-v7a
+      env:
+        COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_python3_sqlite_openssl.py apk --sdk-dir $ANDROID_SDK_HOME --ndk-dir $ANDROID_NDK_HOME --requirements libffi,sdl2,pyjnius,kivy,python3,openssl,requests,sqlite3,setuptools' --arch=arm64-v8a
     - <<: *testing
-      name: Python 3 arm64-v8a
-      env: COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_python3_sqlite_openssl.py apk --sdk-dir $ANDROID_SDK_HOME --ndk-dir $ANDROID_NDK_HOME --requirements libffi,sdl2,pyjnius,kivy,python3,openssl,requests,sqlite3,setuptools' --arch=arm64-v8a
+      name: Python 3 armeabi-v7a
+      os: osx
+      osx_image: xcode11  # since xcode1.3, python3 is the default interpreter
+      before_script:
+        # installs java 1.8, android's SDK/NDK and p4a
+        - make -f ci/makefiles/osx.mk
+        - export JAVA_HOME=/Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home
+        # Run a background process (like we do with linux tests)
+        - while sleep 540; do echo "==== Still running (travis, don't kill me) ===="; done &
+      script:
+        - >
+          cd testapps && python3 setup_testapp_python3_sqlite_openssl.py apk
+          --sdk-dir $HOME/.android/android-sdk
+          --ndk-dir $HOME/.android/android-ndk
+          --requirements libffi,sdl2,pyjnius,kivy,python3,openssl,requests,sqlite3,setuptools
+          --arch=armeabi-v7a
     - <<: *testing
-      name: Python 2 basic
+      name: Python 2 armeabi-v7a (with numpy)
       env: COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_python2_sqlite_openssl.py apk --sdk-dir $ANDROID_SDK_HOME --ndk-dir $ANDROID_NDK_HOME --requirements sdl2,pyjnius,kivy,python2,openssl,requests,sqlite3,setuptools,numpy'
     - <<: *testing
       name: Rebuild updated recipes
--- a/Dockerfile.py2
+++ b/Dockerfile.py2
@@ -37,64 +37,17 @@ ENV RETRY="retry -t 3 --"
 RUN curl https://raw.githubusercontent.com/kadwanev/retry/1.0.1/retry \
     --output /usr/local/bin/retry && chmod +x /usr/local/bin/retry
 
-ENV ANDROID_NDK_HOME="${ANDROID_HOME}/android-ndk"
-ENV ANDROID_NDK_VERSION="17c"
-ENV ANDROID_NDK_HOME_V="${ANDROID_NDK_HOME}-r${ANDROID_NDK_VERSION}"
-
-# get the latest version from https://developer.android.com/ndk/downloads/index.html
-ENV ANDROID_NDK_ARCHIVE="android-ndk-r${ANDROID_NDK_VERSION}-linux-x86_64.zip"
-ENV ANDROID_NDK_DL_URL="https://dl.google.com/android/repository/${ANDROID_NDK_ARCHIVE}"
-
-# download and install Android NDK
-RUN ${RETRY} curl --location --progress-bar --insecure \
-        "${ANDROID_NDK_DL_URL}" \
-        --output "${ANDROID_NDK_ARCHIVE}" \
-    && mkdir --parents "${ANDROID_NDK_HOME_V}" \
-    && unzip -q "${ANDROID_NDK_ARCHIVE}" -d "${ANDROID_HOME}" \
-    && ln -sfn "${ANDROID_NDK_HOME_V}" "${ANDROID_NDK_HOME}" \
-    && rm -rf "${ANDROID_NDK_ARCHIVE}"
-
-
-ENV ANDROID_SDK_HOME="${ANDROID_HOME}/android-sdk"
-
-# get the latest version from https://developer.android.com/studio/index.html
-ENV ANDROID_SDK_TOOLS_VERSION="4333796"
-ENV ANDROID_SDK_BUILD_TOOLS_VERSION="28.0.2"
-ENV ANDROID_SDK_TOOLS_ARCHIVE="sdk-tools-linux-${ANDROID_SDK_TOOLS_VERSION}.zip"
-ENV ANDROID_SDK_TOOLS_DL_URL="https://dl.google.com/android/repository/${ANDROID_SDK_TOOLS_ARCHIVE}"
-
-# download and install Android SDK
-RUN ${RETRY} curl --location --progress-bar --insecure \
-        "${ANDROID_SDK_TOOLS_DL_URL}" \
-        --output "${ANDROID_SDK_TOOLS_ARCHIVE}" \
-    && mkdir --parents "${ANDROID_SDK_HOME}" \
-    && unzip -q "${ANDROID_SDK_TOOLS_ARCHIVE}" -d "${ANDROID_SDK_HOME}" \
-    && rm -rf "${ANDROID_SDK_TOOLS_ARCHIVE}"
-
-# update Android SDK, install Android API, Build Tools...
-RUN mkdir --parents "${ANDROID_SDK_HOME}/.android/" \
-    && echo '### User Sources for Android SDK Manager' \
-        > "${ANDROID_SDK_HOME}/.android/repositories.cfg"
-
-# Download and accept Android licenses (JDK necessary!)
-RUN ${RETRY} apt -y install -qq --no-install-recommends openjdk-8-jdk \
-    && apt -y autoremove
-RUN yes | "${ANDROID_SDK_HOME}/tools/bin/sdkmanager" "build-tools;${ANDROID_SDK_BUILD_TOOLS_VERSION}" > /dev/null
-RUN yes | "${ANDROID_SDK_HOME}/tools/bin/sdkmanager" "platforms;android-27" > /dev/null
-
-# Set avdmanager permissions (executable)
-RUN chmod +x "${ANDROID_SDK_HOME}/tools/bin/avdmanager"
-
-
 ENV USER="user"
 ENV HOME_DIR="/home/${USER}"
+ENV ANDROID_HOME="${HOME_DIR}/.android"
 ENV WORK_DIR="${HOME_DIR}" \
     PATH="${HOME_DIR}/.local/bin:${PATH}"
 
 # install system dependencies
 RUN ${RETRY} apt -y install -qq --no-install-recommends \
         python virtualenv python-pip wget lbzip2 patch sudo \
-    && apt -y autoremove
+    && apt -y autoremove \
+    && apt -y clean
 
 # build dependencies
 # https://buildozer.readthedocs.io/en/latest/installation.html#android-on-ubuntu-16-04-64bit
@@ -113,6 +66,10 @@ RUN ${RETRY} apt -y install -qq --no-install-recommends \
     && apt -y autoremove \
     && apt -y clean
 
+# Install Java, set JAVA_HOME (to accept android's SDK licenses) and clean
+RUN ${RETRY} apt -y install -qq --no-install-recommends openjdk-8-jdk \
+    && apt -y autoremove && apt -y clean
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64
 
 # prepare non root env
 RUN useradd --create-home --shell /bin/bash ${USER}
@@ -124,8 +81,11 @@ RUN echo "%sudo ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers
 
 WORKDIR ${WORK_DIR}
 COPY --chown=user:user . ${WORK_DIR}
-RUN chown --recursive ${USER} ${ANDROID_SDK_HOME}
+RUN mkdir ${ANDROID_HOME} && chown --recursive ${USER} ${ANDROID_HOME}
 USER ${USER}
+
+# Download and install android's NDK/SDK
+RUN make -f ci/makefiles/android.mk target_os=linux
 
 # install python-for-android from current branch
 RUN virtualenv --python=python venv \
--- a/Dockerfile.py3
+++ b/Dockerfile.py3
@@ -17,8 +17,6 @@
 
 FROM ubuntu:18.04
 
-ENV ANDROID_HOME="/opt/android"
-
 # configure locale
 RUN apt update -qq > /dev/null && apt install -qq --yes --no-install-recommends \
     locales && \
@@ -37,57 +35,9 @@ ENV RETRY="retry -t 3 --"
 RUN curl https://raw.githubusercontent.com/kadwanev/retry/1.0.1/retry \
     --output /usr/local/bin/retry && chmod +x /usr/local/bin/retry
 
-ENV ANDROID_NDK_HOME="${ANDROID_HOME}/android-ndk"
-ENV ANDROID_NDK_VERSION="17c"
-ENV ANDROID_NDK_HOME_V="${ANDROID_NDK_HOME}-r${ANDROID_NDK_VERSION}"
-
-# get the latest version from https://developer.android.com/ndk/downloads/index.html
-ENV ANDROID_NDK_ARCHIVE="android-ndk-r${ANDROID_NDK_VERSION}-linux-x86_64.zip"
-ENV ANDROID_NDK_DL_URL="https://dl.google.com/android/repository/${ANDROID_NDK_ARCHIVE}"
-
-# download and install Android NDK
-RUN ${RETRY} curl --location --progress-bar --insecure \
-        "${ANDROID_NDK_DL_URL}" \
-        --output "${ANDROID_NDK_ARCHIVE}" \
-    && mkdir --parents "${ANDROID_NDK_HOME_V}" \
-    && unzip -q "${ANDROID_NDK_ARCHIVE}" -d "${ANDROID_HOME}" \
-    && ln -sfn "${ANDROID_NDK_HOME_V}" "${ANDROID_NDK_HOME}" \
-    && rm -rf "${ANDROID_NDK_ARCHIVE}"
-
-
-ENV ANDROID_SDK_HOME="${ANDROID_HOME}/android-sdk"
-
-# get the latest version from https://developer.android.com/studio/index.html
-ENV ANDROID_SDK_TOOLS_VERSION="4333796"
-ENV ANDROID_SDK_BUILD_TOOLS_VERSION="28.0.2"
-ENV ANDROID_SDK_TOOLS_ARCHIVE="sdk-tools-linux-${ANDROID_SDK_TOOLS_VERSION}.zip"
-ENV ANDROID_SDK_TOOLS_DL_URL="https://dl.google.com/android/repository/${ANDROID_SDK_TOOLS_ARCHIVE}"
-
-# download and install Android SDK
-RUN ${RETRY} curl --location --progress-bar --insecure \
-        "${ANDROID_SDK_TOOLS_DL_URL}" \
-        --output "${ANDROID_SDK_TOOLS_ARCHIVE}" \
-    && mkdir --parents "${ANDROID_SDK_HOME}" \
-    && unzip -q "${ANDROID_SDK_TOOLS_ARCHIVE}" -d "${ANDROID_SDK_HOME}" \
-    && rm -rf "${ANDROID_SDK_TOOLS_ARCHIVE}"
-
-# update Android SDK, install Android API, Build Tools...
-RUN mkdir --parents "${ANDROID_SDK_HOME}/.android/" \
-    && echo '### User Sources for Android SDK Manager' \
-        > "${ANDROID_SDK_HOME}/.android/repositories.cfg"
-
-# Download and accept Android licenses (JDK necessary!)
-RUN ${RETRY} apt -y install -qq --no-install-recommends openjdk-8-jdk \
-    && apt -y autoremove
-RUN yes | "${ANDROID_SDK_HOME}/tools/bin/sdkmanager" "build-tools;${ANDROID_SDK_BUILD_TOOLS_VERSION}" > /dev/null
-RUN yes | "${ANDROID_SDK_HOME}/tools/bin/sdkmanager" "platforms;android-27" > /dev/null
-
-# Set avdmanager permissions (executable)
-RUN chmod +x "${ANDROID_SDK_HOME}/tools/bin/avdmanager"
-
-
 ENV USER="user"
 ENV HOME_DIR="/home/${USER}"
+ENV ANDROID_HOME="${HOME_DIR}/.android"
 ENV WORK_DIR="${HOME_DIR}" \
     PATH="${HOME_DIR}/.local/bin:${PATH}"
 
@@ -114,6 +64,10 @@ RUN ${RETRY} apt -y install -qq --no-install-recommends \
     && apt -y autoremove \
     && apt -y clean
 
+# Install Java and set JAVA_HOME (to accept android's SDK licenses)
+RUN ${RETRY} apt -y install -qq --no-install-recommends openjdk-8-jdk \
+    && apt -y autoremove && apt -y clean
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64
 
 # prepare non root env
 RUN useradd --create-home --shell /bin/bash ${USER}
@@ -127,8 +81,11 @@ RUN pip2 install --upgrade Cython==0.28.6
 
 WORKDIR ${WORK_DIR}
 COPY --chown=user:user . ${WORK_DIR}
-RUN chown --recursive ${USER} ${ANDROID_SDK_HOME}
+RUN mkdir ${ANDROID_HOME} && chown --recursive ${USER} ${ANDROID_HOME}
 USER ${USER}
+
+# Download and install android's NDK/SDK
+RUN make -f ci/makefiles/android.mk target_os=linux
 
 # install python-for-android from current branch
 RUN virtualenv --python=python3 venv \
--- a/ci/makefiles/android.mk
+++ b/ci/makefiles/android.mk
@@ -1,0 +1,68 @@
+# Downloads and installs the Android SDK depending on supplied platform: darwin or linux
+
+# We must provide a platform (darwin or linux) and we need JAVA_HOME defined
+ifndef target_os
+    $(error target_os is not set...aborted!)
+endif
+
+# Those android NDK/SDK variables can be override when running the file
+ANDROID_NDK_VERSION ?= 17c
+ANDROID_SDK_TOOLS_VERSION ?= 4333796
+ANDROID_SDK_BUILD_TOOLS_VERSION ?= 28.0.2
+ANDROID_HOME ?= $(HOME)/.android
+ANDROID_API_LEVEL ?= 27
+
+ANDROID_SDK_HOME=$(ANDROID_HOME)/android-sdk
+ANDROID_SDK_TOOLS_ARCHIVE=sdk-tools-$(target_os)-$(ANDROID_SDK_TOOLS_VERSION).zip
+ANDROID_SDK_TOOLS_DL_URL=https://dl.google.com/android/repository/$(ANDROID_SDK_TOOLS_ARCHIVE)
+
+ANDROID_NDK_HOME=$(ANDROID_HOME)/android-ndk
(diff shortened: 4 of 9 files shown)


TheSin-:
No problem I'm happy to help, I'm very very new to python so I'm just fumbling around.

I'll build a new docker with this patch and test it, hopefully it won't take me too long to test it.

JonasT:
I am wondering, don't we sometimes call "python2" or "python3" directly in other places? E.g. I do that here: https://github.com/kivy/python-for-android/pull/1937

It seems weird that this call wouldn't work, after all it also does on Windows, actually. And wouldn't it also for an actual Cygwin-compiled program too?

Why is Cygwin involved here anyway? Shouldn't we target macOS without Cygwin? This situation confuses me a little

pythonforandroid/python.py
```diff
@@ -382,6 +383,22 @@ class HostPythonRecipe(Recipe):
     '''The default url to download our host python recipe. This url will
     change depending on the python version set in attribute :attr:`version`.'''
 
+    @property
+    def python_bin(self):
```
inclement: I have a vague feeling that it isn't necessarily obvious `python_bin` will be a name only, not a full path. I'd consider making it an internal property (i.e. just call it `_python_bin`). It's a minor thing though.

inclement:
I'd quite like to get this into the next release (sometime in the next few days), or at the very least: we could add a simple log that confirms the python exe name to give confidence in the change. It would then be easy to pull the result from any user's log.

TheSin-:
I’m hopefully going to run a test after dinner tonight.

TheSin-:
did not break, but also did not fix it.  I'm trying to dig deeper but I'm pretty new to docker and even newer to python-for-android...

TheSin-:
does this help

```
_PYTHON_PROJECT_BASE=/home/buildozer/build/.buildozer/android/platform/build/build/other_builds/python3-libffi-openssl-sqlite3/armeabi-v7a__ndk_target_21/python3/android-build _PYTHON_HOST_PLATFORM=linux-arm PYTHONPATH=/home/buildozer/build/.buildozer/android/platform/build/build/other_builds/python3-libffi-openssl-sqlite3/armeabi-v7a__ndk_target_21/python3/Lib _PYTHON_SYSCONFIGDATA_NAME=_sysconfigdata_m_linux_ python -S -m sysconfig --generate-posix-vars ;\
if test $? -ne 0 ; then \
	echo "generate-posix-vars failed" ; \
	rm -f ./pybuilddir.txt ; \
	exit 1 ; \
fi
/bin/sh: 1: python: not found
generate-posix-vars failed
Makefile:595: recipe for target 'pybuilddir.txt' failed
make: *** [pybuilddir.txt] Error 1
```

TheSin-:
so this is the line in the python3 Makefile

```
PYTHON_FOR_BUILD=_PYTHON_PROJECT_BASE=$(abs_builddir) _PYTHON_HOST_PLATFORM=$(_PYTHON_HOST_PLATFORM) PYTHONPATH=$(shell test -f pybuilddir.txt && echo $(abs_builddir)/`cat pybuilddir.txt`:)$(srcdir)/Lib _PYTHON_SYSCONFIGDATA_NAME=_sysconfigdata_$(ABIFLAGS)_$(MACHDEP)_$(MULTIARCH) python
```

so it can be modified, or make an alias could be made so that python calls actually call python.exe?

TheSin-:
So I'm continuing to investigate, I know I'm posting lots but I keep hoping that I'll post the right thing for someone that understands and knows the build process.

oddly enough I do not have python, but the docker image does have python3 and python3.6 interpreters.  But the python3 configure script seems to find python.

```
if test "$cross_compiling" = yes; then
    { $as_echo "$as_me:${as_lineno-$LINENO}: checking for python interpreter for cross build" >&5
$as_echo_n "checking for python interpreter for cross build... " >&6; }
    if test -z "$PYTHON_FOR_BUILD"; then
        for interp in python$PACKAGE_VERSION python3 python; do
            which $interp >/dev/null 2>&1 || continue
            if $interp -c "import sys;sys.exit(not '.'.join(str(n) for n in sys.version_info[:2]) == '$PACKAGE_VERSION')"; then
                break
            fi
            interp=
        done
        if test x$interp = x; then
            as_fn_error $? "python$PACKAGE_VERSION interpreter not found" "$LINENO" 5
        fi
```

for some reason this code int eh configure returns `python`

```
configure:2917: checking for python interpreter for cross build
configure:2930: result: python
```

```
buildozer@7fad8d7bf18a:~/build/.buildozer/android/platform/build/build/other_builds/python3-libffi-openssl-sqlite3/armeabi-v7a__ndk_target_21/python3$ which python3
/usr/bin/python3
buildozer@7fad8d7bf18a:~/build/.buildozer/android/platform/build/build/other_builds/python3-libffi-openssl-sqlite3/armeabi-v7a__ndk_target_21/python3$ which python
buildozer@7fad8d7bf18a:~/build/.buildozer/android/platform/build/build/other_builds/python3-libffi-openssl-sqlite3/armeabi-v7a__ndk_target_21/python3$ 
```

and there is no option for it to use the hostpython3 which is built just before it, which would then be ../../../../hostpython3/desktop/hostpython3/native-build/python.exe

but patching for that would be insane, any other ideas?  I still have no idea how it's finding python vs python3 :\. Maybe it's being passed in via configure param or ENV?

TheSin-:
found the issue, it's the in the python configure script.

```
for interp in python$PACKAGE_VERSION python3 python; do
     which $interp >/dev/null 2>&1 || continue
```

it continues cause it's not found.  BUT interp is already set, so by default it will finish the loop with interp='python' even if it's not found.

So no matter what it'll think it found python when in reality it didn't find anything cause it needs to look for python.exe

TheSin-:
Testing with this setup, no idea if I'm out to lunch here but figured I'd help try to solve it, I still have the original PR on as well as this patch.

```
--- a/pythonforandroid/recipes/python3/__init__.py	2019-08-06 02:39:19.595367919 +0000
+++ b/pythonforandroid/recipes/python3/__init__.py	2019-08-06 02:38:58.098543863 +0000
@@ -1,6 +1,7 @@
 import sh
 from pythonforandroid.python import GuestPythonRecipe
+from pythonforandroid.patching import is_darwin
 
 
 class Python3Recipe(GuestPythonRecipe):
@@ -27,6 +28,9 @@
     if sh.which('lld') is not None:
         patches = patches + ["patches/remove-fix-cortex-a8.patch"]
 
+    if is_darwin():
+        patches = patches + ["patches/fix-interpreter-for-darwin.patch"]
+
     depends = ['hostpython3', 'sqlite3', 'openssl', 'libffi']
     conflicts = ['python3crystax', 'python2']
 
```

fix-interpreter-for-darwin.patch
```
diff --git a/configure b/configure
--- a/configure	2019-08-06 02:35:17.000000000 +0000
+++ b/configure	2019-08-06 02:36:22.000000000 +0000
@@ -2868,7 +2868,7 @@
 # pybuilddir.txt will be created by --generate-posix-vars in the Makefile
 rm -f pybuilddir.txt
 
-for ac_prog in python$PACKAGE_VERSION python3 python
+for ac_prog in python$PACKAGE_VERSION python3 python.exe python
 do
   # Extract the first word of "$ac_prog", so it can be a program name with args.
 set dummy $ac_prog; ac_word=$2
@@ -2917,7 +2917,7 @@
     { $as_echo "$as_me:${as_lineno-$LINENO}: checking for python interpreter for cross build" >&5
 $as_echo_n "checking for python interpreter for cross build... " >&6; }
     if test -z "$PYTHON_FOR_BUILD"; then
-        for interp in python$PACKAGE_VERSION python3 python; do
+        for interp in python$PACKAGE_VERSION python3 python.exe python; do
 	    which $interp >/dev/null 2>&1 || continue
 	    if $interp -c "import sys;sys.exit(not '.'.join(str(n) for n in sys.version_info[:2]) == '$PACKAGE_VERSION')"; then
 	        break
```

TheSin-:
okay bigger issue, because I'm running this in a docker on Darwin, is_darwin() does not return true but docker is still bleeding enough through to get python.exe and case insensitivity.  Maybe host python defaults to python.exe cause it detect the fs being case insensitive?

inclement:
@TheSin- That's great debugging, thanks!

TheSin-:
Thanks I just hope it helps.  It's an odd issue but it'll be nice to have it solved I think.

opacam:
@inclement, so sorry for the delay in the response in here, I was on vacations until yesterday, later I will force push the changes which will fix this issue once and for all (I hope...)

@TheSin-, so many **thanks for your debugging session and for the pyton3's patch** :+1: (I will include that in the next `force push` I will make).Honestly, I don't know how the hell you managed to test this OSX build with docker (as far as I know it's not possible to do so, unless you use qemu inside a docker image...but I may be wrong, so the darwin patch not applied as you commented because the kernel isn't the one expected nor the android's NDK/SDK are the Mac OS ones)...nevertheless... your conclusions are right and has been crucial to fix this :smile:

TheSin-:
yeah it took me a bit to figure it all out since I'm so new to all this, docker and python-for-android, but it's a great tool so I wanted to help fix it for everyone.  I'm glad I was able to help and once the push is available I'll give it all an other shot and let you know.  I still have it all setup and ready.

opacam:
@inclement, I force pushed some changes and added a new commit, now I'm pretty sure that the build success, because I created a CI test for Mac OS X which will force the `Python 3 armeabi-v7a` test to be performed with Mac OSX instead of linux.  Honestly, I'm not sure if we must do this test (I don't like to dance with the apple's rhytm...)...we should probably decide this with the team (@AndreMiras, you are the missing piece since @Jtc0de is already in here... and I suspect that you will want to say something about this OSX test for CI)...so guys...let me know what do you think about this...

Anyway, at least now we are sure that the build for Mac OS X succeeds :smile:...it will work the generated apk? ...I guess that we will find out...:crossed_fingers:

@TheSin-, I think with the added CI test it may not be needed to test it as you did, but I will be happy to read and comment your reflexions :smile: if you test it...we may found other issues related with Mac OSX build...

**note:** I patched python2 for Mac OSX as @TheSin- did, but didn't test the build...teorically it should work...

pythonforandroid/python.py
```diff
@@ -404,22 +438,25 @@ def build_arch(self, arch):
...
-            info('Skipping {name} ({version}) build, as it has already '
-                 'been completed'.format(name=self.name, version=self.version))
+            shprint(sh.make, '-j', str(cpu_count()), '-C', build_dir)
+
+        # Since we got different python's executable names depending on the
+        # build platform, we add a a debug message which prints the generated
+        # files/folders, in case that we need that...
+        python_dir = join(self.get_build_dir(arch.arch), self.build_subdir)
+        info("Files/folders in python's build directory are:")
+        shprint(sh.ls, python_dir)
```
opacam: @inclement, [about the generated python.exe for Mac OSX](https://github.com/kivy/python-for-android/pull/1951#pullrequestreview-270513660)

I read it in the [build docs for python](https://github.com/python/cpython/blob/3.7/README.rst#build-instructions). Now I also added an `ls command` to the hostpython build so we see the generated files/folders. I checked it by enabling a debug build in ci for OSX in another branch and `python.exe` is there...:thinking:...should we remove those debug lines?

pythonforandroid/python.py
opacam: The circumstances it may be different now since the last force push, now we use this property in `PythonRecipe` to determine the right hostpython location (https://github.com/kivy/python-for-android/blob/84adc33df4d32326744ca06357c8aed5d487378d/pythonforandroid/recipe.py#L781). If you still thing that we must set this property private let me know and I will change it, no problem :wink:

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -23,9 +24,10 @@ class Python3Recipe(GuestPythonRecipe):
     name = 'python3'
 
     patches = ["patches/fix-ctypes-util-find-library.patch"]
-
```
opacam: Aaaa...I removed this line break...I think that improves the reading....

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -23,9 +24,10 @@ class Python3Recipe(GuestPythonRecipe):
...
     if sh.which('lld') is not None:
-        patches = patches + ["patches/remove-fix-cortex-a8.patch"]
+        patches += ["patches/remove-fix-cortex-a8.patch"]
```
opacam: shortened a little this line...to be more readable again and match the ones I added below

pythonforandroid/python.py
```diff
@@ -382,6 +390,32 @@ class HostPythonRecipe(Recipe):
...
+        platform since, as explained in python docs, some platforms generates a
+        different executable name to avoid conflicts with case insensitive File
+        systems (like Mac Os X).
+
+        .. seealso::
+            https://github.com/python/cpython/blob/3.7/README.rst#build-instructions
+
+        '''
+        if is_darwin():
+            return 'python.exe'
```
TheSin-: I'd like to go back to my comments, this will break some Darwin installs.

This patch should not be based on darwin tests at all, darwin can has both insensitive or sensitive for both HFS+ and AFS, so in the case of a casesensitive darwin install, this will break.

A new test for FS case sensitivity should be make and used for both he patches and the binary name IMHO

is_darwin() should be is_caseinsensitivefs()

TheSin-:
here is the results of my current tests, but I need to revisit a bit of it.  Other then what I already commented on.

```
_PYTHON_PROJECT_BASE=/home/buildozer/build/.buildozer/android/platform/build/build/other_builds/python3-libffi-openssl-sqlite3/armeabi-v7a__ndk_target_21/python3/android-build _PYTHON_HOST_PLATFORM=linux-arm PYTHONPATH=/home/buildozer/build/.buildozer/android/platform/build/build/other_builds/python3-libffi-openssl-sqlite3/armeabi-v7a__ndk_target_21/python3/Lib _PYTHON_SYSCONFIGDATA_NAME=_sysconfigdata_m_linux_ python -S -m sysconfig --generate-posix-vars ;\
if test $? -ne 0 ; then \
	echo "generate-posix-vars failed" ; \
	rm -f ./pybuilddir.txt ; \
	exit 1 ; \
fi
/bin/sh: 1: python: Permission denied
```

```
$ ls python3/android-build/
Makefile      Objects   config.log     libpython3.7m.a   python-config
Makefile.pre  Parser    config.status  libpython3.7m.so  python-config.py
Misc          Programs  crtbegin_so.o  libpython3.so     python.exe
Modules       Python    crtend_so.o    pyconfig.h        python.exe-gdb.py
```

so it did build it, but it's trying to run python instead of python.exe and causing the build to fail.  I'm hading home but I'll diagnose more once I'm there.

OptimusGREEN:
Hi guys,

Tested this (first time in months) and it built ok. my app is however crashing out at startup with no code changes so wondering if its something here.

here's my logcat

https://pastebin.com/raw/nX6VfmAC

looks like somthing to do with lib python so

opacam:
aaa...that's good news @OptimusGREEN , we have the build...

but I still not convinced of the work I made in here, I'm testing in another branch a different approach (considering the last comments of @TheSin- ) and it will simplify the whole stuff...I will push it later...

regarding the runtime error...what kivy version do you use...`master`?

if that is the case I would recommend you to stick to p4a version...there was some conversation on discord (yesterday) and the error I see it's pretty similar (hard to tell with this kind of libc errors...)

If not...well...we have another problem...

OptimusGREEN:
yeah its kivy master and my p4a fork rebased at this commit as didn't think it had been merged to p4a proper.

opacam:
Ok , guys, travis seems happy with all python versions and OS tested (darwin and linux). So we should have a successful build with Mac OSX.

The thing works by making a copy of our compiled hostpython binary (@TheSin-, [your first choice](https://github.com/kivy/python-for-android/pull/1951#issuecomment-518476640)...thanks again :smile:) in such a way that will reflect the major version (because the python's configure scripts considers that possibility...so we avoid the patching, except for `python2`'s recipe, which haves a wrong version of python :disappointed:).

So...@TheSin- , no more [is_darwin](https://github.com/kivy/python-for-android/pull/1951#pullrequestreview-276167513) :wink:...let see how it goes your docker tests now :grimacing: ...could you do that please?

@OptimusGREEN, it would be great that you could test it now, but with kivy's version `1.11.1` (the p4a's default)...because we know that version is working...the issue you got at runtime it maybe kivy's related (I'm not sure of this...haven't had time to test it)

TheSin-:
I'm on it, I'm just securing a new shelving unit to my wall and I'll get a build started, thanks for all your hard work on this I'm very excited to be able to do my full development natively!

TheSin-:
100% Thumbs up from me, built and ran perfectly!!!  Great job!

OptimusGREEN:
I used Kivy 1.11.1 this time and I seemed to have the same issue (it maybe isn't but looks like it to me) so maybe my issue isn't related to this? It would be strange if it wasn't though as I haven't changed any code since I last successfully built on a linux vm a few months back.

here's my logcat 
https://pastebin.com/raw/qbDkhJXf

opacam:
> 100% Thumbs up from me, built and ran perfectly!!! Great job!

@TheSin- , **¡¡¡thanks for all your work!!!** in here :smile:, I consider this a collaboration work :+1:!!!...in matter fact...you did the hard work in here.

> here's my logcat

@OptimusGREEN, the log you posted isn't useful (at least to me...). This `libc` error at runtime would probably require create a gdb debugable build...but this will require another PR which isn't merged yet. But before that...this error could also be caused due a previous  p4a builds you got in your system (we have a problem in there...we are working on that also). If you didn't clean up the build environment before the last attempt you made, then do it and try it again...so...did you cleaned up your build environment before your tests? (using buildozer: `buildozer android clean`).

**Note:** If after the clean up and rebuilt still doesn't work, I would try to isolate the problem by building a simple apk (I would recommend to use one of our testapps: [setup_testapp_python3_sqlite_openssl.py](https://github.com/kivy/python-for-android/blob/develop/testapps/setup_testapp_python3_sqlite_openssl.py). If that apk doesn't work and fails with the same error you got...well...we  probably have a bug with Mac OS X, but if that apk works, the problem may be in some recipe do you use or in your app's code...

**As a side notes:**
  -  @OptimusGREEN, it would be useful to have the debug's log of the build...with buildozer it will be:  `buildozer android debug --debug`
  - BTW, if you install the last buildozer version from master branch, you can [easily test any p4a's PR](https://github.com/kivy/buildozer/pull/940) by editing your `buildozer.spec` file, so for this one, you would use:
    ```
    p4a.fork = opacam
    p4a.branch = hotfix-macos-build
    ```
    just remember to comment or eddit  those lines again once you are done or this merged, since I supose that you don't want to get stuck in here :wink:

.travis.yml
```diff
@@ -69,15 +69,59 @@ jobs:
...
+        - yes | $ANDROID_SDK_HOME/tools/bin/sdkmanager "build-tools;$ANDROID_SDK_BUILD_TOOLS_VERSION" > /dev/null
+        - yes | $ANDROID_SDK_HOME/tools/bin/sdkmanager "platforms;android-$ANDROID_API_LEVEL" > /dev/null
+        - pip3 install --upgrade Cython==0.28.6
+        - pip3 install -e .
+      before_script:
+        # overrides `testing` for Mac OS since we don't use docker
+        - while sleep 540; do echo "==== Still running (travis, don't kill me) ===="; done &
+      script:
+        - cd testapps/
+        - python3 setup_testapp_python3_sqlite_openssl.py apk --sdk-dir $ANDROID_SDK_HOME --ndk-dir $ANDROID_NDK_HOME --requirements libffi,sdl2,pyjnius,kivy,python3,openssl,requests,sqlite3,setuptools --arch=armeabi-v7a --debug
```
opacam: Augh...I forgot to remove `--debug` from here...for the record then... [here it is the travis build](https://travis-ci.org/kivy/python-for-android/jobs/573210450) for **Mac OSX**

Removing `--debug`...

OptimusGREEN:
yeah i had removed my .buildozer directory altogether.

Nice tips with the p4a.fork thing. i didn't know bout that, really handy.

I'll give it another try

OptimusGREEN:
Ok so tried again and build was successful as previous but again the app crashes on launch with the same libpython37.so error.

output from build https://paste.ee/p/W9xz8

I also tried building on linux and everything was fine and app ran ok so there is defo still a mac issue. It might be a library or something but def only happening on my mac.

my requirements:

requirements = python3,kivy==1.11.1,pyopenssl,openssl,urllib3,android,pyjnius,dropbox,requests==2.20.0,chardet,certifi,idna

AndreMiras:
Impressive work! I've been busy lately and will still be for a little while.
From a quick look, well I like that we introduce CI tests if Apple users is a significant portion of our user base. What hurts me however is to see the `.travis.yml` file growing that much overtime. And I do mean "to see the .travis.yml file". I mean probably moving some of the logic to dedicated bash or Python script would probably peace my mind. I feel like it could help by keeping things a bit more isolated, but I'm not sure if it makes sense and if it doesn't introduce more overhead.
Thanks again for the effort, I'll try to take a deeper look throughout the week.

.travis.yml
```diff
@@ -69,15 +69,59 @@ jobs:
...
+        # download/extract android's NDK
+        - curl --location --progress-bar --insecure $ANDROID_NDK_DL_URL --output $ANDROID_NDK_ARCHIVE
+        - mkdir -p $ANDROID_NDK_HOME_V
+        - unzip -q $ANDROID_NDK_ARCHIVE -d $HOME
+        - ln -sfn $ANDROID_NDK_HOME_V $ANDROID_NDK_HOME
+        - rm -rf $ANDROID_NDK_ARCHIVE
+        # download/extract android's SDK
+        - curl --location --progress-bar --insecure $ANDROID_SDK_TOOLS_DL_URL --output $ANDROID_SDK_TOOLS_ARCHIVE
+        - mkdir -p $ANDROID_SDK_HOME
+        - unzip -q $ANDROID_SDK_TOOLS_ARCHIVE -d $ANDROID_SDK_HOME
```
AndreMiras: It feels like this whole download thing could go to some helper scripts maybe under `ci/`. And if not too painful, it could also be OS agnostic. That could also actually be part of some `Makefile` that could be easily reused inside or outside CI. For instance something like https://github.com/AndreMiras/dockerfiles/blob/51a88e3/buildozer_android/android_sdk.mk

.travis.yml
```diff
@@ -69,15 +69,59 @@ jobs:
...
+        - rm -rf $ANDROID_SDK_TOOLS_ARCHIVE
+        - ls $HOME
```
AndreMiras: Minor: is this for debugging purpose?

pythonforandroid/python.py
```diff
@@ -382,6 +391,30 @@ class HostPythonRecipe(Recipe):
...
     change depending on the python version set in attribute :attr:`version`.'''
 
+    @property
+    def _exe_name(self):
+        '''
+        Returns the name of the python executable depending on the version.
+        '''
+        if not self.version:
+            raise BuildInterruptingException(
+                'The hostpython recipe must has set version'
```
AndreMiras: Minor: s/has/have/

pythonforandroid/python.py
```diff
@@ -404,22 +437,28 @@ def build_arch(self, arch):
...
-
-                shprint(sh.make, '-j', str(cpu_count()), '-C', build_dir)
-        else:
-            info('Skipping {name} ({version}) build, as it has already '
-                 'been completed'.format(name=self.name, version=self.version))
-
-        self.ctx.hostpython = join(build_dir, 'python')
+        with current_directory(recipe_build_dir):
+            # Configure the build
+            with current_directory(build_dir):
```
AndreMiras: Why do we switch current directory twice in a row? Is it because the second time it's a relative path?

pythonforandroid/recipes/python2/__init__.py
```diff
@@ -32,7 +32,9 @@ class Python2Recipe(GuestPythonRecipe):
                'patches/fix-gethostbyaddr.patch',
                'patches/fix-posix-declarations.patch',
                'patches/fix-pwd-gecos.patch',
-               'patches/fix-ctypes-util-find-library.patch']
+               'patches/fix-ctypes-util-find-library.patch',
+               'patches/fix-interpreter-version.patch',
+    ]
```
AndreMiras: Wait are we still spending time trying to fix Python2 related things?
Well now it's done, but I wouldn't have bothered

opacam:
@AndreMiras and @inclement, **thanks for your reviews** :smile:

...I was hoping that you guys don't want this **Mac OS X** test :laughing:. I already expressed my feelings about mac, so the main goal was only to be sure that the thing worked (at least at build time). But since there is interest on have this test, I agree with you guys, the **Mac OS X** logic has to be moved somewhere else, isolating the android part (so we can refactor as well the android part of our docker files). I liked the @AndreMiras's idea of `Makefiles` so I've been working on that, later I will force push the changes :wink:

ci/makefiles/android.mk
```diff
@@ -0,0 +1,78 @@
...
+
+# We must provide a platform (darwin or linux) and we need JAVA_HOME defined
+ifndef target_os
+    $(error target_os is not set...aborted!)
+endif
+
+# Those android NDK/SDK variables can be override when running the file
+ifndef ANDROID_NDK_VERSION
+    ANDROID_NDK_VERSION=17c
+endif
```
AndreMiras: What about the shorthand default? e.g.
```
ANDROID_NDK_VERSION ?= 17c
```

Dockerfile.py3
```diff
@@ -127,9 +80,12 @@ RUN pip2 install --upgrade Cython==0.28.6
 
 WORKDIR ${WORK_DIR}
 COPY --chown=user:user . ${WORK_DIR}
-RUN chown --recursive ${USER} ${ANDROID_SDK_HOME}
+RUN mkdir ${ANDROID_HOME} && chown --recursive ${USER} ${ANDROID_HOME}
 USER ${USER}
 
+# Download and install android's NDK/SDK
+RUN make -f ci/makefiles/android.mk target_os=linux
```
AndreMiras: Nice! I'm just wondering, this layer would be cached entirely then correct? This is fine, I wasn't sure if it would be cached or not

ci/makefiles/android.mk
```diff
@@ -0,0 +1,78 @@
...
+
+download_android_ndk:
+	curl --location --progress-bar --continue-at - \
+	$(ANDROID_NDK_DL_URL) --output $(ANDROID_NDK_ARCHIVE)
+
+# Extract android SDK and remove the compressed file
+extract_android_sdk:
+	mkdir -p $(ANDROID_SDK_HOME) \
+	&& unzip -q $(ANDROID_SDK_TOOLS_ARCHIVE) -d $(ANDROID_SDK_HOME) \
+	&& rm -f $(ANDROID_SDK_TOOLS_ARCHIVE)
```
AndreMiras: Minor for later.
I'm wondering if we do want to clean it or use it as a target for not re-downloading it next time. So basically the idea wold be to have something like:
```mk
$(ANDROID_SDK_TOOLS_ARCHIVE):
	curl --location --progress-bar --continue-at - \
	$(ANDROID_SDK_TOOLS_DL_URL) --output $(ANDROID_SDK_TOOLS_ARCHIVE

download_android_sdk: $(ANDROID_SDK_TOOLS_ARCHIVE)
```
That way running the target twice won't run the download twice if already present. Same story for other targets/dependencies

ci/makefiles/android.mk
```diff
@@ -0,0 +1,78 @@
+# Downloads and installs the Android SDK depending on supplied platform: darwin or linux
```
AndreMiras: Nice DRY file! Pretty useful :clap:

Dockerfile.py3
opacam: > ...this layer would be cached entirely then correct? This is fine, I wasn't sure...

I hope so...but I don't know for sure...so...:no_mouth:

ci/makefiles/android.mk
opacam: I was thinking in keeping docker's layer smallest as possible...since we don't to re use the removed files right now...so ...maybe better to change/test this if we need it, right? :wink:

ci/makefiles/android.mk
opacam: Also changed other entries from `android.mk` and `osx.mk` files :wink:

@AndreMiras, thank you for the tip!!!  :smile:

pythonforandroid/recipes/python2/__init__.py
opacam: > Wait are we still spending time trying to fix Python2 related things?

Well...we still support it right?...2020 is almost here...:smile:

pythonforandroid/python.py
opacam: > Why do we switch current directory twice in a row? Is it because the second time it's a relative path?

I'm not pretty sure of this, since I only removed one indentation level, but, apart of relative paths, I think that it's because we need to write `config.status` inside `native-build` folder...also all the thing makes sense in one context because it's a part of the build process of  python.

Anyway...maybe @inclement can add some words or correct me. He did  the original **hostpython3/python3** recipes :smile: (I only refactored a little, some months back, to have the python2 version of the thing based on the work He made...but the concept and logic are the @inclement's ones :wink:)

ci/makefiles/android.mk
AndreMiras: Yes let's keep it that way :smile:

.travis.yml
```diff
@@ -65,19 +60,42 @@ jobs:
         # case that the travis log doesn't produce any output for more than 10 minutes
         - while sleep 540; do echo "==== Still running (travis, don't kill me) ===="; done &
       script:
-        - docker run -e CI -e TRAVIS_JOB_ID="$TRAVIS_JOB_ID" -e TRAVIS_BRANCH="$TRAVIS_BRANCH" p4a /bin/sh -c "$COMMAND"
+        - >
+          docker run
+          -e CI
+          -e TRAVIS_JOB_ID="$TRAVIS_JOB_ID"
+          -e TRAVIS_BRANCH="$TRAVIS_BRANCH"
```
AndreMiras: Just for your info, you can simplify to `-e TRAVIS_BRANCH` since variables are the same you can skip the `=` part. Just like we did for `-e CI`

Dockerfile.py2
```diff
@@ -37,64 +37,21 @@ ENV RETRY="retry -t 3 --"
...
 
 # install system dependencies
 RUN ${RETRY} apt -y install -qq --no-install-recommends \
         python virtualenv python-pip wget lbzip2 patch sudo \
-    && apt -y autoremove
+    && apt -y autoremove \
+    && apt -y clean
+
+# Install Java, set JAVA_HOME (to accept android's SDK licenses) and clean
+RUN ${RETRY} apt -y install openjdk-8-jdk && apt -y autoremove && apt -y clean
```
AndreMiras: Just wondering, what about `--no-install-recommends`

ci/makefiles/osx.mk
```diff
@@ -0,0 +1,23 @@
+# installs java 1.8, android's SDK/NDK, cython and p4a
+
+# The following variable/s can be override when running the file
+ANDROID_HOME ?= $(HOME)/.android
+
+all: install_java upgrade_cython install_android_ndk_sdk install_p4a
+
+install_java:
```
AndreMiras: This is to think about later, but installing system requirements could actually be part Linux Makefile as well

.travis.yml
```diff
@@ -65,19 +60,42 @@ jobs:
...
     - <<: *testing
-      name: Python 3 arm64-v8a
-      env: COMMAND='. venv/bin/activate && cd testapps/ && python setup_testapp_python3_sqlite_openssl.py apk --sdk-dir $ANDROID_SDK_HOME --ndk-dir $ANDROID_NDK_HOME --requirements libffi,sdl2,pyjnius,kivy,python3,openssl,requests,sqlite3,setuptools' --arch=arm64-v8a
+      name: Python 3 armeabi-v7a
+      os: osx
+      osx_image: xcode11  # since xcode1.3, python3 is the default interpreter
+      before_script:
+        # installs java 1.8, android's SDK/NDK and p4a
+        - make -f ci/makefiles/osx.mk
+        - export JAVA_HOME=/Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home
```
AndreMiras: Just wondering, is needed for building the app in the `script` part below?

.travis.yml
```diff
@@ -65,19 +60,42 @@ jobs:
...
+        # Run a background process to make sure that travis will not kill our tests in
+        # case that the travis log doesn't produce any output for more than 10 minutes
+        - while sleep 540; do echo "==== Still running (travis, don't kill me) ===="; done &
```
AndreMiras: That might not be needed for the OSX build actually, because I don't think it would run more than 20 or 30 minutes without log printing.
`travis_wait` waits 20 minutes by default and can be increased up to 30 minutes by passing time as parameter.

Dockerfile.py2
```diff
@@ -37,64 +37,21 @@ ENV RETRY="retry -t 3 --"
...
     PATH="${HOME_DIR}/.local/bin:${PATH}"
 
 # install system dependencies
 RUN ${RETRY} apt -y install -qq --no-install-recommends \
         python virtualenv python-pip wget lbzip2 patch sudo \
-    && apt -y autoremove
+    && apt -y autoremove \
+    && apt -y clean
+
+# Install Java, set JAVA_HOME (to accept android's SDK licenses) and clean
```
AndreMiras: I lost track of this accept part, but it actually is in the Makefile (`update_android_sdk`). Maybe at some point the system dep could also move to the Makefile. I don't know if that would be a good idea or not

ci/makefiles/android.mk
```diff
@@ -0,0 +1,68 @@
...
+ANDROID_NDK_DL_URL=https://dl.google.com/android/repository/$(ANDROID_NDK_ARCHIVE)
+
+$(info Target install OS is          : $(target_os))
+$(info Android SDK home is           : $(ANDROID_SDK_HOME))
+$(info Android NDK home is           : $(ANDROID_NDK_HOME))
+$(info Android SDK download url is   : $(ANDROID_SDK_TOOLS_DL_URL))
+$(info Android NDK download url is   : $(ANDROID_NDK_DL_URL))
+$(info Android API level is          : $(ANDROID_API_LEVEL))
+$(info Android NDK version is        : $(ANDROID_NDK_VERSION))
+$(info JAVA_HOME is                  : $(JAVA_HOME))
```
AndreMiras: Great idea to log this!

pythonforandroid/python.py
AndreMiras: Thanks, I was just curious, maybe if I get to curious I'll play with it later. But let's get this PR merged first :smile:

Dockerfile.py2
opacam: Better, so I changed this line with `-qq --no-install-recommends` (so we keep our logs smaller) :+1:

.travis.yml
opacam: Yes, we need it because the default java version for travis's `OS X` image (xcode11) has another [version of java](https://docs.travis-ci.com/user/reference/osx/#macos-version)

See also:  kivy/buildozer#945

.travis.yml
opacam: I maintained the background process...I think that could be useful..., but I removed one line of the above comment (since we already commented before)...but we can remove it if you feel this is useless...just let me know :wink:

PR #1952, opened 2019-08-03
Unit tests Recipe.download_file() partly

AndreMiras:
Follow up of #1946.
Increases coverage by testing part of `Recipe.download_file()` handling
https schema. Next up would be to handle more schemas in the tests.
Also addressed @opacam comments from previous pull requests.

### DIFF
--- a/tests/test_recipe.py
+++ b/tests/test_recipe.py
@@ -1,4 +1,5 @@
 import os
+import pytest
 import types
 import unittest
 import warnings
@@ -18,6 +19,10 @@ def patch_logger_info():
 
 def patch_logger_debug():
     return patch_logger('debug')
+
+
+def patch_urlretrieve():
+    return mock.patch('pythonforandroid.recipe.urlretrieve')
 
 
 class DummyRecipe(Recipe):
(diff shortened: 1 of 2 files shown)


tests/test_recipe.py
```diff
@@ -122,3 +140,39 @@ def test_download(self):
...
+        Verifies `urlretrieve()` is being called on https downloads.
+        """
+        recipe, filename = self.get_dummy_python_recipe_for_download_tests()
+        url = recipe.url
+        with (
+                patch_urlretrieve()) as m_urlretrieve, (
+                tempfile.TemporaryDirectory()) as temp_dir:
+            recipe.ctx.setup_dirs(temp_dir)
+            assert recipe.download_file(url, filename) == filename
+        assert m_urlretrieve.call_args_list == [mock.call(
```
inclement: Nicer to put mock.call on the next line with everything else.

AndreMiras: Thanks addressed :+1:

inclement:
Thanks!

Issue #1949, opened 2019-08-01
How to overcome:? #error "LONG_BIT definition appears wrong for platform (bad gcc/glibc config?)."


c0ntradicti0n:
I tried to package my app, that has a few dependencies, with buildozer and p4a a few times, but I run again and again into some errors. 

First I get an error, that <crypt.h> is not there. I found this workaround [https://github.com/kivy/python-for-android/issues/1575](https://github.com/kivy/python-for-android/issues/1575), that seems to work, because that brings me to the next error.
So I change the 
`~/.local/share/python-for-android/build/other_builds/hostpython3/desktop/hostpython3/native-build/pyconfig.h`
and comment the line with `#define HAVE_CRYPT_H 1` and come to:

The next error seems to complain, that my architecture doesn't fit. I don't know, if its the architecture of the NDK (linux x86_64), the compiling machine (linux x86_64) or the android target -- I assume, that is armeabi?  

I declared the dependencies for python for android in the config with: 
`--requirements=python3,kivy,regex,more_itertools,twisted`

Is that way correct or what I have to do more to get some dependencies compiled?
Is it true, that one needs to use python-for-android, when having some other dependencies as numpy instead of buildozer?
I used that pip archive for installing python for android. should I use the master branch instead?

I used Android SDK 28 and Android NDK r17c. These logs are:

`
$ p4a apk --log_level = 2
[INFO]:    Reading .p4a configuration
[WARNING]: --ndk-version is deprecated and no longer necessary, the value you passed is ignored
[WARNING]: $ANDROIDNDKVER is deprecated and no longer necessary, the value you set is ignored
[INFO]:    Will compile for the following archs: armeabi-v7a
[INFO]:    Getting Android API version from user argument: 29
[INFO]:    Available Android APIs are (28, 29)
[INFO]:    Requested API target 29 is available, continuing.
[INFO]:    Getting NDK dir from from user argument
[INFO]:    Found NDK revision 17.2.4988734
[INFO]:    Getting NDK API version (i.e. minimum supported API) from user argument
[INFO]:    Found virtualenv at /home/finn/CorpusCookApp/venv/bin/virtualenv
[INFO]:    ccache is missing, the build will not be optimized in the future.
[INFO]:    Found the following toolchain versions: ['4.9']
[INFO]:    Picking the latest gcc toolchain, here 4.9
[INFO]:    No existing dists meet the given requirements!
[INFO]:    No dist exists that meets your requirements, so one will be built.
[...]

PR #1958, opened 2019-08-15
Fix and update regex's recipe

opacam:
Closes: #1949

### DIFF
--- a/pythonforandroid/recipes/regex/__init__.py
+++ b/pythonforandroid/recipes/regex/__init__.py
@@ -3,8 +3,8 @@ from pythonforandroid.recipe import CompiledComponentsPythonRecipe
 
 class RegexRecipe(CompiledComponentsPythonRecipe):
     name = 'regex'
-    version = '2017.07.28'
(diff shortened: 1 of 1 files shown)


pythonforandroid/recipes/regex/__init__.py
```diff
@@ -3,10 +3,15 @@
 
 class RegexRecipe(CompiledComponentsPythonRecipe):
     name = 'regex'
-    version = '2017.07.28'
-    url = 'https://pypi.python.org/packages/d1/23/5fa829706ee1d4452552eb32e0bfc1039553e01f50a8754c6f7152e85c1b/regex-{version}.tar.gz'
+    version = '2019.06.08'
+    url = (
+        'https://files.pythonhosted.org/packages/6f/4e/'
```
opacam: I'm not sure if we should split the url line as I did, to keep us below 80 characters (I think that We already did that in some recipes...:thinking:...I don't remember now)...anyway...We could also use a `noqa` marker instead.

AndreMiras: Anyway I would not put that has since it's often changing. There's a hack to make it work, try that:
```python
url = 'https://pypi.python.org/packages/source/r/regex/regex-{version}.tar.gz'
```

PR #1963, opened 2019-08-22
Bootstrap backend for Android Library project

### LINKED ISSUE #1957: Building an Android Library
Is it possible to build an Android Library (AAR) with Python-For-Android?

There is a python package which I want to use as a library in existing Android application. I tried to integrate it in Kivy demo app and it is working. However I can't find a way to produce AAR instead of APK.

xuhcc:
Added new bootstrap backend for packaging python service as Android library.

Resolves #1957

I'm not familiar with python-for-android codebase, so I tried to keep changes to existing modules to a minimum. Some files (e.g. jtar classes) are simply copied from the service_only bootstrap. I'm not sure what's the best approach here. Maybe we can move them to `bootstraps/common`?


pythonforandroid/bootstraps/library/build/templates/Service.tmpl.java
```diff
@@ -0,0 +1,113 @@
+package {{ args.package }};
```
AndreMiras: Are you sure this file is needed?

AndreMiras:
Looking good overall. I'd still like to dig more before we can merge. I'm just pretty sad to see we still need quite a lot of copy pasta for adding new bootstrap. We were supposed to mitigate it via https://github.com/kivy/python-for-android/pull/1496 and https://github.com/kivy/python-for-android/pull/991 but we still have a lot to do it seems.

inclement:
I'm not clear from the code alone, what exactly is new/different in this bootstrap compared to (say) the service_only one? Would it be possible to make the aar thing a build option instead of a bootstrap?

pythonforandroid/bootstraps/library/build/templates/Service.tmpl.java
xuhcc: This file contains code for setting service properties (like its entry point) and code for unpacking python bundle. I'm not sure that it is needed, but I don't know how to do these things without it.

xuhcc:
@inclement 

The main difference is that in `service_only` bootstrap services depend on [PythonActivity](https://github.com/kivy/python-for-android/blob/develop/pythonforandroid/bootstraps/service_only/build/src/main/java/org/kivy/android/PythonActivity.java) class. You need to start activity to extract python bundle and only after that you can start services.

For `library` bootstrap I moved the `unpackData` method to service class itself and added custom `AssetExtract` class which depend on `Context` instead of `Activity`.

I think it is possible to make it a build option for `service_only` bootstrap, but I'm not sure how to do it correctly.

inclement:
Thanks for the clarifications @xuhcc. My current thinking is that this would be preferable to add as functionality to one or more existing bootstraps, and to call it using a new `p4a aar` command instead of overloading apk (this would be much nicer in general than starting to add if statements based on the bootstrap type). However, I'm saying this from a position of not having really checked the details of how it needs to work.

PR #1969, opened 2019-08-25
[WIP][LIBS - PART VI] Rework of protobuf_cpp

opacam:
This forms part of the `STL` library refactor (#1947), so depends on it.

In here we do:
  - inherit from `CppCompiledComponentsPythonRecipe` Because depends on android's STL library
  - make use of the base class methods for library recipes
  - Split build_arch into proper methods
  - Shorten some long lines (to be PEP8 friendly)
  - make generated library shared

**Not merge before** #1947

### DIFF
--- a/ci/constants.py
+++ b/ci/constants.py
@@ -67,7 +67,6 @@ BROKEN_RECIPES_PYTHON3 = set([
     'ffpyplayer',
     # requires `libpq-dev` system dependency e.g. for `pg_config` binary
     'psycopg2',
-    'protobuf_cpp',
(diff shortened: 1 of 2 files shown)


inclement:
Looks like travis gets a believable-looking build error, any idea what's up with that?

opacam:
Thanks @AndreMiras :smile:

@inclement, I think that the problem you mentioned was caused because of the lack of some dependencies (I remember seeing some gcc error so I suspect that I branched from a point where the NDK r19 thing was not merged). The rebase solved the CI build issue so... let's merge this :smile:

PR #1975, opened 2019-08-31
Added missing recommendations command and cleaned up code

inclement:
Added this missing command, which existed in the argument parser but wasn't actually implemented.

Not sure what format we really want, but just having the command available is probably an improvement for now.

I thought about removing the argument entirely, but it does seem useful to be able to print the p4a recommendations easily.

### DIFF
--- a/pythonforandroid/recommendations.py
+++ b/pythonforandroid/recommendations.py
@@ -226,9 +226,9 @@ def print_recommendations():
     """
     Print the main recommended dependency versions as simple key-value pairs.
     """
-    print(f'Min supported NDK version: {MIN_NDK_VERSION}')
-    print(f'Recommended NDK version: {RECOMMENDED_NDK_VERSION}')
-    print(f'Min target API: {MIN_TARGET_API}')
-    print(f'Recommended target API: {RECOMMENDED_TARGET_API}')
-    print(f'Min NDK API: {MIN_NDK_API}')
-    print(f'Recommended NDK API: {RECOMMENDED_NDK_API}')
+    print('Min supported NDK version: {}'.format(MIN_NDK_VERSION))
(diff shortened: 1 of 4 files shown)


inclement:
Also updated minimum python version supported to 3.6, as originally agreed in https://github.com/kivy/python-for-android/pull/1918 - it's my mistake that I left it at 3.4 in that PR. This minimum version is now essential since I added f-strings to the code.

pythonforandroid/recommendations.py
```diff
@@ -218,3 +220,14 @@ def check_python_version():
     ):
 
         raise BuildInterruptingException(PY_VERSION_ERROR_TEXT)
+
+def print_recommendations():
```
AndreMiras: Linter unhappy :grimacing: 
```
pythonforandroid/recommendations.py:224:1: E302 expected 2 blank lines, found 1
```

pythonforandroid/recommendations.py
```diff
@@ -187,7 +189,7 @@ def check_ndk_api(ndk_api, android_api):
 
 
 MIN_PYTHON_MAJOR_VERSION = 3
-MIN_PYTHON_MINOR_VERSION = 4
+MIN_PYTHON_MINOR_VERSION = 6
```
AndreMiras: Nice nice :smile:

inclement:
This one is also failing only on osx with the pyconfig.h error...not sure what's wrong, although it seems fairly clear this PR is not the culprit.

inclement:
Tests failing only due to macos issue also present on trunk.

inclement:
Thanks!
Looks like we'll have to do something else about this issue breaking the develop branch tests :/

AndreMiras:
For records, this got fixed recently via https://github.com/kivy/python-for-android/pull/2063

PR #1976, opened 2019-09-01
wip

tshirtman:


inclement:
I guess this is to support launcher interaction? Is it tested to work? I'm not certain there couldn't be some weird interactions with when/how sdl2 normally sets orientation.
It would be nice to abstract this out a bit differently, with a single "overrideDefaultsFromIntent" function call that puts everything in one place.

PR #1982, opened 2019-09-12
Proof of concept - A bunch of tests for library recipes

opacam:
The main idea is to test as many recipes as we can with the less code possible and without creating any file/directory so our tests can be performed as fast as possible (all this tests will only add between 2 and 3 seconds to our CI tests and will cover almost 100% of the code for each tested recipe)

To do so, we create a couple of modules:
  - `tests.recipe_ctx`: allow us to create a proper `Context` to test our recipes
  - `tests.recipe_lib_test`: which holds some base classes to be used to test a recipe depending on the build method used. For starters we introduce two kind of base classes:
    + `BaseTestForMakeRecipe`: To test an standard library build (this will iinclude the recipes which requires the classical build commands`configure/make`)
    + `BaseTestForCmakeRecipe`: To test an library recipe which is compiled with `cmake`

We also refactor the existing recipes tests, so we can remove some lines in there...the ones that creates a `Context`.

**Notes:**
  - I'm aware that we could go a little further with the testing...but for starters I thing it's fine...we can do it later
  - This will cover most of the library recipes, but still we will have to cover a little more
  - I tried to create a generic tests depending on the build system, I'm sure that we could do something similar with `PythonRecipe`s, `BootstrapNDKRecipe`s...)
  - This could be split into multiple `dependant` PRs but, first let's see if the general idea is good enough ...then we will deal with the details if convenient :smile:
  -  @AndreMiras, sorry man, this time I made use of `Python Inheritance`...I just feel that is the best we can do in this case, but I'm always open to other solutions :wink:

### DIFF
--- a/pythonforandroid/recipes/jpeg/__init__.py
+++ b/pythonforandroid/recipes/jpeg/__init__.py
@@ -2,7 +2,6 @@ from pythonforandroid.recipe import Recipe
 from pythonforandroid.logger import shprint
 from pythonforandroid.util import current_directory
 from os.path import join
-from os import environ, uname
 import sh
 
 
@@ -35,10 +34,9 @@ class JpegRecipe(Recipe):
                     '-DCMAKE_POSITION_INDEPENDENT_CODE=1',
                     '-DCMAKE_ANDROID_ARCH_ABI={arch}'.format(arch=arch.arch),
                     '-DCMAKE_ANDROID_NDK=' + self.ctx.ndk_dir,
-                    '-DCMAKE_C_COMPILER={toolchain}/bin/clang'.format(
-                        toolchain=env['TOOLCHAIN']),
-                    '-DCMAKE_CXX_COMPILER={toolchain}/bin/clang++'.format(
-                        toolchain=env['TOOLCHAIN']),
+                    '-DCMAKE_C_COMPILER={cc}'.format(cc=arch.get_clang_exe()),
+                    '-DCMAKE_CXX_COMPILER={cc_plus}'.format(
+                        cc_plus=arch.get_clang_exe(plus_plus=True)),
                     '-DCMAKE_BUILD_TYPE=Release',
                     '-DCMAKE_INSTALL_PREFIX=./install',
                     '-DCMAKE_TOOLCHAIN_FILE=' + toolchain_file,
@@ -53,17 +51,6 @@ class JpegRecipe(Recipe):
                     '-DENABLE_STATIC=1',
                     _env=env)
             shprint(sh.make, _env=env)
-
-    def get_recipe_env(self, arch=None, with_flags_in_cc=False):
-        env = environ.copy()
-
-        build_platform = '{system}-{machine}'.format(
-            system=uname()[0], machine=uname()[-1]).lower()
-        env['TOOLCHAIN'] = join(self.ctx.ndk_dir, 'toolchains/llvm/'
-                                'prebuilt/{build_platform}'.format(
-                                    build_platform=build_platform))
-
-        return env
 
 
 recipe = JpegRecipe()
--- a/tests/recipes/recipe_lib_test.py
+++ b/tests/recipes/recipe_lib_test.py
@@ -1,0 +1,110 @@
+from unittest import mock
+from tests.recipes.recipe_ctx import RecipeCtx
+
+
+class BaseTestForMakeRecipe(RecipeCtx):
+    """
+    An unittest for testing any recipe using the standard build commands
+    (`configure/make`).
+
+    .. note:: Note that Some cmake recipe may need some more specific testing
+        ...but this should cover the basics.
+    """
+
+    recipe_name = None
+    recipes = ["python3", "kivy"]
+    recipe_build_order = ["hostpython3", "python3", "sdl2", "kivy"]
+    expected_compiler = (
+        "{android_ndk}/toolchains/llvm/prebuilt/linux-x86_64/bin/clang"
+    )
+
+    sh_command_calls = ["./configure"]
+    """The expected commands that the recipe runs via `sh.command`."""
+
+    extra_env_flags = {}
+    """
+    This must be a dictionary containing pairs of key (env var) and value.
+    """
+
+    def __new__(cls, *args):
+        obj = super(BaseTestForMakeRecipe, cls).__new__(cls)
+        if obj.recipe_name is not None:
+            print(f"We are testing recipe: {obj.recipe_name}")
+            obj.recipes.append(obj.recipe_name)
+            obj.recipe_build_order.insert(1, obj.recipe_name)
+            return obj
+
+    @mock.patch("pythonforandroid.recipe.Recipe.check_recipe_choices")
+    @mock.patch("pythonforandroid.build.ensure_dir")
+    @mock.patch("pythonforandroid.archs.glob")
+    @mock.patch("pythonforandroid.archs.find_executable")
+    def test_get_recipe_env(
+        self,
+        mock_find_executable,
+        mock_glob,
+        mock_ensure_dir,
+        mock_check_recipe_choices,
+    ):
+        """
+        Test that get_recipe_env contains some expected arch flags and that
+        some internal methods has been called.
+        """
+        mock_find_executable.return_value = self.expected_compiler.format(
+                android_ndk=self.ctx._ndk_dir
+        )
+        mock_glob.return_value = ["llvm"]
+        mock_check_recipe_choices.return_value = sorted(
+            self.ctx.recipe_build_order
+        )
+
+        # make sure the arch flags are in env
+        env = self.recipe.get_recipe_env(self.arch)
+        for flag in self.arch.arch_cflags:
+            self.assertIn(flag, env["CFLAGS"])
+        self.assertIn(
+            f"-target {self.arch.target}",
+            env["CFLAGS"],
+        )
+
+        for flag, value in self.extra_env_flags.items():
+            self.assertIn(value, env[flag])
+
+        # make sure that the mocked methods are actually called
+        mock_glob.assert_called()
+        mock_ensure_dir.assert_called()
+        mock_find_executable.assert_called()
+        mock_check_recipe_choices.assert_called()
+
+    @mock.patch("pythonforandroid.util.chdir")
+    @mock.patch("pythonforandroid.build.ensure_dir")
+    @mock.patch("pythonforandroid.archs.glob")
+    @mock.patch("pythonforandroid.archs.find_executable")
+    def test_build_arch(
+        self,
+        mock_find_executable,
+        mock_glob,
+        mock_ensure_dir,
+        mock_current_directory,
+    ):
+        mock_find_executable.return_value = self.expected_compiler.format(
+                android_ndk=self.ctx._ndk_dir
+        )
+        mock_glob.return_value = ["llvm"]
+        with mock.patch(
+            f"pythonforandroid.recipes.{self.recipe_name}.sh.Command"
+        ) as mock_sh_command, mock.patch(
+            f"pythonforandroid.recipes.{self.recipe_name}.sh.make"
+        ) as mock_make:
+            self.recipe.build_arch(self.arch)
+
+            # make sure that the mocked methods are actually called
+            mock_glob.assert_called()
+            mock_ensure_dir.assert_called()
+            mock_current_directory.assert_called()
+            mock_find_executable.assert_called()
+            for command in self.sh_command_calls:
+                self.assertIn(
+                    mock.call(command),
+                    mock_sh_command.mock_calls,
+                )
+            mock_make.assert_called()
--- a/tests/recipes/recipe_lib_test.py
+++ b/tests/recipes/recipe_lib_test.py
@@ -108,3 +108,44 @@ class BaseTestForMakeRecipe(RecipeCtx):
                     mock_sh_command.mock_calls,
                 )
             mock_make.assert_called()
+
+
+class BaseTestForCmakeRecipe(BaseTestForMakeRecipe):
+    """
+    An unittest for testing any recipe using `cmake`. It inherits from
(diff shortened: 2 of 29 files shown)


AndreMiras:
Absolutely awesome :two_hearts: I can't wait to review!
I actually wanted to do so for so long, I started in a local branch and never finished. I'll try to review ASAP and most likely start now.

pythonforandroid/recipes/jpeg/__init__.py
```diff
@@ -35,10 +34,9 @@ def build_arch(self, arch):
                     '-DCMAKE_POSITION_INDEPENDENT_CODE=1',
                     '-DCMAKE_ANDROID_ARCH_ABI={arch}'.format(arch=arch.arch),
                     '-DCMAKE_ANDROID_NDK=' + self.ctx.ndk_dir,
-                    '-DCMAKE_C_COMPILER={toolchain}/bin/clang'.format(
-                        toolchain=env['TOOLCHAIN']),
-                    '-DCMAKE_CXX_COMPILER={toolchain}/bin/clang++'.format(
-                        toolchain=env['TOOLCHAIN']),
+                    '-DCMAKE_C_COMPILER={cc}'.format(cc=arch.get_clang_exe()),
+                    '-DCMAKE_CXX_COMPILER={cc_plus}'.format(
+                        cc_plus=arch.get_clang_exe(plus_plus=True)),
```
AndreMiras: So you actually did find a bug already while covering it with your tests? :clap:

tests/recipes/recipe_lib_test.py
```diff
@@ -0,0 +1,151 @@
...
+    This must be a dictionary containing pairs of key (env var) and value.
+    """
+
+    def __new__(cls, *args):
+        obj = super(BaseTestForMakeRecipe, cls).__new__(cls)
+        if obj.recipe_name is not None:
+            print(f"We are testing recipe: {obj.recipe_name}")
+            obj.recipes.append(obj.recipe_name)
+            obj.recipe_build_order.insert(1, obj.recipe_name)
+            return obj
```
AndreMiras: I guess you're using the `__new__` over the `__init__()` for a reason, but I have to ask. Why can't this be dealt with  it? Something like:
```python
def __init__(self, *args, **kwargs):
    if self.recipe_name is not None:
...
```

tests/recipes/recipe_lib_test.py
```diff
@@ -0,0 +1,151 @@
...
+    extra_env_flags = {}
+    """
+    This must be a dictionary containing pairs of key (env var) and value.
+    """
+
+    def __new__(cls, *args):
+        obj = super(BaseTestForMakeRecipe, cls).__new__(cls)
+        if obj.recipe_name is not None:
+            print(f"We are testing recipe: {obj.recipe_name}")
+            obj.recipes.append(obj.recipe_name)
```
AndreMiras: It always confuse me how Python can deal with class attributes vs instance attributes.
For instance initially `recipes` is a class attribute, right? Then we access it from the instance to alter it. Would this modify also the class attribute or just for this instance? :smile:

tests/recipes/recipe_lib_test.py
```diff
@@ -0,0 +1,151 @@
...
+        mock_glob.return_value = ["llvm"]
+        with mock.patch(
+            f"pythonforandroid.recipes.{self.recipe_name}.sh.Command"
+        ) as mock_sh_command, mock.patch(
+            f"pythonforandroid.recipes.{self.recipe_name}.sh.make"
+        ) as mock_make:
+            self.recipe.build_arch(self.arch)
+
+            # make sure that the mocked methods are actually called
+            mock_glob.assert_called()
```
AndreMiras: Only the call needs to be in the context manager. The rest can be outside

tests/recipes/recipe_lib_test.py
```diff
@@ -0,0 +1,151 @@
...
+        mock_glob.return_value = ["llvm"]
+        with mock.patch(
+            f"pythonforandroid.recipes.{self.recipe_name}.sh.make"
+        ) as mock_make, mock.patch(
+            f"pythonforandroid.recipes.{self.recipe_name}.sh.cmake"
+        ) as mock_cmake:
+            self.recipe.build_arch(self.arch)
+
+            # make sure that the mocked methods are actually called
+            mock_glob.assert_called()
```
AndreMiras: Same here, you can take this one level down (if you want). It doesn't need to be in the context manager

tests/recipes/test_freetype.py
```diff
@@ -0,0 +1,10 @@
+import unittest
+from tests.recipes.recipe_lib_test import BaseTestForMakeRecipe
+
+
+class TestFreetypeRecipe(BaseTestForMakeRecipe, unittest.TestCase):
```
AndreMiras: Why do we need to inherit from `TestCase` suddenly?

tests/recipes/test_gevent.py
```diff
@@ -1,21 +1,11 @@
 import unittest
 from mock import patch
-from pythonforandroid.archs import ArchARMv7_a
-from pythonforandroid.build import Context
-from pythonforandroid.recipe import Recipe
+from tests.recipes.recipe_ctx import RecipeCtx
 
 
-class TestGeventRecipe(unittest.TestCase):
+class TestGeventRecipe(RecipeCtx, unittest.TestCase):
```
AndreMiras: Also here, I think we can kill the `TestCase` inheritance, can't we?

tests/recipes/test_libgeos.py
```diff
@@ -0,0 +1,29 @@
...
+        self,
+        mock_find_executable,
+        mock_glob,
+        mock_ensure_dir,
+        mock_current_directory,
+        mock_makedirs,
+    ):
+        # We overwrite the base test method because we
+        # want to avoid any file/directory creation
+        super(TestLibgeosRecipe, self).test_build_arch()
```
AndreMiras: I think we can actually ditch all theses Python2 style `super()` call and use the Python3 one if we wanted. At least in unit tests

AndreMiras:
My only regret, is I didn't see the coverage report appearing in GitHub, I wanted to see that satisfying coverall +x% reported here.
Can't wait to see that merged anyway

opacam:
@AndreMiras, thanks for the review :smile:

About this coverage thing...I already noticed that...and I don't know what the hell happen, but I remember that at the same time that I pushed the changes I noticed that the coveralls page went down, so maybe the report wasn't send back to github with the coverage report... because the coveralls job for the last commit is done at:

https://coveralls.io/jobs/53165250

Apparently we will increases the coverage by a `3.3%` so we will pass the barrier of `60%` :tada:

Anyway, I will address your comments which will require a couple of commits, which I hope that brings the coverage report back in here :crossed_fingers:

tests/recipes/recipe_lib_test.py
```diff
@@ -0,0 +1,159 @@
...
+            f"pythonforandroid.recipes.{self.recipe_name}.sh.make"
+        ) as mock_make:
+            self.recipe.build_arch(self.arch)
+
+            for command in self.sh_command_calls:
+                self.assertIn(
+                    mock.call(command),
+                    mock_sh_command.mock_calls,
+                )
+            mock_make.assert_called()
```
AndreMiras: FYI actually this check can also be outside the context manager. The only thing that needs to be in the context manager is the call that you want to be mocked, hence `self.recipe.build_arch(self.arch)`

tests/recipes/recipe_lib_test.py
```diff
@@ -0,0 +1,159 @@
...
+        # we mock/test it inside a Context Manager
+        with mock.patch(
+            f"pythonforandroid.recipes.{self.recipe_name}.sh.make"
+        ) as mock_make, mock.patch(
+            f"pythonforandroid.recipes.{self.recipe_name}.sh.cmake"
+        ) as mock_cmake:
+            self.recipe.build_arch(self.arch)
+
+            mock_cmake.assert_called()
+            mock_make.assert_called()
```
AndreMiras: Same here it could also be outside. I know it can feel weird at first because we assume the variable is scoped to the that bock, but not. Unlike javascript `let` this Python variable is function scoped, just like javascript `var`.

AndreMiras:
Travis is completely lost it seems. Not even the CI is reporting to the ticket :thinking:

opacam:
Mmmmm...yes it's odd, I just pushed my develop branch and no build is triggered (and it should...I was 1 commit behind)...

I have a bad feeling on this...

It maybe related with this new github feature `Actions`?

I did not enabled this feature in my personal account, neither in here... but I see the `Actions` tab in here (not in my fork)...so maybe someone enabled this beta feature for python-for-android repo?

tests/recipes/recipe_lib_test.py
opacam: Thanks man, you are right :smile:, let me change this and try if a `force-push` with a rebase on top of develop branch triggers the build...

AndreMiras:
> It maybe related with this new github feature Actions?

Yes maybe, I can't tell :confused:

opacam:
Mmmm...well then...let's try to merge this and see what happens...ok?

AndreMiras:
Yes go for it :boom:

opacam:
Ok, then I will squash and merge :+1:

opacam:
Finally, the travis builds appeared (with two hours of delay...more or less)...:<
so it seems that there is some delay between pushes and travis requests...it looks like that travis was **lazy/busy**...

Since this delay had side effects for us, we loosed the travis/coveralls report in this github request, but they are here:

  - [Travis build succeed](https://travis-ci.org/kivy/python-for-android/builds/585043392)
  - [Coveralls report](https://coveralls.io/jobs/53250743): **+3.3%**

@AndreMiras, thanks for the reviews, patience and approvals :smile:, tomorrow I will properly answer the unanswered questions (you touched some sensitive points that will require some well thought words)...as well I think that I have some pending words in issues section...I also will deal with that...vacations leaved me out of game, I made an [halt and catch fire](https://en.wikipedia.org/wiki/Halt_and_Catch_Fire)... :relaxed:

pythonforandroid/recipes/jpeg/__init__.py
opacam: Aaa...this is more a little refactor (since at the time we changed JPEG to be built with Cmake, the function `arch.get_clang_exe` does not existed, we were forced to got these values in `get_recipe_env`, since our default compiler was `gcc`, but with the NDK-r19 migration thing, we use `clang`, so we can make use of  `arch.get_clang_exe` and remove some lines :smile:)

tests/recipes/recipe_lib_test.py
opacam: Well, [class attributes are supposed to be shared for all instances](https://www.geeksforgeeks.org/class-instance-attributes-python/)...so  I guess that the answer is yes, we are also modifying class attributes (the thing is that recipes will grow up as long as we perform the tests...but shouldn't affect the test itself at all)

tests/recipes/recipe_lib_test.py
opacam: Aaa... I faced several issues along the way with this: duplicated tests, tests not initialized...:laughing:..finally using `__new__` is what it worked and I forgot about this lines until now...

I just made a test with `__init__`  and it also can work so I'm considering to change `__new__` with `__init__` (I think it's better to not mess with `__new__` if we can do it with `__init__`)

...so...@AndreMiras, let's go with `__init__`?

tests/recipes/test_freetype.py
opacam: We actually need it because nor `RecipeCtx` nor `BaseTestForMakeRecipe` inherits from `unittest.TestCase`

tests/recipes/test_gevent.py
opacam: The same than [this](https://github.com/kivy/python-for-android/pull/1982/files#r324457377)

tests/recipes/recipe_lib_test.py
AndreMiras: Exactly, in this case why not making it explicit and access it through `cls`? That would avoid confusion. Do you see what I mean? This is rather nitpicking, but at the same time I personally feel it's important to get the habit about thinking/clarifying theses when possible since it can cause nasty side effects sometimes

tests/recipes/recipe_lib_test.py
AndreMiras: > I think it's better to not mess with `__new__` if we can do it with `__init__`

That recaps my feeling pretty well. I'm not against `__new__` in general, but it needs to be clearly justified. But yes again if you can't make it work consistently with `__init__` then I'm go for it.

tests/recipes/test_freetype.py
AndreMiras: `<nitpick-troll>`
What is `TestCase` providing that you need in this case? Would tests fail or not run if we drop it? :trollface:
We don't even use setup/teardown here. Yes I know it's hard to say `unittest.TestCase` goodbye :joy: 
Honestly, feel free to not address that comment, I'm just challenging your habits :wink: 
`</nitpick-troll>`

opacam: Jajajaja...No man, we need this or the tests won't run...we are using `setUp/teardown` for all library recipes we introduced/modified, since all of them inherits from `RecipeCtx`...or directly or via `BaseTestForMakeRecipe/BaseTestForCmakeRecipe`

...this fuc*** `Python Inheritance` makes the things hard to track sometimes...:laughing:

tests/recipes/recipe_lib_test.py
opacam: Yes you are right, I'm preparing another PR to address all this we recently talk that cannot be done here. Later I will push it.

**Thanks again** @AndreMiras, your reviews always makes the code better :+1:

tests/recipes/test_freetype.py
AndreMiras: Damn it, you got the point :joy: 
But yes definitely OOP sometimes :sweat_smile: 
Thanks man

PR #1984, opened 2019-09-15
Recipes tests enhancements

opacam:
This is a follow up of #1982

In here we do:
  - Move from `__new__` to `__init__` for `BaseTestForMakeRecipe` ([discussed at here](https://github.com/kivy/python-for-android/pull/1982#discussion_r324419892))
  - Convert cls attributes to instance attributes ([discussed at here](https://github.com/kivy/python-for-android/pull/1982#discussion_r324419973))
  -  Remove `mock` dependency for py3 tests (Because, as per Python 3.3+, `mock` it's a built-in-module). (Unfortunately we still will have a couple of `import mock` entries that we cannot remove until we completely remove python2 from our tests)

### DIFF
--- a/tests/recipes/recipe_lib_test.py
+++ b/tests/recipes/recipe_lib_test.py
@@ -12,8 +12,6 @@ class BaseTestForMakeRecipe(RecipeCtx):
     """
 
     recipe_name = None
-    recipes = ["python3", "kivy"]
-    recipe_build_order = ["hostpython3", "python3", "sdl2", "kivy"]
     expected_compiler = (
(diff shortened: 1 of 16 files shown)


opacam:
Thanks for the approval @AndreMiras :heart: ,

I see that we have a problem with coveralls, the last three PRs does not have the github test for coverage...despite having [the job done in coveralls](https://coveralls.io/jobs/53260618)...:thinking:

AndreMiras:
Yes that's coverall thing is a bit annoying :confused:

PR #1986, opened 2019-09-16
Made p4a use per-arch dist build dirs

inclement:
The approach taken here is to specifically *not* change the dist name, which should remain the same as what the user specified, but to use different folders for different arch builds of the same dist.

This is distinct to (say) the NDK API, where we don't allow multiple versions built for the same dist name, because we can expect that multiple archs is a normal thing to want while multiple NDK APIs for the same project is usually a mistake (and if not, can be resolved by specifying different dist names).

### DIFF
--- a/pythonforandroid/distribution.py
+++ b/pythonforandroid/distribution.py
@@ -42,13 +42,19 @@ class Distribution(object):
         return str(self)
 
     @classmethod
-    def get_distribution(cls, ctx, name=None, recipes=[],
-                         ndk_api=None,
-                         arch_name=None,
-                         force_build=False,
-                         extra_dist_dirs=[],
-                         require_perfect_match=False,
-                         allow_replace_dist=True):
+    def get_distribution(
+            cls,
+            ctx,
+            *,
+            arch_name,  # required keyword argument: there is no sensible default
+            name=None,
+            recipes=[],
+            ndk_api=None,
+            force_build=False,
+            extra_dist_dirs=[],
+            require_perfect_match=False,
+            allow_replace_dist=True
+    ):
         '''Takes information about the distribution, and decides what kind of
         distribution it will be.
 
@@ -67,6 +73,7 @@ class Distribution(object):
         arch_name : str
             The target architecture name to compile against, included in the dist because
             it cannot be changed later during APK packaging.
+            This is a required, keyword-only argument.
         recipes : list
             The recipes that the distribution must contain.
         force_download: bool
@@ -97,12 +104,13 @@ class Distribution(object):
                 d for d in possible_dists if
                 (d.name == name) and (arch_name in d.archs)]
 
-            # There should only be one folder with a given dist name *and* arch.
-            # We could check that here, but for compatibility let's let it slide
-            # and just record the details of one of them. We only use this data to
-            # possibly fail the build later, so it doesn't really matter if there
-            # was more than one clash.
-            folder_match_dist = possible_dists[0]
+            if possible_dists:
+                # There should only be one folder with a given dist name *and* arch.
+                # We could check that here, but for compatibility let's let it slide
+                # and just record the details of one of them. We only use this data to
+                # possibly fail the build later, so it doesn't really matter if there
+                # was more than one clash.
+                folder_match_dist = possible_dists[0]
 
         # 1) Check if any existing dists meet the requirements
         _possible_dists = []
@@ -155,7 +163,7 @@ class Distribution(object):
                     req_recipes=', '.join(recipes),
                     dist_recipes=', '.join(folder_match_dist.recipes)))
 
-        assert len(folder_match_dist) < 2
+        assert len(possible_dists) < 2
 
         # If we got this far, we need to build a new dist
         dist = Distribution(ctx)
@@ -171,7 +179,10 @@ class Distribution(object):
         dist.name = name
         dist.dist_dir = join(
             ctx.dist_dir,
-            generate_dist_folder_name(name, [arch_name])
+            generate_dist_folder_name(
+                name,
+                [arch_name] if arch_name is not None else None,
+            )
         )
         dist.recipes = recipes
         dist.ndk_api = ctx.ndk_api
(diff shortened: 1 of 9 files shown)


pythonforandroid/toolchain.py
```diff
@@ -304,7 +305,7 @@ def __init__(self):
                   '(default: {})'.format(default_storage_dir)))
 
         generic_parser.add_argument(
-            '--arch', help='The archs to build for, separated by commas.',
+            '--arch', help='The arch to build for.',
```
AndreMiras: I'm wondering, are we breaking our CLI API here? Wasn't the arch split via `split_argument_list()` later?

inclement: Yes, a bit, but I did this because as far as I can tell passing multiple archs won't actually work right now. I'm trying to avoid removing the internal flexibility (hence keeping lists of archs even though we only handle one), but it was simplest to make the cli match what we can actually support for now. I'll confirm this before finalising the PR.

inclement:
Comments welcome, but yep, it really is a WIP for now. I should have time to fill in the gaps in the next few days.

pythonforandroid/toolchain.py
AndreMiras: Yes that makes complete sense, thanks. I was also confused seeing it was handling multiple archs from the CLI.

inclement:
I think this is good to go now, barring any further test changes.

I'm not entirely happy with the way the internal variables are handled, but that's wholly a result of how they're already handled :p. I cleared up some of the details, and I'll probably make another PR to further improve things.

pythonforandroid/distribution.py
```diff
@@ -79,15 +86,23 @@ def get_distribution(cls, ctx, name=None, recipes=[],
 
         existing_dists = Distribution.get_distributions(ctx)
 
-        possible_dists = existing_dists
+        possible_dists = existing_dists[:]
```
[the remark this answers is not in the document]
inclement: Looks like it's just out of date, fixed, thanks.

pythonforandroid/distribution.py
```diff
@@ -42,12 +42,19 @@ def __repr__(self):
...
-    def get_distribution(cls, ctx, name=None, recipes=[],
-                         ndk_api=None,
-                         force_build=False,
-                         extra_dist_dirs=[],
-                         require_perfect_match=False,
-                         allow_replace_dist=True):
+    def get_distribution(
+            cls,
+            ctx,
+            *,
```
AndreMiras: Just curious why enforcing keyword arguments and why from here and not from `ctx` for instance?

pythonforandroid/distribution.py
```diff
@@ -60,6 +67,13 @@ def get_distribution(cls, ctx, name=None, recipes=[],
         name : str
             The name of the distribution. If a dist with this name already '
             exists, it will be used.
+        ndk_api : int
+            The NDK API to compile against, included in the dist because it cannot
+            be changed later during APK packaging.
+        arch_name : str
+            The target architecture name to compile against, included in the dist because
+            it cannot be changed later during APK packaging.
+            This is a required, keyword-only argument.
```
AndreMiras: I'm tempted to say this goes without saying `This is a required, keyword-only argument`. Specifically `keyword-only argument` is the case for all except for `cls`

tests/test_distribution.py
```diff
@@ -51,8 +53,10 @@ def setUp_distribution_with_bootstrap(self, bs, **kwargs):
             self.ctx,
             name=kwargs.pop("name", "test_prj"),
             recipes=kwargs.pop("recipes", ["python3", "kivy"]),
+            arch_name=self.TEST_ARCH,
             **kwargs
         )
+        print('distribution when made is', self.ctx.bootstrap.distribution)
```
AndreMiras: Is this some leftover debug line?

tests/test_distribution.py
```diff
@@ -154,10 +158,12 @@ def test_get_distributions(
         ]
 
         dists = self.ctx.bootstrap.distribution.get_distributions(self.ctx)
+        print('dists are', dists)
```
AndreMiras: Debug print here?

pythonforandroid/distribution.py
```diff
@@ -236,3 +266,23 @@ def pretty_log_dists(dists, log_func=info):
...
+    ----------
+    base_dist_name : str
+        The core distribution identifier string
+    arch_names : list of str
+        The architecture compile targets
+    """
+    if arch_names is None:
+        arch_names = ["no_arch_specified"]
+
+    return '{}__{}'.format(
```
AndreMiras: Dumb question: Why the double underscore?
Should we copy/paste that same function on buildozer or can you think a way of reusing it? e.g. maybe exposing it in the CLI so buildozer can use it?

pythonforandroid/distribution.py
inclement: Partly because it's simpler not to break the existing code where ctx is the first argument, and partly because I think it's a fine api style to have ctx by an implicit essential argument that should come first while everything else has no particular order to it. ctx is also a consistent p4a concept with the same name everywhere.

pythonforandroid/distribution.py
inclement: Fair, will change.

pythonforandroid/distribution.py
inclement: Double underscore is just my habit for separating things where each entry can plausibly contain underscores (as the dist name certainly can).

inclement:
Thanks for the review, will resolve before merging.

pythonforandroid/distribution.py
AndreMiras: OK I understand, thanks for explaining :+1:

pythonforandroid/distribution.py
AndreMiras: OK that makes complete sense, thank you!

inclement:
Changes made, should be good now. I'll roll the buildozer changes into my existing arch cleanup PR.

inclement:
Merged!

Will do another p4a release once buildozer is fixed with this.

PR #1993, opened 2019-09-29
Fixes libiconv & libzbar configure host

AndreMiras:
Having `--host=x86` was making the configure script fail at enabling
shared library build on `x86`. This was resulting in failing to copy
`libiconv.so` and `libzbar.so` on `x86` arch.


AndreMiras:
Hi @jtc0de :wave: 
Any idea why the `TestGetSystemPythonExecutable.test_virtualenv` is failing? The error is `ModuleNotFoundError: No module named \'pytoml\'`

AndreMiras:
I've just rebased on `upstream/develop` which fixes #1994

AndreMiras:
I deleted the branch my mistake, but this PR is still valid, re-opening.
@inclement could you take a look, build was green but I guess I may have to trigger again now :confused:

PR #1995, opened 2019-09-29
Fixes test_virtualenv and test_venv failing, closes #1994

### LINKED ISSUE #1994: TestGetSystemPythonExecutable.test_virtualenv test fail
The `TestGetSystemPythonExecutable.test_virtualenv` and `TestGetSystemPythonExecutable.test_venv` tests started failing all of a sudden.
Error was:
```
ModuleNotFoundError: No module named \'pytoml\'\n'
```
This ca be reproduced in local via:
```sh
pytest tests/test_pythonpackage_basic.py::TestGetSystemPythonExecutable::test_virtualenv
```

AndreMiras:
Recent pep517 release seems to break tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ data_files = []
 install_reqs = [
     'appdirs', 'colorama>=0.3.3', 'jinja2', 'six',
     'enum34; python_version<"3.4"', 'sh>=1.10; sys_platform!="nt"',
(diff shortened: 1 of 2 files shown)


AndreMiras:
I think this fix is good enough for now, since it affects only the `setup.py` p4a feature. Also without this fix all upcoming PR would break the build

PR #2000, opened 2019-10-09
Fix libcurl with openssl

tito:
Currently (dunno from when, it was working when i added the recipe), the libcurl doesn't use ssl despite the `--with-ssl` option given, because it is not being able to link with libcrypto.

By adding the flags / libs to the environ like others recipes, everything works again.


opacam:
@tito, thanks, **this looks good** :smile:, just the linting complains, I think that if you `rebase` this branch in top of develop branch we will have a green tick mark, it's the same issue that happened [here](https://github.com/kivy/python-for-android/pull/1993#issuecomment-536298735) .

Issue #1988, opened 2019-09-19
Importing numpy yields: TypeError: add_docstring() argument 2 must be str, not None


brentpicasso:
### Versions

* Python: 2.7
* OS: Linux
* Kivy: 1.11.1
* Cython: 0.28.5

### Description

importing numpy generates the following error when executing on Android:

09-19 11:35:14.025 21284 21846 I python  :  TypeError: add_docstring() argument 2 must be str, not None
09-19 11:35:14.025 21284 21846 I python  : Python for android ended.

Repro: 

helloworld app, with numpy imported. 

### buildozer.spec

Command:
```
buildozer android debug
```

Spec file:
```
[app]

# (str) Title of your application
title = Podium

# (str) Package name
package.name = podium

# (str) Package domain (needed for android/ios packaging)
package.domain = com.autosportlabs

# (str) Source code where the main.py live
source.dir = .

# (list) Source files to include (let empty to include all the files)
source.include_exts = py,png,jpg,kv,atlas,ttf,json,svg,cfg,obj,mtl,glsl,zip,sql

# (list) Source files to exclude (let empty to not exclude anything)
#source.exclude_exts = spec

# (list) List of directory to exclude (let empty to not exclude anything)
source.exclude_dirs = tests, bin, .git, install, build_tools

# (list) List of exclusions using pattern matching
#source.exclude_patterns = license,images/*/*.jpg

# (str) Application versioning (method 1)
version.regex = __version__ = ['"](.*)['"]
version.filename = %(source.dir)s/main.py

# (str) Application versioning (method 2)
# version = 1.2.0

# (list) Application requirements
# comma seperated e.g. requirements = sqlite3, kivy, pygments

requirements = python2, kivy==1.11.1, numpy, scipy, openssl, pyopenssl, android, websocket-client, pygments, raven, pyjnius, https://github.com/autosportlabs/plyer/archive/master.zip, sqlalchemy, contextlib2, podium_api==0.0.13, https://github.com/nskrypnik/kivy3/zipball/master
# ffpyplayer==v4.1.0, ffpyplayer_codecs, 

# (list) Garden requirements
#garden_requirements =

# (str) Presplash of the application
presplash.filename = resource/images/app_splash_screen.png

# (str) Icon of the application
icon.filename = resource/images/app_icon_256x256.png

# (str) Supported orientation (one of landscape, portrait or all)
orientation = all

# (bool) Indicate if the application should be fullscreen or not
fullscreen = 1


#
[...]

PR #2003, opened 2019-10-14
Fix error on py2  when `import numpy`

opacam:
Also:
  - make use of python3 `super()` for numpy recipe

Closes: #1988


opacam:
Aaa.. you are right @AndreMiras, people should move to python3, but we still have python2 in here, so I have the feeling that we should give a solution, or we will fill p4a issues about the same error, anyway feel free to merge this yourself or close it I give this to your judgement :wink:

PR #2006, opened 2019-10-20
twisted: updated to 19.7.0 + removed inceremental.path

goffi-contrib:
Last version is building fine with p4a. Also incremental.patch doesn't
seem needed anymore (using Python 3), I could build without it.


goffi-contrib:
> Thanks for the PR, it looks like the patch might be needed at least on Ubuntu somehow, see build error on CI:
> https://travis-ci.org/kivy/python-for-android/jobs/600355532
> 
> ```
> distutils.errors.DistutilsError: Could not find suitable distribution for Requirement.parse('incremental>=16.10.1')
> ```

indeed, I've restored the patch.

Issue #2009, opened 2019-10-20
Leverage Docker image caching


AndreMiras:
The idea would be to speed up docker build (currently 10 minutes) by using the `--cache-from` flag.
Thing is we also need to `docker push` our successful builds ourselves as the DockerHub autobuild seems to mess up with `COPY` command leading to 100% cache bust when using this image.

Julian-O:
@AndreMiras: Is this still an issue?

AndreMiras:
We haven't solved it in p4a, but I think it's better documented in stackoverflow now.
https://stackoverflow.com/questions/54574821/docker-build-not-using-cache-when-copying-gemfile-while-using-cache-from
Also it's probably not critical and could be closed

Issue #2008, opened 2019-10-20
Revamp .travis.yml file


AndreMiras:
Our Travis file got more complex over time. Also some of the tests run directly on Travis host, making it difficult to reproduce some glitches locally.
I'd like to improve that by relying more on Docker and Makefile.
The idea would be run tests as much as possible on a Docker container, but also using Makefile. That way command being triggered on Travis should be exactly the same as the one running in local testing.

AndreMiras:
Closing as recent #2011 PR and follow up changes improved it. Also #2025 benefited from it as we could leverage some of our Makefile targets.
I still think there's room for improvements e.g. some of the tox related section, but we have enough for now to close the task

PR #2011, opened 2019-10-27
Travis CI revamp part 1, refs #2008

AndreMiras:
Revamping Travis and Docker setup introducing a `Makefile`.
The idea is to move the CI complexity from .travis.yml to `Makefile`.
That makes a single entry point via `make` command and reproducible
builds via Docker.
It makes it easy to run some commands outside docker, such as:
```sh
make testapps/python3/armeabi-v7a
```
Or the same command inside docker:
```sh
make docker/run/make/testapps/python3/armeabi-v7a
```
This pull request also starts introducing some docker layer cache
optimization as needed by #2009 to speed up docker pull/push and
rebuilds from cache.
It also introduces other Docker images good practices like ordering
dependencies alphabetically or always enforcing `apt update` prior
install, refs:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/
Subsequent pull requests would simplify the process furthermore and
leverage the cache to speed up builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,10 @@
 sudo: required
 
-dist: xenial  # needed for more recent python 3 and python3-venv
-
 language: generic
 
 stages:
-  - lint
-  - test
+  - unit tests
+  - build testapps
 
 services:
   - docker
@@ -18,8 +16,8 @@ before_install:
 
 jobs:
   include:
-    - &linting
-      stage: lint
+    - &unittests
+      stage: unit tests
       language: python
       python: 3.7
       before_script:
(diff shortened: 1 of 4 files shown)


AndreMiras:
I'm pretty happy with the level of simplification of the `.travis.yml` file. I think there's still room for improvements in the `tox` part, but let's see that after we merge that one.
I could also leverage some docker caching, which seems to speed up each `docker build` slightly (~2 minutes). This is because we still get some cache bust due to a bug/feature with the Docker `COPY` command. This can be overcome to speed things furthermore, by pushing the image from Travis on merge to develop. But let's address it in subsequent pull request.
I also think we can speed up docker build, pull and push to some nice extent by optimizing the image, e.g. using multi-stage builds (e.g. lighter image for CI), also refs https://github.com/kivy/python-for-android/issues/2012

.travis.yml
```diff
@@ -36,66 +34,35 @@ jobs:
...
         - coveralls
-
-    - &testing
-      stage: test
-      before_script:
-        # build docker image
-        - docker build --tag=p4a --file Dockerfile.py3 .
-        # Run a background process to make sure that travis will not kill our tests in
-        # case that the travis log doesn't produce any output for more than 10 minutes
-        - while sleep 540; do echo "==== Still running (travis, don't kill me) ===="; done &
```
opacam: I suspect that we still may need this lines, not for the basic tests we make but for `rebuild_updated_recipes` whenever we touch some of the most time demanding recipes (`icu`, `boost+libtorrent`...). Anyway, we can leave it out, for now, and reintroduce it later in case that we need it...if not...better for us...less code to maintain and more readable.

.travis.yml
```diff
@@ -36,66 +34,35 @@ jobs:
...
-      script:
-        - >
-          cd testapps && python3 setup_testapp_python3_sqlite_openssl.py apk
-          --sdk-dir $HOME/.android/android-sdk
-          --ndk-dir $HOME/.android/android-ndk
-          --requirements libffi,sdl2,pyjnius,kivy,python3,openssl,requests,sqlite3,setuptools
-          --arch=armeabi-v7a
-    - <<: *testing
+      script: make testapps/python3/armeabi-v7a PYTHON_WITH_VERSION=python3
```
opacam: Do we need to pass `PYTHON_WITH_VERSION=python3`?

We default to python3, so maybe we can remove it right?

Dockerfile.py3
```diff
@@ -37,37 +40,52 @@ RUN curl https://raw.githubusercontent.com/kadwanev/retry/1.0.1/retry \
...
+    python3-dev \
+    python3-pip \
+    python3-venv \
+    sudo \
+    unzip \
+    virtualenv \
+    wget \
+    zip \
+    zlib1g-dev \
+    zlib1g:i386 \
```
opacam: Nice :smile:...far more readable sorted and each dependency in one line :+1:

Makefile
```diff
@@ -0,0 +1,78 @@
...
+	$(TOX) -- tests/ --ignore tests/test_pythonpackage.py
+	@if test -n "$$CI"; then .tox/py$(PYTHON_MAJOR_MINOR)/bin/coveralls; fi; \
+
+rebuild_updated_recipes: virtualenv
+	$(PYTHON) ci/rebuild_updated_recipes.py
+
+testapps/python2/armeabi-v7a: virtualenv
+	. $(ACTIVATE) && cd testapps/ && \
+    python setup_testapp_python2_sqlite_openssl.py apk --sdk-dir $(ANDROID_SDK_HOME) --ndk-dir $(ANDROID_NDK_HOME) \
+    --requirements sdl2,pyjnius,kivy,python2,openssl,requests,sqlite3,setuptools,numpy
```
opacam: **Nitpick:** I would put each argument in one line,  maybe even sorted, as you did with the `apt packages` (the same for other tests)

.travis.yml
AndreMiras: Good point for the demanding recipes. Yes I would try without for now as it's more readable, but I can definitely try with some of theses recipes on my fork to see if it builds for too long without providing any output

.travis.yml
AndreMiras: Good question, on osx I had to put it because the python3.6 binary itself didn't exist, making it fail when generating the virtualenv: `virtualenv --python=$(PYTHON_WITH_VERSION) $(VIRTUAL_ENV)`.
It's just than in osx they don't seem to symlink it

Dockerfile.py3
AndreMiras: Yes definitely and it's a part of the good practices:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#sort-multi-line-arguments

Makefile
AndreMiras: Yes that could make sense too. I'm not sure how the `--requirements` argument would digest spaces and line breaks between the commas.
I hope you don't mind I try to address this point in subsequent PR as I'm excited for merging this one :sweat_smile:

PR #2013, opened 2019-10-28
WIP Push Docker image and build from it in sub-stages

### LINKED ISSUE #2009: Leverage Docker image caching

AndreMiras:
This is a follow-up for https://github.com/kivy/python-for-android/pull/2011 it also supersedes it as it's built on top. See description below:

The `docker build` is now running alongside the `tox` testing as part of
the `pre checks` stage. It also fully leverages the docker image cache.
The newly built image is pushed to be available for subsequent build
stages and pull requests.
Two environment variables `DOCKER_USERNAME` and `DOCKER_PASSWORD` got
added to the Travis UI for making image pushing possible. Both `push`
and `pull` features fail silently with `|| true` as they should be
optional for a valid build.
`TRAVIS_PULL_REQUEST` and `TRAVIS_BRANCH` are built-in Travis variables.
Note that we're passing variables to `Makefile` explicitly on `make`
call rather than using the `--environment-overrides` flag. While it's
more verbose, it makes it easier to follow what's being used and when.
With this change, `Python 3 arm64-v8a` the longest `testapps` build that
was taking ~26 minutes to complete is now taking ~20 minutes when
hitting docker layers cache (most of the time). The shorter build
`Rebuild updated recipes` was taking ~9 minutes and is now down to ~3
minutes.
Closes #2009

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,9 @@
+# used by coveralls.io, refs:
+# https://coveralls-python.readthedocs.io/en/latest/usage/tox.html#travisci
+CI
+TRAVIS
+TRAVIS_BRANCH
+TRAVIS_JOB_ID
+TRAVIS_PULL_REQUEST
+# used for running UI tests
+DISPLAY
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,10 @@
 sudo: required
 
-dist: xenial  # needed for more recent python 3 and python3-venv
-
 language: generic
 
 stages:
-  - lint
-  - test
+  - unit tests
+  - build apps
 
 services:
   - docker
@@ -18,8 +16,8 @@ before_install:
 
 jobs:
   include:
-    - &linting
-      stage: lint
+    - &unittests
+      stage: unit tests
       language: python
       python: 3.7
       before_script:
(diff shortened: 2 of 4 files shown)


AndreMiras:
Build flow and time before the PR:
image
image

And after (on my fork):
image
Note that all the builds are going faster by 6 minutes except the osx one which doesn't run in Docker.
There's still room for improvements after this one gets merged.
Not only we're building faster, but also this code change makes it easier to play with docker locally the same way Travis is doing.

AndreMiras:
Just changed to `WIP` as it seems we cannot access secure variables from pull requests due to security restrictions https://docs.travis-ci.com/user/pull-requests/#pull-requests-and-security-restrictions
That's a real pity, but it makes sense it's disabled. I'll try to modify the pull request to take it into account

AndreMiras:
OK so let's not deal with `docker push` for on-going pull request for now since `docker login` is not possible in pull requests by Travis security restrictions.
The only thing we may do is to `docker push` only on merge to `develop`. But we can deal with this in a completely different pull request.
I adjusted #2011 to leverage the cache to some good enough level for now.

PR #2015, opened 2019-11-01
make the virtual keyboard accept various languages

gottadiveintopython:
Currently, Japanese (and Chinese) keyboard doesn't show up on Android 9. (I don't know about other versions, but at least it *does* show up on  Android 5.0.3). This PR fix it.

Please take a look at [this discussion](https://bugzilla.libsdl.org/show_bug.cgi?id=4775).

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/libsdl/app/SDLActivity.java
+++ b/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/libsdl/app/SDLActivity.java
@@ -2002,7 +2002,7 @@ class DummyEdit extends View implements View.OnKeyListener {
     public InputConnection onCreateInputConnection(EditorInfo outAttrs) {
         ic = new SDLInputConnection(this, true);
 
-        outAttrs.inputType = InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD;
+        outAttrs.inputType = InputType.TYPE_CLASS_TEXT;
         outAttrs.imeOptions = EditorInfo.IME_FLAG_NO_EXTRACT_UI
                 | EditorInfo.IME_FLAG_NO_FULLSCREEN /* API 11 */;
 


pythonforandroid/bootstraps/sdl2/build/src/main/java/org/libsdl/app/SDLActivity.java
```diff
@@ -2002,7 +2002,7 @@ public boolean onKeyPreIme (int keyCode, KeyEvent event) {
     public InputConnection onCreateInputConnection(EditorInfo outAttrs) {
         ic = new SDLInputConnection(this, true);
 
-        outAttrs.inputType = InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD;
+        outAttrs.inputType = InputType.TYPE_CLASS_TEXT | InputType.TYPE_TEXT_VARIATION_NORMAL;
```
Franccisco: Does ```TYPE_TEXT_VARIATION_NORMAL``` really have to be explicit? According to Android documentation, it is the default variation of ```TYPE_CLASS_TEXT```.

gottadiveintopython: Maybe, it doesn't have to be. I'll test it when I have time.

Franccisco: Considering that:

- [TYPE_CLASS_TEXT](https://developer.android.com/reference/android/text/InputType.html#TYPE_CLASS_TEXT) equals 1.

-  [TYPE_TEXT_VARIATION_NORMAL](https://developer.android.com/reference/android/text/InputType.html#TYPE_TEXT_VARIATION_NORMAL) equals 0.

Then ```1 | 0 = 1```, so I think it's a good idea to remove ```TYPE_TEXT_VARIATION_NORMAL```:

```suggestion
        outAttrs.inputType = InputType.TYPE_CLASS_TEXT;
```

By the way, keyboards with predictive text have a non-desirable behavior with this PR: written text (containing just letters) does not appear in TextInput. Only symbols, numbers, and text selected from suggested words are visible.

Test it and tell me if you can reproduce the bug, please. I'll try to fix it.

Franccisco: Considering that:

- [TYPE_CLASS_TEXT](https://developer.android.com/reference/android/text/InputType.html#TYPE_CLASS_TEXT) equals 1.

-  [TYPE_TEXT_VARIATION_NORMAL](https://developer.android.com/reference/android/text/InputType.html#TYPE_TEXT_VARIATION_NORMAL) equals 0.

Then ```1 | 0 = 1```, so I think it's a good idea to remove ```TYPE_TEXT_VARIATION_NORMAL```:

```suggestion
        outAttrs.inputType = InputType.TYPE_CLASS_TEXT;
```

By the way, keyboards with predictive text have a non-desirable behavior with this PR: written text (containing just letters) does not appear in TextInput. Only symbols, numbers, and text selected from suggested words are visible.

Test it and tell me if you can reproduce the bug, please. I'll try to fix it.

mjuchno: I can reproduce the bug.

I do not see any difference in behavior with or without TYPE_TEXT_VARIATION_NORMAL

When prediction is provided, the text does not appear in the TextInput until it is selected from the bar or "confirmed" with space character. Only some keyboard, like Swiftkey, selects the word on space character, others do not (Gboard?).
In most of the cases, on backspace, the text is first deleted from this invisible "buffer" and then from the TextInput.
Symbols and numbers behavior varies depending on a keyboard and whether the keyboard tries to predict numeric input.

Swipe on Swiftkey keyboard inserts the word twice, Gboard does not inserts anything (or inserts a space character? maybe it has something to do with the space character behavior)

gottadiveintopython:
>  keyboards with predictive text have a non-desirable behavior with this PR

I confirmed that. The cause probably is TextInput doesn't listen `on_textedit`event. I ran the following code:

```python
from kivy.app import App
from kivy.lang import Builder
from kivy.factory import Factory

KV_CODE = '''
BoxLayout:
    TextInput:
    Button:
'''

class SampleApp(App):
    def build(self):
        return Builder.load_string(KV_CODE)
    def on_start(self):
        from kivy.core.window import Window
        Window.bind(
            on_textedit=lambda __, text: print('on_textedit:', text),
            on_textinput=lambda __, text: print('on_textinput:', text),
            )

if __name__ == '__main__':
    SampleApp().run()
```

1. I typed `e`. The text didn't appear but `on_textedit: e` was printed.
1. Then I typed `r`. The text still didn't appear but `on_textedit: er` was printed.
1. Then I selected `error`, one of suggestions. `error` appeared and `on_textinput: error` and `on_textinput:` were printed.

If you look at the [source](https://github.com/kivy/kivy/blob/master/kivy/uix/textinput.py), You can see there is no code that listening `on_textedit`. Only `on_textinput` is listened. That's why the text doesn't appear until the user finalized(commited).

Franccisco:
Nice! I tested it and, as you told, ```on_textedit``` is the responsible. I'll take a look on how to implement it on ```textinput.py``` and ```focus.py```.

Also, I realised that with ```TYPE_CLASS_TEXT``` you can insert emojis, but after doing it, the app crashes with:

```python
File "kivy/core/window/_window_sdl2.pyx", line 682, in kivy.core.window._window_sdl2._WindowSDL2Storage.poll
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xed in position 0: invalid continuation byte
```

I tried with multiple emojis and always getting that error.

gottadiveintopython:
Well, `SDL_TEXTEDITING` brings two arguments, which currently being ignored. It might be related to that error.
https://github.com/kivy/kivy/pull/6498/files
I'll test it tomorrow.

But in the first place, Roboto doesn't have glyphs for emojis?

gottadiveintopython:
@Franccisco I can't reproduce the error. Emojis appear as tofu but no error. It might depends on the OS. Which version do you use? Mine is Fire OS 7, which derived from Android 9.

Franccisco:
@gottadiveintopython Yes, it depends on the OS version. In older Android versions, the app crashes.

Well, there are two issues that needs fix at the moment: keyboard suggestions and emojis.

Franccisco:
@gottadiveintopython The tofu you're watching is because the font used by Kivy (Roboto, unless you changed it) does not support emojis. There are some fonts that support them (I think Noto is useful in this case, as it supports emojis and is a font intended to support all languages). As far as I know, Kivy only supports regular, italic and bold variations of a font (https://github.com/kivy/kivy/blob/b1c643c8a6d3d4750dc2fa9a0232354dfcecd3a1/kivy/core/text/__init__.py#L291), so this file may need a modification too.

Franccisco:
@gottadiveintopython The tofu you're watching is because the font used by Kivy (Roboto, unless you changed it) does not support emojis. There are some fonts that already do (I think Noto is useful in this case, as it supports emojis and is a font intended for all languages). As far as I know, Kivy only allows to register the regular, italic and bold variations of a font (https://github.com/kivy/kivy/blob/b1c643c8a6d3d4750dc2fa9a0232354dfcecd3a1/kivy/core/text/__init__.py#L291), so this file may need a modification too.

gottadiveintopython:
close the pr due to #2057

Azag3077:
> @gottadiveintopython The tofu you're watching is because the font used by Kivy (Roboto, unless you changed it) does not support emojis. There are some fonts that already do (I think Noto is useful in this case, as it supports emojis and is a font intended for all languages). As far as I know, Kivy only allows to register the regular, italic and bold variations of a font (https://github.com/kivy/kivy/blob/b1c643c8a6d3d4750dc2fa9a0232354dfcecd3a1/kivy/core/text/__init__.py#L291), so this file may need a modification too.
Hi....

Do you mean I have to register my downloaded font before I can use the emoji in it?

PR #2019, opened 2019-11-11
pygame recipe

robertpfeiffer:
As discussed on the Discord, this is a recipe to build apps based on SDL2-based pygame.

It currently references https://github.com/pygame/pygame/tree/android, a branch that will continue to live until we have an official stable release of pygame based on SDL2 with android support. Is that OK? Should recipes only refer to tarballs/tagged snapshots/releases, or is it allowed to have a moving target in there?

Simple examples have been tested by other pygame users (it doesn't *just* build on my own machine) but some pygame functionality is still untested, and some dependencies like freetype, postmidi and libjpeg are currently not part of the build. It's usable, but not complete.

Your feedback is appreciated.


inclement:
Nice work, I'm pleased to see the recipe is pretty simple in the end!

> It currently references https://github.com/pygame/pygame/tree/android, a branch that will continue to live until we have an official stable release of pygame based on SDL2 with android support. Is that OK? Should recipes only refer to tarballs/tagged snapshots/releases, or is it allowed to have a moving target in there?

In general we'd like to tag specific versions, but for something like this it's reasonable to have a moving target. I think the obvious options are:

* leave this as a PR until you feel some stable target is ready
* merge with a recipe name like "pygame2beta" that will go away once a stable target is available
* merge with the name "pygame2" and just assume nobody will use it much until it's stable

I don't really mind which we do. I wouldn't normally like the last one, but good pygame support is a big target so it's reasonable to compromise for ease of use.

pythonforandroid/recipes/pygame2/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+                sdl_mixer_includes="-I"+join(self.ctx.bootstrap.build_dir, 'jni', 'SDL2_mixer'),
+                jpeg_includes="-I"+jpeg_inc_dir,
+                png_includes="-I"+png_inc_dir,
+                freetype_includes=""
+            )
+            open("Setup", "w").write(setup_file)
+
+    def get_recipe_env(self, arch):
+        env = super(Pygame2Recipe, self).get_recipe_env(arch)
+        if 'sdl2' in self.ctx.recipe_build_order:
```
inclement, line 50: Won't this always be True, or in any case you always want to set the cross compile and android env vars?

pythonforandroid/recipes/pygame2/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+from pythonforandroid.toolchain import current_directory
+
+
+class Pygame2Recipe(CompiledComponentsPythonRecipe):
+
+    version = '2.0.0-dev7'
+    url = 'https://github.com/pygame/pygame/archive/android-2.0.0-dev7.tar.gz'
+
+    site_packages_name = 'pygame'
+    name = 'pygame2'
```
inclement, line 12: Are you going to officially make the new release pygame2, with e.g. a different name in pip? If so then fine, but if not it's probably more obvious to just call the recipe "pygame". Happy to go with your preference though.

pythonforandroid/recipes/pygame2/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+
+    depends = ['sdl2', 'sdl2_image', 'sdl2_mixer', 'sdl2_ttf', 'setuptools', 'jpeg', 'png']
+    call_hostpython_via_targetpython = False  # Due to setuptools
+    install_in_hostpython = False
+
+    def prebuild_arch(self, arch):
+        super(Pygame2Recipe, self).prebuild_arch(arch)
```
AndreMiras, line 19:
```suggestion
        super().prebuild_arch(arch)
```

AndreMiras:
Sorry for missing that one. This is also looking good to me. I think we could make it to the tree pretty easily by:

1. rebasing fixing merge conflicts
2. remove the parts already addressed in https://github.com/kivy/python-for-android/pull/2132 even though it's not yet merged they don't strongly depend on one another
3. addressing the couple of minor comments inclement and I added on the pygame recipe itself

If you're busy I'd be happy to help out

inclement:
Looks good to me. The one failure looks like maybe a network issue, @AndreMiras do you know about it?

AndreMiras:
Yes it's seems to be a network hiccups on the macOS build. I think we're covered by the other builds. I'll restart the failing build now.
I haven't verify if the changes aren't conflicting with the recently merged #2132 
Best to make sure would be a rebasing.
Things look good for a merge otherwise

inclement:
Looks like the test failure is due to a test target not existing. @AndreMiras is this due to a recent change? If it's something simple I'm can push the change and merge.

AndreMiras:
Yes it's due to the recent changes. Basically it's still running on an old .travis.yml, but somehow it picked up the new `Makefile`. It's like if half of upstream/develop (specially https://github.com/kivy/python-for-android/pull/2159) was merged back into the branch.
If it was rebased, things like https://github.com/kivy/python-for-android/pull/2132 would not appear in the diff of this PR like we currently see. I would be happy to spend 10 minutes with @robertpfeiffer to walk him through it, but I didn't see him on Discord.
Anyway I think this is safe to (squash &) merge

pythonforandroid/recipes/pygame/__init__.py
```diff
@@ -0,0 +1,56 @@
+from pythonforandroid.recipe import CompiledComponentsPythonRecipe
+from os.path import join
+from pythonforandroid.toolchain import current_directory
+
+
+class Pygame2Recipe(CompiledComponentsPythonRecipe):
+
+    version = '2.0.0-dev7'
+    url = 'https://github.com/pygame/pygame/archive/android-2.0.0-dev7.tar.gz'
```
AndreMiras, line 9: Usually the version should be configurable
```suggestion
    url = 'https://github.com/pygame/pygame/archive/android-{version}.tar.gz'
```

pythonforandroid/recipes/pygame/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+                sdl_image_includes="-I"+join(self.ctx.bootstrap.build_dir, 'jni', 'SDL2_image'),
+                sdl_mixer_includes="-I"+join(self.ctx.bootstrap.build_dir, 'jni', 'SDL2_mixer'),
+                jpeg_includes="-I"+jpeg_inc_dir,
+                png_includes="-I"+png_inc_dir,
+                freetype_includes=""
+            )
+            open("Setup", "w").write(setup_file)
+
+    def get_recipe_env(self, arch):
+        env = super(Pygame2Recipe, self).get_recipe_env(arch)
```
AndreMiras, line 49: Nitpicking
```suggestion
        env = super().get_recipe_env(arch)
```

AndreMiras:
Merged via https://github.com/kivy/python-for-android/pull/2164 thanks again

PR #1997, opened 2019-10-05
Fixes ffmpeg build on ndk 19

misl6:
Targets issue #1996 .

- [x] ffmpeg builds
- [x] ffmpeg builds w/ openssl
- [x] ffmpeg builds w/ libx264
- [x] ffmpeg builds w/ libshine
- [x] disabled `libshine` due to the missing symbol error and added a comment for it

Known issues:
- if `libshine` is enabled `libavcodec` is throwing an exception during runtime: `ffpyplayer - ImportError: dlopen failed: cannot locate symbol "shine_set_config_mpeg_defaults" referenced by "/data/app/myapp==/lib/arm/libavcodec.so"..` ( disabled `libshine` as a workaround)

Tested over a `https/hls` stream on both `arm64-v8a` and `armeabi-v7a` .

PR #2025, opened 2019-11-17
Introduce github-actions (push & pull_requests)

opacam:
**I just want to share some of our travis tests but for `gh-actions`:**

In here we perform the same tox tests that we do in travis but for **linux** and **MacOs** (for **python** versions **3.6** and **3.7**...so we use a matrix) plus an **flake8 test** (which will make stop all the other tests if not succeed). 

Regarding testsapp, we only perform one of the tests (Python3 for arm64-v8a in ubuntu with docker), so we make less tests than travis regarding testapps. The good thing of this is that we run the tox tests in MacOs, which make me notice that we had wrong some of our tests so we solve this in here.

**Why introduce `gh-actions`?**

Well, because it's one more asset that we can use, and this is only one small part of what can be achieved with `gh-actions`. IMHO, I think that we can benefit from `gh-actions`, I wouldn't remove travis tests at all, but I would use `gh-actions` for any operation that requires to use passwords, since it's more easy to deal than with travis. In matter fact, I already used a `github.secret` (`COVERALLS_REPO_TOKEN`, to send coveralls report), that we may need to setup for kivy (I'm not sure of this...it maybe already added to kivy organization...).

**Note:** I leave this PR in `WIP` status because:
  - we never talked about introducing  `gh-actions` for p4a...maybe the kivy/p4a team doesn't want this at all...we will see...
  -  with this we will send the coveralls report with `travis` and with `gh-actions`...probably we only want to send the reports with one of the services...so this PR may require some adjustments...
  - While doing this (some weeks ago), I felt that `gh-actions` was a little clumsy...`gh-actions` was still in beta status...now are official but it still may be a little clumsy...I suppose that it will get better with time...)

**Some references:**
  - https://github.com/features/actions
  - Github actions video presentation: https://www.youtube.com/watch?v=E1OunoCyuhY

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -44,6 +44,8 @@ jobs:
     - name: Send Coverage to coveralls
       run: |
         coveralls
+      env:
+        COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
 
   build:
     name: Build testapp
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,4 +1,4 @@
-name: Unit tests
+name: Unit tests & Build Testapp
 
 on: ['push', 'pull_request']
 
@@ -49,3 +49,17 @@ jobs:
         python -m coveralls
       env:
         COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
+
+  build:
+    name: Build testapp
+    needs: [flake8]
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout python-for-android
+      uses: actions/checkout@master
+    - name: Pull docker image
+      run: |
+        make docker/pull
+    - name: Build apk for Python 3 arm64-v8a
+      run: |
+        make docker/run/make/testapps/python3/arm64-v8a
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -54,3 +54,17 @@ jobs:
     - name: Build apk for Python 3 arm64-v8a
       run: |
         make docker/run/make/testapps/python3/arm64-v8a
+
+  rebuild_updated_recipes:
+    name: Test updated recipes
+    needs: [flake8]
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout python-for-android
+      uses: actions/checkout@master
+    - name: Pull docker image
+      run: |
+        make docker/pull
+    - name: Rebuild updated recipes
+      run: |
+        make docker/run/make/rebuild_updated_recipes
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -15,11 +15,10 @@ jobs:
       with:
         python-version: 3.7
     - name: Run flake8
-      # we use exactly the same exceptions than tox.ini file
       run: |
         python -m pip install --upgrade pip
-        pip install -U flake8
-        flake8 --ignore=E123,E124,E126,E226,E402,E501,W503,W504 pythonforandroid/ tests/ ci/
+        pip install tox>=2.0
+        tox -e pep8
 
   linting:
     name: Pytest [Python ${{ matrix.python-version }} | ${{ matrix.os }}]
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -35,19 +35,15 @@ jobs:
       uses: actions/setup-python@v1.1.0
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
+    - name: Tox Python 3
+      # ignores test_pythonpackage.py since it runs for too long
       run: |
         python -m pip install --upgrade pip
-        pip install -U virtualenv pytest pytest-cov backports.tempfile pyOpenSSL coveralls
-        pip install -e .
-    - name: Tests
+        pip install tox>=2.0 pyOpenSSL coveralls
+        tox -e py3 -- tests/ --ignore tests/test_pythonpackage.py
+    - name: Send Coverage to coveralls
       run: |
-        python -m pytest tests/ --ignore tests/test_pythonpackage.py --cov pythonforandroid/ --cov-branch
-    - name: Coveralls
-      run: |
-        python -m coveralls
-      env:
-        COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
+        coveralls
 
   build:
     name: Build testapp
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -20,7 +20,7 @@ jobs:
         pip install tox>=2.0
         tox -e pep8
 
-  linting:
+  test:
     name: Pytest [Python ${{ matrix.python-version }} | ${{ matrix.os }}]
     needs: flake8
     runs-on: ${{ matrix.os }}
@@ -35,17 +35,11 @@ jobs:
       uses: actions/setup-python@v1.1.0
       with:
         python-version: ${{ matrix.python-version }}
-    - name: Tox Python 3
-      # ignores test_pythonpackage.py since it runs for too long
+    - name: Tox tests
       run: |
         python -m pip install --upgrade pip
-        pip install tox>=2.0 pyOpenSSL coveralls
-        tox -e py3 -- tests/ --ignore tests/test_pythonpackage.py
-    - name: Send Coverage to coveralls
-      run: |
-        coveralls
-      env:
-        COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
+        pip install tox>=2.0
+        make test
 
   build:
     name: Build testapp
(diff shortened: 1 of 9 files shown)


opacam:
The CI tests for `gh-actions` for this PR can be seen at my fork, [here the one for the last commit](https://github.com/opacam/python-for-android/commit/23d35a65009f4e7196896c9ccdd032a066ed05f1/checks?check_suite_id=314964116)

.github/workflows/push.yml
```diff
@@ -0,0 +1,65 @@
...
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1.1.0
+      with:
+        python-version: 3.7
+    - name: Run flake8
+      # we use exactly the same exceptions than tox.ini file
+      run: |
+        python -m pip install --upgrade pip
+        pip install -U flake8
+        flake8 --ignore=E123,E124,E126,E226,E402,E501,W503,W504 pythonforandroid/ tests/ ci/
```
AndreMiras: why not a `tox -e flake8`?

.github/workflows/push.yml
```diff
@@ -0,0 +1,65 @@
...
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -U virtualenv pytest pytest-cov backports.tempfile pyOpenSSL coveralls
+        pip install -e .
+    - name: Tests
+      run: |
+        python -m pytest tests/ --ignore tests/test_pythonpackage.py --cov pythonforandroid/ --cov-branch
```
AndreMiras: I'm missing why we don't run these through tox since it's already configured with dependencies and run commands? Is running tox cumbersome with github-actions?

.github/workflows/push.yml
```diff
@@ -0,0 +1,65 @@
...
+on: ['push', 'pull_request']
+
+jobs:
+
+  flake8:
+    name: Flake8 tests
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout python-for-android
+      uses: actions/checkout@master
```
AndreMiras: Shouldn't the branch be `develop` here and the other places? Or this is not related to the p4a branch?

.github/workflows/push.yml
```diff
@@ -0,0 +1,65 @@
...
+        pip install -U virtualenv pytest pytest-cov backports.tempfile pyOpenSSL coveralls
+        pip install -e .
+    - name: Tests
+      run: |
+        python -m pytest tests/ --ignore tests/test_pythonpackage.py --cov pythonforandroid/ --cov-branch
+    - name: Coveralls
+      run: |
+        python -m coveralls
+      env:
+        COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
```
AndreMiras: For Travis this isn't required somehow because Coveralls has some "magic integration" that makes it seamless. Is it for sure required with github-actions?

tests/recipes/recipe_lib_test.py
```diff
@@ -13,7 +14,7 @@ class BaseTestForMakeRecipe(RecipeCtx):
 
     recipe_name = None
     expected_compiler = (
-        "{android_ndk}/toolchains/llvm/prebuilt/linux-x86_64/bin/clang"
+        "{android_ndk}/toolchains/llvm/prebuilt/{system}-x86_64/bin/clang"
```
AndreMiras: So basically unit test was failing in osx? Maybe it could be a PR on its own with these fixes

.github/workflows/push.yml
opacam: Aaa...first of all, I'm not an gh-actions expert at all, but, I will try to put some light in here, so here we go:

  - `gh-actions` works based on events (In this PR we listen and react to `push` and `pull_requests` **github events**...so we kind of imitate **travis behaviour**). But, there is a lot of events that we can use, just [check this](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows) if you want to get more detailed info about events.
  - `gh-actions` is not just a name, it's a set of repos: https://github.com/actions. Those repos can be used by any other repo.

Said this, when we call the line:

  ```uses: actions/checkout@master```

What are we doing is executing the code located at: https://github.com/actions/checkout, which simply will run the proper git commands, so our workflow can access the content of our repo. The `@master` just point to the version that we use **for the specific github action** (master branch in this case but it could be a release...).


Ok, hope that this clarifies a little this particular `gh-actions` thing :wink:

.github/workflows/push.yml
opacam: Aaa...yes...that should work, honestly, I had a different idea about flake8 tests (and never finished that) ...let me explain...

With some `gh-actions`, we can interact with the web ui interface of github, so, I was thinking that it would be great that the detected PEP8 errors could be notified automatically at PRs ...mmm...better check this action to get a better idea: https://github.com/suo/flake8-github-action . So... I did not implement this because that flake8 action cannot be configured (at least for now, but it could be another options...needs investigation)

So, said this, **we could give it a try to the tox command (I will try this when I have some time)**, but I would keep one eye on this flake8 action mentioned above, because we could have some improvements here...specially thinking on PR reviewers...it seems that the `red cross` in travis tests, somehow and sometimes, is not seen, right? :laughing:

.github/workflows/push.yml
AndreMiras: Yes it clarifies, I was exactly not sure if this branch was our or the tooling one, thanks :+1:

.github/workflows/push.yml
opacam: Yes, we need it: https://github.com/opacam/python-for-android/runs/317382295#step:5:6

AndreMiras: Thank you for trying out. I also link the actual log output here "for posterity":
```
Not on Travis or CircleCI. You have to provide either repo_token in .coveralls.yml or set the COVERALLS_REPO_TOKEN env var.
  File "/opt/hostedtoolcache/Python/3.7.5/x64/lib/python3.7/site-packages/coveralls/cli.py", line 61, in main
    service_name=options['--service'])
  File "/opt/hostedtoolcache/Python/3.7.5/x64/lib/python3.7/site-packages/coveralls/api.py", line 58, in __init__
    self.ensure_token()
  File "/opt/hostedtoolcache/Python/3.7.5/x64/lib/python3.7/site-packages/coveralls/api.py", line 67, in ensure_token
    self.config_filename))
coveralls.exception.CoverallsException: Not on Travis or CircleCI. You have to provide either repo_token in .coveralls.yml or set the COVERALLS_REPO_TOKEN env var.
##[error]Process completed with exit code 1.
```

.github/workflows/push.yml
```diff
@@ -0,0 +1,62 @@
...
+        python-version: ${{ matrix.python-version }}
+    - name: Tox Python 3
+      # ignores test_pythonpackage.py since it runs for too long
+      run: |
+        python -m pip install --upgrade pip
+        pip install tox>=2.0 pyOpenSSL coveralls
+        tox -e py3 -- tests/ --ignore tests/test_pythonpackage.py
+    - name: Send Coverage to coveralls
+      run: |
+        coveralls
```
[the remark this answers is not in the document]
opacam: Yeah, I like it, this way the dilemma of deciding when we should send the report to coveralls it will be solved, since `CI` environment variable will only exist for travis, so no report will be send with gh-actions unless we change that in the `makefile` by the gh-actions equivalent (which is `GITHUB_ACTIONS`...for the record :wink:).

Thanks again @AndreMiras :smile:

opacam:
BTW, I think that we should include the `rebuild updated recipes` test in here, to try to solve the errors we have with some recipes in travis like at #1997 ...what do you think?

opacam:
As expected, the test for `rebuild updated recipes`  [is failling with travis](https://travis-ci.org/opacam/python-for-android/jobs/616015004#L940-L943) with:

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```

Ahahaha, but it [succeed with gh-actions](https://github.com/opacam/python-for-android/runs/317480493) :smile:

.github/workflows/push.yml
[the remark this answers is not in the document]
AndreMiras: Nice it looks simpler. Good to know for the env variable

opacam:
Ok then, I think that we can merge this already, but before, here I leave the [last gh-actions CI test](https://github.com/opacam/python-for-android/commit/0a939087001b9c8c0a11acd463a62d6bef8a054c/checks?check_suite_id=325621209), corresponding to the last commit (remember that all the `gh-actions` tests for this PR are in my fork)

Please @AndreMiras, do the merge for this one (or give me the order to do the squash/merge and I will do it), since you had some concerns about this `gh-actions` thing, I would like having your consent to merge this one (btw...whatever is your decision, is fine to me :wink:)

AndreMiras:
I'm happy with it that's why it's approved, I let you the pleasure to hit that (squash &) merge button :+1: 
Thanks again for your time investigating it

opacam:
Done, [here we have it](https://github.com/kivy/python-for-android/actions), also accessible via the :arrow_forward: `Actions` tab at github interface :smile:

PR #2031, opened 2019-11-30
Enhance distribution reuse: `android_api`

opacam:
In here we solve a problem we have when changing `android_api` if we already have a build distribution that we can reuse. The problem is that the changes are not reflected in the final APK, because there is some hardcoded `android_api` in some of the distribution files. So in here we will update those files so when we perform the build of the APK with gradle it will change the necessary stuff with the new `android_api`.

We also:
  - Store the `android_api` value to `dist_info.json` file (and update it if we reuse the dist with a new `android_api`)
  - Prettify `dist_info.json` (to be easier to read it for us)
  - Reduce complexity of `Distribution.get_distributions`
  - Add unittests for the added code

How to test this
--------------------
Well, it will require two builds, the first one with one `android_api` and then another with a new `android_api`. Once the second build complete you will see that the distribution file: `AndroidManifest.xml` has the value of the second  build (btw, we never modify this file in this PR, this file is modified by gradle). You can check around line 22, for something like:
```
    <uses-sdk android:minSdkVersion="21" android:targetSdkVersion="28" />
```

The value of `android:targetSdkVersion` should be the one you used with `android_api` in your second build.

How will look like the `dist_info.json` file
----------------------------------------------------
```
{
    "android_api": 28,
    "archs": [
        "arm64-v8a"
    ],
    "bootstrap": "sdl2",
    "dist_name": "bdisttest_python3_sqlite_openssl_googlendk",
    "hostpython": "/home/opacam/.local/share/python-for-android/build/other_builds/hostpython3/desktop/hostpython3/native-build/python3",
    "ndk_api": 21,
    "python_version": "3.7",
    "recipes": [
        "hostpython3",
        "libffi",
        "openssl",
        "sdl2_image",
        "sdl2_mixer",
        "sdl2_ttf",
        "sqlite3",
        "python3",
        "sdl2",
        "setuptools",
        "requests",
        "six",
        "pyjnius",
        "android",
        "kivy",
        "peewee"
    ],
    "use_setup_py": false
}
```

As you see, the keys are sorted and indented with 4 spaces :wink:...imho, far readable than all in one line, right?

Notes
--------
Some parts of the code where taken from #1926, so this PR is a portion of what we presented there, but refreshed and enhanced.


inclement:
Thanks @opacam, that's a great improvement. Just to let you know I've seen this, but probably won't be able to review it for a few days.

tests/test_distribution.py
```diff
@@ -76,6 +90,126 @@ def test_properties(self):
...
+    def test_get_dist_info(self, mock_open):
+        """Test that method
+        :meth:`~pythonforandroid.distribution.Distribution.get_dist_info`
+        calls the proper methods and returns the proper value."""
+        self.setUp_distribution_with_bootstrap(
+            Bootstrap().get_bootstrap("sdl2", self.ctx)
+        )
+        mock_open.side_effect = [
+            mock.mock_open(read_data=json.dumps(dist_info_data)).return_value
+        ]
```
AndreMiras: Using mock open this way is somehow bugging me, I guess I'm not used to it.
How about something more conventional?
```python
      def test_get_dist_info(self):
          """Test that method
          :meth:`~pythonforandroid.distribution.Distribution.get_dist_info`
          calls the proper methods and returns the proper value."""
          self.setUp_distribution_with_bootstrap(
          )
          mock_open = mock.mock_open(read_data=json.dumps(dist_info_data))
          with mock.patch("pythonforandroid.distribution.open", mock_open):
              dist_info = self.ctx.bootstrap.distribution.get_dist_info("/fake_dir")
          mock_open.assert_called_once_with("/fake_dir/dist_info.json", "r")
          self.assertIsInstance(dist_info, dict)
```
It's really passing the mock_open.side_effect with the other `return_value` that is bugging me, it feels unnatural.

tests/test_distribution.py
```diff
@@ -76,6 +90,126 @@ def test_properties(self):
...
+        dist_info = self.ctx.bootstrap.distribution.get_dist_info("/fake_dir")
+        mock_open.assert_called_once_with("/fake_dir/dist_info.json", "r")
+        self.assertIsInstance(dist_info, dict)
```
AndreMiras: Would it make sense to go as far as:
```python
self.assertEqual(dist_info, dist_info_data)
```

opacam:
> Thanks @opacam, that's a great improvement. Just to let you know I've seen this, but probably won't be able to review it for a few days.

@inclement, no problem at all, this way we can polish this thing thanks to the @AndreMiras review :wink:

@AndreMiras, I will take care of your comments ASAP...I  did a quick look to your comments and I see that we can do a little refactoring, nice !! :satisfied:

Fak3:
Just a thought - could this dist-info.json also be used to hold the environment variables for debugging purposes? Currently they only exist in the temp file during build, so some time ago I raised a PR to keep this file permanently inside the dist dir: https://github.com/kivy/python-for-android/pull/1902

PR #2034, opened 2019-12-05
Fix vibration for testapps

opacam:
Method `vibrate(milliseconds)` was deprecated in API level 26,
now we should use ` vibrate(android.os.VibrationEffect)`.

I labeled this `WIP` because I thing that we should consider the possibility to use `plyer` instead of using the more low level way of `pyjnius`, because we are repeating the code a lot with all the testapps (except for the flask testapp...which uses the same solution but written slightly differently). But the vibration fix for `plyer` it was merged recently, and we still not have a release with it, so it will force us to put a github requirement in our testapp setup.py files and pin it to a commit or master branch...so...
@inclement and @AndreMiras, what do you think about this, `plyer` or `pyjnius`?

**Note:**
  - This solution is based on the kivy/plyer#523 and was adapted to our needs
  - I tested the `testapp_flask` and the standard one `testapp_sqlite_openssl` and this last one, you could also test it, without building it, via the corresponding job, you can download the testapps built with gh-actions via the `Artificats` button :smile:

**See also:** https://developer.android.com/reference/android/os/Vibrator#vibrate(long)

### DIFF
--- a/testapps/testapp/main.py
+++ b/testapps/testapp/main.py
@@ -123,21 +123,34 @@ class TestApp(App):
 
     def test_pyjnius(self, *args):
         try:
-            from jnius import autoclass
+            from jnius import autoclass, cast
         except ImportError:
             raise_error('Could not import pyjnius')
             return
-
         print('Attempting to vibrate with pyjnius')
-        # PythonActivity = autoclass('org.renpy.android.PythonActivity')
-        # activity = PythonActivity.mActivity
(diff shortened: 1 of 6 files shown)


opacam:
Yeah, I agree with you, we should give a little more love to our testing tools, let's merge this for now, since as you said, fixes the deprecated behavior.

Thanks @AndreMiras !! :satisfied:

PR #2036, opened 2019-12-07
update pyjnius recipe

tshirtman:


tshirtman:
I must confess i only tested the sdl2 patch, and it worked, i edited them both manually, but i don't see a reason the other one wouldn't work. I did test an app with pyjnius and it worked fine.

PR #2044, opened 2019-12-24
Python 3.8 support on Android

inclement:
Needs cleaning up so not for merging yet (in particular want to retain the py3.7 patches for py3.7 builds), but this seems to work.


opacam:
**Nice work** :+1:

Thanks @inclement, I will review it when you finish the last touches, aaa...and it would be great to have the *python 3.8 patch* uploaded so we can to test this :smile:

jgavris:
+1 ...If you can upload the patch for 3.8 we're more than willing to test it.

inclement:
Doing some testing now, but this is intended to be basically final (barring bugfixes).

pythonforandroid/patching.py
```diff
@@ -69,3 +70,15 @@ def is_ndk(ndk):
...
         return recipe.ctx.ndk == ndk
     return is_x
+
+def is_version_gt(version):
+    def is_x(recipe, **kwargs):
+        return LooseVersion(recipe.version) > version
+
+def is_version_gt(version):
+    def is_x(recipe, **kwargs):
+        return LooseVersion(recipe.version) < version
```
AndreMiras: The `is_version_gt()` function is defined twice, I guess one of the two should be `is_version_lt()`

inclement:
I've tested that this is working. There's a minor inconvenience that you have to remember to change both the python3 and hostpython3 versions if targeting something different to the default, but that's something we can fix later.

pythonforandroid/patching.py
```diff
@@ -69,3 +70,17 @@ def is_ndk(ndk):
...
     return is_x
+
+def is_version_gt(version):
+    def is_x(recipe, **kwargs):
+        return LooseVersion(recipe.version) > version
+
+def is_version_lt(version):
+    def is_x(recipe, **kwargs):
+        return LooseVersion(recipe.version) < version
+    return is_x
```
opacam: I think that we could remove this two methods, since we don't use it anywhere...or maybe you were planning to use that in a near future...in that case just fixing the [the gh-actions pep8 error](https://github.com/kivy/python-for-android/pull/2044/checks?check_run_id=444633138#step:4:16) should be enough to make happy travis and gh-actions.

**Note:** if you remove the methods, you should also probably remove the import line: `from distutils.version import LooseVersion`

pythonforandroid/recommendations.py
```diff
@@ -187,7 +187,7 @@ def check_ndk_api(ndk_api, android_api):
 
 
 MIN_PYTHON_MAJOR_VERSION = 3
-MIN_PYTHON_MINOR_VERSION = 4
+MIN_PYTHON_MINOR_VERSION = 6
```
opacam: Aaaa...you made this change at #1975, see [AndreMiras comment](https://github.com/kivy/python-for-android/pull/1975/files#r319776002)...anyway, leave it if you want, just remember to remove it from there :wink:

pythonforandroid/python.py
```diff
@@ -438,10 +438,19 @@ def build_arch(self, arch):
...
-            # seems to be the normal and expected procedure.
-            shprint(sh.cp, join('Modules', 'Setup.dist'),
-                    join(build_dir, 'Modules', 'Setup'))
+            # Create the Setup file. This copying from Setup.dist is
+            # the normal and expected procedure before Python 3.8, but
+            # after this the file with default options is already named "Setup"
+            setup_dist_location = join('Modules', 'Setup.dist')
+            if exists(setup_dist_location):
+                shprint(sh.cp, join('Modules', 'Setup.dist'),
+                        join(build_dir, 'Modules', 'Setup'))
```
opacam: I think that you forgot to change `join('Modules', 'Setup.dist')` with the variable `setup_dist_location`...right?

```
            setup_dist_location = join('Modules', 'Setup.dist')
            if exists(setup_dist_location):
                shprint(sh.cp, setup_dist_location,
                        join(build_dir, 'Modules', 'Setup'))
```

pythonforandroid/recommendations.py
inclement: It doesn't matter where this change comes from, it just needs to be made.

pythonforandroid/patching.py
inclement: I'd prefer to leave these in, since the patching.py is a set of generic comparison tools that we might want to use in the future, but I've added some blank lines to make pylint happy.

inclement:
Thanks for the reviews. Looks like the only test failure is now in travis on macOS, with the `error: [Errno 2] No such file or directory: '/usr/local/include/python3.8/pyconfig.h'` that I've seen other people mention before, presumably because the Travis builds are using 3.7. I've changed the travis.yml to use 3.8 to see if that works, but does anyone know why osx (and only osx) is using the system's Python.h in the build?

inclement:
Looks like 3.8 isn't available on macos with travis anyway, so updating the travis.yml doesn't do anything useful. Let's leave it at 3.7 for now, it's fine to run p4a under this version and we can sort out multi-version testing later if we want.

This leaves the question of what to do with the failing macos build. Does anyone expect to commit a fix for the Python.h problem any time soon (or maybe I already missed one?)?

inclement:
Just ran the tests on the develop branch again, to establish that the macos pyconfig.h issues is present there as well. I think this PR can therefore be merged, the tests are only failing in one place, the same as on trunk.

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -18,15 +19,24 @@ class Python3Recipe(GuestPythonRecipe):
         :class:`~pythonforandroid.python.GuestPythonRecipe`
     '''
 
-    version = '3.7.1'
+    version = '3.8.1'
```
AndreMiras: Why bumping the default version to something "cutting edge"? I think it's great that we support it, but maybe it's a bit early to bump it as a default at the same time?

inclement:
On inspection, I don't think it's actually breaking the build for osx, 
since that build is also failing for the same reason on develop. Not 
sure what's up there - some osx environment change? I've seen users 
reporting similar issues there.

About bumping the version, I was taking the view that if this actively 
breaks something then that's probably something we want to know about 
anyway. Really it shouldn't cause any issues. I don't have a strong 
preference if you want to leave it at 3.7.1.

AndreMiras:
Re the broken osx build, I just saw from your other PR, the osx build is broken no matter what https://github.com/kivy/python-for-android/pull/1975
So I understand now this is not related to this one either and I misunderstood that :+1:

inclement:
Thanks. I'm going to go ahead and merge this, we can always change the default version later if preferred.

PR #1779, opened 2019-04-03
[WIP] Bump to dev SDL2 for touch fix & extract .java from SDL2 tarball

JonasT:
Bumps SDL2 to an experimental version to fix SDL_TOUCH_MOUSEID support with `SDL_HINT_ANDROID_SEPARATE_MOUSE_AND_TOUCH=1` ( https://bugzilla.libsdl.org/show_bug.cgi?id=4572 ) and fix the SDL2 bootstrap containing copied `.java` files inside our python-for-android source tree - instead they will now be copied from the SDL2 tarball directly, making it unnecessary to remember to sync the java files as long as our `SDLActivity.java` patch still cleanly applies

PR #2057, opened 2020-01-28
Add keyboard input type customization for SDL2

### LINKED ISSUE #796: Numeric keyboard doesn't work with SDL2
Not sure if this needs a change in Kivy, or p4a, or both.

See [here](https://github.com/kivy/python-for-android/issues/768) for original report and more info.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/35282962-numeric-keyboard-doesn-t-work-with-sdl2?utm_campaign=plugin&utm_content=tracker%2F11371&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F11371&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Franccisco:
This is the continuation of kivy/kivy#6716. With these changes, 
applications using SDL2 as window provider will be able to change the 
keyboard input type. This is achieved replacing the default value of 
outAttrs.inputType on SDLActivity.java for keyboardInputType. 
PythonActivity.java now has changeKeyboard, a function that basically 
allows the switching of the input type, modifying keyboardInputType to a 
given value. This function is called in Kivy through the android module.

Closes #2015, Fixes #796

### DIFF
--- a/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonActivity.java
+++ b/pythonforandroid/bootstraps/sdl2/build/src/main/java/org/kivy/android/PythonActivity.java
@@ -31,6 +31,7 @@ import android.os.AsyncTask;
 import android.os.Bundle;
 import android.os.PowerManager;
 import android.util.Log;
+import android.view.inputmethod.InputMethodManager;
 import android.view.SurfaceHolder;
 import android.view.SurfaceView;
 import android.view.View;
@@ -421,7 +422,7 @@ public class PythonActivity extends SDLActivity {
     /** Whether main routine/actual app has started yet **/
     protected boolean mAppConfirmedActive = false;
     /** Timer for delayed loading screen removal. **/
-    protected Timer loadingScreenRemovalTimer = null; 
+    protected Timer loadingScreenRemovalTimer = null;
 
     // Overridden since it's called often, to check whether to remove the
     // loading screen:
@@ -431,7 +432,7 @@ public class PythonActivity extends SDLActivity {
         considerLoadingScreenRemoval();
         return result;
     }
-   
+
     /** Confirm that the app's main routine has been launched.
      **/
     @Override
@@ -484,7 +485,7 @@ public class PythonActivity extends SDLActivity {
     public void removeLoadingScreen() {
         runOnUiThread(new Runnable() {
             public void run() {
-                if (PythonActivity.mImageView != null && 
+                if (PythonActivity.mImageView != null &&
                         PythonActivity.mImageView.getParent() != null) {
                     ((ViewGroup)PythonActivity.mImageView.getParent()).removeView(
                         PythonActivity.mImageView);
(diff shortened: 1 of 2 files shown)


Fak3:
Does it fix #796?

Franccisco:
It should:
https://github.com/kivy/kivy/blob/56d5b7ae665cee5210aa75e23582edbd7ca6c046/kivy/core/window/_window_sdl2.pyx#L548

chavarinbonbyn:
Which version contains the fix ?

Franccisco:
@chavarinbonbyn None yet. It's a work in progress (for more details, see kivy/kivy#6716).

inclement:
Thanks, this change seems good, but I'm wary about changing our SDLActivity code because I thought we were nowadays pulling this straight from the SDL project without changes. If we want to modify it, it would be ideal to do so only via the PythonActivity subclass.

Does anyone else have any thoughts about it?

AndreMiras:
I didn't realise that, then it's a very valid point.
I think we should clarify by maybe dynamically pulling this code just like we pull other recipe and patch when needed. @inclement would this make sense and how achievable this would be to you?

inclement:
It should be pretty straightforward, but I'm not sure what is the "correct" way to do it. That might only be because I haven't looked at it much for a while - I actually thought we were dynamically extracting the SDL2 deps in their entiretly.

Fak3:
The #1779 pr removes SDL files from p4a source, but that pr has not not been merged yet

inclement:
Thanks @Fak3, I really am behind!

Franccisco:
> Thanks, this change seems good, but I'm wary about changing our SDLActivity code because I thought we were nowadays pulling this straight from the SDL project without changes.

I know it is somehow a breaking change because I modified SDLActivity.

> If we want to modify it, it would be ideal to do so only via the PythonActivity subclass.

Do you guys have any suggestion on how to achieve this?

By the way, if you merge this pull request as it is, you'll find some bugs when using it with Kivy. It's described in kivy/kivy#6716 and #2015. The numeric keyboard is working as expected, but any TYPE_CLASS_TEXT keyboard without a PASSWORD variation isn't.

In summary, the bugs are:

* It seems the predictive text in TYPE_CLASS_TEXT keyboards is not working because of a missing implementation of on_textedit in the TextInput class.

* In old android versions the app will crash when inserting an emoji, and a tofu appears in newer android versions. It might be a font issue.

Franccisco:
Hello!

I wanted to try a different approach, but I realized the
SDLActivity.java file does not exist (the pull request #2113 is related
to it). In the new approach I tried I still need to modify the
SDLActivity file, so how do I proceed?

dwmoffatt:
@Franccisco What about changing the changeKeyboard in PythonActivity to modify mTextEdit directly instead of it doing a resetInput on it? 

In theory, this could work if the changes take effect without needing to do a resetInput on the mTextEdit

matham:
https://github.com/kivy/python-for-android/pull/2405 can close this?

AndreMiras:
Definitely, thanks for the heads up

PR #2063, opened 2020-02-16
virtualenv 20 breaks the osx build

AndreMiras:


AndreMiras:
Build green again :tada: merging

PR #2064, opened 2020-02-18
fixed patch's name to apply correctly

HirotsuguMINOWA:
fixed description of patch's name to apply correctly


AndreMiras:
We could also have renamed the patch file itself to keep its name explicit as it's actually removing the cortex flag. But I think this is equally fine, merging

PR #2065, opened 2020-02-20
Moved python3 patches to correct locations specified by recipe

inclement:


AndreMiras:
I think this was addressed via https://github.com/kivy/python-for-android/pull/2064 right?

inclement:
Oh yes, thanks.

PR #2067, opened 2020-02-24
[recipes] Update Pillow to v7.0.0

opacam:
Tested at runtime with an app of mine for `arm64.v8a` and `Python 3.8.1` (the default for p4a develop branch at the time of writing) and it seems to work fine.

**Note:** I tested it with an updated version of `freetype` and `harfbuzz` libraries (which I will create, a separate PR later) but it should also work with the current libraries we have at develop branch.


AndreMiras:
Macos build is also green now, feel free to merge when you're ready @opacam

PR #2068, opened 2020-02-24
[recipes] Update freetype & add zlib support

opacam:
Tested at runtime with an app of mine for `arm64.v8a` and `Python 3.8.1` (the default for p4a develop branch at the time of writing) using the updated Pillow recipe #2067 alongside an updated `harfbuzz` recipe (which I will publish later in a separate PR)

### DIFF
--- a/pythonforandroid/recipes/freetype/__init__.py
+++ b/pythonforandroid/recipes/freetype/__init__.py
@@ -24,7 +24,7 @@ class FreetypeRecipe(Recipe):
         https://sourceforge.net/projects/freetype/files/freetype2/2.5.3/
     """
 
-    version = '2.5.5'
+    version = '2.10.1'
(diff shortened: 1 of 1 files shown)


pythonforandroid/recipes/freetype/__init__.py
```diff
@@ -35,9 +35,6 @@ def get_recipe_env(self, arch=None, with_harfbuzz=False):
                 'harfbuzz', self.ctx
             ).get_build_dir(arch.arch)
             freetype_install = join(self.get_build_dir(arch.arch), 'install')
-            env['CFLAGS'] = ' '.join(
-                [env['CFLAGS'], '-DFT_CONFIG_OPTION_USE_HARFBUZZ']
-            )
```
opacam: Removed this macro because it's already defined with the latest version of `freetype`, and,  if we keep it, will throw a warning during the compilation

inclement:
Thanks!

PR #2069, opened 2020-02-24
[recipes] Update harfbuzz to v2.6.4

opacam:
Tested at runtime with an app of mine for `arm64-v8a` and `Python 3.8.1` (the default for p4a develop branch at the time of writing) using the updated Pillow recipe #2067 alongside an updated freetype recipe #2068


opacam:
Tried to restart gh-actions failed build but it seems that we are affected by actions/checkout#23

See also: https://github.com/kivy/python-for-android/runs/464566374?check_suite_focus=true#step:2:246

inclement:
Thanks!

PR #2046, opened 2019-12-27
Unify most of the test apps into `on device unit test app`

opacam:
So depending on the requested recipes at built time, a different test app will be ran (as well as different unit tests). We also set this *unified test app* the default for our `CI` tests (since we remove the `testapp_sqlite3_openssl` ... alongside with some more test apps).

There are 3 app modes:
  - **kivy** (if kivy detected in `app_requirements.txt`)
  - **flask** (if flask detected in `app_requirements.txt`)
  - **no-gui** (if above cases are not taken)

In all the cases mentioned the unittests will be run. In the two first use cases (kivy and flask) the result of the unit tests can be seen via the app. If an `no-gui` app is built, then the results must be checked via `adb logcat`.

Also be aware that this PR adds **more artifacts for gh-actions** via an `matrix`, so now we will have `on device unit test app` for:
   - **Python3 & arm64-v8a** *(tested on device)*
   - **Python3 & armeabi-v7a** *(tested on device)*
   - **Python3 & x86_64**
   - **Python3 & x86**
   - ~~**Python2 & arm64-v8a** *(tested on device)*~~
   - ~~**Python2 & armeabi-v7a** *(tested on device)*~~
   - ~~**Python2 & x86_64**~~

**A quick view of the situation:**
   - the removed `testapp_keyboard` and  `testapp_service`, now lives in the `on device unit test app`
   - the testapps for `pillow`, `matplotlib` and `encryption` has been moved into the unit tests (so we can also test those recipes via the flask app)
  - the testapp for `flask` has been also moved/reworked into the new `on device unit test app`, this allow us to share the resources with the kivy app (which are the same).
  - Furthermore,  we generate a file at build time `app_requirements.txt` which will hold the supplied recipes at build time. This file will allow us to test these recipes via our unittests. Just noting that the recipes we want to test, must be explicitly mentioned. Also should be mentioned that this feature its only available for **direct build with p4a** (when `setup_test_app.py` is ran)...so **buildozer builds will build an basic test app**, no matter what recipes are introduced into `buildozer.spec` file, because the  `app_requirements.txt` will only be created if the `setup_test_app.py` file is ran.

**Notes:**
  - @inclement, I included you as author of this thing (alongside myself), I think that we are the authors of the removed test apps and you are also the author of the base code for the `on device unit test app` (@AndreMiras and *anyone looking into this*, I think that you are not involved with these test apps, please, let me know if I'm mistaken and I will include you as author as well).
[...]

PR #2073, opened 2020-02-29
Playing with `on_device_unit_test_app` #2046

opacam:
**This is not to be reviewed nor merged**, of course commented if you want, but the purpose of this is to play a little with some recipes and the `on_device_unit_test_app` (#2046).

Some of the commits I will make in here, will be to test some active PRs. But I also will make some other commits which may be future PRs...if we like them. BTW, each `gh-actions` run should produce four apks, one for each arch, but probably it won't be the case...we will have failures on some/all archs (but this is fine...since is the first step to find a future solution). Also You should know that the generated artifacts via `gh-actions` are `.zip` compressed files with the apk inside them.

**So, this is what I'm planning to do in here:**
  - [x] Don't stop builds in case that one of the matrix jobs fails
  - [x] Test the current `Pillow` recipe, v5.2.0
  - [x] Test the updated `Pillow` recipe, v7.0.0 (#2067)
  - [x] Test current `matplotlib`'s recipe (I think that we have it broken alongside with some direct dependencies...due to the NDKr19 upgrade we made some months ago)
  - [x] Add `KiwisolverTestCase` for `on_device_unit_test_app` and test it
  - [x] Test `kiwisolver` fix (#2075)
  - [x] Test updated `matplotlib`'s recipe (#2076)
  - [x] Test the updated `Pillow` and `Matplotlib` recipes with updated libraries (#2068 and #2069)
  - [x] Test the updated `numpy` recipe (#2077)

**Notes:**
  - I will completely ignore `travis` results in here, because this is to test `gh-actions` generated artifacts
  - The `planning` may change along the way, so I will try to keep it updated
  - This tests will be useful because we are testing with latest develop branch (at the time of writing), so we are testing with our current default python version `3.8.1` (#2044)


opacam:
Hahahaha, first bugs found, for archs `x86` and `x86_64`...when building `harfbuzz` recipe in both cases:

BTW, the first commit I made here 9fca4b0 (not included in #2046), allow us to  go on if some arch fails to build...:thinking:...I think that we should include that in #2046, since it will be difficult to test at runtime all the generated artifacts (precisely those failed archs are the ones that I cannot test at runtime, so I don't usually test the build in my local machine...and it will probably be the same case for other members/contributors, since those kind of CPUs, AFAIK are not common on smartphones...). 

About the errors:
   - [x86_64](https://github.com/kivy/python-for-android/pull/2073/checks?check_run_id=477087948#step:4:26500): it looks like that we should include `-fPIC` to our `CFLAGS`, probably we should do that at class level at `Archx86_64`
  - [x86](https://github.com/kivy/python-for-android/pull/2073/checks?check_run_id=477087961#step:4:26579): This is not so clear to me...:thinking:

Anyway, let's go on and test the updated Pillow recipe #2067 and see what happens...

opacam:
Ok, as expected, same `harfbuzz` errors for mentioned archs: `x86` and `x86_64` for the updated `Pillow` recipe (v7.0.0), so let's merge the updated one, since it cannot broke what is already broken and, probably, it has better `Python 3.8.1` compatibility :wink:

I tested the generated apks at runtime for the updated `Pillow` recipe, both works with my android device running Android 10.

[Here are the generated apks, zip file compressed](https://github.com/kivy/python-for-android/actions/runs/47405021)

BTW, I never uploaded an image of how it will look like the `on device unit test app`, here you have the one I got with one of those generated apk via `gh-actions` (only showing one screen of the app in portrait mode, split into two screenshots):

Screenshot_20200229-233121
Screenshot_20200229-233153

AndreMiras:
It looks amazing, thanks for sharing :clap:

opacam:
Ok, `matplotlib` recipe doesn't even have a chance to be built, fails installing `kiwisolver`, [we may needed a `cppy` recipe](https://github.com/kivy/python-for-android/pull/2073/checks?check_run_id=477737509#step:4:35068)...so let's fix this...

To do so, I will create a couple of PRs, which then I will merge the corresponding branch into here. This way, we can play a little more with this, having the fixes ready to review into the proper PR and the testing in here, since the `matplotlib` fix will probably fail the CI tests, because it will depend of the `kiwisolver` fix.

opacam:
Ok, it seems that we managed to succed to build the matplotlib recipe, except for Archs: `x86_64` and `x86`. But this is not strange, remeber that, at this point, we never removed the `Pillow` recipe nor the `harfbuzz` library which already has been failling for the mentioned archs at earlier tests, so maybe we still have the chance to succeed at build stage for the `Pillow` and `matplotlib` recipes for the failed archs. 

A couple of screenshots to show you...extracted from one of the [generated apks from gh-actions CI test](https://github.com/kivy/python-for-android/actions/runs/47692275) (both apks works at runtime), now we also will see a `matplotlib` image besides the image from the `Pillow` tests:

Screenshot_20200301-175423
Screenshot_20200301-175444

About the errors:
  - [x86_64](https://github.com/kivy/python-for-android/pull/2073/checks?check_run_id=477969984#step:4:40862): looks like that zlib headers are not found when compiling Pillow...:thinking:...this is new...but remember that I added one extra commit here 7d61143 which affects this arch, so it seems that we go further than before, but not enough... 
  - [x86](https://github.com/kivy/python-for-android/pull/2073/checks?check_run_id=477969988#step:4:33829): the same linker error than before ...as expected ([mentioned here](https://github.com/kivy/python-for-android/pull/2073#issuecomment-592998163))
  - `travis error with macos`: also this is new, seems that we have a [build error with numpy and macos](https://travis-ci.org/kivy/python-for-android/jobs/656950155#L16598), at linking stage...

Ok then, let's be less demanding and let's test the thing by removing harfbuzz. I expect a failure for `x86_64` but maybe we succeed with `x86` and travis will probably fail again...

opacam:
Ok, seems that we improve the situation if we don't use `harfbuzz`, we only failed to build the `x86_64` arch...[Here are the generated apks](https://github.com/kivy/python-for-android/actions/runs/47752336). And below, images of the last runtime test I made (only arm64-v8a), but now in landscape mode...yes..the `on_device_unit_test_app`, by default, use the orientation based on the phone orientation :smile: :

Screenshot_20200301-200522
Screenshot_20200301-201050

Well then, let's test it now with the updated `freetype` recipe #2068

opacam:
Well it seems that the updated `freetype` recipe plays well with `Pillow` and `matplotlib`, we got the same failure than before for `x86_64`.

Tested one of the [generated apps](https://github.com/kivy/python-for-android/actions/runs/47775142) at runtime, in this case I tested `arm64-v8a` and it works as expected :smile:

Screenshot_20200301-232439

So, let's go further and test the updated `harfbuzz` recipe (#2069)...

opacam:
Yeah, the same situation that with the old `freetype` and `harfbuzz` libraries. [Here are the generated apks](https://github.com/kivy/python-for-android/actions/runs/47830438). In this case I tested, at runtime, arch `armeabi-v7a`, works fine.

So, so far we have this situation:
  - `harfbuzz` is broken for `x86` (I don't know what is happening here...needs investigation)
  - `Pillow` doesn't build for `x86_64` (not find zlib headers)
  - `numpy` doesn't compile for MacOS (at least in our CI tests)
  - `x86_64` may need some extra `CFLAGS` at Arch level (7d61143)

So now I will remove `harfbuzz` again, but also `Pillow` so we should see if we can get a succesful `matplotlib` build for all archs.

opacam:
More thoughts:
  - Successfully built all archs for `matplotlib` recipe (without `harfbuzz`) :smile:
  - MacOs build with `travis` is still failing (of course...nothing made to change that)
  - Yeah, `rebuild_updated_recipes` has no problem now with `matplotlib` recipe :tada:. BTW, if you look to the commit I made eb5e35a , you will see that the `matplotlib` recipe has hardcoded version for `freetype` library, and for `png` as well. That is not ideal and we probably want to get the version dynamically...anyway doesn't seem to affect the `matplotlib` compilation/installation....I suppose that `matplotlib` build only checks for a minimal version...still...before the merging of #2076, I'm planning to do a couple more of commits to address these things :wink:
  -  I cannot test, at runtime,  [the generated apks](https://github.com/kivy/python-for-android/actions/runs/48432064) for `x86` and `x86_64`...will work at runtime? :no_mouth:

opacam:
Ok, we have `matplotlib` recipe merged into `develop` branch, with the issues mentioned before already fixed in there. BTW, thanks for the fast review @AndreMiras :+1: :)

So now I want to do one last test in here, the `numpy` update #2077, because `matplotlib` recipe depends on it...so better not to broke that, right? So...let's continue the investigation :mag_right:

opacam:
Yeah, The new `numpy` recipe plays nice with the matpltolib recipe for `armeabi-v7a` and `arm64-v8a` ([Here are the generated apks](https://github.com/kivy/python-for-android/actions/runs/50161192))

BTW, the error we got are [the expected ones](https://travis-ci.org/github/kivy/python-for-android/builds/658715939?utm_medium=notification&utm_source=github_status):
  - MacOsX test app build fails for `travis`
  - Python2 fails because the new `numpy` recipe is not compatible with it.

**So, as a conclusion for the `numpy` test:**
The new `numpy` recipe is not compatible with `python2`. Given that [python2 will not be maintained anymore](https://pythonclock.org/), and the fact that we have the `numpy` recipe posted and approved without merging for several days without anyone complaining, I think that the show must go on, so let's merge `numpy` into `develop` branch :smile:

iruriksl:
The issue is still open as I faced it while building for armeabi-v7a & arm64-v8a android arch as the results are also crossed in checks liink.

Is their any work around for mentioned arch?

Issue #1900, opened 2019-06-29
Matplotlib recipe depends on local environment


inclement:
See https://travis-ci.org/kivy/python-for-android/jobs/551991431#L5947-L5958 for failing build in ci environment.

The recipe works for me but is probably getting things like the freetype headers from the wrong places (i.e. system versions). It needs fixing to correct this, and removing from the broken recipes list in `ci/constants.py`.

AndreMiras:
I'm sharing the relevant part of the log here in case the Travis one gets deleted.
```
[INFO]:    Building matplotlib for armeabi-v7a
[INFO]:    matplotlib apparently isn't already in site-packages
[INFO]:    Building compiled components in matplotlib
[INFO]:    -> directory context /home/user/.local/share/python-for-android/build/other_builds/matplotlib-python3/armeabi-v7a__ndk_target_21/matplotlib
stty: 'standard input': Inappropriate ioctl for device
[INFO]:    -> running python setup.py build_ext -v
Exception in thread background thread for pid 11361:
Traceback (most recent call last):
  File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/home/user/venv/lib/python3.6/site-packages/sh.py", line 1540, in wrap
    fn(*args, **kwargs)
  File "/home/user/venv/lib/python3.6/site-packages/sh.py", line 2459, in background_thread
    handle_exit_code(exit_code)
  File "/home/user/venv/lib/python3.6/site-packages/sh.py", line 2157, in fn
    return self.command.handle_command_exit_code(exit_code)
  File "/home/user/venv/lib/python3.6/site-packages/sh.py", line 815, in handle_command_exit_code
    raise exc
sh.ErrorReturnCode_1: 
  RAN: /home/user/.local/share/python-for-android/build/other_builds/hostpython3/desktop/hostpython3/native-build/python setup.py build_ext -v
  STDOUT:
============================================================================
Edit setup.cfg to change the build options
BUILDING MATPLOTLIB
            matplotlib: yes [3.0.3]
                python: yes [3.7.1 (default, Jun 28 2019, 23:44:36)  [GCC
                        7.4.0]]
              platform: yes [linux]
REQUIRED DEPENDENCIES AND EXTENSIONS
                 numpy: yes [not found. pip may install it below.]
[...]

Issue #1750, opened 2019-03-12
System requirements checks proposal


AndreMiras:
### Versions

* Python: N/A
* OS: Linux
* Kivy: N/A
* Cython: N/A

### Description

We often receive bug reports that ends up being related to a missing system dependency.
Often the stacktrace is not obvious on what's missing.
For instance see:

- https://github.com/kivy/python-for-android/issues/1710, https://github.com/kivy/python-for-android/issues/1858, https://github.com/kivy/buildozer/issues/841 (libtinfo5 and deps)
- https://github.com/kivy/python-for-android/issues/1576 (libffi-dev and other deps)
- https://github.com/kivy/python-for-android/issues/1900 (matplotlib requires libfreetype6-dev, pkg-config and libpng12-dev)
- will update this list as I come across similar issues

In order to save developers troubleshooting time, I think we could make some optional system requirements prechecks using e.g. pkg-config and raise a warning detected missing system dependency.
This would be a first approach in tempting to mitigate these issues.
Also at first we could start with a per-recipe specific approach until we find a maybe a more generic way e.g. `Recipe.system_dependencies` member or something.

inclement:
Quoting Phenus on discord:

> The buildozer instructions are missing 4 or 5 dependencies
> zip, automake, autoconf and libltdl-dev

PR #2076, opened 2020-03-01
Fix `matplotlib` and update to `v3.1.3`

opacam:
This PR depends on a working `kiwisolver` and `cppy` recipes (#2075).

**Closes:** #1900, kivy/buildozer#1003

**Notes:** I will merge the branch of this PR alongside the branch of #2075 into #2073 so we can see if the thing is working with the `on_device_unit_test_app`.

**See also:**
  - [this comment](https://github.com/kivy/python-for-android/pull/2073#issuecomment-593102184)
  - #1750

### DIFF
--- a/pythonforandroid/recipes/matplotlib/__init__.py
+++ b/pythonforandroid/recipes/matplotlib/__init__.py
@@ -54,11 +54,15 @@ class MatplotlibRecipe(CppCompiledComponentsPythonRecipe):
             # read template file into buffer
             with open(pc_template_file) as template_file:
                 text_buffer = template_file.read()
-            # set the library absolute path
+            # set the library absolute path and library version
+            lib_recipe = self.get_recipe(lib_name, self.ctx)
             text_buffer = text_buffer.replace(
-                'path_to_built',
-                self.get_recipe(lib_name, self.ctx).get_build_dir(arch.arch),
+                'path_to_built', lib_recipe.get_build_dir(arch.arch),
             )
+            text_buffer = text_buffer.replace(
+                'library_version', lib_recipe.version,
+            )
+
             # write the library pc file into our defined dir `PKG_CONFIG_PATH`
             pc_dest_file = join(pkg_config_path, lib_to_pc_file[lib_name])
             with open(pc_dest_file, 'w') as pc_file:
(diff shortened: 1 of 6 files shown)


pythonforandroid/recipes/matplotlib/__init__.py
```diff
@@ -1,26 +1,94 @@
...
 
     call_hostpython_via_targetpython = False
 
+    def generate_libraries_pc_files(self, arch):
+        """
+        Create *.pc files for libraries that `matplotib` depends on.
+
+        Because, for unix platforms, the mpl install script uses `pkg-config`
+        to detect libraries installed in non standard locations (our case...
+        well...we don't event install the libraries...so we must trick a little
```
AndreMiras: `we don't event` -> `we don't even`

pythonforandroid/recipes/matplotlib/mpl_android_fixes.patch
```diff
@@ -1,29 +1,64 @@
...
--        return ['numpy>=1.10.0']
-+        return ['numpy==1.16.4']  # to match p4a's target version
-
+-        return ['numpy>=1.11']
++        return []  # don't need it for p4a, due to above changes
+ 
      def get_install_requires(self):
--        return ['numpy>=1.10.0']
+-        return ['numpy>=1.11']
 +        return ['numpy==1.16.4']  # to match p4a's target version
```
AndreMiras: I have some doubts with this line. Does it really have to be hard pinned to the current default p4a numpy version?
And if it does, should we at some point try to pick this up dynamically?

opacam: > I have some doubts with this line. Does it really have to be hard pinned to the current default p4a numpy version?

I'm not sure, I'm doing a local test...we will see

> And if it does, should we at some point try to pick this up dynamically?

I think so, but it will add more complexity to the recipe...

AndreMiras: Yes I was also thinking about the complexity. I think we can address that later if needed.
You have my :heavy_check_mark: feel free to merge if it builds :green_heart:

opacam:
Thanks for the `Approval` @AndreMiras :smile:

**Note:** I updated the first post, since with the changes made in here we can close some issues.

opacam:
I think that this is ready to :twisted_rightwards_arrows:

**let's do it!!**

:smile_cat:

PR #2077, opened 2020-03-02
Update `numpy` to v1.18.1 (add `cython` recipe)

opacam:
We add the `cython` recipe because the last version of `numpy` (at time of writing v1.18.1), requires `cython` installed as a Python module during the build stage of the `numpy` recipe.

Also this will make the `numpy` recipe only available for `Python3` since it requires python>=3.5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,12 +47,12 @@ jobs:
       after_success:
         - coveralls
     - &testapps
-      name: Python 3 arm64-v8a
+      name: Python 3 arm64-v8a (with numpy)
       stage: build testapps
       before_script: make docker/pull
       script: make docker/run/make/testapps/python3/arm64-v8a
     - <<: *testapps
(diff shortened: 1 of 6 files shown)


opacam:
Seems that the new `numpy` recipe [is also failing for MacOs](https://travis-ci.org/kivy/python-for-android/jobs/657095323#L8687)...similar error as the old `numpy` version tested at #2046 ([see this comment](https://github.com/kivy/python-for-android/pull/2073#issuecomment-593126362))...anyway I just wanted to test in here, imho, it is not a good idea rely such test on MacOs build, so I will move it.

**Note:**: Just noting that this cannot break what is already broken, but we still will lose `numpy` for `Python2`...well...I don't think that this fact is so important considering [that it won't be maintained anymore](https://pythonclock.org/)...we should remove it, probably sooner than later

PR #2088, opened 2020-03-17
Playing with `on_device_unit_test_app` #2046 - Episode II

opacam:
**This is not to be reviewed nor merged**, of course commented if you want, but the purpose of this is to play a little more with some new recipes and the `on_device_unit_test_app` (#2046)...so I understand this as some kind of continuation of #2076 (just to have a new start point with the current develop branch).

I'm not sure if we have to include all the changes I advance to you in here, but sure that there are some new recipes that can be very useful. I made this to make it work [mplfinance](https://github.com/matplotlib/mplfinance), a kind of `matplotlib`'s extension. To do so I faced several issues along the way, which forced me to introduce a couple of libraries that will allow us to build some python compiled modules that currently aren't compiled: `_bz2.cpython-38.so` and `_lzma.cpython-38.so`. Also I was forced to create a [Pandas](https://github.com/pandas-dev/pandas) recipe...this one is interesting right?

Anyway, If you guys find this interesting, I will create separate PRs...so give me your thoughts about this:

   - should we include libraries: `bz2` and `lzma`?
   - should we enable the built of some extra python modules: `_bz2.cpython-38.so` and `_lzma.cpython-38.so` (depends on previous steps)?
   - should we add a `pandas` recipe (depends on previous steps)?
   - should we introduce `mplfinance` recipe (again...depends on previous steps)?

I think that the first three points are worth it...the last one probably too (it doesn't seem hard to maintain).


pythonforandroid/recipes/mplfinance/__init__.py
```diff
@@ -0,0 +1,18 @@
+
```
AndreMiras: Typically this recipe as well as `pandas` could have been added in another dedicated PR, right?
This two recipe alone are simple enough to review, but in general if should try to reduce huge PR when possible to reviewers don't get scared/discouraged by the size of it or lose focus during the review. Also if large PR receive a lot of comments then the contributor might as well get discouraged or simply side tracked and drop it.
By the way line one is empty :smile:

pythonforandroid/recipes/mplfinance/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+from pythonforandroid.recipe import PythonRecipe
+
+
+class MplfinanceRecipe(PythonRecipe):
+
+    # there is not an official release yet...
+    # so pinned to last commit at the time of writing
+    version = 'cde5008'
+    url = 'https://github.com/matplotlib/mplfinance/archive/{version}.zip'
```
AndreMiras: Jugging from PyPI there're some alpha releases we could use
```suggestion
    version = '0.12.3a3'
    url = 'https://pypi.python.org/packages/source/m/mplfinance/mplfinance-{version}.tar.gz'
```

testapps/on_device_unit_tests/README.rst
```diff
@@ -4,18 +4,38 @@ On device unit tests
...
+  - `no gui`: if neither of above cases is taken
+
 The main tests are for the recipes built in the apk. Each module (or
 other tool) is at least imported and subject to some basic check.
 
-This app is experimental, it doesn't yet support things like testing
-only the requirements you ask for (so if you build with requirements
-other than those specified, the tests may fail). It also has no gui
-yet, the results must be checked via logcat.
+This test app can be build via `setup.py` or via buildozer. In both
```
AndreMiras: `can be build` -> `can be built`

testapps/on_device_unit_tests/setup.py
```diff
@@ -0,0 +1,92 @@
+"""
+This is the `setup.py` file for the `on device unit test app`.
+
+In this module we can control how will be built our test app. Depending on
+our requirements we can build an kivy, flask or a non-gui app. We default to an
```
AndreMiras: `an kivy` -> `a kivy`

testapps/on_device_unit_tests/setup.py
```diff
@@ -0,0 +1,92 @@
...
+kivy app, since the python-for-android project its a sister project of kivy.
```
AndreMiras: `its a sister` -> `is a sister`

inclement:
Since I'm a little late, what's the status of this, does anything still need comments?

opacam:
Actually we can close this because  this was only a draft and, the important stuff, we have it already merged in separate PRs, anyhow...there maybe some interesting tests for the `on_device_unit_tests` and one recipe for `mplfinance`, but not to be done here...so closing this and...
**¡¡thanks to be here guys!!**
:smile:

PR #2091, opened 2020-03-22
Update quickstart.rst

BornForFever:


AndreMiras:
Ahah just realise this was pull requested into the wrong branch :smile: It should be targeting `develop`

PR #2092, opened 2020-03-22
Android library

zworkb:
# p4a support for service-backend by packaging an .aar library

This PR is built on #1963 submitted by @xuhcc who did great work in this direction, but unfortunately seems to have not the time [to continue the work](https://github.com/kivy/python-for-android/issues/1957#issuecomment-587058496), so most credit goes to @xuhcc.

In the first turn I took his work and made it work from the current p4a develop (including a buildozer integration, see corresponding PR kivy/buildozer#1037)

I am willing to bring this to a form so that it can be merged and am open for discussions and happy for some good advice...

Most promising was [inclement's proposal](https://github.com/kivy/python-for-android/pull/1963#issuecomment-526864800) in which he proposed to pimp up the `service_only` bootstrap and move the build steps into an `aar` command.

If no other advice is coming I would go that way, since the service_only bootstrap seems to fit best here, but I don't want to break anything.

I have set up a [test project](https://github.com/zworkb/p4a-library-demo) where this can be tested.

UPDATE:

I added support for inheritance in bootstraps, so that a bootstrap inheriting from another bootstrap only needs to ad the files that differ from the original into it's bootstrap dir.

So we avoid file duplication.

what has been done so far:

- ported @xuhcc's PR to recent p4a master
- added python3.8 support to pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonUtil.java
- openssl recipe switched to version '1.1.1e' because in 1.1.1 there was a print without ()
- bootstraps dirs are cumulatively copied based on their inheritance, can be arbitrarily deep
- add `symlink` param to copy_files, when set the copy target are symlinked
- support for the `aar` directive in buildozer
- create a 'p4a aar' command, so that lots of cluttering conditionals can be moved away from `toolchain.apk()`
- began to remove ant support (@inclement allowed me to do so)
- renamed `library` bootstrap to `service_library`, because that describe it better
- test setup `setup_testlib_service.py`

@inclement's change requests are implemented now
PR is ready to merge now


rnixx:
Awesome! Bundling a python service as aar and deliver it as part of another application in the end feels more like a real world usecase than delivering a python service standalone as apk, especially since the more rescticted service changes in android API.

Referring the bootstrap. the change looks like this can easily be done conditionally in the service only bootstrap.

Since it looks pretty much aar generation is orthogonal to bootstraps, would be cool if a python activity and the other bootstraps if it makes sense also can be bundled as aar. I have no idea what this implies on in code though.

pythonforandroid/bootstrap.py
```diff
@@ -138,20 +141,22 @@ def name(self):
...
-        self.common_dir = self.get_common_dir()
-        copy_files(join(self.bootstrap_dir, 'build'), self.build_dir)
-        copy_files(join(self.common_dir, 'build'), self.build_dir,
-                   override=False)
-        if self.ctx.symlink_java_src:
-            info('Symlinking java src instead of copying')
-            shprint(sh.rm, '-r', join(self.build_dir, 'src'))
-            shprint(sh.mkdir, join(self.build_dir, 'src'))
-            for dirn in listdir(join(self.bootstrap_dir, 'build', 'src')):
-                shprint(sh.ln, '-s', join(self.bootstrap_dir, 'build', 'src', dirn),
```
AndreMiras: This is the reason the `test_bootstrap_prepare_build_dir_with_java_src()` unit test is failing. It's because we no longer call `sh.ln` in this `prepare_build_dir()` function. So it's fine, we just need to remove the mock in the unit test for this one. So basically you need to edit the `test_bootstrap_prepare_build_dir_with_java_src()` test and remove the `sh.ln` patching as well as the assert we're doing with it and we should be all good

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -274,7 +274,7 @@ def compile_dir(dfn, optimize_python=True):
 def make_package(args):
     # If no launcher is specified, require a main.py/main.pyo:
     if (get_bootstrap_name() != "sdl" or args.launcher is None) and \
-            get_bootstrap_name() != "webview":
+            get_bootstrap_name() not in ["webview", "service_library"]:
```
rnixx: This look like it should be a setting on the bootstrap.

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -529,8 +529,9 @@ def make_package(args):
         aars=aars,
         jars=jars,
         android_api=android_api,
-        build_tools_version=build_tools_version
-        )
+        build_tools_version=build_tools_version,
+        is_library=get_bootstrap_name() == 'service_library',
```
rnixx: This also looks like a candidate for a setting on the bootstrap class

pythonforandroid/bootstraps/common/build/whitelist.txt
```diff
@@ -1 +0,0 @@
-# put files here that you need to un-blacklist
```
inclement: Is this an unintended deletion?

pythonforandroid/build.py
```diff
@@ -484,7 +484,9 @@ def set_archs(self, arch_names):
             ', '.join([arch.arch for arch in self.archs])))
 
     def prepare_bootstrap(self, bs):
-        bs.ctx = self
+        if bs:
+            # XXX: this condition fixes tox ci tests, dont know when in reality we have no bootstrap
```
inclement: It sounds like this is because prepare_bootstrap is being passed None when it shouldn't be. I would prefer to keep the code with prepare_bootstrap expecting a real argument, and the tests passing one if necessary, but I haven't looked at the details of the failure - do you think there's some reason that wouldn't be practical?

pythonforandroid/recipes/android/__init__.py
```diff
@@ -41,7 +41,7 @@ def prebuild_arch(self, arch):
         is_webview = bootstrap_name == 'webview'
         is_service_only = bootstrap_name == 'service_only'
 
-        if is_sdl2 or is_webview or is_service_only:
+        if is_sdl2 or is_webview or is_service_only or bootstrap_name == 'service_library':
```
inclement: Prefer adding this to the list of conditions above - or in fact, you could just replace the whole thing with `bootstrap_name in ["sdl2", "sdl2_gradle", "webview", "service_only", "service_library"]`. Even the sdl2_gradle there is probably an obselete name.

pythonforandroid/bootstrap.py
```diff
@@ -138,20 +141,22 @@ def name(self):
...
-                        join(self.build_dir, 'src'))
+
+        # get all bootstrap names along the __mro__, cutting off Bootstrap and object
```
inclement: This introspection feels like it should be its own method.

pythonforandroid/bootstrap.py
```diff
@@ -138,20 +141,22 @@ def name(self):
...
+        classes = self.__class__.__mro__[:-2]
+        bootstrap_names = [cls.name for cls in classes] + ['common']
+        bootstrap_dirs = [
+            join(self.ctx.root_dir, 'bootstraps', bootstrap_name)
+            for bootstrap_name
+            in reversed(bootstrap_names)
```
inclement: Prefer to merge this with the previous line, it's an unnecessary line break.

pythonforandroid/bootstraps/common/build/build.py
inclement: Could you make this `is_library=(get_bootstrap_name() == 'service_library'),`, I think it's clearer.

inclement: It's set up this way because the dist is nominally independent of the main p4a code, although that's an old design decision that isn't really true any more. So, this is fine for now, but I agree it would be nice to consolidate the whole api.

pythonforandroid/bootstrap.py
```diff
@@ -138,20 +141,22 @@ def name(self):
...
+        ]
+        # now do a cumulative copy of all bootstrap dirs
+        for bootstrap_dir in bootstrap_dirs:
+            copy_files(join(bootstrap_dir, 'build'), self.build_dir, symlink=self.ctx.symlink_java_src)
```
inclement: If we're optionally symlinking everything now, could you rename `symlink_java_src` to `symlink_bootstraps` or similar?

inclement:
This looks good to me, just a few minor comments in the review.

pythonforandroid/bootstraps/common/build/whitelist.txt
zworkb: oops, yes it seems so, should I undo that or are you doing it during merge?

inclement: Please do undo it

pythonforandroid/build.py
zworkb: I agree, thats the reason why I put in that comment. prepare_bootstrap now raises a TypeError when bootstrap is None, fixed tests to reflect that

pythonforandroid/bootstrap.py
zworkb: renamed everythong to symlink_bootstrap_files

pythonforandroid/bootstraps/common/build/build.py
zworkb: done

pythonforandroid/bootstrap.py
zworkb: done

pythonforandroid/recipes/android/__init__.py
zworkb: done

pythonforandroid/bootstraps/common/build/whitelist.txt
zworkb: done

pythonforandroid/bootstrap.py
zworkb: done

inclement:
Thanks for all the rebase effort!

PR #2097, opened 2020-03-24
:sparkles: Compression libraries - episode III: add support for libbz2 & liblzma to python3

opacam:
This is the third part and last of a set of recipes that will add add support for libbz2 & liblzma to python3 recipe, allowing us to build some extra python modules:
  - `_bz2.cpython-38.so`
  - `_lzma.cpython-38.so`

Also I put up-to-date the python3 recipe documentation :pencil:

These libraries are required by some python packages, like `pandas`, which I plan to create a PR later :wink:

**Notes:**
  - tested at #2088
  - **this PR should not be merged before previous PRs, named `Compression libraries...`**: #2095 and #2096

### DIFF
--- a/pythonforandroid/recipes/python3/__init__.py
+++ b/pythonforandroid/recipes/python3/__init__.py
@@ -6,14 +6,29 @@ from pythonforandroid.patching import version_starts_with
 
 class Python3Recipe(GuestPythonRecipe):
     '''
-    The python3's recipe.
+    The python3's recipe
+    ^^^^^^^^^^^^^^^^^^^^
 
-    .. note:: This recipe can be built only against API 21+. Also, in order to
-        build certain python modules, we need to add some extra recipes to our
-        build requirements:
+    The python 3 recipe can be built with some extra python modules, but do so,
+    we need some libraries. Per default, we ship the python3 recipe with some
+    common libraries, defined at ``depends``.We also support some optional
+    libraries, which are less common that the ones defined at ``depends``, so
+    we added them as optional dependencies (``opt_depends``).
 
-            - ctypes: you must add the recipe for ``libffi``.
+    Below you have a relationship between the python modules and the recipe
+    libraries::
 
+        - _ctypes: you must add the recipe for ``libffi``.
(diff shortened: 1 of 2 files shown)


pythonforandroid/recipes/python3/__init__.py
```diff
@@ -6,14 +6,29 @@
...
 class Python3Recipe(GuestPythonRecipe):
     '''
-    The python3's recipe.
+    The python3's recipe
+    ^^^^^^^^^^^^^^^^^^^^
 
-    .. note:: This recipe can be built only against API 21+. Also, in order to
-        build certain python modules, we need to add some extra recipes to our
-        build requirements:
+    The python 3 recipe can be built with some extra python modules, but do so,
```
AndreMiras: `but do so` -> `to do so`

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -6,14 +6,29 @@
...
+    we need some libraries. Per default, we ship the python3 recipe with some
```
AndreMiras: `Per default` -> `By default`

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -6,14 +6,29 @@
...
+    common libraries, defined at ``depends``.We also support some optional
```
AndreMiras: I would say `defined in` but I suck at grammar. Also missing a space between the dot and "We"

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -6,14 +6,29 @@
...
+    libraries, which are less common that the ones defined at ``depends``, so
```
AndreMiras: `defined in`

pythonforandroid/recipes/python3/__init__.py
opacam: I'm not sure about this but what I'm sure is that your English is a lot better than mine...so let's go for `in`, thanks!!! :smile:

PR #2100, opened 2020-03-25
:sparkles: Add `pandas` recipe

opacam:
**Notes:**
  - tested at #2088
  - the version of pandas tested above is not the same the we finally introduced in here, because the last days, the pandas team released a bug fix release, so I updated the version :wink:
  - Since the version tested at #2088 is not the same that I put in here, I tested it locally (at build and at runtime) with an updated `pytz` recipe, which I will post later in a separate PR.

### DIFF
--- a/pythonforandroid/recipes/pandas/__init__.py
+++ b/pythonforandroid/recipes/pandas/__init__.py
@@ -1,0 +1,35 @@
+from os.path import join
+
+from pythonforandroid.recipe import CppCompiledComponentsPythonRecipe
+
+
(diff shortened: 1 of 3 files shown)


AndreMiras:
Build is now green

opacam:
@AndreMiras, Thanks to restart the build and the approval!! :smile:

Let's merge the `pandas` recipe now :tada:

inclement:
Didn't see this PR at the time, but seeing the message in the log was a nice surprise - awesome work!

PR #2101, opened 2020-03-25
:arrow_up: Update `pytz` to version `2019.3`

opacam:
Because is one of the dependencies of `pandas` recipe #2100,
and despite that the old version seems to work fine,
I think it's better to go for to the latest available version.


opacam:
@tshirtman, **Thanks !!**, let's merge this :smile:

PR #2103, opened 2020-03-26
Update docker.rst

yingshaoxo:
I don't know if the old way works, but this new way works for me.

### DIFF
--- a/doc/source/docker.rst
+++ b/doc/source/docker.rst
@@ -66,3 +66,34 @@ already have Docker preinstalled and set up.
 5. Find the ``.APK`` file on this location::
 
        ls -lah testapps
+
+Docker Method2
+======
+
+Here's a more practical and understandable way, it requires you to install SDK and NDK manully first.
+
+1. Download an example::
+
+       git clone https://github.com/yingshaoxo/kivy-diary.git
+       cd kivy-diary
+
+2. Prepare your SDK and NDK::
+
+       Use Android Studio SDK Manager to install it. Here I installed SDK 29 and NDK 21.
+
+3. Run it (you need to change the first SDK path)::
+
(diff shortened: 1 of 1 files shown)


inclement:
Thanks @yingshaoxo, I have no objection to this (but made a small wording change). I'll defer to @AndreMiras as the docker expert though.

AndreMiras:
Thanks @yingshaoxo for the pull request.
I would rather not rely on a repo that's not under the kivy organisation.
So I think it would be better if the method were more generic and not relying on `yingshaoxo/kivy-diary`.
If I understood correctly, the method you're presenting is about postponing the one-liner we have to getting a shell to the container in `--interactive` mode to later run the p4a command within it. Or is it something else?
If it's the case, then I think adding a note saying so should be enough.
Also we could add to clarify that all `p4a` commands are then available within this shell e.g `p4a apk --private ...`

yingshaoxo:
@AndreMiras 

Yeah, you're right. One line of docker command won't be enough, often we need to get into the container to do more stuff to make things done.

I use `yingshaoxo/kivy-diary` there is because when I write that tutorial, I suddenly realize that I couldn't make an clear explanation without an existing compilable kivy repository. (I have to show them how to set the right Env variables and how to use p4a.) You guys could use your own if you have one. It's not a problem.

The initial purpose for me to write that tutorial is :

1. I think the traditional way of installation for compiling kivy app is too complicate to make. A simple point would be you have to install JDK-8 to use p4a. But it's out of date. And it definitely would conflict with modern software in Linux. The only reasonable way for compiling kivy is to use Docker.

2. For those people in China, they're facing a big problem from the GFW. So it can be very difficult for them to install the whole SDK and NDK within the docker image. Meanwhile, the docker container is cheap, it can be messed up or destroyed easily. That's why local SDK and NDK with Android Studio matters.

In the end, I'm just a contributor, I do want I can to make things simpler, as for whether you take it, it depends on you. But if Python wants to exceed from the competition of JavaScript, it must be a more friendly cross-platfrom programming language.

PR #2106, opened 2020-03-28
:fire: Move to python3 `super` calls

opacam:


opacam:
BTW, I made this after #2015, so I recommend to merge it after that. You can see the tox error complaining about [python2 entry point test](https://github.com/kivy/python-for-android/pull/2106/checks?check_run_id=541495076#step:4:58)...anyhow once we have that merged we should be fine

inclement:
Thanks!

PR #2105, opened 2020-03-28
:fire: Drop Python 2 support

opacam:
In this PR, we drop support for Python 2...so no more `python2` recipe.

Quick view of what we do:
  - remove `python2` recipe
  - remove `hostpython2` recipe
  - remove python2 testapps
  - remove python2's specific code
  - Adapt recipes to the new situation
  - Adapt our tests to the new situation
  - Adapt our CI test to the new situation
  - Update docs

**Notes**
  - we will not have a green tick mark for our `rebuild_updated_recipes` CI test, since we have to touch a lot of recipes and some of them will create a [dependencies conflict](https://travis-ci.org/github/kivy/python-for-android/jobs/668007165#L469).

PR #2109, opened 2020-03-28
Updates quickstart.rst "Installing Dependencies"

AndreMiras:
- orders dependencies bullets alphabetically
- updates Arch Linux dependencies

### DIFF
--- a/doc/source/quickstart.rst
+++ b/doc/source/quickstart.rst
@@ -61,33 +61,35 @@ Installing Dependencies
 
 p4a has several dependencies that must be installed:
 
-- git
 - ant
-- python (We recommend using python3.)
+- autoconf (for ffpyplayer_codecs recipe)
+- ccache (optional)
+- cmake (required for some native code recipes like jpeg's recipe)
 - cython (can be installed via pip)
-- a Java JDK (e.g. openjdk-8)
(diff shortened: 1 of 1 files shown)


doc/source/quickstart.rst
```diff
@@ -82,12 +85,11 @@ install most of these with::
     sudo apt-get update
     sudo apt-get install -y build-essential ccache git zlib1g-dev python2.7 python2.7-dev libncurses5:i386 libstdc++6:i386 zlib1g:i386 openjdk-8-jdk unzip ant ccache autoconf libtool
```
opacam: Maybe we should update python version for debian/ubuntu:
```suggestion
    sudo apt-get install -y build-essential ccache git zlib1g-dev python3.7 python3.7-dev libncurses5:i386 libstdc++6:i386 zlib1g:i386 openjdk-8-jdk unzip ant ccache autoconf libtool
```
**Note:** I updated this line at #2105, but I will fix the merge conflicts, if you want to do it in here, no problem (I wil have to do it anyway)

AndreMiras: Done, I went with python3 and python3-dev meta package rather than explicit full version. I hope that's fine

PR #2113, opened 2020-03-30
Bump to SDL2 2.0.10 & extract .java from SDL2 tarball: merge conflicts fixed

inclement:
This is the code from https://github.com/kivy/python-for-android/pull/1779 with (minor) merge conflicts fixed.


inclement:
If the tests pass, I intend to merge this right away.

opacam:
The [flake 8 errors](https://travis-ci.org/github/kivy/python-for-android/jobs/668837611#L266-L280) now are fixed by #2114, a new merge/rebase of develop branch should fix that

pythonforandroid/bootstrap.py
```diff
@@ -137,6 +137,23 @@ def name(self):
...
+            src_dir = join(sdl2_build_dir, "SDL", "android-project",
+                           "app", "src", "main", "java",
+                           "org", "libsdl", "app")
+            target_dir = join(self.dist_dir, 'src', 'main', 'java', 'org',
+                              'libsdl', 'app')
+
+            # Do actual copying:
+            info('Copying in SDL2 .java files from: ' + str(src_dir))
+            if not os.path.exists(target_dir):
+                os.makedirs(target_dir)
```
AndreMiras: We could save one line by simulating the `--parents` flag:
```suggestion
            os.makedirs(target_dir, exist_ok=True)
```

inclement:
Unsurprisingly there are some minor test failures, but the app builds seem to work fine. I'll see if I can fix it quickly.

I'm finding that test_pythonpackage.py runs out my ram within a few seconds, which didn't happen in the past. Not sure what's up there.

pythonforandroid/bootstrap.py
etc0de: I think I wrote it like this to maintain Python 2 compatibility. So yes, for 3.x-only shortening it like this makes sense :+1:

AndreMiras:
I actually often skip the `test_pythonpackage.py` in local because for me it always ran for ages

opacam:
If I remember well...the problem of these tests are that  we don't  mock at all the file/dir reading/creation and this has a serious impact to the tests performance. So, as @AndreMiras, I almost always skip these tests as well.

Anyway, I also tested one of the [generated apks](https://github.com/kivy/python-for-android/actions/runs/67617605) for `on_device_unit_test_app`, the one for `arm64-v8a`, and it works ok, actually I think that it works better, since the kivy's keyboard features now I think that are working as expected.

etc0de:
For what it's worth, `test_pythonpackage.py` tests with pip because pip has no stable interface and likes to break things, and there is no libpip or other library-level alternative. So this is kind of an external problem that cannot be easily solved unless pip decides to do so.

inclement:
Got sidetracked on this but still trying to prioritise it over anything else.

I fixed a bug with the sdl2 patching, on runs after the first one it was partially-applying the patch before deciding the patch wasn't needed, breaking the java code. I've made it do a dry run instead, which seems to resolve that issue.

I want to do a bit more testing, but as far as I know it's working fine now.

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -524,24 +524,29 @@ def make_package(args):
...
+                          str(patch_path))
                 else:
                     raise e
+            else:
+                # The dry run worked, so do the real thing
+                subprocess.check_output(patch_command)
+
 
 
 def parse_args_and_make_package(args=None):
```
opacam, line 552: We have an extra blank line here that causes [flake8 to fail](https://github.com/kivy/python-for-android/pull/2113/checks?check_run_id=631276014#step:4:23):
```suggestion
                subprocess.check_output(patch_command)


def parse_args_and_make_package(args=None):
```

inclement:
Fixed tox, I think, by fixing recipe cache clearing.

inclement:
Fixed another bug, in test_recipe.py. This one only occurs if you run the tests from an unexpected directory, as the test tries to listdir a directory that doesn't exist because the exists call has been overridden. To keep things simple, I changed the test state so that this directory is not in the search path, but I might revisit this in another PR.

PR #2116, opened 2020-03-31
Update recommended NDK version to 19c

inclement:
Probably newer versions are also fine, but making this change in particular because the buildozer download url for 19b no longer exists. I think the 19c one should work fine though.


inclement:
Noticed while looking at https://stackoverflow.com/questions/60951864/need-help-creating-apk-with-buildozer/60952224?noredirect=1#comment107843733_60952224

PR #2117, opened 2020-04-01
Bumps openssl to 1.1.1f

AndreMiras:
I haven't tested it yet, as I understood the bump is required to fix runtime issues


inclement:
Could you actually raise this against master? I think that's the right way to do things, merge into master (it's a post-release fix, version number should be previous_version.post0), then merge into release-2020.03.30, then merge into develop.

AndreMiras:
OK :+1:  https://github.com/kivy/python-for-android/pull/2118

PR #2123, opened 2020-04-01
Merge master into develop

inclement:
The merge conflicts were pretty trivial, so I just resolved them in favour of the current develop state.

What do you think about merging this? Then presumably we could merge it into the release-2020.03.30 branch as well to resolve the conflicts there.

@AndreMiras

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ years. These **unified python recipes** require a **minimum target api level of 
 api level below 21, you should use an older version of python-for-android
 (<=0.7.1).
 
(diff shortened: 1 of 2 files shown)


inclement:
Looking at the log, it includes the merges of previous releases. I wonder if I've been doing that wrong, failing to merge them both to master and to develop in the correct way.

inclement:
I *suspect* the merging back to develop thing might be what I was doing wrong before. Let's pay careful attention when pushing the next release.

I think it might be necessary to re-merge into the release-2020.03.30 branch, but that should be easy to, I'll tackle it shortly.

PR #2126, opened 2020-04-02
Fixes code block directives

AndreMiras:
- s/codeblock/code-block/
- fixes code block being nested because of previous literal block `::`

See screenshot attached, to the left before the change, to the right after the change:
image


AndreMiras:
Yes indeed I didn't rebase that one. In theory I should, but I'll simply merge instead :stuck_out_tongue: 
Thanks for the review

PR #2132, opened 2020-04-04
Adding more assets

robertpfeiffer:
Add an `--add-asset` option to `p4a apk`.
`--add-asset assets/foo` will copy the local directory `./assets/foo` to `assets/assets/foo` in the apk. This can used multiple times, to add multiple files and directories. `--add-asset /home/user/path/to/image.png:images/image.png` will copy the file to `assets/images/image.png`.

Probably not as useful with kivy, but very useful with SDL2 and SDL2-based frameworks, as SDL2 RWops use paths relative to the `assets/` directory.


pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -304,6 +305,15 @@ def make_package(args):
                 tar_dirs.append(python_bundle_dir)
         if get_bootstrap_name() == "webview":
             tar_dirs.append('webview_includes')
+        if hasattr(args, "assets") and args.assets is not None:
```
inclement, line 308: I think argparse should make this simpler, isn't it an empty list by default or something so you can just do `for asset in args.assets:`?

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -239,9 +239,10 @@ def make_package(args):
     assets_dir = "src/main/assets"
 
     # Delete the old assets.
-    try_unlink(join(assets_dir, 'public.mp3'))
-    try_unlink(join(assets_dir, 'private.mp3'))
```
inclement, line 243: Looks like try_unlink isn't used any more, could you delete it?

pythonforandroid/toolchain.py
```diff
@@ -494,6 +494,10 @@ def add_parser(subparsers, *args, **kwargs):
         # However, it is also needed before the distribution is finally
         # assembled for locating the setup.py / other build systems, which
         # is why we also add it here:
+        parser_apk.add_argument(
+            '--add-asset', dest='assets',
+            action="append",
+            help=('Put this in the assets folder'))
```
inclement, line 500: Unnecessary () around the string

inclement, line 500: Could you document the : splitting thing in the help?

pythonforandroid/toolchain.py
```diff
@@ -582,6 +586,14 @@ def add_parser(subparsers, *args, **kwargs):
...
         if hasattr(args, "build_mode") and args.build_mode == "release":
             args.unknown_args += ["--release"]
+        if hasattr(args, "assets") and args.assets is not None:
+            # Pass this value on to the internal bootstrap build.py:
+            for asset in args.assets:
+                if ":" in asset:
+                    asset_src, asset_dest = asset.split(":")
+                else:
+                    asset_src = asset_dest = asset
+                args.unknown_args += ["--asset", os.path.abspath(asset_src)+":"+asset_dest]
```
inclement, line 596: Isn't this just unnecessary? The unknown_args include everything that wasn't parsed by the first p4a cmdparser, so if you just don't have a `--add-asset` in that argumentparser it will get passed through anyway. It seems you're pulling out the arguments, then adding them again (just with --asset instead of --add-asset)?

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -239,9 +239,10 @@ def make_package(args):
...
+    if os.path.exists(assets_dir):
+        shutil.rmtree(assets_dir)
     ensure_dir(assets_dir)
+    open(os.path.join(assets_dir, ".gitkeep"), 'a').close()
```
inclement, line 245: Is this needed? Even if you're adding the output to git, it seems like this stuff should be handled at that later stage (although it's acceptable if there's some good reason for it).

robertpfeiffer: I asked about this on Discord. As it stands now, it's fine if the assets directory is completely deleted and recreated each time, but the behaviour *used* to be that the assets directory was kept so I tried to be consistent with that.

pythonforandroid/toolchain.py
robertpfeiffer: AFAIK, the build.py is executed in a different working directory, so we need to convert the relative paths to absolute before all that. The unknown_args part is not the problem, `os.path.abapath` is.

pythonforandroid/bootstraps/common/build/build.py
robertpfeiffer: I thought I tried that and it didn't work. Maybe I misconfigured argparse.

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -239,9 +239,10 @@ def make_package(args):
     assets_dir = "src/main/assets"
 
     # Delete the old assets.
-    try_unlink(join(assets_dir, 'public.mp3'))
-    try_unlink(join(assets_dir, 'private.mp3'))
+    if os.path.exists(assets_dir):
+        shutil.rmtree(assets_dir)
```
AndreMiras, line 243: I'm allergic to `if` :smile: 
```suggestion
    shutil.rmtree(assets_dir, ignore_errors=True)
```

pythonforandroid/bootstraps/common/build/build.py
AndreMiras: This is when we should coverage test this part as well. Not that I'm demanding it now, but I wish we had that already covered

pythonforandroid/toolchain.py
```diff
@@ -494,6 +494,10 @@ def add_parser(subparsers, *args, **kwargs):
         # However, it is also needed before the distribution is finally
         # assembled for locating the setup.py / other build systems, which
         # is why we also add it here:
+        parser_apk.add_argument(
```
AndreMiras, line 497: Could you cover test your change in `tests/test_toolchain.py`?
There're already some examples in this file and as I understand what you're doing in `toolchain.py` unit testing the behaviour should be fairly easy

pythonforandroid/toolchain.py
inclement: Ah, got it, that's fine then, we can leave that code pattern for fixing later. Thanks for the clarification.

pythonforandroid/bootstraps/common/build/build.py
inclement: Got it. I think it should be fine to ditch it and let the directory be deleted if you want, but if leaving this in could you add a comment noting this reasoning?

PR #2138, opened 2020-04-11
Update common bootstrap code after switch to Py 3.8.

kollivier:


kollivier:
Sorry, have been really busy with updating our p4a app over the past couple weeks and am now just getting back to this, but I see someone else has applied a fix for this, so this PR is no longer needed!

PR #2145, opened 2020-04-20
Fix a problem with the way we were calling virtualenv

zurgeg:
This PR fixes a problem where any buildozer build errors with `Can not import virtualenv.__main__; virtualenv is not a package` as I tested and when I made this modification it worked.

### DIFF
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -767,7 +767,7 @@ def run_pymodules_install(ctx, modules, project_dir=None,
     venv = sh.Command('virtualenv')
     with current_directory(join(ctx.build_dir)):
         shprint(venv,
-                '-p={}'.format(
+                '-p=python{}'.format(
                     ctx.python_recipe.major_minor_version_string.
                     partition(".")[0]
                     ),
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -764,10 +764,10 @@ def run_pymodules_install(ctx, modules, project_dir=None,
             project_has_setup_py(project_dir) and not ignore_setup_py:
         info('Will process project install, if it fails then the '
              'project may not be compatible for Android install.')
-    venv = sh.Command('virtualenv')
+    venv = sh.Command('python3')
     with current_directory(join(ctx.build_dir)):
         shprint(venv,
-                '-p python{}'.format(
+                '-m venv'.format(
                     ctx.python_recipe.major_minor_version_string.
                     partition(".")[0]
                     ),
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -775,7 +775,7 @@ def run_pymodules_install(ctx, modules, project_dir=None,
                         ),
                     'venv'
                    )
-        except:
+        except ModuleNotFoundError:
             shprint(venv,
                     '-m virtualenv --python={}'.format(
                         ctx.python_recipe.major_minor_version_string.
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -765,7 +765,6 @@ def run_pymodules_install(ctx, modules, project_dir=None,
         info('Will process project install, if it fails then the '
              'project may not be compatible for Android install.')
     venv = sh.Command('python3')
-    
     with current_directory(join(ctx.build_dir)):
         try:
             shprint(venv,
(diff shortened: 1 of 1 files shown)


zurgeg:
Just testdrove the PRs changes on my local build machine, seems to work.

AndreMiras:
Thanks you @zurgeg for the initiative.
The CI is failing `/home/user/app/venv/bin/python3: No module named  virtualenv --python=3`.
We're probably just missing the system dependency. This is not yet clear to me why as the module is part of the `setup.py` which we `pip install -e` from the `Makefile`.
This needs to be investigated/fixed before we can merge

zurgeg:
It worked for me, dunno why it doesn't work on Travis CI.

zurgeg:
I think it is caused by there being 2 different versions, `venv` in >=Python 3.6 and `virtualenv` in  <Python3.6 (I think)

zurgeg:
Finally just got those builds to work (I think). It had to do with a single whitespace on a line.

zurgeg:
For some reason the `try: except:` block is not trying...

zurgeg:
@AndreMiras turns out my first build was lucky... I just experienced your same issue, changed python3 to virtualenv (Should work anywhere), and it started working.

opacam:
@zurgeg ,
> For some reason the `try: except:` block is not trying...

Try changing:

`ModuleNotFoundError` by `sh.ErrorReturnCode_1`

AndreMiras:
Thanks for investing time into this as I haven't had time lately.
Why can't we make venv being a p4a dep using setup.py then rather than hoping for it to be available?

opacam:
mmm...I think that `venv` is [already integrated into python standard modules](https://docs.python.org/3/library/venv.html) (since v3.3)... so is  not necessary to touch `setup.py` right?

zurgeg:
Oh yeah, I'll do that, also doing what I think works for all Python versions.

zurgeg:
Finally got those builds to work! So apparently no matter if you use ```venv``` or ```virtualenv``` you can still use the ```virtualenv``` command.

AndreMiras:
But wait now you removed the `--python {version}` parameter and only using plain `virtualenv` it means system will pick the default Python version which can be 2.7 (e.g. Ubuntu 18.04) right? Also I can't think of how this is solving the original problem on buildozer you mentioned :thinking:

zurgeg:
Oh... Hold on, I though it was --python=..., so I thought that was problematic, hold on...

AndreMiras:
Also you know you can try things locally e.g. using Docker so you don't have to wait for the CI which is probably slower than your local build

zurgeg:
Just fixed your issue @AndreMiras!

pythonforandroid/build.py
```diff
@@ -764,17 +764,15 @@ def run_pymodules_install(ctx, modules, project_dir=None,
...
-    venv = sh.Command(ctx.virtualenv)
+    venv = sh.Command('virtualenv')
     with current_directory(join(ctx.build_dir)):
         shprint(venv,
-                '--python=python{}'.format(
-                    ctx.python_recipe.major_minor_version_string.
-                    partition(".")[0]
-                    ),
-                'venv'
+            '-p python{}'.format(
```
AndreMiras, line 770: If I understand correctly this didn't change, but we only changed `venv = sh.Command('virtualenv')` then maybe don't introduce a change here if we don't end up fixing that line. I mean rollback the full arg `--python=` and the indentation so it's clear what the actual fix is

zurgeg: Okay, doing that now...

zurgeg: For some reason it gives ```path =python3 not found from --python=python3```.

zurgeg:
Also @opacam I don't think using ```venv``` is a good idea because otherwise we would be dropping support for under 3.3, even though 3.3<= isn't used much, it isn't a good idea to just cut it.

zurgeg:
Okay, going way back to my ```try... except block``` idea... I think I got it to work thanks to @opacam it should provide optimal capability (especially because Python 2 still works) with older versions!

AndreMiras:
This is properly painful right? Honestly we don't mind much about Python 3.3, it's not like it wasn't already a nightmare to support 3.6+ and multiple platforms.
opacam made a not bad attempt to ditch virtualenv in his on-going PR and this should also (hopefully) simplify code/dependencies https://github.com/kivy/python-for-android/pull/2152

zurgeg:
Yeah, I think the other PR is better, closing this one when it’s merged.

AndreMiras:
Thanks for bring this up to our attention and helping out debugging this complex one :hearts:

inclement:
Looks like this was closed by merging #2152, but thanks @zurgeg for the PR.

PR #2147, opened 2020-04-20
Twisted recipe

pavelsof:
Update of the recipe for Twisted so that it removes the unit tests in the `prebuild_arch` hook. This should decrease the size from 11 down to 4 MB for the uncompressed APK directory.

### DIFF
--- a/pythonforandroid/recipes/twisted/__init__.py
+++ b/pythonforandroid/recipes/twisted/__init__.py
@@ -1,3 +1,6 @@
+import os
+import shutil
+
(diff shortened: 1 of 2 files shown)


AndreMiras:
Thanks for the fix. I've edited the commit message for slightly more explicit.
Out of curiosity, is there a bug report upstream and what's the stance of them regarding it?

pavelsof:
Thank you!

I looked around a bit, but I cannot find a ticket about this in Twisted's tracker or any info about whether they actually have an opinion or this is just how their directory structure happened to be :/

PR #2151, opened 2020-04-24
Update setup.py: add classifiers & python_requires

opacam:


setup.py
```diff
@@ -88,6 +88,7 @@ def recursively_include(results, directory, patterns):
       description='Android APK packager for Python scripts and apps',
       long_description=long_description,
       long_description_content_type='text/markdown',
+      python_requires="==3.*,>=3.6.0",
```
AndreMiras, line 91: Nice! Just curious isn't this redundant? Don't we only need `python_requires="==3.*,>=3.6.0"`.
If you added it in case of Python 4, I would say keep it simple

opacam:
Mmmm...interesting...I did some research...


I think that pip search doesn't care about python version...for instance...I maintain a Python3 only package, never compatible with Python2. This project is named `Cohen3` and [here](https://pypi.org/project/Cohen3/0.8.0/) you have the first version I published at pipy. You will se that is classified as Python3, but here you have the pip2 search:
```
opacam@DEPOBOX:~$ pip2 search Cohen3
DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
Cohen3 (0.9.2)  - Cohen3 - DLNA/UPnP Media Server
```
 
And if you try to install a python3 only package, using the branch of this PR:
```
opacam@DEPOBOX:~$ pip2 install https://github.com/opacam/python-for-android/archive/feature-update-classifiers.zip --user
DEPRECATION: Python 2.7 reached the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 is no longer maintained. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
Collecting https://github.com/opacam/python-for-android/archive/feature-update-classifiers.zip
  Downloading https://github.com/opacam/python-for-android/archive/feature-update-classifiers.zip
     / 2.0 MB 167 kB/s
ERROR: Package 'python-for-android' requires a different Python: 2.7.18 not in '==3.*,>=3.6.0'
```

AndreMiras:
I think pip do care when searching and your project didn't have the Python classifier on the setup so I think that's why it appeared https://github.com/opacam/Cohen3/blob/0.8.0/setup.py
Either way I don't really mind as it was more a heads up and I think Django really does it for this reason

PR #2152, opened 2020-04-24
 Use python3's venv instead of virtualenv

opacam:
Here:
  - we remove one p4a dependency `virtualenv`
  - move from `pytoml` to `toml`, because `pytoml` is [deprecated](https://pypi.org/project/pytoml/)
  - remove unneeded code related with virtualenv

**Notes:**
  - close #2145

### DIFF
--- a/tests/test_pythonpackage_basic.py
+++ b/tests/test_pythonpackage_basic.py
@@ -6,7 +6,6 @@ while the other additional ones aren't (for build time reasons).
 """
 
 import os
-import pytest
 import shutil
 import sys
 import subprocess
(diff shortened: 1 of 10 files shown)


AndreMiras:
We probably also want to ditch `virtualenv<20` from `install_reqs` on setup.py.
But I think this looks good enough for a merge

AndreMiras:
I'm not so sure yet why the macOS build is failing on Travis only with this error.
https://travis-ci.org/github/kivy/python-for-android/jobs/679209347
```
/Applications/Xcode-11.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ranlib: file: libpython3.8.a(pymath.o) has no symbols
gcc     -Wl,-stack_size,1000000  -framework CoreFoundation -o python.exe Programs/python.o libpython3.8.a -ldl   -framework CoreFoundation    
gcc     -Wl,-stack_size,1000000  -framework CoreFoundation -o Programs/_testembed Programs/_testembed.o libpython3.8.a -ldl   -framework CoreFoundation    
./python.exe -E -S -m sysconfig --generate-posix-vars ;\
	if test $? -ne 0 ; then \
		echo "generate-posix-vars failed" ; \
		rm -f ./pybuilddir.txt ; \
		exit 1 ; \
	fi
 CC='gcc' LDSHARED='gcc -bundle -undefined dynamic_lookup    ' OPT='-DNDEBUG -g -fwrapv -O3 -Wall' 	_TCLTK_INCLUDES='' _TCLTK_LIBS='' 	./python.exe -E /Users/travis/.python-for-android/build/other_builds/hostpython3/desktop/hostpython3/setup.py  build
running build
running build_ext
error: [Errno 2] No such file or directory: '/usr/local/include/python3.8/pyconfig.h'
make[1]: *** [sharedmods] Error 1
  STDERR:
make: *** [testapps/armeabi-v7a] Error 1
The command "make testapps/armeabi-v7a PYTHON_WITH_VERSION=python3" exited with 2.
```

opacam:
> We probably also want to ditch virtualenv<20 from install_reqs on setup.py...

This is [already done](https://github.com/opacam/python-for-android/blob/feature-update-deps/setup.py#L23-L27) at commit fdd66d3 :wink:

> I'm not so sure yet why the macOS build is failing on Travis only with this error...

Ok, I've been looking into a solution to this problem with macOs and it's clearly related with virtualenv/venv. When trying to run p4a inside  a virtualenv/venv the hostpython's build scripts gets confused on about to locating the `pyconfig.h` file. So **the solution is to not use virtualenv or venv when trying to build from a macOs**...:thinking:...of course is a partial solution (it's more like a shortcut than a fix), but will solve the problem for now...probably this should be addressed at CPython.

**Note:**  `kivy-ios` seems to has the same problem, kivy/kivy-ios#401

tests/test_pythonpackage_basic.py
```diff
@@ -340,6 +302,10 @@ def test_venv(self):
                 os.path.join(test_dir, "venv", "bin", "pip"),
                 "install", "-U", "pep517<0.7.0"
             ])
+            subprocess.check_output([
+                os.path.join(test_dir, "venv", "bin", "pip"),
+                "install", "-U", "toml"
+            ])
```
AndreMiras, line 308: This is a bit out of scope of this PR, but I'm thinking why don't we `pip install -e` then?

opacam: Well I suppose that you are speaking of `pip install -e .` right?

In this case...why should we? we don't need all p4a dependencies for this test, and the venv will be destroyed after is passed. Also, considering that we are speaking of one of the slowest tests of p4a, IMHO, better to try to save some time :wink:

AndreMiras: makes sense :+1:

PR #2153, opened 2020-04-25
Show loading screen while unpacking for webview bootstrap

kollivier:
This is a port of the SDL bootstrap's UnpackTask fix to the webview bootstrap, so that the loading screen can be displayed during unpack instead of a plain black window.


kollivier:
Note that although it shows a lot of lines changed, a majority of the changes are just reformatting caused by moving the actual UI and python loading code into `UnpackTask.onPostExecute`. (And needing to access `PythonActivity` via `PythonActivity.mActivity` since `this` became `UnpackTask` when the code got moved.

PR #2155, opened 2020-04-25
Compile version SDK should match target SDK so webview can support Android versions 5-10.

kollivier:
How to reproduce this problem on current develop:

1. Compile with `usesClearTextTraffic` property in `AndroidManifest.xml`, and find it requires a min SDK of 26 to compile since it was added in that SDK. 
2. Run the built apk on a device with Android 5-7 and get missing symbol errors. 

We've most commonly seen ones about `DT_HASH`, but there are several that could appear, based on NDK backwards-incompatible changes listed here:

https://android.googlesource.com/platform/bionic/+/master/android-changes-for-ndk-developers.md

This PR changes the `compileSDKVersion` to match the `targetSDKVersion` in order to allow us to use features like `usesClearTextTraffic` while still maintaining compatible down to the min SDK version (21).

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -422,17 +422,20 @@ main.py that loads it.''')
         )
 
     # Find the SDK directory and target API
-    with open('project.properties', 'r') as fileh:
-        target = fileh.read().strip()
-    android_api = target.split('-')[1]
-    try:
-        int(android_api)
-    except (ValueError, TypeError):
-        raise ValueError(
-            "failed to extract the Android API level from " +
-            "build.properties. expected int, got: '" +
-            str(android_api) + "'"
-        )
+    if 'ANDROIDAPI' in os.environ:
+        android_api = os.environ['ANDROID_API']
+    else:
+        with open('project.properties', 'r') as fileh:
+            target = fileh.read().strip()
+        android_api = target.split('-')[1]
+        try:
+            int(android_api)
+        except (ValueError, TypeError):
+            raise ValueError(
+                "failed to extract the Android API level from " +
+                "build.properties. expected int, got: '" +
+                str(android_api) + "'"
+            )
     with open('local.properties', 'r') as fileh:
         sdk_dir = fileh.read().strip()
     sdk_dir = sdk_dir[8:]
--- a/pythonforandroid/bootstrap.py
+++ b/pythonforandroid/bootstrap.py
@@ -163,7 +163,7 @@ class Bootstrap:
 
         with current_directory(self.build_dir):
             with open('project.properties', 'w') as fileh:
-                fileh.write('target=android-{}'.format(self.ctx.android_api))
+                fileh.write('target=android-{}'.format(self.ctx.ndk_api))
 
     def prepare_dist_dir(self):
         ensure_dir(self.dist_dir)


pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -422,17 +422,20 @@ def make_package(args):
...
-    except (ValueError, TypeError):
-        raise ValueError(
-            "failed to extract the Android API level from " +
-            "build.properties. expected int, got: '" +
-            str(android_api) + "'"
-        )
+    if 'ANDROIDAPI' in os.environ:
+        android_api = os.environ['ANDROID_API']
+    else:
+        with open('project.properties', 'r') as fileh:
```
kollivier, line 428: I kept this code in place but I'm not sure if there will ever be a case where `ANDROIDAPI` is not set during a build. Also, are `project.properties` included or used anywhere in gradle builds?

inclement:
I'm a little confused with this one.

> C/C++ compilation should use the NDK/min SDK version since it doesn't…
> … weak link, not target SDK.

Not sure what this title + continuation is trying to say exactly?

> This PR changes the -target argument for C/C++ compilation to use tne ndk_api value rather than the android_api value.

It doesn't look like this affects any of the C/C++ compilation because build.py is run only after everything is compiled. It does affect the compileSdkVersion in the build.gradle, which looks like probably a correct change, and it also makes the targetSdkVersion equal to that old api number. However, it's highly deliberate that the compile version and targetSdkVersion don't currently match, because as far as I was aware that's a reasonable thing to do. It is quite important because we specifically compile against an old api (21), so that we're compatible with anything after that, but the app should still use a recent targetSdkVersion so that it's allowed on the Play store etc.

Are you saying that this mismatch is not allowed? Has something changed in Android 10 in particular to cause problems? Or have I misremembered the names of the different variables.

kollivier:
Sorry, yeah, the way I wrote it was confusing, and was based off my memory. There was a point where I thought the C/C++ compilation was the problem, but in fact it is the min == compile SDK requirement that led to our problem. I'll update the title of the ticket to be less confusing, and have provided some explanation below to make it clearer what this change is about.

The specific issue we ran into was when we added the `usesClearTextTraffic` flag to the manifest, as is done here:

https://github.com/kivy/python-for-android/commit/5d329bf49ae37c988dd4caa5f1192cd7cf6146e9

Using this is a hard requirement for allowing HTTP traffic on Android 8.0+. For this to compile, we needed to specify a compile SDK that knows about the `usesClearTextTraffic` property, which meant using compile SDKs for 8.0 or above. However, currently, that means we need to set our min SDK to the 8.0 version as well, even though we're trying to support versions down to 5.0 (and maybe 4.4 if we can figure that out!). Essentially, in the current code, we need to choose between only supporting Android versions 5-7, or only supporting versions >= 8.

With `minSDKVersion=21` and `compileSDKVersion=29`, though, we can use any feature in that series of releases. If it's a manifest change (e.g. `usesClearTextTraffic`), it will get ignored on old versions but used on newer ones. If it's a Java API, we just need to be careful to do a (pseudocode) `if (android_version >= 8.0) { /*... 8.0+ specific code here ... */ }` to use the API. 

There's a very long and detailed article about this here:

https://medium.com/androiddevelopers/picking-your-compilesdkversion-minsdkversion-targetsdkversion-a098a0341ebd

But the TL;DR is that your `compileSDKVersion` should match your `targetSDKVersion` so that you can incorporate anything required by versions > min (like `usesClearTextTraffic`), and let the min SDK determine the lowest version you can support.

How I came up with that confusing title: The C/C++ compilation is and was actually doing the right thing, but initially I was confused about that because upping the SDK version to support `usesClearTextTraffic` resulted in runtime crashes about missing symbols on devices < 8.0. I eventually sorted out the real problem but apparently had this stuck on the brain. ;-)

inclement:
Thanks for the explanation. I understand the general goal, and we definitely want to get it right, but as far as I can see we already set the compileSdkVersion to match the targetSdkVersion and your PR doesn't affect that. Instead, your PR gets the android_api from the $ANDROID_API environment variable, and makes it so that if it's ever read from project.properties (although as you note this may never happen) it will actually retrieve the wrong value and target an old api.

I'm not sure if we're currently updating project.properties after the first write, but for projects built with correct api numbers from the start it doesn't seem like this changes anything.

Am I missing something?

kollivier:
No, sorry, you're right, this is totally my bad. It's 'fixing' a problem that I created in the first place by changing what we output to `project.properties`. Looking at the state of current develop, it seems right to me, but that leaves me wondering how we ran into the issue in the first place. 

I'm going to close out this PR, and check through my code changes again to see if there's some other change that I am missing. If I find something, I'll post it, otherwise sorry for wasting your time with this!

inclement:
Thanks for checking! It's totally possible (maybe likely even) that you hit some buggy code path that led to bad results, so do feel free to open another PR if you find it. p4a used to not handle this right at all, and I fixed it a while back to set the variables the current way, but something could definitely have been missed.

PR #2156, opened 2020-04-25
Get --add-source working for dirs in Gradle builds

kollivier:
I had trouble making the file patterns work using `include` (no errors, but files didn't get included). However, this approach works for specifying source directories, which was our use case. Figured I'd post what we have now and see if anyone knows how to fix the patterns case, or knows if that case is used often?

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -381,8 +381,8 @@ main.py that loads it.''')
             if ':' in spec:
                 specdir, specincludes = spec.split(':')
                 print('WARNING: Currently gradle builds only support including source '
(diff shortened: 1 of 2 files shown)


inclement:
I've previously tried this and had the same problems with getting patterns working, although I never tried the case without them so it's great to see that works.

Would you mind adding a warning during the templating, notifying the user that their pattern will be ignored (if they passed one)? But other than that, I'm very happy to merge.

kollivier:
Good call, thanks! Just added a warning, would this work?

inclement:
@kollivier Looks good, fixed a couple of typos, will merge when the tests pass.

inclement:
Thanks again!

PR #2159, opened 2020-04-28
Fixes hostpython build with macOS venv

AndreMiras:
Running the freshly compiled host python binary from another Python
process (e.g. sh module or os.system()) may get wrongly detected as
if it was ran directly from the venv.
This happens when `pyvenv.cfg` is present (e.g. venv/pyvenv.cfg).
It makes `sysconfig.is_python_build()` returns `False` instead of `True`
since it's using the compiled interpreter (e.g. ./build-dir/python).
https://docs.python.org/3/library/sysconfig.html#sysconfig.is_python_build
https://github.com/python/cpython/blob/v3.8.2/Lib/sysconfig.py#L127
This is because the `sys._home` attribute is used during the detection.
The issue was first seen in macOS venv which generates the `pyvenv.cfg`.
To compare both behaviours, try the following within and without a venv:
```python
import os
os.system("./build-dir/python -E -c 'import sysconfig; print(sysconfig._sys_home)'")
```
One would return `/usr/local/bin` and the other `None`

Refs:
- https://github.com/kivy/kivy-ios/issues/401
- https://github.com/kivy/python-for-android/pull/2063

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,14 +47,6 @@ testapps/%: virtualenv
     python setup.py apk --sdk-dir $(ANDROID_SDK_HOME) --ndk-dir $(ANDROID_NDK_HOME) \
     --arch=$($@_APP_ARCH)
 
-testapps-no-venv/%:
-	pip3 install Cython==0.28.6
-	pip3 install -e .
-	$(eval $@_APP_ARCH := $(shell basename $*))
-	cd testapps/on_device_unit_tests/ && \
-    python3 setup.py apk --sdk-dir $(ANDROID_SDK_HOME) --ndk-dir $(ANDROID_NDK_HOME) \
-    --arch=$($@_APP_ARCH)
-
 clean:
 	find . -type d -name "__pycache__" -exec rm -r {} +
 	find . -type d -name "*.egg-info" -exec rm -r {} +
--- a/pythonforandroid/recipes/hostpython3/__init__.py
+++ b/pythonforandroid/recipes/hostpython3/__init__.py
@@ -4,6 +4,7 @@ from multiprocessing import cpu_count
 from os.path import exists, join, isfile
 
 from pythonforandroid.logger import shprint
+from pythonforandroid.patching import is_version_lt
 from pythonforandroid.recipe import Recipe
 from pythonforandroid.util import (
     BuildInterruptingException,
@@ -34,6 +35,10 @@ class Hostpython3Recipe(Recipe):
     url = 'https://www.python.org/ftp/python/{version}/Python-{version}.tgz'
     '''The default url to download our host python recipe. This url will
     change depending on the python version set in attribute :attr:`version`.'''
+
+    patches = (
+        ('patches/pyconfig_detection.patch', is_version_lt("3.8.3")),
+    )
 
     @property
     def _exe_name(self):
--- a/pythonforandroid/recipes/python3/__init__.py
+++ b/pythonforandroid/recipes/python3/__init__.py
@@ -8,7 +8,7 @@ from os.path import dirname, exists, join, isfile
 from shutil import copy2
 
 from pythonforandroid.logger import info, warning, shprint
-from pythonforandroid.patching import version_starts_with
+from pythonforandroid.patching import version_starts_with, is_version_lt
 from pythonforandroid.recipe import Recipe, TargetPythonRecipe
 from pythonforandroid.util import (
     current_directory,
@@ -55,6 +55,8 @@ class Python3Recipe(TargetPythonRecipe):
     name = 'python3'
 
     patches = [
+        ('patches/pyconfig_detection.patch', is_version_lt("3.8.3")),
+
         # Python 3.7.1
         ('patches/py3.7.1_fix-ctypes-util-find-library.patch', version_starts_with("3.7")),
         ('patches/py3.7.1_fix-zlib-version.patch', version_starts_with("3.7")),
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,7 @@ jobs:
         # installs java 1.8, android's SDK/NDK and p4a
         - make -f ci/makefiles/osx.mk
         - export JAVA_HOME=/Library/Java/JavaVirtualMachines/adoptopenjdk-8.jdk/Contents/Home
-      script: make testapps-no-venv/armeabi-v7a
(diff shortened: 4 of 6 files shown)


pythonforandroid/build.py
```diff
@@ -764,7 +764,7 @@ def run_pymodules_install(ctx, modules, project_dir=None,
         shprint(host_python, '-m', 'venv', 'venv')
 
         # Prepare base environment and upgrade pip:
-        base_env = copy.copy(os.environ)
+        base_env = dict(copy.copy(os.environ))
```
AndreMiras, line 767: This has nothing to do with the macOS venv fix, but it suddenly started failing (develop branch included). So I included the fix in the PR but dedicated commit

Makefile
```diff
@@ -47,14 +47,6 @@ testapps/%: virtualenv
...
 
-testapps-no-venv/%:
-	pip3 install Cython==0.28.6
-	pip3 install -e .
-	$(eval $@_APP_ARCH := $(shell basename $*))
-	cd testapps/on_device_unit_tests/ && \
-    python3 setup.py apk --sdk-dir $(ANDROID_SDK_HOME) --ndk-dir $(ANDROID_NDK_HOME) \
-    --arch=$($@_APP_ARCH)
-
 clean:
```
opacam, line 50: :fire: Glad that we can remove this!

pythonforandroid/build.py
opacam: :+1: but you created a separate PR right #2160? nevermind, it's fine to me!

pythonforandroid/recipes/hostpython3/__init__.py
```diff
@@ -35,6 +35,8 @@ class Hostpython3Recipe(Recipe):
     '''The default url to download our host python recipe. This url will
     change depending on the python version set in attribute :attr:`version`.'''
 
+    patches = ('patches/pyconfig_detection.patch',)
```
opacam, line 38: is the new patch working/needed for Python 3.7?

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -55,6 +55,8 @@ class Python3Recipe(TargetPythonRecipe):
     name = 'python3'
 
     patches = [
+        ('patches/pyconfig_detection.patch'),
+
```
opacam, line 59: Do we need to patch python3 recipe as well? I understand that the problem is with hostpython and MacOs during build stage...:thinking:

AndreMiras: Yes I also thought, but then I saw the build failing with the exact same issue so I decided to add it there as well

pythonforandroid/recipes/hostpython3/__init__.py
AndreMiras: Good question, I've only verified for 3.8 to be honest. I can apply it to 3.8 only to be safe

rubendibattista:
This has been fixed also upstream, apparently: https://bugs.python.org/issue40261

[I went on trying to understand what was happening]

AndreMiras:
Very nice to see @rubendibattista !
So I could have saved a day of my life by just waiting 3.8.3 to be released :joy: 
Then it also means this patch needs to be applied to 3.8.2 only for now. Will update

opacam:
Nice, so in Python 3.8.3 this should be solved :smile:

**¡¡¡thanks for the info!!!**

rubendibattista:
@AndreMiras at least for you was just one day! I spent one week after giving up... I reached that situation of crazyness where I was patching `CPython` directly. :)

rubendibattista:
@opacam and 3.7.8

rubendibattista:
However this is the change they're applying upstream: https://github.com/python/cpython/commit/5765acaf64cc2c52ce8a35de9407faddf6885598

Care to see if it solves the problem also here?

opacam:
Maybe it will solve the problem, but I don't have a Mac to test...anyway...if you want to try, [the following patch](https://gist.github.com/opacam/49e9395ef036560f30d53f7666e1f2f8) will modify hostpython/python recipes to use the branch you mentioned from CPython :wink:

AndreMiras:
Opacam the problem with this approach is it hardcodes the version and people won't be able to use different versions. Plus I'd rather not rely on something not yet release like so.
@rubendibattista I haven't tested that patch you shared. Good thing it's an upstream one so it should be properly tested. Basically I would apply only `Modules/main.c` change.
I'll rebase this PR and address some of the comments and then create another PR where I apply `Modules/main.c` patch only and let's see if that compiles

pythonforandroid/recipes/hostpython3/__init__.py
```diff
@@ -35,6 +36,10 @@ class Hostpython3Recipe(Recipe):
     '''The default url to download our host python recipe. This url will
     change depending on the python version set in attribute :attr:`version`.'''
 
+    patches = (
+        ('patches/pyconfig_detection.patch', is_version_lt("3.8.3")),
```
AndreMiras, line 40: Now applying to anything lower than 3.8.3. Since this issue was also impacting some recent version of the 3.7 branch I thought it would be easier to go that way than being ultra ranging inside MINOR and PATCH with a complex expression (given MAJOR.MINOR.PATCH)

rubendibattista:
@AndreMiras the patch on  `pythonw.c` is not needed? 

In any case I could test this. I'll do it ASAP

AndreMiras:
OK the patch got rejected, it doesn't seem to apply properly I haven't investigated much, but i think my solution is good enough until 3.8.3 is released.
Let me know what you guys think

AndreMiras:
@rubendibattista which patch on `pythonw.c` you mean the code comment? Because the only upstream fix I see is `Modules/main.c` the rest is unit testing and code comments, right?
Yes please give it a try and let me know

rubendibattista:
@AndreMiras Ok, let's go with your initial PR. I'll try to see if I can adapt the upstream patch to current versions of Python.

AndreMiras:
I found out why the upstream patch didn't apply. It was for the 3.7 branch :sweat_smile: 
The 3.8 one is https://github.com/python/cpython/pull/19110
As for now the fix seems to work, but I will give the upstream patch a try on kivy-ios.
Thanks again for the heads up

rubendibattista:
🔝

PR #2164, opened 2020-05-01
Adds pygame recipe

AndreMiras:
Edit: this authored by @robertpfeiffer I've just copy/pasted his description in the commit message, rebased and made few edits to give an example of what the PR could look like.

As discussed on the Discord, this is a recipe to build apps based on
SDL2-based pygame.

It currently references https://github.com/pygame/pygame/tree/android,
a branch that will continue to live until we have an official stable
release of pygame based on SDL2 with android support.

Simple examples have been tested by other pygame users (it doesn't just
build on my own machine) but some pygame functionality is still untested,
and some dependencies like freetype, postmidi and libjpeg are currently
not part of the build. It's usable, but not complete.

### DIFF
--- a/pythonforandroid/recipes/pygame/__init__.py
+++ b/pythonforandroid/recipes/pygame/__init__.py
@@ -1,0 +1,64 @@
+from os.path import join
+
+from pythonforandroid.recipe import CompiledComponentsPythonRecipe
+from pythonforandroid.toolchain import current_directory
+
+
+class Pygame2Recipe(CompiledComponentsPythonRecipe):
+    """
+    Recipe to build apps based on SDL2-based pygame.
+
+    .. warning:: Some pygame functionality is still untested, and some
+        dependencies like freetype, postmidi and libjpeg are currently
+        not part of the build. It's usable, but not complete.
+    """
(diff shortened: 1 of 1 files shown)


pythonforandroid/recipes/pygame/__init__.py
```diff
@@ -0,0 +1,56 @@
+from pythonforandroid.recipe import CompiledComponentsPythonRecipe
+from os.path import join
+from pythonforandroid.toolchain import current_directory
```
opacam, line 3: **nitpick:** Maybe we can sort the imports

AndreMiras: Addressed, I wish we were linting using `isort` across the code base

AndreMiras:
Addressed the docstring and iport sorting. I've also updated the commit message as we no longer reference a moving branch but a release candidate

robertpfeiffer:
Looks good to me, I think. I'll make another PR after the next pygame dev release.

AndreMiras:
Thanks for the feedback and for the effort, merging :rocket:

PR #2177, opened 2020-05-07
Updates release documentation

AndreMiras:
- Links both Travis and GitHub CI build systems
- Updates the PyPI release process (handled by the CI)

### DIFF
--- a/doc/source/contribute.rst
+++ b/doc/source/contribute.rst
@@ -55,8 +55,8 @@ At this point, wait for reviewer approval and conclude any discussion that arise
 - Merge the release branch to the ``master`` branch.
 - Also merge the release branch to the ``develop`` branch.
 - Tag the release commit in ``master``, with tag ``vYYYY.MM.DD``. Include a short summary of the changes.
-- Create the release distributions: ``python3 setup.py sdist bdist_wheel``
-- Upload the release to pypi: ``python3 -m twine upload``.
+- Release distributions and PyPI upload should be `handled by the CI
+  <https://github.com/kivy/python-for-android/blob/v2020.04.29/.travis.yml#L60-L70>`_.
 - Add to the Github release page (see e.g. `this example <https://github.com/kivy/python-for-android/releases/tag/v2019.06.06>`__):
   - The python-for-android README summary
   - A short list of major changes in this release, if any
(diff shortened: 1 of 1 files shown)


AndreMiras:
Also one point I'd like to discuss.
In the checklist we used the markdown format, but the documentation is restructured text. So I'm not sure what we want here. Do we want it as a code block so it can be copy pasted in each GitHub release PR or do we want it to be properly rendered in rst format?
https://github.com/kivy/python-for-android/blob/2b33eae/doc/source/contribute.rst#release-checklist

doc/source/contribute.rst
```diff
@@ -68,11 +68,13 @@ At this point, wait for reviewer approval and conclude any discussion that arise
...
+  - [ ] [Travis build](https://travis-ci.org/kivy/python-for-android)
+  - [ ] [GitHub Action](https://github.com/kivy/python-for-android/actions)
 - [ ] Run the tests locally via `tox`: this performs some long-running tests that are skipped on Travis.
 - [ ] Build and run the [on_device_unit_tests](https://github.com/kivy/python-for-android/tree/master/testapps/on_device_unit_tests) app using buildozer. Check that they all pass.
 - [ ] Build and run the following [testapps](https://github.com/kivy/python-for-android/tree/master/testapps) for arch `armeabi-v7a` and `arm64-v8a`:
-  - `python3 setup_testapp_python3_sqlite_openssl.py apk`
+  - [ ] `python3 setup_testapp_python3_sqlite_openssl.py apk`
     - [ ] `armeabi-v7a`
     - [ ] `arm64-v8a`
 - [ ] Check that the version number is correct
```
opacam, line 80: I would modify these lines, since, IMHO, the `on_device_unit_tests` app has the same functionality than `testapp_python3_sqlite_openssl` plus a little more (we can check device orientation,if service is working, kivy's keyboard optionts...). So, I would suggest to test `on_device_unit_tests`, because we also could avoid to do the local build (since we have it done in an isolated system with github actions). Still, I would put the command needed to perform a local build for a given arch, just in case that github actions fails or who knows...

Also I would suggest to put the command relative to the project (not inside `testapps`) since I think that, if you take an programming IDE like pycharm, the default terminal path is at project level, so it will make easier to copy/paste the command and run it from inside an IDE without entering the `testapps` direcctory. Also I think that is more explicit.

```
- [ ] Build (or download from github actions) and run the following [testapps](https://github.com/kivy/python-for-android/tree/master/testapps/on_device_unit_tests) for arch `armeabi-v7a` and `arm64-v8a`:
  - [ ] on_device_unit_tests
    - [ ] `armeabi-v7a` (cmd: `python3 testapps/on_device_unit_tests/setup.py  --arch=armeabi-v7a --debug`)
    - [ ] `arm64-v8a` (cmd: `python3 testapps/on_device_unit_tests/setup.py  --arch=arm64-v8a --debug`)
```

Anyhow,  I would suggest to have @inclement approval for the suggested changes I made (he has far better vision than mine and this is very sensitive stuff...), for your changes, no problem at all,so you have my approval.

**Thanks** :smile:

AndreMiras:
Thanks @opacam I've addressed your comments.
I also decided to fix the rendering to really be reStructuredText so it renders OK in our docs. See how it renders [on my fork](https://github.com/AndreMiras/python-for-android/blob/feature/release_documentation_update/doc/source/contribute.rst#release-checklist) vs how it renders [currently](https://github.com/kivy/python-for-android/blob/2b33eae/doc/source/contribute.rst#release-checklist)

doc/source/contribute.rst
```diff
@@ -68,13 +68,24 @@ At this point, wait for reviewer approval and conclude any discussion that arise
...
+
+- [ ] Build and run the `on_device_unit_tests <https://github.com/kivy/python-for-android/tree/master/testapps/on_device_unit_tests>`_ app using buildozer. Check that they all pass.
+
+- [ ] Build (or download from github actions) and run the following `testapps <https://github.com/kivy/python-for-android/tree/master/testapps/on_device_unit_tests>`_ for arch `armeabi-v7a` and `arm64-v8a`:
+
+  - [ ] on_device_unit_tests
+
+    - [ ] `armeabi-v7a` (cmd: `python3 testapps/on_device_unit_tests/setup.py  --arch=armeabi-v7a --debug`)
+
+    - [ ] `arm64-v8a` (cmd: `python3 testapps/on_device_unit_tests/setup.py  --arch=arm64-v8a --debug`)
```
opacam, line 87: Mmmm..my mistake in here, I realized that I gived to you a wrong command, **sorry** :grimacing:, this is more appropiated:


```suggestion
    - [ ] `armeabi-v7a`::

          cd testapps/on_device_unit_tests && PYTHONPATH=. python3 setup.py apk  --ndk-dir=<your-ndk-dir> --sdk-dir=<your-sdk-dir> --arch=armeabi-v7a --debug

    - [ ] `arm64-v8a`::

          cd testapps/on_device_unit_tests && PYTHONPATH=. python3 setup.py apk  --ndk-dir=<your-ndk-dir> --sdk-dir=<your-sdk-dir> --arch=arm64-v8a --debug
```

AndreMiras: I actually had some doubt as I thought I tried in the past with `cd`-ing and we couldn't.
Anyway, now I'm think we should maybe keep this simple and not give the command example at all

inclement: I like including the command example because the intention of this doc is to make preparing a release as absolutely thoughtless as possible. Perhaps it would be nice to move it to a makefile command that builds the release targets and prints some clear information about where they were output (but of course don't feel the need to do that here).

opacam: Ei guys I fixed [the command](https://github.com/kivy/python-for-android/pull/2177#discussion_r422480998): PYTHONPATH should include p4a sources and testapp sources:

```PYTHONPATH=.:../../```

That is why didn't work for you @AndreMiras

AndreMiras:
Thanks guys for the feedback.
Indeed @inclement I thought maybe that would be it regarding the markdown, so let's code format it instead.
Good idea for the issue template. Things is most of the issues we create are not for releasing and users creating bugs would get even more confused with the process :smile: 
Anyway, I'll put back the examples as suggested by @opacam if you both like it and also format it to code block so it can be copy pasted

inclement:
@AndreMiras I haven't looked at this much before, but it looks [here](https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository) like the template would just be a dropdown option, which would be quite clear.

AndreMiras:
Yeah I will give it a try then. But using triple backquotes when sharing the bug log is not that complex, yet most of bug reports don't have it. Same thing for avoiding duplicated bugs creation :sweat_smile: 
I'll update this PR still and look into the release creation one later

AndreMiras:
I won't look into the dedicated issue template just now. So let's merge this minor improvement until we try out for a better solution

PR #2179, opened 2020-05-08
Narrows some context manager scopes

AndreMiras:
Tries to keep the `current_directory` context manager as scoped as
possible. Also use the `contextlib.suppress` context manager when
appropriated as it's more concise.

### DIFF
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -11,6 +11,7 @@ import re
 import sh
 import shutil
 import subprocess
+from contextlib import suppress
 
 from pythonforandroid.util import (
     current_directory, ensure_dir,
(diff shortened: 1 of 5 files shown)


pythonforandroid/build.py
```diff
@@ -621,10 +622,8 @@ def run_setuppy_install(ctx, project_dir, env=None):
             ),
             "freeze"
         ], env=copy.copy(env))
-        try:
+        with suppress(AttributeError):
             constraints = constraints.decode("utf-8", "replace")
-        except AttributeError:
-            pass
```
opacam, line 627: :fire: two less  lines of code  without sacrificing readability :smile:

AndreMiras: Yes I love that one :)
But it cannot really be used nicely if we have a large comment block above the `pass`, so yes I decided to introduced it only when we had a simple except ...: pass

PR #2182, opened 2020-05-09
Reuse common AssetExtract.java

AndreMiras:
Also removes unused `mActivity` member.
Note we changed the constructor input parameter to accept a `Context`
object which is a parent class of the previous `Activity` one.
Changes were tested on both kivy and service_only bootstraps.


pythonforandroid/bootstrap.py
```diff
@@ -132,9 +132,6 @@ def get_build_dir(self):
     def get_dist_dir(self, name):
         return join(self.ctx.dist_dir, name)
 
-    def get_common_dir(self):
```
AndreMiras, line 135: no longer needed since https://github.com/kivy/python-for-android/pull/2092

PR #2183, opened 2020-05-09
Use common Hardware.java

AndreMiras:
All 3 `Hardware.java` seemed to be the same (minus some formatting).


pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonService.java
```diff
@@ -16,8 +16,6 @@
 
 import org.kivy.android.PythonUtil;
 
-import org.renpy.android.Hardware;
```
AndreMiras, line 19: Wasn't actually used in this file.
We may introduce some linting to the Java files as well once we're done deleting duplicated code

PR #2185, opened 2020-05-09
Removes java concurrency/ folder

AndreMiras:
None of the class seems to be used.


inclement:
Is there any chance there's pyjnius code using this?

AndreMiras:
I haven't think about that one, I'll look into it, thanks

AndreMiras:
It got introduced in https://github.com/kivy/python-for-android/pull/578 but I still don't have it clear why :smile:

AndreMiras:
yes agree, if it breaks something someone should complain. Isn't it how software development works? :smile:

PR #2196, opened 2020-05-13
:white_check_mark: Add tests for hostpython3 recipe

opacam:
This will cover **100%** of the code for the hostpython3 recipe.
Also we migrate `os.path.exists` and `os.path.isfile` to `pathlib.Path`'s implementation because when we mock these functions, it seems that property `return_value` is being ignored.

### DIFF
--- a/tests/recipes/test_hostpython3.py
+++ b/tests/recipes/test_hostpython3.py
@@ -1,0 +1,164 @@
+import unittest
+
+from os.path import join
+from unittest import mock
+
+from pythonforandroid.recipes.hostpython3 import (
+    HOSTPYTHON_VERSION_UNSET_MESSAGE, SETUP_DIST_NOT_FIND_MESSAGE,
+)
+from pythonforandroid.util import BuildInterruptingException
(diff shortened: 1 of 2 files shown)


tests/recipes/test_hostpython3.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def test_property__exe_name_no_version(self):
+        hostpython_version = self.recipe.version
+        self.recipe._version = None
+        with self.assertRaises(BuildInterruptingException) as e:
+            py_exe = self.recipe._exe_name  # noqa: F841
+        self.assertEqual(e.exception.args[0], HOSTPYTHON_VERSION_UNSET_MESSAGE)
+        # restore recipe's version or we will get failures with other test,
+        # since we share `self.recipe with all the tests of the class
+        self.recipe._version = hostpython_version
```
AndreMiras, line 27: I think this is probably fine, but usually a safe way to handle theses things is through context managers.
e.g. https://github.com/kivy/python-for-android/pull/1948
I would say let's keep it simple as you did, but just wanted to share the use of this old PR

Issue #2020, opened 2019-11-12
Support for androidx


tito:
[AndroidX](https://developer.android.com/jetpack/androidx)  is a new set of libraries that replaces the old Android "supports" libraries, and more.

We don't use supports libraries directly (if i recall correctly), but to activate `androidx` package, it requires a new file named `gradle.properties` with:

```
android.useAndroidX=true
android.enableJetifier=true
```

misl6:
It's starting to be quite a priority, in order to support the latest features on Google Play Services.

RobertFlatt:
`gradle.properties` is an extra file we don't have to generate.
This can be implemented in `bootstraps/common/build/templates/build.tmpl.gradle`

Insert this block as the last item in the android{} block:
```
   ext {
      useAndroidX=true
      enableJetifier=true
   }
```
And if like me you are using "androidx.core.content.FileProvider", in buildozer.spec add:
 `  android.gradle_dependencies = "androidx.core:core:1.3.2"`

Other androidx classes may be in other packages https://maven.google.com/web/index.html

**This requires android.api=28  or greater**

I have I have used this to build and run an app implementing Android's file sharing feature.
I have not seen any side effects yet!

misl6:
Fixed via https://github.com/kivy/python-for-android/pull/2385

PR #2200, opened 2020-05-14
Content providers

pavelsof:
This PR should make it possible to add [content providers](https://developer.android.com/guide/topics/manifest/provider-element) to the Android manifest. Two options are added to the common `build.py` script:

- `--content-providers` is practically identical to the `--intent-filters` option, it expects the path to an XML file the contents of which are included in the `AndroidManifest.xml`.
- `--add-xml-resource` can be used to copy one or more files over to the `src/main/res/xml` subdir of the build dir. This is needed because `<provider>` elements in the Android manifest usually have to refer to another XML file that defines the paths the app is allowed to "export".

If approved, I will also prepare a PR for the Buildozer repo to add the `buildozer.spec` counterparts.

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -511,6 +511,11 @@ main.py that loads it.''')
         join(res_dir, 'values/strings.xml'),
         **render_args)
 
+    # extra resources
+    if args.add_resource:
+        for file_path in args.add_resource:
+            shutil.copy(file_path, join(res_dir, basename(file_path)))
+
     if exists(join("templates", "custom_rules.tmpl.xml")):
         render(
             'custom_rules.tmpl.xml',
@@ -717,6 +722,9 @@ tools directory of the Android SDK.
                           'filename containing xml. The filename should be '
                           'located relative to the python-for-android '
                           'directory'))
+    ap.add_argument('--add-resource', dest='add_resource', action='append',
+                    help=('Copy this file at this path to the src/main/res '
+                          'subdirectory of the build.'))
     ap.add_argument('--with-billing', dest='billing_pubkey',
                     help='If set, the billing service will be added (not implemented)')
     ap.add_argument('--add-source', dest='extra_source_dirs', action='append',
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -511,10 +511,13 @@ main.py that loads it.''')
         join(res_dir, 'values/strings.xml'),
         **render_args)
 
-    # extra resources
-    if args.add_resource:
-        for file_path in args.add_resource:
-            shutil.copy(file_path, join(res_dir, basename(file_path)))
+    # extra xml resources
+    if args.add_xml_resource:
+        xml_res_dir = join(res_dir, 'xml')
+        if not exists(xml_res_dir):
+            os.mkdir(xml_res_dir)
+        for file_path in args.add_xml_resource:
+            shutil.copy(file_path, join(xml_res_dir, basename(file_path)))
 
     if exists(join("templates", "custom_rules.tmpl.xml")):
         render(
@@ -722,8 +725,9 @@ tools directory of the Android SDK.
                           'filename containing xml. The filename should be '
                           'located relative to the python-for-android '
                           'directory'))
-    ap.add_argument('--add-resource', dest='add_resource', action='append',
-                    help=('Copy this file at this path to the src/main/res '
+    ap.add_argument('--add-xml-resource', dest='add_xml_resource',
+                    action='append',
+                    help=('Copy this file to the src/main/res/xml '
                           'subdirectory of the build.'))
     ap.add_argument('--with-billing', dest='billing_pubkey',
                     help='If set, the billing service will be added (not implemented)')
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -372,6 +372,10 @@ main.py that loads it.''')
         with open(args.intent_filters) as fd:
             args.intent_filters = fd.read()
 
+    if args.content_providers:
+        with open(args.content_providers) as f:
+            args.content_providers = f.read()
+
     if not args.add_activity:
         args.add_activity = []
 
@@ -703,6 +707,12 @@ tools directory of the Android SDK.
                           'the discovered ndk_api in the dist'))
     ap.add_argument('--intent-filters', dest='intent_filters',
                     help=('Add intent-filters xml rules to the '
+                          'AndroidManifest.xml file. The argument is a '
+                          'filename containing xml. The filename should be '
+                          'located relative to the python-for-android '
+                          'directory'))
+    ap.add_argument('--content-providers', dest='content_providers',
+                    help=('Add providers xml rules to the '
                           'AndroidManifest.xml file. The argument is a '
                           'filename containing xml. The filename should be '
                           'located relative to the python-for-android '
--- a/pythonforandroid/bootstraps/service_only/build/templates/AndroidManifest.tmpl.xml
+++ b/pythonforandroid/bootstraps/service_only/build/templates/AndroidManifest.tmpl.xml
@@ -92,6 +92,10 @@
             </intent-filter>
         </receiver>
         {% endif %}
+
+        {%- if args.content_providers -%}
+        {{- args.content_providers -}}
+        {%- endif -%}
     </application>
 
 </manifest>
--- a/doc/source/buildoptions.rst
+++ b/doc/source/buildoptions.rst
@@ -86,6 +86,8 @@ options (this list may not be exhaustive):
   include multiple jar files, pass this argument multiple times.
 - ``--intent-filters``: A file path containing intent filter xml to be
   included in AndroidManifest.xml.
+- ``--content-providers``: A file path containing provider xml to be
+  included in AndroidManifest.xml.
 - ``--service``: A service name and the Python script it should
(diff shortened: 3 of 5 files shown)


pavelsof:
Just opened kivy/buildozer#1119 as an example of how the two new `build.py` options could be used by Buildozer.

Please consider following up on either this PR here or on #1922, so that there is an easy way to setup content providers in a Kivy app.

AndreMiras:
Thanks @pavelsof for the PR, looks clean.
I'm thinking, would it make sense to make the flag name more generic e.g. `--extra-application-xml`?
Because your application actually looks generic enough to cover other cases.
I'm wondering what @inclement thinks about it :thinking:

pythonforandroid/bootstraps/service_only/build/templates/AndroidManifest.tmpl.xml
```diff
@@ -92,6 +92,10 @@
             </intent-filter>
         </receiver>
         {% endif %}
+
+        {%- if args.content_providers -%}
```
inclement, line 96: Why {%- rather than just {% (and same elsewhere)?

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -707,6 +719,16 @@ def parse_args_and_make_package(args=None):
                           'filename containing xml. The filename should be '
                           'located relative to the python-for-android '
                           'directory'))
+    ap.add_argument('--content-providers', dest='content_providers',
+                    help=('Add providers xml rules to the '
+                          'AndroidManifest.xml file. The argument is a '
+                          'filename containing xml. The filename should be '
+                          'located relative to the python-for-android '
+                          'directory'))
+    ap.add_argument('--add-xml-resource', dest='add_xml_resource',
```
inclement, line 728: Didn't we add something else recently for moving stuff into the src/main/res folder?
I'm wondering if what we really want is a generic "copy this file/folder into src/main/res" argument, rather than many arguments for different subfolders.

Fak3:
I believe this is more generic version of my pr #1922. Maybe this one is better, as it is more flexible. But please note that I had to add more Gradle build flags for this to work. (See my original pr)

pythonforandroid/bootstraps/service_only/build/templates/AndroidManifest.tmpl.xml
pavelsof: No other reason apart from not being familiar with Jinja and copy-pasting the lines that render the intent filters. I will fix it.

pavelsof:
@AndreMiras, I agree, it would be better if the option name makes it clear that it can be used to insert arbitrary XML in the `<application>` element. Should I change it?

@Fak3, thank you for jumping in! Indeed, just as you did, I also had to add `com.android.support:support-v4:26.1.0` as a dependency in Gradle; but I used the `android.gradle_dependencies` option in `buildozer.spec`. Also, I did not have to enable multidex.

pythonforandroid/bootstraps/common/build/build.py
pavelsof: Sure, I can replace it with a more generic argument to copy files/directories into `src/main/res`.

AndreMiras:
Thanks guys for the discussion!
@pavelsof if #1922 would work for you use case and @Fak3 is willing to resume his work on it then we could go for it instead.
If we were to integrate yours, then yes I would say let's make the parameters a bit more generic and add unit tests.
Otherwise if we really need both then we would go for both.

pavelsof:
@AndreMiras, yes, #1922 works for me. It would be slightly easier to use as well, because a Kivy developer would only need the paths XML, but not the `<provider>` boilerplate.

RobertFlatt:
This would be great if this was signed off and 'dozer supported the args, I just did it by hand and inserted :

```
    <provider
        android:name="android.support.v4.content.FileProvider"
        android:authorities="{{ args.package }}.fileprovider"
        android:grantUriPermissions="true"
        android:exported="false">
        <meta-data
            android:name="android.support.FILE_PROVIDER_PATHS"
            android:resource="@xml/provider_paths" />
    </provider>
```
Also I used
`android.gradle_dependencies = "com.android.support:support-core-utils:27.0.0"`

I tested build and execution on api 27, 29, and 30 all good

Arbitrary html would be a good thing in general. 
And this case androidx support will require a change the `android:name` field value

RobertFlatt:
FYI, the steps to move from the depreciated packages to androidx https://github.com/kivy/python-for-android/issues/2020#issuecomment-735021748

pavelsof:
Ping! This is a polite reminder/request to please merge this PR. I have been using my [python-for-android](https://github.com/pavelsof/python-for-android) and [Buildozer](https://github.com/pavelsof/buildozer) forks for a [pet Kivy app](https://github.com/pavelsof/mobile-wormhole), and there have not been problems for what is now almost a year. Content providers constitute an important thing in Android and Kivy should be able to work with them.

Fak3:
> Ping! This is a polite reminder/request to please merge this PR.

I believe we have to fix #2385 first, it would not work without it

pavelsof:
It works (at least for me) with the older, probably deprecated by now, `android.support.v4.content` way. But I agree, AndroidX would be much better.

What can we do to prod the core maintainers to say what is wrong with either PR? :confused:

RobertFlatt:
> What can we do to prod the core maintainers to say what is wrong with either PR? 😕

I wish I knew the answer to that question 🤔

If https://github.com/kivy/python-for-android/pull/2385 where approved (sigh) then support here for both versions might look like this:

```
	{% if args.enable_androidx %}
	android:name="androidx.core.content.FileProvider"
	{% else %}
	android:name="android.support.v4.content.FileProvider"
	{% endif %}	
```

RobertFlatt:
As an enhancement I suggest a default `provider_paths.xml`
This would be a lot easier for the script kiddies.

For example the attached `provide_paths.xml` that provides all the default directories in both local storage directories.
So a custom` provider_paths.xml` would only be required if the user makes a custom directory, this is easy to explain.
[provider_paths.txt](https://github.com/kivy/python-for-android/files/5863136/provider_paths.txt) (renamed from .xml to .txt to keep Github happy).

File belongs in `pythonforandroid/bootstraps/sdl2/build/src/main/res/xml` which must be created.
I have not checked if an explict copy is required in the build.

darpan5552:
Instead of `Fileprovider` there can be other conditions also where we need to add extra entries to `Manifest`. Furthermore there is no guarantee that in future there will not be more such important requirements.

Why not keep everything thing as `xml` file as the boilerplate code added to manifest also follows xml syntax.

Adding two parameters `--extend-manifest <something.xml>` and `--add-xml-resource <something 2.xml>` should do the thing.

RobertFlatt:
Is this what you want?
https://github.com/kivy/python-for-android/blob/develop/pythonforandroid/bootstraps/sdl2/build/templates/AndroidManifest.tmpl.xml#L45

Also there is at least one xml related PR https://github.com/kivy/python-for-android/pull/2330 that sits and rots :(

darpan5552:
> Is this what you want?
> https://github.com/kivy/python-for-android/blob/develop/pythonforandroid/bootstraps/sdl2/build/templates/AndroidManifest.tmpl.xml#L45 
Yes, i wanted this. It would be better if it also integrates addition of `intent-filters` in it. I know both are at different levels in xml tree but there are not too much use cases that in can't be handled on p4a side.

> Also there is at least one xml related PR #2330 that sits and rots :(

#2330 is also a very specific case where he wants to insert intent filters manually but automating the modification of manifest file.
In contrast, i am on the side of `No Automation Policy` as there can be large cases for such automation. A better way could be to allow the user to write modification and we just push the modifications in its place.

Fak3:
I think the most generic thing to address all cases will be to allow p4a use custom manifest template file, via an option --manifest=/path/to/mymanifest.xml.tmpl

darpan5552:
There are some things important in default manifest file. So for using a
custom manifest file, a user has to study the manifest lying under the hood.
I argue that if he can go there, then what is the use of providing options,
he can directly do additions right there only.

HeRo002:
I just asked "Has content providers been incorporated in p4a?" on https://groups.google.com/g/kivy-users/c/7VVYdkA0gjQ - and received this answer by [RobertFlatt](https://github.com/RobertFlatt) : 

  I think the whole content provider thing faded to irrelevance, turned out that the MediaStore can provide a uri.

  I wrote this as a test case:  https://github.com/Android-for-Python/Share-Send-Example

Fak3:
So how this can help with plyer camera, which requires fileprovider to work?

RobertFlatt:
@Fak3 You want to save an image to shared storage on Android >= 10? 
Write a temp image to local storage and then copy that to the MediaStore.

I wrote a Camera Widget that on Android can save screen shots to public storage (for photos the widget uses native image capture that handles this). https://github.com/Android-for-Python/Camera4Kivy  

The save and copy code is here <https://github.com/Android-for-Python/Camera4Kivy/blob/main/src/camera4kivy/preview_camerax.py#L224-L253>

Fak3:
I want it to be compatible with Android 7

RobertFlatt:
You don't say what "it" is, but if "it" is save a file then on Android <10 shared storage is a file system.

Fak3:
I want to capture a photo with external camera app. On Android 7 it throws an error when you just try to give an external app a filesystem path to save an image to. It only works if you properly define fileprovider in your Android manifest.

Fak3:
See kivy/plyer#549

RobertFlatt:
Yep, Plyer Camera looks like it has had bit rot for several years. The design approach is obsolete.

I see that you would have that issue.

Fak3:
Working approach for Android 7 is still needed. It will take years for users to switch to Android 10

PR #2208, opened 2020-05-19
Add possibility to add a backup rules xml file

tcdude:
This PR enables users to pass a new argument to p4a to enable
custom backup rules through the means of an XML file as documented
in the official [Android docs](https://developer.android.com/guide/topics/data/autobackup#IncludingFiles).

I also noticed that the "allowBackup" flag would do nothing when userd
with either `webview` or `service_only` bootstraps, where the 
`AndroidManifest.tmpl.xml` would set it to a hardcoded "true" instead of 
the user-defined value.

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -474,7 +474,7 @@ main.py that loads it.''')
     if args.backup_rules:
         res_xml_dir = join(res_dir, 'xml')
         ensure_dir(res_xml_dir)
-        shutil.copy(args.backup_rules, res_xml_dir)
+        shutil.copy(join(args.private, args.backup_rules), res_xml_dir)
         args.backup_rules = split(args.backup_rules)[1][:-4]
 
     # Render out android manifest:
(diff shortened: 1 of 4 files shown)


Jorilx:
This would be very nice to have, thanks!
I had to specify the filename as `../../../../../../backup_rules.xml` though, is this expected? Thanks again!

tcdude:
It is expected, since I duplicated behavior from an existing command line argument. I copied the help text, advising that the path needs to be relative to `python-for-android`. You can also use absolute paths though, which would make it probably easier to read...

I have too little insight into the inner workings of p4a to implement a better way, but I'm happy to update this PR if someone tells me how to make this relative to the project root or similar.

tshirtman:
I do think making it relative to either the current working directory (`os.getcwd()`) when the p4a command is called, *or* from the source directory we tell p4a to package (should be `args.private`?). using either `os.path.join(path, args.backup_rules)` or `pathlib.Path(path) / args.backup_rules`.

Jorilx:
I have a PR ready to be submitted for adding support for `--backup-rules` to buildozer :)

tcdude:
I went for `args.private`... Looks like I've made a little mess with my latest commit... hope that doesn't create too much trouble to the maintainers when merging this PR...

Sorry

Jorilx:
I'm not an expert by any means, but maybe it would be easier to just close this PR and open another "clean" one?

tshirtman:
nothing wrong with force pushing if you rebased/reset the branch to clean up, the PR will update itself if you push to the same branch.

tcdude:
thanks @tshirtman that seems to have done the trick then :)

Let me know if there are further changes to be included, before this can be merged

tshirtman (approved):
LGTM

PR #2209, opened 2020-05-20
:sparkles: Add `opencv_extras` recipe

### LINKED ISSUE #2206: My despair at trying to simply import the opencv face class (python-for-android)
I've been trying this for weeks, I was told it would be straightforward and that if I wanted to import the face class I could simply "add that option to the recipe".  But this is somehow starting to seem more difficult than I first thought. Has anyone ever managed to compile any of the extra modules (contrib) through p4a by changing the opencv recipe? If this is indeed too difficult, could I use pyjnius, as only 10 lines of code in my entire app is dedicated to detecting facial landmarks.

opacam:
This recipe allows us to build extra modules from the `opencv_contrib` repository. It depends on `opencv` recipe and all the build of the modules will be performed inside opencv build directory.

This closes #2206

**Note:** I've not tested this recipe at runtime but it builds fine for `arm64-v8a` in my local tests, plus generates the extra opencv libraries (which are a lot so probably will increase the size of the apk considerably).

**See also:** https://github.com/opencv/opencv_contrib

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ testapps-with-numpy/%: virtualenv
 	$(eval $@_APP_ARCH := $(shell basename $*))
 	. $(ACTIVATE) && cd testapps/on_device_unit_tests/ && \
     python setup.py apk --sdk-dir $(ANDROID_SDK_HOME) --ndk-dir $(ANDROID_NDK_HOME) \
-    --requirements libffi,sdl2,pyjnius,kivy,python3,openssl,requests,sqlite3,setuptools,numpy,opencv,opencv_extras \
+    --requirements libffi,sdl2,pyjnius,kivy,python3,openssl,requests,sqlite3,setuptools,numpy \
     --arch=$($@_APP_ARCH)
 
 testapps/%: virtualenv
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ testapps-with-numpy/%: virtualenv
 	$(eval $@_APP_ARCH := $(shell basename $*))
 	. $(ACTIVATE) && cd testapps/on_device_unit_tests/ && \
     python setup.py apk --sdk-dir $(ANDROID_SDK_HOME) --ndk-dir $(ANDROID_NDK_HOME) \
-    --requirements libffi,sdl2,pyjnius,kivy,python3,openssl,requests,sqlite3,setuptools,numpy \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -38,7 +38,7 @@ testapps-with-numpy/%: virtualenv
 	$(eval $@_APP_ARCH := $(shell basename $*))
 	. $(ACTIVATE) && cd testapps/on_device_unit_tests/ && \
     python setup.py apk --sdk-dir $(ANDROID_SDK_HOME) --ndk-dir $(ANDROID_NDK_HOME) \
-    --requirements libffi,sdl2,pyjnius,kivy,python3,openssl,requests,sqlite3,setuptools,numpy \
+    --requirements libffi,sdl2,pyjnius,kivy,python3,openssl,requests,sqlite3,setuptools,numpy,opencv,opencv_extras \
```
AndreMiras, line 41: Do we really want to have opencv and extras build each time? Or is it just for temporary testing the PR?

opacam: hahaha....this was only to test the build...I will reverse it later (that is why I labeled this PR `wip` and put the emoji :construction: in my last commit).

Here are the artifacts for each arch:

https://github.com/kivy/python-for-android/actions/runs/110275077

Btw, it almost doubles the size of the apk...anyway, it may be useful to someone and it doesn't look like hard to maintain.

AndreMiras: Great that what's I thought, but wasn't sure :+1: 
I have nothing against merging, ping me for the final review

SonofPoseidon1982:
Thanks and well done Opacam. I will keep you up to date my progress.

pythonforandroid/recipes/opencv/__init__.py
```diff
@@ -32,6 +32,7 @@ class OpenCVRecipe(NDKRecipe):
         'libopencv_dnn.so',
         'libopencv_imgcodecs.so',
         'libopencv_photo.so'
+        # todo: add opencv_extras libraries
```
AndreMiras, line 35: what's the story on that one?

opacam:
Well, the [travis error is expected](
https://travis-ci.org/github/kivy/python-for-android/jobs/692129659#L1564)...this recipes compiles an insane number of libraries and it takes too long, anyway, we know that it builds thanks to previous tests with gh-actions, so let's merge this!!

@AndreMiras, thanks for the review!! :smile:

@SonofPoseidon1982, thanks for the testing and feedback!! :+1:

PR #2210, opened 2020-05-20
allow specify --numeric-version with APK_NUMERIC_VERSION env

zebra-lucky:
- add `default=os.environ.get('APK_NUMERIC_VERSION', None)` to `--numeric-version` argument

### DIFF
--- a/doc/source/buildoptions.rst
+++ b/doc/source/buildoptions.rst
@@ -58,6 +58,9 @@ options (this list may not be exhaustive):
 - ``--private``: The directory containing your project files.
 - ``--package``: The Java package name for your project. e.g. ``org.example.yourapp``.
 - ``--name``: The app name.
+- ``--numeric-version``: The numeric version number of the project.
+  If not given and no ``APK_NUMERIC_VERSION`` env is set, this is
+  automatically computed from the version.
 - ``--version``: The version number.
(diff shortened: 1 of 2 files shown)


AndreMiras:
Thanks @zebra-lucky for the PR.
The code looks OK, but apart from not having unit tests I have mixed feelings about this feature :thinking: 
It basically introduces inconsistencies in the way we deal with arguments has it would be the only argument of this file that comes with "hybrid" way of setting it up.
I suppose this is a follow-up from https://github.com/kivy/buildozer/pull/1079, but I'm wondering why didn't we do with a normal `buildozer.spec` file parameter rather than with environment variable?
I'm not completely against the feature however, just need to think and discuss it further.
I'm wondering what @inclement thinks about it too

zebra-lucky:
> I suppose this is a follow-up from kivy/buildozer#1079, but I'm wondering why didn't we do with a normal buildozer.spec file parameter rather than with environment variable?

For me env approach is more flexible and versionCode can be calculated with use of other build env variables (`APP_ANDROID_ARCH=arm64-v8a` as example).

І'l look how to add unit testing tonight.

obfusk:
buildozer now supports setting `APP_ANDROID_NUMERIC_VERSION`.

PR #2211, opened 2020-05-20
[WIP] :package: Make our default build  as `release`

### LINKED ISSUE #2207: apk size growing
### Versions

* Python:  3.8.2
* OS:   Win10/WSL/Ubuntu20.04
* Kivy: 1.11.1
* Cython: 0.29.17

### Description
Building "Hello World" in an empty directory:

buildozer init        
buildozer android debug

Results in an apk that is 28MB

p4a 2020.4.29 (NDK 19c, SDK 29, armeapi-v7a ) also gives 28MB
p4a 2020.4.29 (NDK 19c, SDK 29, arm64-v8a ) gives 21MB

Not long ago this was 12MB 

Is 28MB the new normal?

### buildozer.spec

Command:
```sh
// REPLACE ME: buildozer command ran? e.g. buildozer android debug
```

Spec file:
```
// REPLACE ME: Paste your buildozer.spec file here
```

### Logs

```
// REPLACE ME: Paste the build ouput containing the error
```

opacam:
To make a build in debug mode you must use `--debuggable` (which is actually the word that AndroidManifest.xml uses to perform a debuggable build)

This closes #2207

**Important:** To actually perform a full debuggable build, you must be sure to remove any existant release built, since some libraries (the ones that we use the `ndk_build` command) won't be compiled again, unless we remove the old compilation. Also noting, that this is the same for the other way around (If you performed a debuggable build and you want a release built, then better remove the old compiled files)

**Note**: I labeled this `wip` because I think that our github actions artifacts should be build as `debuggable`, otherwise it would require to sign the apks to be able to install it on our devices...or maybe...should  we [sign them](https://github.com/marketplace/actions/sign-android-release)?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ docker/run/make/%: docker/build
 docker/run/make/with-artifact/%: docker/build
 	$(eval $@_APP_ARCH := $(shell basename $*))
 	docker run --name p4a-latest --env-file=.env $(DOCKER_IMAGE) make $*
-	docker cp p4a-latest:/home/user/app/testapps/on_device_unit_tests/bdist_unit_tests_app__$($@_APP_ARCH)-debug-1.1-.apk ./apks
+	docker cp p4a-latest:/home/user/app/testapps/on_device_unit_tests/bdist_unit_tests_app__$($@_APP_ARCH)-release-unsigned-1.1-.apk ./apks
 	docker rm -fv p4a-latest
 
 docker/run/shell: docker/build
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -460,7 +460,7 @@ main.py that loads it.''')
         "service": service,
         "service_names": service_names,
         "android_api": android_api,
-        "debug": "debug" in args.build_mode,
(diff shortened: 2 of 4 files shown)


opacam:
To compare the apks size, the artifacts from:
  - latest commit of develop's branch, [around **40MB**](https://github.com/kivy/python-for-android/actions/runs/107065023)
  - and the ones built with this PR, [less than **20MB**](https://github.com/kivy/python-for-android/actions/runs/110469367)

inclement:
Please don't merge this yet, just looking

inclement:
I think this is the wrong fix, the APK build modes (debug and release) and the .so debug stripping are independent, and should not be conflated.

I think the correct fix is to not include debug symbols by default, even in debug APKs, but have a new p4a argument (perhaps --debuggable) that enables them.

It should be normal and expected behaviour that debug builds are the norm and release builds are the exception.

opacam:
@inclement, thanks for the clarification!!

And for both of you, **thanks for the review!!** :+1:

**Note:** I'm closing this since the branch's name wouldn't match the things that we have to do, so I will open a new PR with a fix that suits our needs and a new branch name :wink:

PR #2220, opened 2020-05-30
:arrow_up: Bumps to Cython==0.29.19

AndreMiras:


AndreMiras:
armv7 artifact is running fine on device, except the `requests` module test that started failing (also develop) I'll investigate that later.
Merging

PR #2226, opened 2020-06-02
Master

AndreMiras:


AndreMiras:
It looks like we didn't do so paste releases? Because it has quite some commits

AndreMiras:
Hm OK actually reading the docs before we were actually merging the release branch also to develop and not master back to develop. We got to clarify this up, I'm not yet sure which one makes more sense. I'd say master to develop is what we should be doing

PR #2234, opened 2020-06-08
Updated gradle plugin version

shashi278:
Description of changes:

* 3.1.4 seems to be pretty outdated and also has some issues with Firebase, 3.5.2 works quite well.
* Got rid of deprecated `compile` method. Replaced them with `implementation`.
<del>* Added extra `classpath` and `apply plugin` as they're required to use firebase(Instead of hardcoding these in the build.gradle file, there could be options to add extra `classpath` and `apply plugin` as per one's requirement).</del> (Because Google services plugin requires a google-services.json file so this plugin must only be added if requires)


AndreMiras (approved):
CI happy so we should be fine, thanks

PR #2246, opened 2020-06-25
Fix runtime psycopg2 error

Progdrasil:
I inadvertently introduced a runtime error in #2245.

For some reason psycopg2 2.8.4 has a linking issue to PQencryptPasswordConn. This seems to have been fixed with 2.8.5 on my machine. So a simple revision bump should be fine.


Progdrasil:
Right sorry about that

Bouchra994:
hi so i must uninstall the 2.8.4 psycopg2 and install the 2.8.5 to fix the the error?

PR #2248, opened 2020-06-27
added --activity-class-name and --activity-package parameters 

vesellov:
to make possible to use custom PythonActivity java class

### DIFF
--- a/pythonforandroid/recipes/android/__init__.py
+++ b/pythonforandroid/recipes/android/__init__.py
@@ -51,6 +51,8 @@ class AndroidRecipe(IncludedFilesBehaviour, CythonRecipe):
             'PY2': 0,
             'JAVA_NAMESPACE': java_ns,
             'JNI_NAMESPACE': jni_ns,
+            'ACTIVITY_CLASS_NAME': self.ctx.activity_class_name,
+            'ACTIVITY_CLASS_NAMESPACE': self.ctx.activity_class_name.replace('.', '/'),
         }
 
         # create config files for Cython, C and Python
--- a/pythonforandroid/recipes/android/src/android/permissions.py
+++ b/pythonforandroid/recipes/android/src/android/permissions.py
@@ -9,6 +9,9 @@ except ImportError:
         raise RuntimeError("pyjnius not available")
 
 
+from android.config import ACTIVITY_CLASS_NAME, ACTIVITY_CLASS_NAMESPACE
+
+
 class Permission:
     ACCEPT_HANDOVER = "android.permission.ACCEPT_HANDOVER"
     ACCESS_COARSE_LOCATION = "android.permission.ACCESS_COARSE_LOCATION"
@@ -431,7 +434,7 @@ class _onRequestPermissionsCallback(PythonJavaClass):
     """Callback class for registering a Python callback from
     onRequestPermissionsResult in PythonActivity.
     """
-    __javainterfaces__ = ['org.kivy.android.PythonActivity$PermissionsCallback']
+    __javainterfaces__ = [ACTIVITY_CLASS_NAMESPACE + '$PermissionsCallback']
     __javacontext__ = 'app'
 
     def __init__(self, func):
@@ -484,7 +487,7 @@ class _RequestPermissionsManager:
     def register_callback(cls):
         """Register Java callback for requestPermissions."""
         cls._java_callback = _onRequestPermissionsCallback(cls.python_callback)
-        python_activity = autoclass('org.kivy.android.PythonActivity')
+        python_activity = autoclass(ACTIVITY_CLASS_NAME)
         python_activity.addPermissionsCallback(cls._java_callback)
 
     @classmethod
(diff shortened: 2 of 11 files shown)


pythonforandroid/toolchain.py
```diff
@@ -376,6 +376,16 @@ def __init__(self):
...
             help='Directory to look for local recipes')
 
+        generic_parser.add_argument(
+            '--activity-package', '--activity_package',
+            dest='activity_package', default='org.kivy.android',
+            help='The java package where activity class is defined')
+
+        generic_parser.add_argument(
+            '--activity-class-name', '--activity_class_name',
+            dest='activity_class_name', default='PythonActivity',
```
AndreMiras, line 386: I'm wondering if both default couldn't be shared with the one defined in `build.py` to keep a single source of truth. Maybe using a const somewhere.
Also matter of taste but I would ditch the underscored version of the argument

pythonforandroid/recipes/android/src/android/permissions.py
```diff
@@ -431,7 +434,7 @@ class _onRequestPermissionsCallback(PythonJavaClass):
     """Callback class for registering a Python callback from
     onRequestPermissionsResult in PythonActivity.
     """
-    __javainterfaces__ = ['org.kivy.android.PythonActivity$PermissionsCallback']
+    __javainterfaces__ = [ACTIVITY_NAMESPACE + '/' + ACTIVITY_CLASS_NAME + '$PermissionsCallback']
```
AndreMiras, line 437: I guess we got that wrong since the beginning , but it seems that both doted and slashed notation are working then, right?

pythonforandroid/recipes/android/__init__.py
```diff
@@ -51,6 +51,9 @@ def prebuild_arch(self, arch):
             'PY2': 0,
             'JAVA_NAMESPACE': java_ns,
             'JNI_NAMESPACE': jni_ns,
+            'ACTIVITY_CLASS_NAME': self.ctx.activity_class_name,
+            'ACTIVITY_PACKAGE': self.ctx.activity_package,
+            'ACTIVITY_NAMESPACE': self.ctx.activity_package.replace('.', '/'),
```
AndreMiras, line 56: How about introducing a `ACTIVITY_FULL_CLASS_NAME` value that is already concatenating both `activity_class_name` and `activity_package`?
e.g.
```python
{
    "ACTIVITY_FULL_CLASS_NAME": f"{self.ctx.activity_package}.{self.ctx.activity_class_name}",
}
```
That way you don't need to carry over the concatenation all over later in the code

pythonforandroid/toolchain.py
vesellov: i will check about unit tests Andre.  i was a bit in the rush when started it.

pythonforandroid/recipes/android/src/android/permissions.py
vesellov: yes. exactly... i was in doubt, but checked with py-jnius docs and it says it must be '/'

pythonforandroid/recipes/android/__init__.py
vesellov: 👍  i will check that!

pythonforandroid/toolchain.py
AndreMiras: Thanks, we have pretty low coverage in both p4a and buildozer, but I'm trying to push for increasing it slowly

pythonforandroid/recipes/android/__init__.py
vesellov: ok... now i just added one input parameter `activity_class_name` which must be a full name of the Java class.

pythonforandroid/toolchain.py
vesellov: added some basic assertion in the existing unit test to check that Context was populated properly with new value of `activity_class_name`

pythonforandroid/toolchain.py
```diff
@@ -376,6 +376,11 @@ def __init__(self):
             dest='local_recipes', default='./p4a-recipes',
             help='Directory to look for local recipes')
 
+        generic_parser.add_argument(
+            '--activity-class-name',
+            dest='activity_class_name', default='org.kivy.android.PythonActivity',
```
AndreMiras, line 381: I wish we had a single source of truth shared with https://github.com/kivy/python-for-android/pull/2248/files#diff-c439e980dc587602fda87459815f09feR450 for the `org.kivy.android.PythonActivity` value, but I can live without

AndreMiras:
CI happy, I've also tried the `bdist_test_app_unittests__armeabi-v7a-debug-1.1.apk` artifact on device, looking good, merging

pythonforandroid/toolchain.py
vesellov: @AndreMiras  you mean that there are two settings now that are storing same value? : `android.entrypoint` and `android.activity_class_name` ?

vesellov:
👍

PR #2275, opened 2020-08-03
build python3 with loadable-sqlite-extensions

obfusk:
I assume most users don't need this (unlike me), but I don't really see any downsides to having it enabled by default.


obfusk:
> Thanks you for the PR, do you know how much the final APK grows from it?

No. But the only thing it does is compile a few extra lines of C (that used to be enabled by default). So I can't imagine it's noticeable.

obfusk:
This is the only change: https://github.com/python/cpython/blob/master/Modules/_sqlite/connection.c#L1077

PR #2278, opened 2020-08-04
`libpython3.8m.so` should not have `m` suffix

davidhewitt:
Hi,

I'm helping a user to diagnose a downstream issue https://github.com/PyO3/pyo3/issues/1077 when compiling a Rust application using p4a. In the linked issue is a lot of background and the OP also contains a link to a minimal example.

Their build with p4a targeting Python 3.8 was failing with link errors. Their workaround was to manually link to `libpython3.8m.so`.

However, I understood the `m` suffix was removed with Python 3.8 after pymalloc builds became ABI-compatible with the mainline: https://docs.python.org/3/whatsnew/3.8.html#build-and-c-api-changes

So I think that `libpython3.8m.so` is a non-standard name for libpython, and propose here the `m` suffix is dropped by p4a.

I am very new to the p4a codebase, so please give me a heads-up for any other sections of the code that should be adjusted for this and I'm happy to round off this PR.

### DIFF
--- a/pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonUtil.java
+++ b/pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonUtil.java
@@ -43,7 +43,7 @@ public class PythonUtil {
         libsList.add("python3.5m");
         libsList.add("python3.6m");
         libsList.add("python3.7m");
-        libsList.add("python3.8m");
+        libsList.add("python3.8");
         libsList.add("main");
         return libsList;
     }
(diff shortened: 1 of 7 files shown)


inclement:
Thanks, and sorry for the delay getting back to you - it seems the p4a developers have been otherwise occupied recently. The basic change looks good to me, but I think it will also be necessary to update the reference to libpython3.8m in pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonUtil.java. That's the only other thing I can think of offhand though.

davidhewitt:
Thanks, I've fixed that and a few other places in the codebase which appeared relevant.

CI failures seem unrelated - download error is presumably spurious and the travis issue looks to be https://github.com/kivy/buildozer/issues/1144

davidhewitt:
ping @inclement is there anything I can do to help see this merged?

AndreMiras (approved):
LGTM thanks for looking into it, addressing inclement comments and squashing everything :muscle: 
Yeah dev have been pretty busy lately so thanks for your efforts and patience

PR #2279, opened 2020-08-04
fix travis

obfusk:


obfusk:
Ironically, one of the builds timed out downloading sqlite3. Seems to be fine otherwise.

obfusk:
>  Could you simply remove the debugging code (ls and which).

I left it in intentionally so it would be easier to fix if the version changes again. But I can remove it of course.

AndreMiras:
Yes I'd rather remove it indeed

obfusk:
Removed.

AndreMiras:
Re the macOS build, we may want to create a tech debt task and disable only the macOS build until this is fixed

PR #2282, opened 2020-08-05
update flask to 1.1.2

obfusk:


obfusk:
I just realized that the old version of `flask` did not yet depend on `click`.
Which means that anyone not pinning `flask` will now also get `click`.
And if `click` is not pinned to an older version, you run into #2264.

AndreMiras:
Thanks for the heads up

PR #2291, opened 2020-08-11
download_file: show error + exponential sleep

obfusk:
I was trying to fix the travis osx build, but the download failure seems to have been replaced by a different problem :(
Might still be a useful modification though.


AndreMiras:
Also I think this build error was fixed by your recent PR https://github.com/kivy/python-for-android/pull/2292 let's merge

obfusk:
> Also I think this build error was fixed by your recent PR #2292 let's merge

Yeah. The download failure seems to have disappeared on its own (for now). And the new problem was fixed by that PR.

PR #2299, opened 2020-08-14
Support Android resources similarly to assets

### LINKED ISSUE #2298: Add support for custom resources
There is currently no way to add resources to the Android resource folder, which prevents instantiating a Drawable (for instance to add a custom notification icon).

lerela:
Fixes #2298 

This PR allows adding a resource to the Android resources folder exactly as an asset. Difference with assets is that resources can subsequently be instantiated as Drawables:

```python
Drawable = jnius.autoclass("{}.R$drawable".format(__pkg_name__)) # __pkg_name__ is tld.org.app
custom_icon = getattr(Drawable, "custom_icon")
```

This is especially useful to pick a notification icon that respect the new Android requirements (PNG with white shape on transparent background) instead of using the app icon, which gets rendered as a white square on many recent devices. I could not find an easy way to do so with an asset.

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -232,6 +232,7 @@ main.py that loads it.''')
             sys.exit(1)
 
     assets_dir = "src/main/assets"
+    res_dir = "src/main/res"
 
     # Delete the old assets.
     shutil.rmtree(assets_dir, ignore_errors=True)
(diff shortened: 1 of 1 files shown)


pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -307,6 +308,14 @@ def make_package(args):
             else:
                 shutil.copytree(realpath(asset_src), join(assets_dir, asset_dest))
 
+        for res in args.resources:
+            res_src, res_dest = res.split(":")
+            if isfile(realpath(res_src)):
+                ensure_dir(dirname(join(res_dir, res_dest)))
```
lerela, line 314: This is obviously redundant with the assets copying logic just above, and could be factorized if needed

robertpfeiffer:
I think you forgot to expose this functionality in toolchain.py.

misl6:
Superseded by https://github.com/kivy/python-for-android/pull/2684

PR #2301, opened 2020-08-19
Add manifestPlaceholders

misl6:
This PR implements the `manifestPlaceholders` usage as described here: https://developer.android.com/studio/build/manifest-build-variables

### DIFF
--- a/pythonforandroid/bootstraps/common/build/templates/build.tmpl.gradle
+++ b/pythonforandroid/bootstraps/common/build/templates/build.tmpl.gradle
@@ -36,6 +36,7 @@ android {
         targetSdkVersion {{ android_api }}
         versionCode {{ args.numeric_version }}
         versionName '{{ args.version }}'
+        manifestPlaceholders = {{ args.manifest_placeholders}}
     }
 
 	{% if debug_build -%}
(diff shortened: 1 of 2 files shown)


pythonforandroid/bootstraps/common/build/templates/build.tmpl.gradle
```diff
@@ -36,6 +36,7 @@ android {
         targetSdkVersion {{ android_api }}
         versionCode {{ args.numeric_version }}
         versionName '{{ args.version }}'
+        manifestPlaceholders = {{ args.manifest_placeholders}}
```
AndreMiras, line 39: I didn't get a chance to check, but I assume this default behaviour you introduced doesn't break anything right?
Could you try to run the auto-compiled test apps on device and confirm it's not breaking?

AndreMiras (approved):
LGTM, nothing against it as long as it doesn't break default setup

pythonforandroid/bootstraps/common/build/templates/build.tmpl.gradle
misl6: It just adds an empty map, so it shouldn't break anything.
BTW, I just downloaded the artifact for arm64-v8a and it works flawlessly.

AndreMiras: Nice thanks for giving a try, let's merge :rocket:

PR #2312, opened 2020-09-11
Call cython via the setuptools entrypoint, fixes #2311

### LINKED ISSUE #2311: Cython is called in incompatible way, not all options are supported (especially -I)
### Checklist

- [X] the issue is indeed a bug and not a support request
- [X] issue doesn't already exist: https://github.com/kivy/python-for-android/issues
- [ ] I have a short, runnable example that reproduces the issue
- [X] I reproduced the problem with the latest development version (`p4a.branch = develop`)
- [X] I used the grave accent (aka backticks) to format code or logs when appropriated

### Versions

- Python: 3.7 (also tried with 3.8)
- OS: Ubuntu 20.04
- Kivy: 1.11.1
- Cython: git master HEAD, 0.29.21, 0.29.19
- OpenJDK: 8u265-b01-0ubuntu2~20.04

### Description

Compiling ZMQ recipe fails because cython is called in an incompatible way.

```sh
/home/rambo/.virtualenvs/py4a_37/bin/python3 -m Cython.Build.Cythonize --help
```

has different command line options from

```sh
/home/rambo/.virtualenvs/py4a_37/bin/cython --help
```

This (at least based on the recipe trying to pass -I options) is the correct script, unfortunately the `Cython.Compiler.Main` module cannot be called as a script (lack the `if __name__ == '__main__'` check so just changing the module in https://github.com/kivy/python-for-android/blob/develop/pythonforandroid/recipe.py#L1095 is not going to solve this.

### buildozer.spec

Using p4a directly

Command:
```sh
p4a apk --private `pwd` --package=fi.advian.hellokv --use-setup-py --name "Hello kivy" --version 0.1 --bootstrap=sdl2 --requirements=python3,kivy --ndk-api=21 --ndk-dir=$HOME/.buildozer/android/platform/android-ndk-r19c --sdk-dir=$HOME/.buildozer/android/platform/android-sdk/
```

### Logs

See https://gist.github.com/rambo/feab17c51a742b3d1f2239295298732b

rambo:
As per #2311, the. running `python -m Cython.Build.Cythonize` is **not** the same thing as running `cython`

### DIFF
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -1089,10 +1089,7 @@ class CythonRecipe(PythonRecipe):
             del cyenv['PYTHONPATH']
         if 'PYTHONNOUSERSITE' in cyenv:
             cyenv.pop('PYTHONNOUSERSITE')
-        python_command = sh.Command("python{}".format(
-            self.ctx.python_recipe.major_minor_version_string.split(".")[0]
-        ))
-        shprint(python_command, "-m", "Cython.Build.Cythonize",
+        shprint(sh.Command("cython"),
                 filename, *self.cython_args, _env=cyenv)
 
     def cythonize_build(self, env, build_dir="."):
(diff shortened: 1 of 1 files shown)


inclement (approved):
Seems reasonable, thanks! I won't merge just yet in case someone else has an opinion, but the tests have passed so it should be fine.

AndreMiras:
Thanks for the PR. I also have nothing against it and test suite is happy

AndreMiras (approved):

PR #2322, opened 2020-09-25
Add option for adding "uses-feature" tags to AndroidManifest.xml

### LINKED ISSUE #958: uses-feature tag in AndroidManifest file
Is it possible to add uses-feature tags in AndroidManifest file ?

i.e:  `<uses-feature android:name="android.hardware.usb.host" />   `

see [https://developer.android.com/guide/topics/connectivity/usb/host.html](https://developer.android.com/guide/topics/connectivity/usb/host.html)

rambo:
fixes #958

Couldn't find any ready made unit test to check for keys in AndroidManifest.xml (permissions for example could have
had one) and don't understand the code deeply enough (yet) to make one from scratch.


rambo:
I have no idea where that attempt to pass --permissions to setup.py in those failing recipe tests, nothing I've touched IMO

Julian-O:
See #958 for an update.

eMPee584:
The update is: merging this PR fixes https://github.com/kivy/buildozer/issues/1715 🎉

PR #2332, opened 2020-10-05
Add recipe for https://docs.aiohttp.org/en/stable/

rambo:
had this in private recipes but would probably be useful for larger audience.

Didn't find any example in tests how to test a CythonRecipe so no unit test :(


AndreMiras (approved):
LGTM, thanks

PR #2334, opened 2020-10-12
add android:usesCleartextTraffic="true" to sdl2

sharkguto:
add android:usesCleartextTraffic="true" to sdl2 for using webview with kivy

```python3

from kivy.app import App
from jnius import autoclass
from kivy.clock import Clock
from android.runnable import run_on_ui_thread
from kivy.uix.widget import Widget

WebView = autoclass("android.webkit.WebView")
WebSettings = autoclass("android.webkit.WebSettings")
WebViewClient = autoclass("android.webkit.WebViewClient")
activity = autoclass("org.kivy.android.PythonActivity").mActivity


@run_on_ui_thread
def create_webview(*args):
    webview = WebView(activity)
    webview.getSettings().setJavaScriptEnabled(True)
    webview.getSettings().setMixedContentMode(WebSettings.MIXED_CONTENT_ALWAYS_ALLOW)
    wvc = WebViewClient()
    webview.setWebViewClient(wvc)
    activity.setContentView(webview)
    webview.loadUrl("http://18.15.178.11:9999")


class Wv(Widget):
    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.__functionstable__ = {}
        Clock.schedule_once(create_webview, 0)


class ServiceApp(App):
    def build(self):
        return Wv()


if __name__ == "__main__":
    ServiceApp().run()

```

PR #2338, opened 2020-10-19
added few input parameters to make possible to use custom java classes and tweak AndroidManifest.xml

vesellov:


vesellov:
Hello guys.... this is the second part of the change I did a while ago in another PR : https://github.com/kivy/python-for-android/pull/2248

Now I am also adding option to specify custom PythonService.java class in case you need to tweak it. For example in my app I need to add additional code to `onDestroy()` method and that I am not able to do without redefining the java class...   once that is merged I will start already prepared PR for buildozer repo

vesellov:
Can someone please review? this is a very useful feature for those who requires deep customization of .java classes

vesellov:
I must also mention that input parameter `--activity-class-name` is already there, but it seems there is a better way to pass it further in the flow. So I am also changing that here.

pythonforandroid/bootstraps/sdl2/build/templates/AndroidManifest.tmpl.xml
```diff
@@ -59,8 +59,11 @@
                  android:icon="@drawable/icon"
                  android:allowBackup="{{ args.allow_backup }}"
                  {% if args.backup_rules %}android:fullBackupContent="@xml/{{ args.backup_rules }}"{% endif %}
+                 {% if args.network_security_config %}android:networkSecurityConfig="{{ args.network_security_config }}"{% endif %}
```
inclement, line 62: I'm wondering if it would be better to have a more generic "additional-application-xml" option by which these new arguments can be added, rather than a new argument for each option. This provides better support for people wanting to use new different arguments in the future, and doesn't clutter up the arguments list so much. What do you think?

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -803,6 +805,14 @@ def parse_args_and_make_package(args=None):
     ap.add_argument('--manifest-placeholders', dest='manifest_placeholders',
                     default='[:]', help=('Inject build variables into the manifest '
                                          'via the manifestPlaceholders property'))
+    ap.add_argument('--network-security-config', dest='network_security_config', default='',
+                    help='Add a Network Security Configuration file path to AndroidManifest.xml')
+    ap.add_argument('--uses-cleartext-traffic', dest='uses_cleartext_traffic', default='',
+                    help='Indicate that app intends to use cleartext network traffic in AndroidManifest.xml')
+    ap.add_argument('--service-class-name', dest='service_class_name', default='org.kivy.android.PythonService',
```
inclement, line 812: I feel like the `'org.kivy.android.PythonService'` string should probably be pulled out to a named constant at the top of the file, and the same for the activity.

inclement (commented):
I think this is fine with me - the reason this is not configurable is more that nobody has cared than that there's any reason not to allow it, if it's useful, and it doesn't seem to add too much extra to think about.

pythonforandroid/bootstraps/sdl2/build/templates/AndroidManifest.tmpl.xml
vesellov: fully agree, @inclement ! 
for example: https://github.com/kivy/python-for-android/pull/2334/files

What if "additional-application-xml" input parameter would be a file path with an extra XML code to be inserted inside `<application ... >` tag?    because I feel like passing a raw XML code inside the parameter is not a good idea...   what do  you think?

pythonforandroid/bootstraps/common/build/build.py
vesellov: good point!

pythonforandroid/bootstraps/sdl2/build/templates/AndroidManifest.tmpl.xml
vesellov: okay... as we discussed in Discord chat I will add a normal string parameter `--extra-manifest-application-arguments` that will be injecting value : `<application {{ args.extra_manifest_application_arguments }} >

pythonforandroid/bootstraps/common/build/build.py
vesellov: done

inclement (approved):
Looks good to me, thanks, and sorry again for the delay approving.

vesellov:
Thanks for help @inclement ! :+1:

PR #2344, opened 2020-10-25
Changes needed to build jpype recipe

tshirtman:


pythonforandroid/recipe.py
```diff
@@ -771,6 +771,7 @@ def should_build(self, arch):
 
         for lib in self.generated_libraries:
             if not exists(join(lib_dir, lib)):
+                print("** Missing library", join(lib_dir, lib))
```
inclement, line 774: Could we log this rather than print?

pythonforandroid/recipes/sdl2/__init__.py
```diff
@@ -29,5 +30,9 @@ def build_arch(self, arch):
                 _env=env
             )
 
+    def should_build(self, arch):
+        return not exists(join(self.ctx.bootstrap.build_dir, "libs", arch.arch, 'libSDL2.so'))
```
inclement, line 34: This will make some people happy!

pythonforandroid/recipe.py
```diff
@@ -820,8 +821,17 @@ class PythonRecipe(Recipe):
     This is almost always what you want to do.'''
 
     setup_extra_args = []
+    '''List of extra arguments to pass to setup.py for all targets'''
+
+    setup_global_extra_args = []
     '''List of extra arguments to pass to setup.py'''
```
inclement, line 827: Not sure I follow here, what is the difference between extra_args and global_extra_args? And why is the one with "global" in the name not the one documented as being passed to all targets?

Thrameos: I believe there was a problem with ordering.   You have two possible positions for extra arguments which mean different things.   

Consider

```
python setup.py --enable-tracing build_ext
```

and 

```
python setup.py build_ext --makefile
```

The first option is a global argument which applies to ever possible target.  The second is an option that is passed to the build_ext script specifically.   This is a problem for JPype is it used both types and some like install wont accept the target specific options.   There are specific options for build_ext such as --makefile, --jar and others that apply to all targets such as install, build, bdist such as --enable-tracing, --enable-coverage, etc.  

The issue is that we need both "target" args (extra_args) and extra args that apply to the "global" spot (global_extra_args).  If you have a better way to represent it I can change.  

Unfortunately as you are probably aware the python setup.py argument system is so poorly documented that most authors don't know how to define either global or target args.  It really should be simple to place args in the setup for the global and add target args in the right place.  But instead you have to define the args in one location, then write methods to handle them in another.  (I would have just made some decorators to make the task easy).  It took me a long time and many hours of puzzling before I realized what they were aiming to do with the position of args (and how to make the script actually recognize them.)

pythonforandroid/recipe.py
Thrameos: Sure.  I wasn't sure how the log system worked.

PR #2346, opened 2020-11-03
Create corna informer.py

bhavya2007:


tshirtman:
This has nothing to do with python for android, good on you for coding stuff, but create your own project if you want to share them, don't pollute unrelated existing ones.

PR #2351, opened 2020-11-09
added missing digest support to recipes

fuzzyTew:
This adds support for all the hash fragment types pypi may use: https://pip.pypa.io/en/stable/reference/pip_install/#hashes-from-pypi .

Allowing cryptographically secure hashes can help system administrators defend against compromise.

Documentation is only added for sha512sum and blake2bsum to maintain backwards compatibility in the interface and stay somewhat concise.  These two algorithms are rated highly.

### DIFF
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -59,6 +59,18 @@ class Recipe(with_metaclass(RecipeMeta)):
     finished correctly.
     '''
 
+    sha512sum = None
+    '''The sha512sum of the source from the :attr:`url`. Non-essential, but
+    you should try to include this, it is used to check that the download
+    finished correctly.
+    '''
+
+    blake2bsum = None
+    '''The blake2bsum of the source from the :attr:`url`. Non-essential, but
+    you should try to include this, it is used to check that the download
+    finished correctly.
+    '''
+
     depends = []
     '''A list containing the names of any recipes that this recipe depends on.
     '''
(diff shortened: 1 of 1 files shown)


AndreMiras (approved):
Very nice!
Looking good to me, thanks :pray: 
Couple of notes for a potential follow up PR.

- add tests to the `TestRecipe` class of [`tests/test_recipe.py`](https://github.com/kivy/python-for-android/blob/0039087/tests/test_recipe.py)
- fix a couple of DRY issues

fuzzyTew:
Thanks so much.

Additional note: It turns out pypi is recommending sha256, and also providing md5 and blake2-256.  Should probably change the 3 algorithm choices here to match those.

pythonforandroid/recipe.py
```diff
@@ -342,16 +354,19 @@ def download(self):
...
-                raise ValueError(
-                    ('Received md5sum from both the {} recipe '
-                     'and its url').format(self.name))
-            url = ma.group(1)
-            expected_md5 = ma.group(2)
-        else:
-            expected_md5 = self.md5sum
+        expected_digests = {}
+        for alg in set(hashlib.algorithms_guaranteed) | set(('md5', 'sha512', 'blake2b')):
+            expected_digest = getattr(self, alg + 'sum') if hasattr(self, alg + 'sum') else None
```
obfusk, line 359: @fuzzyTew fyi you could shorten

```py
getattr(self, alg + 'sum') if hasattr(self, alg + 'sum') else None
```

to

```py
getattr(self, alg + 'sum', None)
```

xloem: Thanks, I didn't see that.  If you have any interest / effort, I may have trouble getting to this for a while.

The biggest issue with my code is that I provided no way to set the digest length, so you can't do 256-bit blake2 which pypi uses.

Might be reasonable also to add a note regarding other security norms of the system not yet being as tight as the new algorithms.

PR #2363, opened 2020-11-29
Pyzint and Tesserocr Recipes

xloem:
I'm actually submitting these together so I can reference the branch on f-droid to publish a prototype app.  Let me know if they need to be broken up.

### DIFF
--- a/pythonforandroid/recipes/libleptonica/__init__.py
+++ b/pythonforandroid/recipes/libleptonica/__init__.py
@@ -1,0 +1,49 @@
+from pythonforandroid.recipe import Recipe
+from pythonforandroid.logger import shprint
+from pythonforandroid.util import current_directory
+from multiprocessing import cpu_count
+import os
+import sh
+
+
+class LibLeptonicaRecipe(Recipe):
+    version = '1.80.0'
+    url = 'https://github.com/DanBloomberg/leptonica/releases/download/{version}/leptonica-{version}.tar.gz'
+    md5sum = 'd640d684234442a84c9e8902f0b3ff36'
+    sha256sum = 'ec9c46c2aefbb960fb6a6b7f800fe39de48343437b6ce08e30a8d9688ed14ba4'
+
+    depends = ['png', 'jpeg']
+    built_libraries = {'libleptonica.so': os.path.join('install', 'lib')}
+
+    def build_arch(self, arch):
+        env = self.get_recipe_env(arch)
+
+        source_dir = self.get_build_dir(arch.arch)
+        build_dir = os.path.join(source_dir, 'build')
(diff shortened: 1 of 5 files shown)


xloem:
note: re tesseract, the [android mobile vision api](https://developers.google.com/vision/) could be wrapped in plyer or somesuch, and in discord somebody mentioned using [android tess-two](https://github.com/alexcohn/tess-two) but [with python](https://www.programmersought.com/article/80571840005/)

AndreMiras (commented):
Thanks for sharing the effort and well done for publishing on f-droid.
It might be a good idea indeed to try to split the PR and see if we can get the CI to build.
I would suggest the following PRs:

1. libzbar & zbarlight
2. tesseract related, potentially broken down in two PR, (libs first)

pythonforandroid/recipes/libzbar/__init__.py
```diff
@@ -8,13 +8,14 @@
 
 class LibZBarRecipe(Recipe):
 
-    version = '0.10'
+    version = '0.23.1'
 
-    url = 'https://github.com/ZBar/ZBar/archive/{version}.zip'
+    url = 'https://github.com/mchehab/zbar/archive/{version}.tar.gz'
```
AndreMiras, line 13: Why is that? It seems to be a less popular and "implicit" fork

xloem: hey, I don't have time to address most of these things right now, but https://github.com/ZBar/ZBar hasn't been updated since 2012.  https://github.com/mchehab/zbar is actively maintained.

AndreMiras: Thanks for the info, do we know if it's the new "official fork", is the API compatible with old version?

xloem: The API seems compatible (I didn't have to change any code to switch to it).  I don't see any evidence that the original maintainer knows it exists, but it's the version released in gentoo and mentioned on wikipedia.  It seems to be part of linuxtv.org .

ZBar 0.10 from 2012 had some bugs with binary data for me, so I use newer versions.

AndreMiras: Indeed, good point for Gentoo, I'm sold :smile: 
https://packages.gentoo.org/packages/media-gfx/zbar
I think it deserves a comment for this URL explaining we are using this fork on purpose and don't want to roll back unless we have good reason. Because sometimes some recipes point to the wrong URL by mistake and someone may think it's the case for this one

xloem: Good plan.

PR #2372, opened 2020-12-15
Add service information for buildozer.spec

xloem:
Since this seems to be where the background service documentation is, I wanted to add how to set them up in buildozer.


AndreMiras (approved):
LGTM, thanks

PR #2374, opened 2020-12-19
Enable androidx

RobertFlatt:
Addresses https://github.com/kivy/python-for-android/issues/2020
AndroidX is a support library that replaces the depreciated Android support library.

This PR adds a build option `--enable-androidx` which (when used) adds a block to `build.tmpl.gradle`
When AndroidX is enabled, a specific AndroidX library can be added using the existing  `gradle_dependencies` option.
Some (or all?) of the latest packages in the AndroidX library depend on `api=28`  or greater.

There is an alternative implementation, setting the same two Gradle flags in a new file named `gradle.properties`. Because of the file management uncertainties introduced in the context of `p4a`, this author will not use this approach.

The previous support library is depreciated. There is an argument that AndroidX should be enabled by default. This would impact users who specify an existing support package in `gradle_dependencies`. And there may be some other unforeseen impact. Alternatively AndroidX could be enabled by `api` version, but this would not be transparent to users. I think AndroidX should be enabled by default, but not yet; so it is enabled by build command option. 

As an example user dependency consider https://github.com/kivy/python-for-android/pull/2200 , it depends on the Android support library but is agnostic to which one. However the user code required does depend on the chosen support library, the change is in a Java class name, and in the name of the `gradle_dependencies` package. 

The need for https://github.com/kivy/python-for-android/pull/2200 is wide spread; for example the `plyer` camera https://github.com/kivy/plyer/blob/master/plyer/platforms/android/camera.py uses a 'file url' which can no longer be used between Activities, so is not functional with api =29 or greater. To prevent user rework after https://github.com/kivy/python-for-android/pull/2200 is incorporated, I suggest that this PR be incorporated at the same time.

### DIFF
--- a/pythonforandroid/bootstraps/common/build/templates/build.tmpl.gradle
+++ b/pythonforandroid/bootstraps/common/build/templates/build.tmpl.gradle
@@ -96,6 +96,13 @@ android {
         }
     }
 
+    {% if args.enable_androidx %}
+    ext {
+        useAndroidX=true
+        enableJetifier=true
+    }
+    {% endif %}
+
 }
 
 dependencies {
--- a/doc/source/buildoptions.rst
+++ b/doc/source/buildoptions.rst
@@ -196,8 +196,7 @@ options (this list may not be exhaustive):
 - ``add-source``: Add a source directory to the app's Java code.
 - ``--compile-pyo``: Optimise .py files to .pyo.
 - ``--resource``: A key=value pair to add in the string.xml resource file.
-- ``--enable-androidx``: If the argument is included, the AndroidX support 
-library is enabled. 
+- ``--enable-androidx``: If the argument is included, the AndroidX support library is enabled. 
 
 
 Requirements blacklist (APK size optimization)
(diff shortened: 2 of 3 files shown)


pythonforandroid/bootstraps/common/build/templates/build.tmpl.gradle
```diff
@@ -96,6 +96,13 @@ android {
         }
     }
 
+    {% if args.enable_androidx %}
+    ext {
```
tshirtman, line 100: Would it be preferable to put the `if`/`endif` *inside* the `ext` block (which would exist, but empty by default, and this way, more options could be added later in that block if needed?

Just a thought, it's certainly not a problem and it can be changed later when we actually have more options to put in there.

tshirtman (approved):
Looks good to me.

RobertFlatt:
@tshirtman ext{} easy to do.

But wait.
A FileProvider works great using androidx (and I developed a content URI api I like), but....
While testing I discovered that pyjnius can't find android.permissions._WHATEVER_ , 
presumably the library not linked. An issue with this PR, I suspect.
Investigating....

RobertFlatt:
The first issue is the version of the Android Gradle plugin specified by p4a 3.1.4 does not support jetifier, so AndroidX compatibility is not implemented. A minimum of 3.2.1 is required, the current version is 4.0.1

Great release notes here https://developer.android.com/studio/releases/gradle-plugin#3-2-0

Investigating....

PR #2376, opened 2020-12-19
Fix master only merges

inclement:
Merge master into develop, to pull in the couple of PRs that have been made straight to master.

I know it's easy to miss, but PRs should go to develop. Not sure if we can protect master more, maybe non-core-devs can be blocked from making PRs to master.

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -485,6 +485,7 @@ main.py that loads it.''')
         "service_names": service_names,
         "android_api": android_api,
         "debug": "debug" in args.build_mode,
+        "native_services": args.native_services
     }
     if get_bootstrap_name() == "sdl2":
         render_args["url_scheme"] = url_scheme
@@ -643,6 +644,9 @@ tools directory of the Android SDK.
     ap.add_argument('--service', dest='services', action='append', default=[],
                     help='Declare a new service entrypoint: '
                          'NAME:PATH_TO_PY[:foreground]')
+    ap.add_argument('--native-service', dest='native_services', action='append', default=[],
+                    help='Declare a new native service: '
+                         'package.name.service')
     if get_bootstrap_name() != "service_only":
         ap.add_argument('--presplash', dest='presplash',
                         help=('A jpeg file to use as a screen while the '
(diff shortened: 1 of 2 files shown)


Fak3:
By the way, why the project needs separate develop branch? People are confused by that, as most of other projects use master branch to track the recent commits.

inclement:
It doesn't "need" it, it's just a convenient workflow. We roughly follow a [git flow](https://nvie.com/posts/a-successful-git-branching-model/)-type model.

Fak3:
I mean... What is exactly you find convenient in having develop branch instead of master?

inclement:
We're making the choice that master represents the latest stable/released state (notionally it will always be production-ready), while develop represents the unstable development state. The names are fairly arbitrary, we're going with the original git-flow names as linked above.

Fak3:
Hm, I thought that there are some practical reasons rather than historical

AndreMiras (approved):
LGTM thanks

AndreMiras:
It is a practical reason @Fak3. You can read the post inclement shared above https://nvie.com/posts/a-successful-git-branching-model/
It basically explain why this is a convenient model

Fak3:
> It is a practical reason @Fak3. You can read the post inclement shared above https://nvie.com/posts/a-successful-git-branching-model/
> It basically explain why this is a convenient model

What is "it"? I skimmed through this article and only left with impression that the naming is completely arbitrary. The kivy, plyer and dozens of other projects call the most recent branch as "master" instead of "develop". And use branch called "stable" or "vX.X" for latest stable code

inclement:
Thanks @AndreMiras 

@Fak3 The naming is indeed arbitrary, except insofar as we selected this choice because we like it.

xloem:
Thanks for merging this.

@Fak3 it sounds like you have personal experience that would give good reasons to do it another way.  The devs would need to know of these reasons to consider them.  It sounds like the lower the barriers to understanding the project are, the more small fixes can come in from other projects.  Changing things like this is usually something that would happen slowly with discussion, because many people have been working with it this way for a long time.

Fak3:
The reason I asked is because this issue revealed to me that I was not the only one surprised by branch naming. If this becomes a significant enough annoyance for new contributors, then it may be worth to consider aligning the branch naming with kivy and most other projects. Another related issue is that git, github and many other projects are changing the "master" to "main", which is covered here: https://github.com/github/renaming. So if the p4a project going to align the naming with the rest of the world, then i would suggest to use "main" for the most recent stuff and "stable" for latest stable release.

AndreMiras:
Thanks for explaining.
Using main and stable just feel as arbitrary to me. Both are documented workflows from different sources and we decided to chose Git-flow so far on p4a.
Regarding pull requesting on the correct branch this is done automatically on the default branch (which is `develop`) when you open a pull request unless someone changes it. It has happened a couple of times, but I don't feel it's something to worry about

PR #2378, opened 2020-12-24
address comments to #2351, bringing hashes in line with pypi

xloem:


obfusk:
> this is a draft

FYI: GitHub supports [draft pull requests](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-pull-requests#draft-pull-requests), so you could convert this PR to a draft.

xloem:
Thanks, I couldn't find that link for the life of me.

pythonforandroid/recipe.py
```diff
@@ -53,22 +53,21 @@ class Recipe(with_metaclass(RecipeMeta)):
     '''A string giving the version of the software the recipe describes,
     e.g. ``2.0.3`` or ``master``.'''
 
-    md5sum = None
-    '''The md5sum of the source from the :attr:`url`. Non-essential, but
-    you should try to include this, it is used to check that the download
-    finished correctly.
+    sha256sum = None
+    '''The sha256sum of the source from the :attr:`url`. As of 2020, pip
+    recommendes use of this hash.  You should try to include this.  It
```
AndreMiras, line 58: typo: recommends

pythonforandroid/recipe.py
```diff
@@ -1195,10 +1176,28 @@ def reduce_object_file_names(self, dirn):
             shprint(sh.mv, filen, join(file_dirname, parts[0] + '.so'))
 
 
-def algsum(alg, filen):
-    '''Calculate the digest of a file.
+def verify_algsum(recipe, algs, filen):
```
AndreMiras, line 1179: thanks for the DRY fix

tests/test_recipe.py
```diff
@@ -115,13 +115,23 @@ def test_download_url_not_set(self):
             mock.call('Skipping test_recipe download as no URL is set')]
 
     @staticmethod
-    def get_dummy_python_recipe_for_download_tests():
+    def get_dummy_python_recipe_for_download_tests(good_digests=[], bad_digests=[]):
```
AndreMiras, line 118: I'd rather avoid mutable default arguments.
https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments
I know it doesn't look as elegant, but to avoid potential debugging headaches I'd recommend doing this instead:
```python
def get_dummy_python_recipe_for_download_tests(good_digests=None, bad_digests=None):
    good_digests = good_digests or []
    bad_digests = bad_digests or []
```

tests/test_recipe.py
```diff
@@ -146,6 +156,26 @@ def test_download_url_is_set(self):
                 'https://www.python.org/ftp/python/3.7.4/Python-3.7.4.tgz')]
         assert m_touch.call_count == 1
 
+    def test_hashes(self):
+        """
+        Verifies digest hashes are enforced.
+        """
```
AndreMiras, line 162: Could be one-line docstring https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

tests/test_recipe.py
```diff
@@ -146,6 +156,26 @@ def test_download_url_is_set(self):
...
+                                   good_digests=digests)
+            recipe.ctx.setup_dirs(temp_dir)
+            recipe.download()
+            for bad_digest in digests:
+                good_digests = digests - set([bad_digest])
+                recipe, filename = self.get_dummy_python_recipe_for_download_tests(
+                                       good_digests=good_digests,
+                                       bad_digests=[bad_digest])
+                recipe.ctx.setup_dirs(temp_dir)
+                with self.assertRaises(ValueError):
```
AndreMiras, line 176: `pytest` helper is also an helpful alternative in this case.
```python
with pytest.raises(ValueError, match="sum does not match expected"):
```

AndreMiras (commented):
Thanks for the follow up PR, it looks pretty good already!
I left a couple of suggestion, but nothing blocking really

PR #2380, opened 2020-12-26
Update __init__.py

kuzeyron:
Version bump for Kivy stable 2.0.0.


obfusk:
You're asking to merge this into `master` instead of `develop` (which is also why the CI fails). I suggest you change that :)

AndreMiras:
oh wait you're pull requesting to `master`, you should keep it to the default branch which is `develop`

PR #2385, opened 2020-12-29
Enable AndroidX

RobertFlatt:
Enable the use of the AndroidX support library, as the current Android support library used by p4a is depreciated.
Implemented as described here  https://developer.android.com/jetpack/androidx/migrate
No significant apk size change, until a `android.gradle_dependencies` is specified


inclement (approved):
Looks good, thanks.

It seems like it might be worth doing an explicit target api version check to help users avoid activating it accidentally, but I guess it doesn't really matter since older targets are generally not useful.

obfusk:
Given the multiple "trigger CI" commits it would have been nice if someone had squashed those before merging :sweat_smile:

PR #2386, opened 2020-12-30
webview: add enableZoom() method

obfusk:


AndreMiras (approved):
Nice, another undocumented feature :sweat_smile:

obfusk:
\o/

I'd be happy to document it :)
Where would be the best place to do that?

AndreMiras:
That's a great question actually :smile: 
We don't have yet specific section for Webview in the documentation, the only thing we have is:
https://python-for-android.readthedocs.io/en/latest/quickstart/#build-a-webview-application

PR #2387, opened 2020-12-30
webview: add mOpenExternalLinksInBrowser field

obfusk:


AndreMiras (approved):
Thanks, LGTM. One question, how do you configure it?
Where/when do you set the `mOpenExternalLinksInBrowser` static attribute to a different value when you want to change the behaviour?
Edit: nevermind, we can load the activity from the Python code and set it there, thanks!

obfusk:
>  Edit: nevermind, we can load the activity from the Python code and set it there, thanks!

Exactly. Doing that in `main.py` before starting the flask app works fine.

I should probably document that somewhere :sweat_smile:

PR #2400, opened 2021-01-04
don't run git pull when not on a branch

obfusk:
When using e.g. a `git+https` url and a `version` that refers to a tag (not a branch), p4a fails when the repo has already been cloned b/c it tries to run `git pull` (which requires a branch and fails with a detached HEAD). This fixes that.


AndreMiras (approved):
LGTM, thanks!

PR #2401, opened 2021-01-05
services: fix  START_STICKY 

mzakharo:
This patchet is fixing a Java stack-trace during a  restart of sticky service by the OS.

Specifically:
```E AndroidRuntime: Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'android.os.Bundle android.content.Intent.getExtras()' on a null object reference```

To reproduce,  one can force a system memory pressure event by allocating a bunch of ram in another process.

Sticky  background services + `ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS`  allows for creating resilient `background` services that get auto-restarted by Android OS under memory/resource pressure, and thus can outlast normal services created using the official [auto-restart documentation](https://python-for-android.readthedocs.io/en/latest/services/#service-auto-restart)

This patchset is inspired by  [this stackoverflow post](https://stackoverflow.com/questions/47362161/kivy-service-stops-when-app-is-closed)

### CODE REVIEW

File: pythonforandroid/bootstraps/common/build/templates/Service.tmpl.java [hunk 1]
```diff
@@ -34,6 +36,44 @@ static public void start(Context ctx, String pythonServiceArgument) {
         ctx.startService(intent);
     }
 
+    @Override
```
AndreMiras, line 39: Something feels off here, I'm not sure :thinking:
We're duplicating and changing logic of code from the `start()` method above, don't we?
So that I understand. This is when the service gets auto-restarted, it goes through the `onStartCommand()` with a null intent, hence we don't have access to the extra parameters. That's why you're preparing the `nativeStart()` parameters from here, currect?
In this case, do we ever need the logic we have in `start()` and `onStartCommand()`? I mean can't we always asume a null intent and have only one logic in one place?
Another thing I'm thinking we could investigate is the `START_REDELIVER_INTENT` flag maybe.

AndreMiras (commented):
Thanks for investigating and fixing. I left few comments/questions, but I think we're already close to getting something we can merge.
Another path that might be worth investigating is the `START_REDELIVER_INTENT` bit.
If I understood correctly, it may behave like the `START_STICKY`, but providing the intent this time. That would make the diff much shorter as just changing `START_STICKY` to `START_REDELIVER_INTENT`. That's something to investigate

pythonforandroid/bootstraps/common/build/templates/Service.tmpl.java
mzakharo: There is one more functionality that is delivered by this PR -> service will auto-restart and will not crash when  main acitivty(app) is closed.  `START_REDELIVER_INTENT` does not seem to support this functionality, so I am keeping `START_STICKY` for now. This PR is battle-tested and has been used for quite some time now, so I am hesitant to make major revisions, maybe soneone can look into `REDELIEVER_INTENT`  at some  later point. 

 I believe `intent` does not always have to be null, in case someone is passing parameters from the app.  I managed to clean up the `run()` a bit, but I am cautious to make any 'invasive' changes, since I am not sure I know of all the usecases this code is used for.

AndreMiras (approved):
Thanks for the clean up and the feedback on the comments.
I think we're OK. I'll leave it there a few days before merging in case someone (@inclement?) has something to say.
This part of the code is not heavily used, nor tested, it has few edge cases that're easy to overlook

mzakharo:
Turns out there are usecases where both foreground + STICKY is useful, in which case, the original CL was inusfficient.

 latest CL addresses  the foreground service support, ads a comment for why we dont turn off sticky services on app termination.

Also, this adds a new `getDefaultIntet` method, which removes logic duplication while starting a service from a custom broadcast receiver.
```java
        ix = ServiceMycustomservice.getDefaultIntent(context, "");
        ix.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
        context.startService(ix);
```

AndreMiras (approved):
Thanks for updating with latest develop change. I usually prefer a rebase tho.
I will squash & merge

PR #2404, opened 2021-01-07
PyZQM recipe needs setuptools, list it explicitly in deps

rambo:
It seems to be no longer automagically available at this build stage (maybe some order of operations got changed)

### DIFF
--- a/pythonforandroid/recipes/pyzmq/__init__.py
+++ b/pythonforandroid/recipes/pyzmq/__init__.py
@@ -13,7 +13,7 @@ class PyZMQRecipe(CythonRecipe):
     version = 'master'
     url = 'https://github.com/zeromq/pyzmq/archive/{version}.zip'
     site_packages_name = 'zmq'
-    depends = ['libzmq']
+    depends = ['setuptools', 'libzmq']
     cython_args = ['-Izmq/utils',
                    '-Izmq/backend/cython',
                    '-Izmq/devices']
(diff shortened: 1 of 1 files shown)


pythonforandroid/recipes/pyzmq/__init__.py
```diff
@@ -13,7 +13,7 @@ class PyZMQRecipe(CythonRecipe):
     version = 'master'
     url = 'https://github.com/zeromq/pyzmq/archive/{version}.zip'
```
AndreMiras, line 14: Could we take the opportunity to pin this version to last release so our builds are more reproducible?
```suggestion
    version = '20.0.0'
    https://github.com/zeromq/pyzmq/archive/v{version}.zip
```

pythonforandroid/recipes/pyzmq/__init__.py
```diff
@@ -13,7 +13,7 @@ class PyZMQRecipe(CythonRecipe):
...
     site_packages_name = 'zmq'
-    depends = ['libzmq']
+    depends = ['setuptools', 'libzmq']
```
AndreMiras, line 16: Probably became a hard dependency in recently released pyzmq v20.0.0

AndreMiras (dismissed):
Thanks for the update, I've added a small suggestion if you don't mind editing the PR?

pythonforandroid/recipes/pyzmq/__init__.py
rambo: Changed

AndreMiras (approved):
Thanks

PR #2406, opened 2021-01-11
Update quickstart.rst macOS brew cask command

yestoday-tv:
-brew cask install
+brew install --cask


AndreMiras (changes requested):
Thanks for the suggestion. We don't PR to master, but to develop

PR #2409, opened 2021-01-22
update setuptools

obfusk:


AndreMiras (approved):
Travis is also hanging here, but looking good to me, thanks

PR #2414, opened 2021-01-26
Avoids 'Permission Denied' error during docker build

thescheff:
During docker build ```docker build --tag=p4a --file Dockerfile .``` I got a **Permission Denied** error. Changing file permission solved the problem for me.


tshirtman:
merged from #2419

PR #2419, opened 2021-02-14
Avoids build error for opencv and bumps version to 4.5.1

thescheff:
Allows build with docker container. Without this changes, I could not build opencv with p4a.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -98,7 +98,7 @@ RUN mkdir ${ANDROID_HOME} && chown --recursive ${USER} ${HOME_DIR} ${ANDROID_HOM
 USER ${USER}
 
 # Download and install android's NDK/SDK
-COPY ci/makefiles/android.mk /tmp/android.mk
+COPY --chown=user:user ci/makefiles/android.mk /tmp/android.mk
 RUN make --file /tmp/android.mk \
     && sudo rm /tmp/android.mk
 
--- a/pythonforandroid/recipes/opencv/__init__.py
+++ b/pythonforandroid/recipes/opencv/__init__.py
@@ -13,7 +13,7 @@ class OpenCVRecipe(NDKRecipe):
         build of most of the libraries of the opencv's package, so we can
         process images, videos, objects, photos...
     '''
-    version = '4.0.1'
+    version = '4.5.1'
     url = 'https://github.com/opencv/opencv/archive/{version}.zip'
     depends = ['numpy']
     patches = ['patches/p4a_build.patch']
@@ -75,6 +75,8 @@ class OpenCVRecipe(NDKRecipe):
                     '-DANDROID_STANDALONE_TOOLCHAIN={}'.format(self.ctx.ndk_dir),
                     '-DANDROID_NATIVE_API_LEVEL={}'.format(self.ctx.ndk_api),
                     '-DANDROID_EXECUTABLE={}/tools/android'.format(env['ANDROID_SDK']),
+                    '-DANDROID_SDK_TOOLS_VERSION=6514223',
+                    '-DANDROID_PROJECTS_SUPPORT_GRADLE=ON',
 
                     '-DCMAKE_TOOLCHAIN_FILE={}'.format(
                         join(self.ctx.ndk_dir, 'build', 'cmake',
--- a/pythonforandroid/recipes/opencv_extras/__init__.py
+++ b/pythonforandroid/recipes/opencv_extras/__init__.py
@@ -15,7 +15,7 @@ class OpenCVExtrasRecipe(Recipe):
     .. seealso:: https://github.com/opencv/opencv_contrib
 
     """
-    version = '4.0.1'
+    version = '4.5.1'
     url = 'https://github.com/opencv/opencv_contrib/archive/{version}.zip'
     depends = ['opencv']
 


Dockerfile
```diff
@@ -98,7 +98,7 @@ RUN mkdir ${ANDROID_HOME} && chown --recursive ${USER} ${HOME_DIR} ${ANDROID_HOM
 USER ${USER}
 
 # Download and install android's NDK/SDK
-COPY ci/makefiles/android.mk /tmp/android.mk
+COPY --chown=user:user ci/makefiles/android.mk /tmp/android.mk
 RUN make --file /tmp/android.mk \
     && sudo rm /tmp/android.mk
```
AndreMiras, line 103: Not entirely sure why we `chown`-ed, but I guess that `sudo` here is no longer needed then

AndreMiras (approved):
Thanks for the update, it looks good to me.
Looks like this could solve https://github.com/kivy/buildozer/issues/1144 right?
I've made a tiny comment if you don't mind updating before we merge
Edit: note that Travis is failing with `error: Can not find Rust compiler`, but that was prior your PR

Dockerfile
thescheff: I ran into a `Permission Denied` error during the docker build. It originated from the make command. I don't like running to much commands in `sudo` so I tried to fix the rights first. After the change it worked.

I ran the Dockerbuild without `sudo` entirely and had no issues. So perhaps the adding of sudo right is entirely unnecessary?

AndreMiras: We should definitely not run the docker client with sudo indeed.
What I meant in my comment is regarding the `sudo` line within the container. Could we remove it since you chown user:user. So basically two lines after your change there's a `sudo rm` simply remove the sudo part

thescheff:
Yes [kivy/buildozer#1144 ](https://github.com/kivy/buildozer/issues/1144) gets the same error I did:

> CMake Error at cmake/android/OpenCVDetectAndroidSDK.cmake:176 (message):
  Android SDK Tools: OpenCV requires Android SDK Tools revision 14 or newer.

And the cmake arguments `-DANDROID_SDK_TOOLS_VERSION=6514223` and `-DANDROID_PROJECTS_SUPPORT_GRADLE=ON` are missing too.

I don't know how I could have created this error: `error: Can not find Rust compiler`. The error occurs before executing the test script. During the `python3.7 -m pip install pyOpenSSL` installation. That doesn't seem connected to these changes.

Dockerfile
thescheff: Yes I could even run dockerbuild & build a test apk with the following changes:
```
...
# with sudo access and no password 
# RUN usermod -append --groups sudo ${USER}
# RUN echo "%sudo ALL=(ALL) NOPASSWD: ALL" >> /etc/sudoers

...

# Download and install android's NDK/SDK
COPY --chown=user:user ci/makefiles/android.mk /tmp/android.mk
RUN make --file /tmp/android.mk \
    && rm /tmp/android.mk
...
```
I'm not sure about further consequences but for my limited test the user didn't even need to be in the `sudoers` group.

obfusk:
> I don't know how I could have created this error: `error: Can not find Rust compiler`. The error occurs before executing the test script. During the `python3.7 -m pip install pyOpenSSL` installation. That doesn't seem connected to these changes.

That should be fixed by #2422 (when it's merged).

ss1969:
> Allows build with docker container. Without this changes, I could not build opencv with p4a.
I tried this fix, and it works with P4A-develop / arm64-v8a / SDK27 / NDK21.
but it will crash into an "compileDebugJavaWithJavac FAILED" error with P4A-develop / armeabi-v7a / SDK24 / NDK21.
Can you show me your environment please?

thescheff:
> > Allows build with docker container. Without this changes, I could not build opencv with p4a.
> > I tried this fix, and it works with P4A-develop / arm64-v8a / SDK27 / NDK21.
> > but it will crash into an "compileDebugJavaWithJavac FAILED" error with P4A-develop / armeabi-v7a / SDK24 / NDK21.
> > Can you show me your environment please?

You used the default Dockerfile? I tested only with the "default" SDK/NDK versions. The following is what the Dockerfile installs (Source: android.mk):
```
ANDROID_NDK_VERSION ?= 19b
ANDROID_SDK_TOOLS_VERSION ?= 6514223
ANDROID_SDK_BUILD_TOOLS_VERSION ?= 29.0.3
ANDROID_HOME ?= $(HOME)/.android
ANDROID_API_LEVEL ?= 27
```
Perhaps this is the problem? 

As for my environment:

I build the Dockerfile from **thescheff:develop** added the following vars in the container:
```
echo "ANDROIDSDK="$HOME/.android/android-sdk"" >> $HOME/.bashrc
echo "ANDROIDNDK="$HOME/.android/android-ndk"" >> $HOME/.bashrc
echo "ANDROIDAPI=27" >> $HOME/.bashrc
echo "ANDROIDNDKVER=19b" >> $HOME/.bashrc
source $HOME/.bashrc
```

And ran the build:
```
p4a apk --private $HOME/myapp \
    --package=org.example.myapp \
    --name="My application" \
    --version=0.1 \
    --bootstrap=sdl2 \
    --requirements=python3,kivy,opencv \
    --blacklist-requirements=sqlite3,openssl \
    --permission CAMERA \
    --sdk-dir $ANDROIDSDK \
    --ndk-dir $ANDROIDNDK \
    --arch armeabi-v7a
```

PR #2424, opened 2021-02-17
webview: flush cookies & improve shouldOverrideUrlLoading

obfusk:


obfusk:
This should make cookies work properly (by default they're flushed to disk only every 5 -- or was it 10? -- minutes, so closing the app within that time results in unsaved cookies).

AndreMiras (approved):
LGTM

PR #2429, opened 2021-02-25
Update __init__.py

Neizvestnyj:
Update sdl2_ttf to version 2.0.15


AndreMiras (approved):
Judging from the changelog it shouldn't introduce retro compatibility issues, but I haven't tested myself.
https://www.libsdl.org/projects/SDL_ttf/
2.0.15:
 * Updated to FreeType version 2.9.1
 * Text rendering functions now use the alpha component of the text colors
 * Added support for characters greater than 0xFFFF (e.g. emoji) in the UTF-8 APIs

PR #2432, opened 2021-02-28
Reintroduce documentation of android module

tito:
Things about BroadcastReceiver, Runnable and onActivityResult
are really important for anybody that tries to use Android apis.
I've added back in api.rst, but if another place is preferred, just
tell me :)


AndreMiras (approved):
LGTM

PR #2445, opened 2021-04-04
Fix ImportError bug: cannot locate symbol "modf"

Neizvestnyj:
Fix this bug `ImportError: dlopen failed: cannot locate symbol "modf"`
https://github.com/kivy/buildozer/issues/1020

### DIFF
--- a/pythonforandroid/recipes/Pillow/__init__.py
+++ b/pythonforandroid/recipes/Pillow/__init__.py
@@ -91,7 +91,7 @@ class PillowRecipe(CompiledComponentsPythonRecipe):
             env['LDFLAGS'] += f' -L{join(webp_install, "lib")}'
         env['LDFLAGS'] += f' -L{ndk_lib_dir}'
         if cflags not in env['CFLAGS']:
-            env['CFLAGS'] += cflags
+            env['CFLAGS'] += cflags + " -lm"
         return env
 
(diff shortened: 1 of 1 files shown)


inclement:
Seems reasonable, thanks, but I think it should ideally be in LDFLAGS.

Neizvestnyj:
> Seems reasonable, thanks, but I think it should ideally be in LDFLAGS.

do you mean so?  `env ['LDFLAGS'] + = f ' -lm'`

inclement:
Yes, but no need to make this one a format string, skip the f.

That said, looking at the recipe it looks like it uses `env['LIBS']` for this a few lines above. Could you check adding -lm in that line works?

Neizvestnyj:
> Yes, but no need to make this one a format string, skip the f.
> 
> That said, looking at the recipe it looks like it uses `env['LIBS']` for this a few lines above. Could you check adding -lm in that line works?

Hmm and really, it should work, should I build a package using this recipe and then create a new pul request?

tshirtman:
you could update this pull request instead of creating a new one, but aside that, yes.

Neizvestnyj:
@tshirtman @inclement I did a test build, everything works

inclement (approved):
Nice, thanks.

PR #2446, opened 2021-04-07
android: add support for adaptive icon/launcher

SomberNight:
see https://developer.android.com/guide/practices/ui_guidelines/icon_design_adaptive

This adds a `--icon-fg` and a `--icon-bg` build option to the sdl2 bootstrap,
that can be used to specify two PNG files to be used as the foreground and background
of an adaptive icon. These will get used if API 26+ is available at runtime.
The existing `--icon` option to specify an icon as a single PNG file is kept, as
- this provides backwards compatibility,
- and even if the new options are used, it might be a good idea to set all three
  options: when running on API <26, the PNG specified via `--icon` gets used.


tshirtman (approved):

tshirtman:
Thanks!

PR #2449, opened 2021-04-21
update sqlite3

obfusk:


pythonforandroid/recipes/sqlite3/__init__.py
```diff
@@ -5,9 +5,9 @@
 
 
 class Sqlite3Recipe(NDKRecipe):
-    version = '3.34.1'
+    version = '3.35.5'
     # Don't forget to change the URL when changing the version
-    url = 'https://www.sqlite.org/2021/sqlite-amalgamation-3340100.zip'
+    url = 'https://www.sqlite.org/2021/sqlite-amalgamation-3350500.zip'
```
AndreMiras, line 10: I know it was already not working in the past, but ideally the `url` should be built from the `version`.
I think the PR is good enough, but we could follow up on that.
I think overriding the `versioned_url()` method should do the trick

AndreMiras (approved):
LGTM, waiting for the build to be green before merging

pythonforandroid/recipes/sqlite3/__init__.py
obfusk: The year in the url isn't something we can derive from the `version` though :(

```
https://www.sqlite.org/2021/sqlite-amalgamation-3350500.zip
                       ^^^^
```

AndreMiras: True that :disappointed:

PR #2452, opened 2021-05-04
fixed redirection for download liblzma

td1803:
liblzma was no longer loaded correctly. redirection not considered.

### DIFF
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -213,7 +213,9 @@ class Recipe(with_metaclass(RecipeMeta)):
             while True:
                 try:
                     # jqueryui.com returns a 403 w/ the default user agent
-                    url_opener.addheaders = [('User-agent', 'Mozilla/5.0')]
+                    # url_opener.addheaders = [('User-agent', 'Mozilla/5.0')]
+                    # Mozilla/5.0 doesnt handle redirection for liblzma
+                    url_opener.addheaders = [('User-agent', 'Chrome/59.0')]
                     urlretrieve(url, target, report_hook)
                 except OSError as e:
                     attempts += 1
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -213,9 +213,8 @@ class Recipe(with_metaclass(RecipeMeta)):
             while True:
                 try:
                     # jqueryui.com returns a 403 w/ the default user agent
-                    # url_opener.addheaders = [('User-agent', 'Mozilla/5.0')]
                     # Mozilla/5.0 doesnt handle redirection for liblzma
-                    url_opener.addheaders = [('User-agent', 'Chrome/59.0')]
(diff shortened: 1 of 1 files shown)


pythonforandroid/recipe.py
```diff
@@ -213,7 +213,9 @@ def report_hook(index, blksize, size):
             while True:
                 try:
                     # jqueryui.com returns a 403 w/ the default user agent
-                    url_opener.addheaders = [('User-agent', 'Mozilla/5.0')]
+                    # url_opener.addheaders = [('User-agent', 'Mozilla/5.0')]
```
AndreMiras, line 216: Why would we keep this line commented rather than completely removing it?

AndreMiras (commented):
Thanks for the PR :pray: 
But maybe then this would impact another package that doesn't handle it with Chrome.
I'm thinking maybe we should take another approach.
Either make this easily overridable per recipe, or hacking it around in liblzma recipe only.
Also what was the exact error you got with liblzma at Python level, was it a `OSError` or something else?
How about when we do it with `curl` what is returned (just for debugging purpose)

td1803:
The html of the Internet page is returned because the redirection does not work. Subsequently, unzip is not possible. The line url_opener.addheaders = [('User-agent', 'Mozilla/5.0')] is new in the develop branch compared to the master branch. It works with the master branch. The comment can of course be omitted.  Traugott DrossHabelschwerdter Allee 17 14195 Berlin Tel.: 0179 462 42 05Email.: ***@***.*** Von: Andre MirasGesendet: Mittwoch, 5. Mai 2021 08:27An: kivy/python-for-androidCc: td1803; AuthorBetreff: Re: [kivy/python-for-android] fixed redirection for download liblzma (#2452) @AndreMiras commented on this pull request.Thanks for the PR 🙏But maybe then this would impact another package that doesn't handle it with Chrome.I'm thinking maybe we should take another approach.Either make this easily overridable per recipe, or hacking it around in liblzma recipe only.Also what was the exact error you got with liblzma at Python level, was it a OSError or something else?How about when we do it with curl what is returned (just for debugging purpose)In pythonforandroid/recipe.py:> @@ -213,7 +213,9 @@ def report_hook(index, blksize, size):              while True:                  try:                      # jqueryui.com returns a 403 w/ the default user agent -                    url_opener.addheaders = [('User-agent', 'Mozilla/5.0')] +                    # url_opener.addheaders = [('User-agent', 'Mozilla/5.0')] Why would we keep this line commented rather than completely removing it?—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.

obfusk:
Setting the user agent to `Wget/1.0` seems to work for both jqueryui (the reason we override the default to begin with) and liblzma. It's not guaranteed to work for everything either but probably the easiest short-term fix.

tshirtman (approved):

AndreMiras (approved):
Yeah let's give it a try.
Thanks @tshirtman for the review

PR #2456, opened 2021-05-19
Creating new recipe for GitPython

FilipeMarch:
My Kivy application is using Github authentication and it is using a lot of [GitPython](https://github.com/gitpython-developers/GitPython/) functions. 
I am trying to create a recipe for it, but I'm struggling to understand the correct way to do this; I have no previous experience with python-for-android and I would really appreciate some help on this. 

```python
from pythonforandroid.recipe import CompiledComponentsPythonRecipe


class GitPythonRecipe(CompiledComponentsPythonRecipe):
    version = '20.1.0'
    url = 'git+https://github.com/gitpython-developers/GitPython'
    depends = ['gitdb', 'typing-extensions']
    build_cmd = 'build'


recipe = GitPythonRecipe()

### DIFF
--- a/pythonforandroid/recipes/GitPython/__init__.py
+++ b/pythonforandroid/recipes/GitPython/__init__.py
@@ -1,0 +1,12 @@
+from pythonforandroid.recipe import CompiledComponentsPythonRecipe
+
+
+class GitPythonRecipe(CompiledComponentsPythonRecipe):
+    version = '20.1.0'
+    url = 'git+https://github.com/gitpython-developers/GitPython'
+    depends = ['gitdb', 'typing-extensions']
+    call_hostpython_via_targetpython = False
+    build_cmd = 'build'
+
+
+recipe = GitPythonRecipe()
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by obfusk** [comment]
It looks like the latest version of GitPython is `3.1.17`, not `20.1.0` and `depends` should be `python_depends`.

**Comment by obfusk** [comment]
Its also better to use `https://github.com/gitpython-developers/GitPython/archive/{version}.zip` instead of a `git+https` url (unless you really need e.g. git submodules).

**Comment by obfusk** [comment]
I haven't tested it, but I think this should work:

```python
from pythonforandroid.recipe import CompiledComponentsPythonRecipe


class GitPythonRecipe(CompiledComponentsPythonRecipe):
    version = '3.1.17'
    url = 'https://github.com/gitpython-developers/GitPython/archive/{version}.zip'
    depends = ['setuptools']
    python_depends = ['gitdb', 'typing-extensions']


recipe = GitPythonRecipe()
```

obfusk:
I assumed it was using C modules since you used a `CompiledComponentsPythonRecipe`, but it doesn't.
That's easy to fix: just use a `PythonRecipe`.

But this is not going to work on Android:

> GitPython needs the git executable to be installed on the system and available in your PATH

misl6:
I agree with @obfusk.
Feel free to re-open the PR if you find a solution for the git executable.

PR #2462, opened 2021-06-11
Move PythonActivityUtil.unpackData to PythonUtil.unpackData

rnixx:
Consolidates unpackData for use outside activity context.


AndreMiras:
I haven't investigated why the tests are failing.
Did you look into it?

rnixx:
looks like travis reached some limit, all other tests are sane now

rnixx:
one question is whether we should remove PythonActivityUtil entirely since it's not used any more in bootstraps or keep for B/C in case someone depends on it in possible customizations.

pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonActivityUtil.java
```diff
@@ -1,110 +1,29 @@
...
+import java.io.File;
 
 import org.renpy.android.ResourceManager;
-import org.renpy.android.AssetExtract;
 
 
 public class PythonActivityUtil {
+
+    // XXX: This class is no longer used in any of p4a bootstraps.
+    //      Keep it here for B/C or remove?
```
AndreMiras, line 14: Yeah if it's not used I'd be tempted to delete indeed

AndreMiras (commented):
Thanks for the PR, LGTM.
I'd indeed remove the PythonActivityUtil if it's not used

rnixx:
Removed PythonActivityUtil. Two tests on travis are failing with docker pull rate limit, so probably unrelated. All other tests are fine.

AndreMiras (approved):
LGTM thanks
Ideally I like PR with squashed commits, but I think that's a matter of taste.
Will try to think to squash upon merge

PR #2470, opened 2021-07-08
Fix bad library found by ctypes for 64-bit arch (#2468)

macdems:
#2468 is a bug that should be addressed also in the `master` branch!


AndreMiras:
Already addressed via https://github.com/kivy/python-for-android/pull/2469
We usually pull request to develop and then use develop to release to master

PR #2436, opened 2021-03-05
setup.py: add *.diff to package_data

obfusk:
fixes #2435

PR #2471, opened 2021-07-15
Add *.diff to manifest and package_data

viblo:
In this https://github.com/kivy/python-for-android/pull/2436 diff files where added to the package_data in setup.py. However, that is not enough. This PR adds the *.diff pattern to a couple of more places needed.


AndreMiras (approved):
LGTM thanks

PR #2481, opened 2021-08-03
This is a recipe for Icecream

Arjun-Somvanshi:
Hey, This is a recipe for Icecream, would really help buildozer give a colored output, let me know if something else is needed. Thanks.

### DIFF
--- a/pythonforandroid/recipes/icecream/__init__.py
+++ b/pythonforandroid/recipes/icecream/__init__.py
@@ -1,0 +1,10 @@
+from pythonforandroid.recipe import PythonRecipe
+
+
+class IcecreamRecipe(PythonRecipe):
+    version = '2.1.1'
+    url = 'https://github.com/gruns/icecream/archive/refs/tags/v{version}.tar.gz'
+    depends = ['setuptools']
+
+
+recipe = IcecreamRecipe()
--- a/pythonforandroid/recipes/icecream/__init__.py
+++ b/pythonforandroid/recipes/icecream/__init__.py
@@ -5,6 +5,7 @@ class IcecreamRecipe(PythonRecipe):
     version = '2.1.1'
     url = 'https://github.com/gruns/icecream/archive/refs/tags/v{version}.tar.gz'
     depends = ['setuptools']
+    call_hostpython_via_targetpython = False
 
 
 recipe = IcecreamRecipe()


Arjun-Somvanshi:
hmm, any idea why it failed/?

Arjun-Somvanshi:
Ah! that seemed to work.

inclement:
This looks like a pure python module, does it actually need a recipe?

Arjun-Somvanshi:
> This looks like a pure python module, does it actually need a recipe?

It's better to have a recipe than including the repository in every project I want to Icecream. It increases the size of the apk too.

Arjun-Somvanshi:
[](https://github.com/mk-fg/pretty-yaml) pyaml seems to have a recipe though it's pure python. I think all external packages should be recipes instead of having users to include them in their projects, it not only will bloat the app, but makes the process of adding packages to your app more complex?

Kulothungan16:
Maybe you should take a look at this tshirman's message about this module https://discord.com/channels/423249981340778496/678695971164520523/843264213425324042

Arjun-Somvanshi:
> Maybe you should take a look at this @tshirtman's message about this module https://discord.com/channels/423249981340778496/678695971164520523/843264213425324042

Yes, as a part of this same thread it is suggested that either one has to clone this repo, and fix the bug as to why it doesn't compile to android or as Turtle suggested making a recipe for it should also solve the problem, and that is what I decided to do.

inclement:
> Yes, as a part of this same thread it is suggested that either one has to clone this repo, and fix the bug as to why it doesn't compile to android and use that in buildozer.spec or as Turtle suggested making a recipe for it should also solve the problem, and that is what I decided to do.

That's fine, but you didn't state this in the first place, it isn't obvious that icecream is broken and needs a trivial recipe to fix an issue with its build.

> pyaml seems to have a recipe though it's pure python. I think all external packages should be recipes instead of having users to include them in their projects,

You don't in general need to include modules manually if they don't have recipes. It sounds like even if icecream has a bug that is fixed by using a recipe, you have a misconception about this.

Arjun-Somvanshi:
Yes, I understand now what you mean now, I have had issues with a few other pure python packages too so this was a misconception. 
Is this pull request needing anything else? Would like someone to review it, if they have the time. :)

tshirtman:
Wouldn’t this recipe need a patch to be actually helpful? But wouldn’t it be better to actually do a pull request on icecream solving the issue, (using the change i suggested in the linked discord thread), so it would benefit not just p4a/buildozer users, but any user of the project that would hit this bug?

Issue #1860, opened 2019-06-11
[enhancement] do not rebuild already built packages


Fak3:
p4a version: git master
os: opensuse tumbleweed

Each time `p4a apk ...` is launched, it tries to rebuild following packages again: android, libffi, sqlite3, jni
While many othes report that its already built an skipped: 
```
[INFO]:    Building png for armeabi-v7a
[INFO]:    png said it is already built, skipping
```
It would be nice to avoid rebuilding long-compiling packages.

inclement:
I just checked, and `android`, `libffi` and `sqlite3` all correctly detected they were already built.

Also worth noting that even if the build is re-run, it isn't from scratch and will reuse the previous build (usually just re-running the last compilation step). If this doesn't appear to be happening, it would be useful to see logs to understand what p4a is doing.

Fak3:
Here is the full log: https://gist.github.com/Fak3/49d63e74d177f8e0d345d8ec99bb5983

Fak3:
jni looks like the most heavy one, it takes about 5 minutes to rebuild on my pc

Fak3:
i was able to prevent rebuilding of sdl2 (jni) by adding this to the sdl2 recipe:
```python
def should_build(self, arch):
        from os.path import exists, join
        libdir = join(
                self.get_build_dir(arch.arch),
                '../..',
                'libs', arch.arch
            )
        libs = 'libhidapi.so libmain.so libSDL2_image.so libSDL2_mixer.so libSDL2.so libSDL2_ttf.so'.split()
        return not all(exists(join(libdir, x)) for x in libs)

```

Julian-O:
Appears to be fixed. Closing.

If the problem persists, please let us know.

PR #2482, opened 2021-08-05
Add should_build method to sdl2 recipe

AndyRusso:
Hello. When making a distribution, sdl2 always builds all over and it takes a lot of time. This pull request changes it, by adding a should_build method to the sdl2 recipe.


AndyRusso:
#1860

dlech:
> When making a distribution, sdl2 always builds all over and it takes a lot of time.

Same here. Would be helpful to get this merged.

PR #2487, opened 2021-09-04
Drop travis-ci.org and use github-actions for pypi release

misl6:
We previously relied on `travis-ci.org` for builds. Now, `travis-ci.org` has been shut down: (https://blog.travis-ci.com/2021-05-07-orgshutdown)

This PR, removes `travis-ci.org` builds, and completely switches to `Github Actions`, also for pypi builds.

TODO:
- [x] Check for secrets


.github/workflows/pypi-release.yml
```diff
@@ -0,0 +1,25 @@
...
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade setuptools wheel twine
+    - name: Build
+      run: |
+        python setup.py sdist bdist_wheel
+        twine check dist/*
+    - name: Publish package
+      if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags')
+      uses: pypa/gh-action-pypi-publish@v1.1.0
```
AndreMiras, line 22: Minor: why not the latest version which is 1.4.2?
There're interesting features along the way like `twine check` running automatically (in 1.2.0), verbose settings for upload debugging (1.4.0) and few bug fixes

AndreMiras (dismissed):
Looking good, thanks!

.github/workflows/pypi-release.yml
misl6: No specific reason for that, just copy-pasted the same version We're using in `kivy-ios` :)
Updated to `v1.4.2`.

PR #2499, opened 2021-10-05
change NDK links go to v19c

leha-code:


doc/source/quickstart.rst
```diff
@@ -122,7 +122,7 @@ the SDK packages needed.
 For Android NDK, note that modern releases will only work on a 64-bit
 operating system. **The minimal, and recommended, NDK version to use is r19b:**
 
- - `Go to ndk downloads page <https://developer.android.com/ndk/downloads/>`_
+ - `Go to ndk downloads page <https://github.com/android/ndk/wiki/Unsupported-Downloads#r19c>`_
```
AndreMiras, line 125: Thanks for the pull request.
I think we also have one more reference above from the same documentation

leha-code: That's just about that you should use any version but r19c is maximum

misl6:
NDK 23b is now the default supported version, I'm closing this PR as is now outdated.

Thank you for taking care of `python-for-android`!

PR #2504, opened 2021-10-12
Makes pep8 happy on sdl2 recipe

misl6:
Fixes pep8.
Issue introduced in #2482 .


misl6:
> Thanks, not sure how I didn't see that in the PR itself 🤔

If you take a closer look, seems that the CI never started 😕 , so it makes sense that looked great for you.

Issue #2497, opened 2021-09-26
Pillow's recipe problem in aab generation


mmsprojects:
Hi, in first I'd thank you all for the development of this tool and @misl6 for the recent implementation about .aab package! I tried and had success building the aab file from @misl6 p4a's and buildozer's branch. 

But I only could get success by removing the Pillow from the requirements of `buildozer.spec`. I thing there's something with Pillow recipe or I'm doing wrong the implementation. I'm passed the bug related by @Guhan-SenSam about opencv. 

I'm using Kivymd and it kind of depends of pillow. 
''''
**[INFO]:    <- directory context /home/ros/APP_MD_ADS/python-for-android
[INFO]:    Building Pillow for armeabi-v7a
[INFO]:    Pillow apparently isn't already in site-packages
[INFO]:    Building compiled components in Pillow
[WARNING]: ndk_platform is deprecated and should be avoided in new recipes
[INFO]:    -> directory context /home/ros/APP_MD_ADS/.buildozer/android/platform/build-arm64-v8a_armeabi-v7a/build/other_builds/Pillow/armeabi-v7a__ndk_target_21/Pillow
[DEBUG]:   -> running python3 setup.py build_ext -v
[DEBUG]:        running build_ext
[DEBUG]:        building 'PIL._imaging' extension
[DEBUG]:        creating build
[DEBUG]:        building 'PIL._imagingft' extension
[DEBUG]:        building 'PIL._imagingmath' extension
[DEBUG]:        building 'PIL._imagingmorph' extension
[DEBUG]:        creating build/temp.linux-x86_64-3.8
[DEBUG]:        creating build/temp.linux-x86_64-3.8
[DEBUG]:        creating build/temp.linux-x86_64-3.8
[DEBUG]:        creating build/temp.linux-x86_64-3.8/src
[DEBUG]:        creating build/temp.linux-x86_64-3.8/src
[...]

PR #2516, opened 2021-11-01
Bump Pillow version to 8.4.0

misl6:
That PR started with the intent to only fix issue #2497, but finished bumping the Pillow version for convenience.
Is marked as WIP, cause:

- [ ] Needs some runtime testing


pythonforandroid/recipes/Pillow/patches/fix-setup.patch
```diff
@@ -1,135 +1,196 @@
---- Pillow-7.0.0/setup.py.orig	2020-01-02 06:19:26.000000000 +0100
```
AndreMiras, line 1: Always fun to read the diff of a patch :smile:

AndreMiras (commented):
looks good to me already, thanks.
I see the title is still WIP, ping me if you need another :eyes: later

aleonov224:
Verified for arm64-v8a, armeabi-v7a. Two AAB application used KivyMD 0.104.2 successfully uploaded to Google Play market. Both are worked.

PR #2520, opened 2021-11-14
Updated version of pygame from 2.0.1 to 2.1.0

CAPTAIN1947:
Changed the version of pygame in the recipe from 2.0.1 (released in Dec 2020) to the latest release 2.1.0 (released in Nov 2021).

### DIFF
--- a/pythonforandroid/recipes/pygame/__init__.py
+++ b/pythonforandroid/recipes/pygame/__init__.py
@@ -13,7 +13,7 @@ class Pygame2Recipe(CompiledComponentsPythonRecipe):
         not part of the build. It's usable, but not complete.
     """
 
-    version = '2.0.1'
+    version = '2.1.0'
     url = 'https://github.com/pygame/pygame/archive/{version}.tar.gz'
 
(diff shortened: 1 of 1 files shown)


misl6:
@CAPTAIN1947
Did you had the chance to test it on runtime?

CAPTAIN1947:
> @CAPTAIN1947 Did you had the chance to test it on runtime?

@misl6 
Yes I tried a few games and examples and they were working fine with 2.1.0.
But I only tested by specifically specifying the version 2.1.0 in the buildozer.spec.
I also took permission from the pygame devs and they said that its ok as 2.1.0 has a lot of changes (the support for python2 and sdl1 is dropped) and it's good if the version is updated.

misl6 (approved):
Thank you!

I didn't have the chance to test it on runtime but looks good, and you posted some additional useful info.

I don't immediately merge it, but I will schedule a follow-up for the next week, so if meanwhile someone shows up and has an additional test case, we can hear another opinion about it.

CAPTAIN1947:
Thanks. :)

PR #2525, opened 2021-11-28
Fix build

correa:
Build fails with AAPT: error: resource drawable/icon not found. Since it moved to mipmap. Update AndroidManifest template with the new path


AndreMiras (approved):
LGTM, related to https://github.com/kivy/python-for-android/pull/2446

PR #2537, opened 2022-01-02
Dockerfile: set ANDROIDSDK and ANDROIDNDK env vars

maxjakob:
since they remain static for each container build.

This simplifies the usage of the docker container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -101,6 +101,8 @@ USER ${USER}
 COPY --chown=user:user ci/makefiles/android.mk /tmp/android.mk
 RUN make --file /tmp/android.mk \
     && sudo rm /tmp/android.mk
+ENV ANDROIDSDK="${ANDROID_HOME}/android-sdk" \
+    ANDROIDNDK="${ANDROID_HOME}/android-ndk"
 
 # install python-for-android from current branch
 COPY --chown=user:user Makefile README.md setup.py pythonforandroid/__init__.py ${WORK_DIR}/


maxjakob:
@AndreMiras Do you think this is useful? It made the container simpler to use for me because I didn't have to worry about these variable. If you don't think this is needed, no hard feelings, we can close this.

Dockerfile
```diff
@@ -101,6 +101,8 @@ USER ${USER}
 COPY --chown=user:user ci/makefiles/android.mk /tmp/android.mk
 RUN make --file /tmp/android.mk \
     && sudo rm /tmp/android.mk
+ENV ANDROIDSDK="${ANDROID_HOME}/android-sdk" \
+    ANDROIDNDK="${ANDROID_HOME}/android-ndk"
```
AndreMiras, line 105: This isn't documented, but I think `ANDROIDSDK` deprecated actually.
Maybe what you're looking for is `ANDROID_SDK_HOME` and `ANDROID_NDK_HOME`?
Also why not grouping higher up with the other environment variables?

AndreMiras (commented):
Thanks for the suggestion, I left a comment, but I've been out of the loop lately so I might be missing things.
Also to make sure I got the context right, you're a user of that container image that you're pulling from DockerHub and you'd like simplify your `docker run` workflow?
If you omit these currently during your `docker run` you get a p4a runtime error I imagine right?

maxjakob:
> Also to make sure I got the context right, you're a user of that container image that you're pulling from DockerHub and you'd like simplify your docker run workflow?

Exactly.

I used the container some years ago and found adding these useful. Very possible that the names have changed and I can't recall the specific error. Given these uncertainties, I'm closing this. Thanks for the feedback!

PR #2541, opened 2022-01-17
Update AndroidManifest.tmpl.xml

drahba:
If an activity, service, or broadcast receiver uses intent filters and doesn't have an explicitly-declared value for android:exported, your app can't be installed on a device that runs Android 12 or higher. (https://developer.android.com/about/versions/12/behavior-changes-12). I added three changes accordingly with the requirements.


misl6:
Superseded by #2551

PR #2544, opened 2022-01-21
refactoring code with For Else

zjzh:
refactoring code with For Else which is more pythonic, concise and efficient; how do you think this change which has practical value？

### DIFF
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -728,7 +728,7 @@ class Recipe(with_metaclass(RecipeMeta)):
             if recipe_file is not None:
                 break
 
-        if not recipe_file:
+        else:
             raise ValueError('Recipe does not exist: {}'.format(name))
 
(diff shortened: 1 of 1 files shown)


tshirtman (approved):

tshirtman:
While i think refactoring should be incidental, when needed, rather than for the sake of it, (or else we end up refactoring needlessly), this is indeed more natural, so i'm fine with it. :)

AndreMiras (approved):

misl6:
Ouch! That was merged into `master` instead of `develop`.

AndreMiras:
Ouch I missed that too :facepalm:  thanks for the heads up.
I imagine we could cherry pick back to develop to have them align for next release.
Or we could simply revert it or do we have other idea?

misl6:
It's a really minor cosmetic change, so doesn't really change anything for users that are targeting `master`.
As we did in past I think is great to cherry-pick it to develop, without reverting the master, then will be aligned during next release.

PR #2554, opened 2022-03-03
tflite-runtime recipe

RobertFlatt:
Also includes pybind11 recipe.

An example using this recipe and some documentation is here: https://github.com/Android-for-Python/c4k_tflite_example , this link also referenced in `tflite-runtime/__init__.py`

This recipe does not build on x86_64 so in this case the build runs but creates nothing, enabling ci testing (I hope). This is documented in  `tflite-runtime/__init__.py` . Though this is not ideal,  it is solely an emulator case and x86 is presumably a substitute.

The platform build dependency occurs because tflite implementation varies by platform. So a "tflite python api" for x86_64 may compile for some platform, but not another. The Android platform "tflite python api" depends on the RUY matrix multiplication library which does not currently compile for x86_64.

### DIFF
--- a/pythonforandroid/recipes/pybind11/__init__.py
+++ b/pythonforandroid/recipes/pybind11/__init__.py
@@ -9,6 +9,6 @@ class Pybind11Recipe(Recipe):
 
     def get_include_dir(self, arch):
         return join(self.get_build_dir(arch.arch), 'include')
-    
-    
+
+
 recipe = Pybind11Recipe()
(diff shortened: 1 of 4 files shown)


AndreMiras (commented):
LGTM, we simply need to get the version right, it shouldn't include the "v"

AndreMiras (approved):
looking good, thanks

RobertFlatt:
Thank you. Example now has local tflite recipe and its reference removed.

PR #2558, opened 2022-03-08
add recipe for freetype-py to not include the prebuilt .so for the wr…

domedave:
…ong architecture


AndreMiras (approved):
LGTM, thanks for keeping the CI green

PR #2560, opened 2022-03-08
Upgrading the flask version to avoid exception at the start of webview application

Prashanth-BC:
The depencies of flask on itsdangerous and itsdangerous is unable to find the name json hence the application fails to start with the following exception.
ImportError: cannot import name 'json' from itsdangerous
This issue needs to fixed by upgrading the version of the flask the recipie.
I have tested with the modified reciepe with the latest version of flask and found that application works after the fix.


AndreMiras (approved):
LGTM

PR #2316, opened 2020-09-16
Added a build.py argument to add arbitrary xml to the <application> section of AndroidManifest.xml

mzakharocsc:
This fasciliates BOOT_COMPLETED, as per https://github.com/kivy/kivy/wiki/Starting-Kivy-App-and-Service-on-bootup-on-Android, as well as  fascilitates api29 storage workaround (as mentioned on Discord `android-support` channel. https://discord.com/channels/423249981340778496/712344698559397895/755856361283387522)

Buildozer update will follow once this is merged in.

PR #2570, opened 2022-03-29
Add extra xml inside application element option

Craulback:


misl6:
Hi @Craulback !

Isn't that one the same of #2316?

Issue #2569, opened 2022-03-27
"Unit test apk" + "Unit test aab" + "Test updated recipes" test jobs should be run also on macOS (both Intel and Apple Silicon)


misl6:
### Checklist

- [ ] the issue is indeed a bug and not a support request
- [x] issue doesn't already exist: https://github.com/kivy/python-for-android/issues
- [ ] I have a short, runnable example that reproduces the issue
- [ ] I reproduced the problem with the latest development version (`p4a.branch = develop`)
- [ ] I used the grave accent (aka backticks) to format code or logs when appropriated

### Versions

- Python: *
- OS: macOS
- Kivy: *
- Cython: *
- OpenJDK: *

### Description

Currently, `Unit test apk`, `Unit test aab` and `Test updated recipes` CI jobs are only run on `ubuntu-latest`.
In order to enhance macOS support for `python-for-android`, these tests should also have a corresponding run on macOS.

misl6:
Fixed via #2574

PR #2566, opened 2022-03-26
[WIP] Add support for Python 3.10 and make it the default while building hostpython3 and python3

misl6:
- Adds support for Python 3.10 in `hostpython3` and `python3` recipes.
- Makes Python 3.10 the default version.

PR #2550, opened 2022-02-05
NDK 23 + Gradle 7 support

misl6:
This PR adds support for NDK 23 and updates the Gradle version to 7.
These are the first steps needed in order to support Apple Silicon M1 without the aid of Rosetta.

Is a WIP and needs some more (upcoming) love.

PR #2574, opened 2022-04-06
macOS CI: ADD APK, AAB & Updated Recipes build

misl6:
- Fixes issue #2569 
- macOS builds are expected to fail ( `hostpython3`, `python3` and `numpy` builds are failing on macOS >= 11) https://github.com/kivy/python-for-android/pull/2566 should fix `python3` and `hostpython3`, but `numpy` needs to be updated accordingly.
- Builds on Apple Silicon are expected to fail until #2550 is ready and merged)
- On macOS (Github hosted) runners `ANDROID_*` env vars override is needed as are already defined https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md#environment-variables-2 (Could be interesting to use the runner-environment contained version when we will migrate to a recent version of the NDK).

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -47,8 +47,8 @@ jobs:
         parallel: true
         flag-name: run-${{ matrix.os }}-${{ matrix.python-version }}
 
-  build_apk:
-    name: Unit test apk
+  ubuntu_build_apk:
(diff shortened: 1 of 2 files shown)


AndreMiras (approved):
Pretty cool thanks!

PR #2576, opened 2022-04-11
Remove trailing - in filename

HyTurtle:
Errant `-` removed from filename


misl6:
Duplicated of the more complete https://github.com/kivy/python-for-android/pull/2581/files

PR #2578, opened 2022-04-14
Bumps libffi to v3.4.2 + adds -fPIC on i686-linux-android

misl6:
- Bumps `libffi` from `v3.3` to `v3.4.2` ( `v3.3` was failing to build via Android-LLVM on macOS as `-flto` flag is present in `CFLAGS` by default (See: libffi/pull/590)
- Added missing `-fPIC` to `Archx86` (i686-linux-android)

~~⚠️ Do not merge it yet as it needs some tests during runtime.~~
✅  Tests passed during runtime on `on_device_unit_tests` (On a `arm64-v8a` device)

### DIFF
--- a/pythonforandroid/archs.py
+++ b/pythonforandroid/archs.py
@@ -264,6 +264,7 @@ class Archx86(Arch):
         '-mssse3',
         '-mfpmath=sse',
         '-m32',
+        '-fPIC',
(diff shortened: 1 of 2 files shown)


AndreMiras (approved):
LGTM thanks!

PR #2330, opened 2020-10-02
Add option for adding files to res/xml without touching manifest

rambo:
usable for example with intent-filters that need extra resources (like `android.hardware.usb.action.USB_DEVICE_ATTACHED`)

PR #2580, opened 2022-04-18
Adds an option to allow add resources to Android res/ directory

DexerBR:
Related to kivy/buildozer#1428

Reference: https://developer.android.com/guide/topics/resources/providing-resources

PRs it can cover (at least partially) :
#2330
#2200 
#1922 
#2299 
#2570 
#2322 

Maybe related issues:
#1964 
#2298 
#2614 

### Important notes:
- It is important that the user defined resource folder follows the file structure of the Android documentation linked above.
- If not followed, p4a will perform the copy/overwrite, according to the documentation:
   - Only folders will be copied to _res_/ directory.
   - Only files will be copied to each of the folders (existing or to be created) in the _res_/ directory.

### Copy/overwrite process steps:

- Clean android resources directory (_res_)
- Create default/required p4a res directories (I believe this still needs to be defined)
  - Current default _res_ directories:
    - _drawable_
    - _drawable-mdpi_
    - _drawable-hdpi_
    - _drawable-xhdpi_
    - _drawable-xxhdpi_
    - _layout_
    - _mipmap_
    - _mipmap-anydpi-v26_
    - _values_
    - _xml_

  - Currently in p4a, some of these directories are required by later code, but the directory is not created if it doesn't exist.
  - I'm not sure if they all need to stay or if it would be nice to include others by default (Although they are not needed by p4a at the moment).
- Copy default/required p4a files (currently, many of them are copied on demand along the code. It might be interesting to keep this in one place).
- Finally, copy directories and files (overwrite where applicable) according to user defined folder (via `add-res` argument) in res folder.

❗Note: Later arguments that change files in the resource folder, such as `icon_fg`, `icon_bg`, may override the `add-res` argument, if the user decides to put an icon in the "custom resource folder", and then pass another file to the `icon_fg` argument, for example. Changes in this behavior may also need to be evaluated.

Some things need tweaking, but overall, the templating process needs an update.

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -4,7 +4,7 @@ from gzip import GzipFile
 import hashlib
 import json
 from os.path import (
-    dirname, join, isfile, realpath,
+    dirname, join, isfile, isdir, realpath,
     relpath, split, exists, basename
 )
 from os import environ, listdir, makedirs, remove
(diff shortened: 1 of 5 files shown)


misl6:
Superseded by https://github.com/kivy/python-for-android/pull/2684

Thank you!

PR #2581, opened 2022-04-18
Remove stray - in output file name

dbnicholson:
The final component in the package name is just the extension, so
prefixing it with a - means the output file ends in `-.<ext>`.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -3,9 +3,9 @@ name: Unit tests & build apps
 on: ['push', 'pull_request']
 
 env:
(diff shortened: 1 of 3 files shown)


AndreMiras:
Hey @dbnicholson thanks for addressing that.
Do you mind rebasing and fixing the conflict.
```sh
git pull --rebase upstream/develop # given that the remote is called upstream
```
Sorry that we took so long to take a look as this ended up conflicting

dbnicholson:
Sorry for the delay. Rebased now. It's actually nicer since the github workflow has the artifact names defined in one place in environment variables.

AndreMiras (approved):
Nice thanks a mil!

PR #2582, opened 2022-04-18
Include HOME in build environment

dbnicholson:
Many of the tools run by p4a store intermediate files in the user's home
directory. Passing the HOME environment variable to the build
environment has 2 positive effects:

1. To completely encapsulate the build, HOME can be set to an alternate
   directory than the user's actual home directory. Many tools such as
   p4a have options to override these paths, but that must be done on a
   case by case basis and it would require that p4a pass through these
   options or environment variables.

2. In containerized environments the user may not be registered in the
   accounts database. If the user's home directory can't be found from
   the HOME environment variable or the accounts database, many tools
   will fail. An example of this is python2.7 when run by `ndk-build`.

I don't know if this is the best place to put this, but it seemed the most common location where the build environment is setup.

### DIFF
--- a/pythonforandroid/archs.py
+++ b/pythonforandroid/archs.py
@@ -109,6 +109,18 @@ class Arch:
     def get_env(self, with_flags_in_cc=True):
         env = {}
 
+        # HOME: User's home directory
+        #
+        # Many tools including p4a store outputs in the user's home
+        # directory. This is found from the HOME environment variable
+        # and falls back to the system account database. Setting HOME
+        # can be used to globally divert these tools to use a different
+        # path. Furthermore, in containerized environments the user may
+        # not exist in the account database, so if HOME isn't set than
+        # these tools will fail.
+        if 'HOME' in environ:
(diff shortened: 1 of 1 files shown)


dbnicholson:
Anything I can do to help this along? Would a test help?

misl6:
Hi @dbnicholson !

- It makes sense, and looks a great addition. ❤️ 
- Maybe the comment can better clarify (as you did on the PR description) why is that needed? (Just to help anyone looking at this line in future)
- A rebase so we can have a full ✅ CI run would be also nice.

Feel free to ping me as soon you're ready!

dbnicholson:
Ok, I'll beef up the comment.

dbnicholson:
Ready for review again. Thanks!

misl6:
> Ready for review again. Thanks!

Oh yeah, showed up on my notifications 4 minutes ago! 😄
The comment looks good!

Let's wait for the CI to complete 🏁

misl6 (approved):
LGTM. Thank you!

PR #2583, opened 2022-04-19
Set PATH using real SDK and NDK directories

dbnicholson:
This is a regression from f7f8cea636714dc4843b8818c553510f99084b6a.
Previously, `self.sdk_dir` and `self.ndk_dir` were passed to
`select_and_check_toolchain_version`.

### DIFF
--- a/tests/test_build.py
+++ b/tests/test_build.py
@@ -1,9 +1,13 @@
+import os
+import sys
 import unittest
 from unittest import mock
 
 import jinja2
 
-from pythonforandroid.build import run_pymodules_install
+from pythonforandroid.build import (
+    Context, RECOMMENDED_TARGET_API, run_pymodules_install,
+)
 from pythonforandroid.archs import ArchARMv7_a, ArchAarch_64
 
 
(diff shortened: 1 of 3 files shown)


misl6:
Ouch, I missed that! Agree about these changes.
What about line `374` ?
Maybe we should add a test or  a comment that cover this case?

dbnicholson:
> Ouch, I missed that! Agree about these changes. What about line `374` ? Maybe we should add a test or a comment that cover this case?

Whoops, missed that line as I was focused on the f-strings. Let me see if I can make a test for this.

dbnicholson:
Updated now with the missing line and a minimal test. It passed locally, but this is my first time in this test suite so I might not be doing the right thing.

misl6:
Unfortunately the tests are failing, can you take care of it?

dbnicholson:
Sure, I'll take a look on Monday.

dbnicholson:
Hmm, it passes when I run only `tests/test_recipe.py`. Some other test must be altering `PATH` without cleaning up.

dbnicholson:
Oh, I see what's going on now. This actually comes from f7f8cea636714dc4843b8818c553510f99084b6a, but my added test exercises it. Prior to that commit, `select_and_check_toolchain_version` would alter `os.environ`. After that commit, the `Context` class attribute `env` is altered.

Since the test I added runs `Context.prepare_build_environment`, the global `Context.env` class attribute is altered for any subsequent test. I can think of 3 possible fixes:

- Change `Context.prepare_build_environment` to alter `os.environ` like it used to.
- Change my new test to mock access to `Context.env` so `prepare_build_environment` doesn't affect subsequent tests
- Change `env` to be a `Context` instance attribute instead of a class attribute. I'd argue this is the most proper, but I don't know if there are other implications of it.
- Change `test_get_recipe_env_with` so it compares against `self.ctx.env['PATH']` instead of `os.environ['PATH']`. I think this change should be made regardless since that's actually what `Arch.get_env` does. Whether `self.ctx.env` and `os.environ` are the same is an orthogonal check.

dbnicholson:
I think this should be fixed now. I changed the new test so it patches `Context.env` and also corrected `test_get_recipe_env_with` so it compares against `ctx.env` instead of `os.environ` even though it's not needed with the patching.

tests/test_recipe.py
```diff
@@ -294,7 +294,7 @@ def test_get_recipe_env_with(
         #  check that the mocks have been called
         mock_ensure_dir.assert_called()
         mock_find_executable.assert_called_once_with(
-            expected_compiler, path=os.environ['PATH']
+            expected_compiler, path=self.ctx.env['PATH']
```
misl6, line 297: Good catch!

tests/test_build.py
```diff
@@ -89,3 +93,40 @@ def test_android_manifest_xml(self):
...
+            user_sdk_dir='sdk',
+            user_ndk_dir='ndk',
+            user_android_api=None,
+            user_ndk_api=None,
+        )
+
+        real_sdk_dir = os.path.join(os.getcwd(), 'sdk')
+        real_ndk_dir = os.path.join(os.getcwd(), 'ndk')
+        assert context.sdk_dir == real_sdk_dir
+        assert context.ndk_dir == real_ndk_dir
```
misl6, line 121: Super-minor, can we add a comment describing why is that needed?
Is very likely that in the future we will not immediately get why that was needed.

misl6 (approved):
LGTM, thank you!

I've just added a minor suggestion.
I will wait to merge, so you can address it, but not consider it as a blocking :)

tests/test_build.py
dbnicholson: Sure.

dbnicholson:
I added the comment, ready to review again.

misl6:
> I added the comment, ready to review again.

Thank you!

misl6 (approved):

PR #2588, opened 2022-04-24
Upload artifacts produced from every build platform, not only ubuntu-latest

misl6:
These changes allows us to test the artefacts produced from `ubuntu-latest`, `macos-latest` and `apple-silicon-m1`.

Builds made on different platforms may lead to a different result. (As an example, `numpy` seems to pass during build-time on `macos-latest`, but later it fails during on-device tests).

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -2,6 +2,10 @@ name: Unit tests & build apps
 
 on: ['push', 'pull_request']
 
+env:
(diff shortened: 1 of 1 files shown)


misl6:
(⚠️ CI is still expected to fail on `macos-latest` and `apple-silicon-m1`)

AndreMiras (approved):
Makes sense, thanks

PR #2590, opened 2022-05-01
Added py3dns recipe

Neizvestnyj:
I found an error in the `py3dns` module for android. And since this library is not on github and it has not been updated for more than 3 years, I decided to copy the library to my repository and fix the bug by adding a patch. This is a fairly popular library, used as a dependency for `validate_email`.

```py
from kivy.app import App
from kivy.uix.button import Button

import DNS
import smtplib


class TestApp(App):
    def build(self):
        return Button(text='Get DNS',
                      on_release=lambda *args: print(self.get_dns_host('gmail.com')),
                      )

    @staticmethod
    def get_dns_host(hostname: str, timeout=5):
        valid_host = None
        try:
            DNS.DiscoverNameServers()
            mx_hosts = DNS.mxlookup(hostname, timeout)
            for mx in mx_hosts:
                smtp = smtplib.SMTP()
                try:
                    smtp.connect(mx[1])
                except smtplib.SMTPConnectError:
                    continue  # try the next MX server in list
                else:
                    valid_host = mx[1]
        except Exception as err:
            print(err)

        return valid_host


TestApp().run()
```

### DIFF
--- a/pythonforandroid/recipes/py3dns/__init__.py
+++ b/pythonforandroid/recipes/py3dns/__init__.py
@@ -4,7 +4,7 @@ from pythonforandroid.recipe import PythonRecipe
 class Py3DNSRecipe(PythonRecipe):
     site_packages_name = 'DNS'
     version = '3.2.1'
-    url = 'https://github.com/Neizvestnyj/py3dns/archive/{version}.zip'
+    url = 'https://launchpad.net/py3dns/trunk/{version}/+download/py3dns-{version}.tar.gz'
     depends = ['setuptools']
     patches = ['patches/android.patch']
     call_hostpython_via_targetpython = False
--- a/pythonforandroid/recipes/py3dns/__init__.py
+++ b/pythonforandroid/recipes/py3dns/__init__.py
@@ -1,0 +1,13 @@
+from pythonforandroid.recipe import PythonRecipe
+
+
+class Py3DNSRecipe(PythonRecipe):
+    site_packages_name = 'DNS'
+    version = '3.2.1'
+    url = 'https://github.com/Neizvestnyj/py3dns/archive/{version}.zip'
+    depends = ['setuptools']
+    patches = ['patches/android.patch']
+    call_hostpython_via_targetpython = False
+
+
+recipe = Py3DNSRecipe()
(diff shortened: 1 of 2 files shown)


pythonforandroid/recipes/py3dns/__init__.py
```diff
@@ -0,0 +1,13 @@
+from pythonforandroid.recipe import PythonRecipe
+
+
+class Py3DNSRecipe(PythonRecipe):
+    site_packages_name = 'DNS'
+    version = '3.2.1'
+    url = 'https://github.com/Neizvestnyj/py3dns/archive/{version}.zip'
```
AndreMiras, line 7: How about using the official launchpad download link instead?
https://launchpad.net/py3dns/trunk/3.2.1/+download/py3dns-3.2.1.tar.gz
```suggestion
    url = 'https://launchpad.net/py3dns/trunk/{version}/+download/py3dns-{version}.tar.gz'
```

pythonforandroid/recipes/py3dns/patches/android.patch
```diff
@@ -0,0 +1,21 @@
...
++++ b/DNS/Base.py
+@@ -49,9 +49,13 @@ defaults= { 'protocol':'udp', 'port':53, 'opcode':Opcode.QUERY,
+             'server': [] }
+ 
+ def ParseResolvConf(resolv_path="/etc/resolv.conf"):
+-    "parses the /etc/resolv.conf file and sets defaults for name servers"
+-    with open(resolv_path, 'r') as stream:
+-        return ParseResolvConfFromIterable(stream)
++    try:
++        "parses the /etc/resolv.conf file and sets defaults for name servers"
```
AndreMiras, line 13: It doesn't matter too much, but docstring could stay where it was instead of being in the try/except block

AndreMiras (approved):
Great start, added a couple of minor comments.
Too bad the lib is not maintained if it's still being used.
By the way have you tried proposing your patch upstream?
Also have you tried to ask validate_email to dump/replace the dependency since it's unmaintained?

pythonforandroid/recipes/py3dns/__init__.py
Neizvestnyj: Hmm, well, I didn't think about it. [py3dns](https://pypi.org/project/py3dns/) works fine and does not require any changes yet. It's just that there is such a bug on android. The last `validate_email` release was in 2015 :)

pythonforandroid/recipes/py3dns/patches/android.patch
Neizvestnyj: done

AndreMiras (approved):
Thanks for the PR and follow up adjustments.
Too bad for validate_email not being too maintained either.
But overall looks good to me.
Will merge after the CI is done

PR #2593, opened 2022-05-13
Prefer avdmanager from cmdline-tools

dbnicholson:
The Android SDK tools are deprecated in favor of the command line tools.
Trying to use avdmanager from the deprecated tools fails on OpenJDK 11
with since it can't find the `javax/xml/bind/annotation/XmlSchema`
class. Try `cmdline-tools/latest/bin/avdmanager` first.

### DIFF
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -40,7 +40,11 @@ def get_ndk_sysroot(ndk_dir):
 
 
 def get_targets(sdk_dir):
-    if exists(join(sdk_dir, 'tools', 'bin', 'avdmanager')):
+    if exists(join(sdk_dir, 'cmdline-tools', 'latest', 'bin', 'avdmanager')):
+        avdmanager = sh.Command(join(sdk_dir, 'cmdline-tools', 'latest', 'bin', 'avdmanager'))
+        targets = avdmanager('list', 'target').stdout.decode('utf-8').split('\n')
+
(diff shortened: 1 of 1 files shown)


dbnicholson:
This could definitely be better by looking for other versions of `cmdline-tools` instead of just latest.

AndreMiras (approved):
LGTM thanks for the PR

HyTurtle:
Slight counterpoint if it hasn't been noticed; in buildozer (will need to find where) its cmdline-tools that are unzipped to the tools location, so may be a slight breaking change caused there.

dbnicholson:
I think that should be fine since it still falls back to that location. Our project has a similar non-standard setup that it promotes, but I think if you follow the way sdkmanager does the installs, this is how it should look.

gruve-p:
This seems to cause issues for several users:
https://stackoverflow.com/questions/75405703/python-for-android-attributeerror-avdmanager-bug-when-building-with-buildozer

https://github.com/Groestlcoin/electrum-grs/actions/runs/4142937521/jobs/7164201570

dbnicholson:
Looks like `sh.Command` broke it's API and started decoding the output. This PR didn't cause that.

dbnicholson:
Yep. https://github.com/amoffat/sh/blob/develop/MIGRATION.md#return-value-now-a-true-string. Probably want to pin that to 1.* for now.

PR #2594, opened 2022-05-14
Handle all the macOS prerequisites (except NDK/SDK) via prerequisites.py

misl6:
✂️ Partially sliced from #2586

- Is still experimental
- Moved `autoconf`, `automake`, `libtool`, `openssl` and `pkg-config`
- Added some generic tests and specific ones.
- Changed the logic around `mandatory` and `installer_is_supported`
- Added `SKIP_PREREQUISITES_CHECK` so we can skip prerequisites checks during recipe tests.
- brew-installable prerequisites can follow a DRY approach, but we should make sure that also doesn't over-complicate the `linux` implementation, so I'll leave the improvements for later.

### DIFF
--- a/pythonforandroid/prerequisites.py
+++ b/pythonforandroid/prerequisites.py
@@ -4,20 +4,20 @@ import sys
 import platform
 import os
 import subprocess
+import shutil
 from pythonforandroid.logger import info, warning, error
 
 
 class Prerequisite(object):
     name = "Default"
-    mandatory = True
-    darwin_installer_is_supported = False
-    linux_installer_is_supported = False
+    mandatory = dict(linux=False, darwin=False)
+    installer_is_supported = dict(linux=False, darwin=False)
 
     def is_valid(self):
         if self.checker():
             info(f"Prerequisite {self.name} is met")
             return (True, "")
-        elif not self.mandatory:
+        elif not self.mandatory[sys.platform]:
             warning(
                 f"Prerequisite {self.name} is not met, but is marked as non-mandatory"
             )
@@ -73,10 +73,7 @@ class Prerequisite(object):
             raise Exception("Unsupported platform")
 
     def install_is_supported(self):
-        if sys.platform == "darwin":
-            return self.darwin_installer_is_supported
-        elif sys.platform == "linux":
-            return self.linux_installer_is_supported
+        return self.installer_is_supported[sys.platform]
 
     def linux_checker(self):
         raise Exception(f"Unsupported prerequisite check on linux for {self.name}")
@@ -96,11 +93,42 @@ class Prerequisite(object):
     def linux_helper(self):
         info(f"No helper available for prerequisite: {self.name} on linux")
 
+    def _darwin_get_brew_formula_location_prefix(self, formula, installed=False):
+        opts = ["--installed"] if installed else []
+        p = subprocess.Popen(
+            ["brew", "--prefix", formula, *opts],
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+        _stdout_res, _stderr_res = p.communicate()
+
+        if p.returncode != 0:
+            error(_stderr_res.decode("utf-8").strip())
+            return None
+        else:
+            return _stdout_res.decode("utf-8").strip()
+
+
+class HomebrewPrerequisite(Prerequisite):
+    name = "homebrew"
+    mandatory = dict(linux=False, darwin=True)
+    installer_is_supported = dict(linux=False, darwin=False)
+
+    def darwin_checker(self):
+        return shutil.which("brew") is not None
+
+    def darwin_helper(self):
+        info(
+            "Installer for homebrew is not yet supported on macOS,"
+            "the nice news is that the installation process is easy!"
+            "See: https://brew.sh for further instructions."
+        )
+
 
 class JDKPrerequisite(Prerequisite):
     name = "JDK"
-    mandatory = True
-    darwin_installer_is_supported = True
+    mandatory = dict(linux=False, darwin=True)
+    installer_is_supported = dict(linux=False, darwin=True)
     min_supported_version = 11
 
     def darwin_checker(self):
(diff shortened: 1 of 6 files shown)


pythonforandroid/prerequisites.py
```diff
@@ -96,11 +93,45 @@ def darwin_helper(self):
...
+class HomebrewPrerequisite(Prerequisite):
+    name = "homebrew"
+    mandatory = dict(linux=False, darwin=True)
+    installer_is_supported = dict(linux=False, darwin=False)
+
+    def darwin_checker(self):
+        if shutil.which("brew"):
+            return True
+        else:
+            return False
```
AndreMiras, line 121:
```suggestion
        return shutil.which("brew")
```

pythonforandroid/prerequisites.py
```diff
@@ -216,13 +247,123 @@ def darwin_installer(self):
...
+class OpenSSLPrerequisite(Prerequisite):
+    name = "openssl@1.1"
+    mandatory = dict(linux=False, darwin=True)
+    installer_is_supported = dict(linux=False, darwin=True)
+
+    def darwin_checker(self):
+        if self._darwin_get_brew_formula_location_prefix("openssl@1.1", installed=True):
+            return True
+        else:
+            return False
```
AndreMiras, line 259: Why don't we inline that one too here and in the other places?
```suggestion
        return self._darwin_get_brew_formula_location_prefix("openssl@1.1", installed=True)
```

tests/test_prerequisites.py
```diff
@@ -0,0 +1,272 @@
...
+        self.mandatory = dict(linux=False, darwin=True)
+        self.installer_is_supported = dict(linux=False, darwin=False)
+        self.prerequisite = HomebrewPrerequisite()
+
+    def test_darwin_checker_is_supported(self):
+        try:
+            self.prerequisite.darwin_checker()
+        except Exception as e:
+            if e.__str__().startswith("Unsupported prerequisite check"):
+                self.fail("Darwin checker should be supported")
```
AndreMiras, line 61: Another option could be to leverage pytest skipif
```suggestion
    pytest.mark.skipif(sys.platform != "darwin", reason="MacOS test only")
    def test_darwin_checker_is_supported(self):
        self.prerequisite.darwin_checker()
```

tests/test_prerequisites.py
```diff
@@ -0,0 +1,272 @@
...
+    AutoconfPrerequisite,
+    AutomakePrerequisite,
+    LibtoolPrerequisite,
+    PkgConfigPrerequisite,
+    CmakePrerequisite,
+    get_required_prerequisites,
+)
+
+
+class PrerequisiteSetUpBaseClass(object):
```
AndreMiras, line 17: Object is already inherited in Python 3
```suggestion
class PrerequisiteSetUpBaseClass:
```

tests/test_prerequisites.py
```diff
@@ -0,0 +1,272 @@
...
+
+    @mock.patch("pythonforandroid.prerequisites.subprocess.check_output")
+    def test_darwin_installer(self, check_output):
+        self.prerequisite.darwin_installer()
+        check_output.assert_called_once_with(["brew", "install", "cmake"])
+
+
+class TestDefaultPrerequisitesCheckandInstall(unittest.TestCase):
+    def setUp(self):
+        pass
```
AndreMiras, line 244: This can probably go

AndreMiras (dismissed):
LGTM thanks :ok_hand: 
One question, is the install running by default then on macOS?
I'm wondering if we want to have this run on demand instead.
I imagine it depends on the users, but while I like things to be easy to get setup, I don't like when scripts install things system wide without my consent.
As an Arch and Gentoo user I may not be the norm here, I thing generally macOS users don't mind too much

tests/test_prerequisites.py
misl6: Agree. Old habits are hard to forget ;)

tests/test_prerequisites.py
misl6: Yeah, now is definitely useless.

tests/test_prerequisites.py
misl6: `pytest.mark.skipif(sys.platform != "darwin", reason="MacOS test only")` skips the test when not on `darwin`.

This one, instead, checks, if the `darwin_checker` is raising an error with `Unsupported prerequisite check` in it, as the `darwin_checker`, is marked as supported on this prerequisite.

AndreMiras: Oh right, what I don't get is why do we bother to try/except/if in this test, it's usually a red flag.
In this case why don't we simply:
```python
assert self.prerequisite.darwin_checker()
```

misl6:
> LGTM thanks 👌
> One question, is the install running by default then on macOS?
> I'm wondering if we want to have this run on demand instead.
> I imagine it depends on the users, but while I like things to be easy to get setup, I don't like when scripts install things system wide without my consent.
> As an Arch and Gentoo user I may not be the norm here, I thing generally macOS users don't mind too much

Here, we ask the user permission to install the prerequisite. We only skip the permission-checking when running non-interactively:
(As a macOS user which previously was on Linux, I'm also into the club  of users that hates when install-scripts do an installation without informing the user )

https://github.com/kivy/python-for-android/blob/eb343a59896b5277ddd6057c4db0ba6573a6a74b/pythonforandroid/prerequisites.py#L55

https://github.com/kivy/python-for-android/blob/eb343a59896b5277ddd6057c4db0ba6573a6a74b/pythonforandroid/prerequisites.py#L35-L51

tests/test_prerequisites.py
misl6: With `assert self.prerequisite.darwin_checker()`, we only check if it fails (which is probably enough ATM), not specifically if it throws a `Unsupported prerequisite check on macOS for ...` exception.

BTW, re-reading the code, considering that we're testing the darwin_checker function:
https://github.com/kivy/python-for-android/blob/eb343a59896b5277ddd6057c4db0ba6573a6a74b/tests/test_prerequisites.py#L63-L69
we can probably get rid of this test?

pythonforandroid/prerequisites.py
misl6: I'm expecting (here and in similar ones), to need a "clean" `boolean` in future, so returning the `sh.which("brew")` ouput which is a string doesn't seem a good idea.

How about:
```python
return shutil.which("brew") != ""
```

tests/test_prerequisites.py
AndreMiras: Ah yeah indeed, I leave that up to you either way is fine to me

pythonforandroid/prerequisites.py
AndreMiras: Oh yeah my bad, you're right.
Yeah in this case you can either:
```python
return shutil.which("brew") is not None
```
Or:
```python
return bool(shutil.which("brew"))
```

AndreMiras (approved):
Thanks for getting back.
And yes indeed for the interactive mode by default, we're all good.
LGTM, I'd love to see the `darwin_checker()` inlined, but it's probably matter of taste / nitpick, feel free to address or not

misl6:
> Thanks for getting back.
> And yes indeed for the interactive mode by default, we're all good.
> LGTM, I'd love to see the `darwin_checker()` inlined, but it's probably matter of taste / nitpick, feel free to address or not

Addressed. Thank you!

Issue #2533, opened 2021-12-27
jnius can't find class  "org.kivy.android.PythonActivity" with webview


foolish-suits:
### Checklist

- [x] the issue is indeed a bug and not a support request
- [x] issue doesn't already exist: https://github.com/kivy/python-for-android/issues
- [x] I have a short, runnable example that reproduces the issue
- [x] I reproduced the problem with the latest development version (`p4a.branch = develop`)
- [ ] I used the grave accent (aka backticks) to format code or logs when appropriated

### Versions

- Python: 3.8.10
- OS: Ubuntu:20.04
- Kivy: 2.0.0
- Cython: 0.29.15
- OpenJDK: java-13-openjdk-amd64

I followed the build env from the Docker in version v2021.09.05.

### Description

I want to use webview as bootstrap and use jnius to access Android API.
I tried the unit test code but build with webview as bootstrap. Build success, running and show the web well, but "vibrate" function failed with 'Didn't find class "org.kivy.android.PythonActivity"'
If I build with SDL2 the vibrate function will work well.

I have tried a lots but no luck, is there any demo or example can work as webview+jnius? Any advice are appreciate.
Thank you!

Command:
```
PYTHONPATH=. python3 -m pythonforandroid.toolchain apk  --private=/home/user/app/testapps/on_device_unit_tests/test_app --dist-name=testapp --package=org.tp4a --name=TestApp --version=0.1 --requirements=sqlite3,libffi,openssl,pyjnius,flask,python3,genericndkbuild --bootstrap=webview --ndk-dir=/home/user/.android/android-ndk-r19b --sdk-dir=/home/user/.android/android-sdk --android-api=27 --arch=armeabi-v7a --permission=INTERNET --permission=VIBRATE  --debug --ignore-setup-py --storage-dir=/home/user/.local/share/python-for-android


```

### Logs
I python  :   File "jnius/jnius_export_func.pxi", line 26, in jnius.jnius.find_javaclass
I python  :   File "jnius/jnius_utils.pxi", line 91, in jnius.jnius.check_exception
I python  : jnius.jnius.JavaException: JVM exception occurred: Didn't find class "org.kivy.android.PythonActivity" on path: DexPathList[[directory "."],nativeLibraryDirectories=[/system/lib, /vendor/lib, /system/lib, /vendor/lib]] java.lang.ClassNotFoundException

```

dbnicholson:
[...]

PR #2597, opened 2022-05-20
Fix webview testapp jnius usage

dbnicholson:
Disable flask threading so that jnius works using the webview bootstrap. See #2533 for more details.


AndreMiras (approved):
LGTM thanks

PR #2598, opened 2022-05-20
Add service to webview test app

dbnicholson:
This will help ensure that services work from the webview bootstrap.


AndreMiras (approved):
pretty cool thanks

PR #2599, opened 2022-05-21
Introduces pkg_config_location in Prerequisite and use OpenSSLPrerequisite().pkg_config_location in hostpython3, so we can support ssl on hostpython3 just out of the box also on macOS

misl6:
✂️ (Partially) from #2586 
- `hostpython3` requires `PKG_CONFIG_PATH` to be set on macOS in order to have `ssl` support (which is needed by `pip`)
- Introduces `pkg_config_location` property in `Prerequisite`.
- Introduced `homebrew_formula_name` in `Prerequisite`, to avoid some DRY violations.

### DIFF
--- a/pythonforandroid/prerequisites.py
+++ b/pythonforandroid/prerequisites.py
@@ -10,6 +10,7 @@ from pythonforandroid.logger import info, warning, error
 
 class Prerequisite(object):
     name = "Default"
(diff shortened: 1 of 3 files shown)


AndreMiras (dismissed):
LGTM

PR #2605, opened 2022-05-26
changed arch.ndk_platform to arch.ndk_lib_dir in `librt` recipe

kengoon:
fix for 
```python
AttributeError: 'ArchARMv7_a' object has no attribute 'ndk_platform'
```


AndreMiras (approved):
"Test updated recipes" are green and did build librt, so I guess we're good.
Thank you :beers:

PR #2606, opened 2022-05-26
removed `usr` and `lib` from ndk library path in `librt` recipe

kengoon:
fix for lxml build error
```sh
ld: error: unable to find library -lrt
clang-12: error: linker command failed with exit code 1 (use -v to see invocation)
error: command '/usr/bin/ccache' failed with exit status 1
```

the above error is using this path
```sh
/home/<user>/.buildozer/android/platform/android-ndk-r23b/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/arm-linux-androideabi/21/usr/lib
```
instead of 
```sh
/home/<user>/.buildozer/android/platform/android-ndk-r23b/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/arm-linux-androideabi/21
```

### DIFF
--- a/pythonforandroid/recipes/librt/__init__.py
+++ b/pythonforandroid/recipes/librt/__init__.py
@@ -19,7 +19,7 @@ class LibRt(Recipe):
         finishes'''
 
     def build_arch(self, arch):
-        libc_path = join(arch.ndk_lib_dir, 'usr', 'lib', 'libc')
+        libc_path = join(arch.ndk_lib_dir, 'libc')
         # Create a temporary folder to add to link path with a fake librt.so:
         fake_librt_temp_folder = join(
             self.get_build_dir(arch.arch),


AndreMiras:
OK linking issue, we missed that from the CI because it would generate runtime errors indeed.
Did you get a chance to test the binary runtime this time to be sure the recipe is OK?

kengoon:
The build is now successful, I am about to test the apk. I will update soon if it works fine

kengoon:
I can confirm that the runtime binary works perfect

AndreMiras (approved):
Great thanks for the fix and the test, let's :ship:

misl6:
Hi @kengoon and @AndreMiras 

I’m not against these changes (And thank you @kengoon for taking care of it), but please consider that are already targeted via #2586 (which will likely exit the WIP status this weekend) by using the brand-new `pythonforandroid.androidndk.AndroidNDK` object and the new versioned properties (See https://android.googlesource.com/platform/ndk/+/master/docs/BuildSystemMaintainers.md#sysroot).

AndreMiras:
Oh right good point @misl6, thanks for the heads up

PR #2612, opened 2022-06-01
Fixing service_library bootstrap + .aar  build.

mzakharo:
This CL contains fixes, building service_library bootstrap with aar target. Tested on Android 12.  My primary goal was to create a way to use flutter with p4a.   

Here is my tutorial on how to marry flutter with p4a:
https://github.com/mzakharo/flutter_with_python_android/blob/main/README.md

Maybe a tutorial like this could be made part of p4a documentation. Comments?


mzakharo:
It took a few tries, but the new test ` Unit test aar ` is now passing

Makefile
```diff
@@ -47,6 +47,13 @@ testapps-with-numpy-aab: virtualenv
     --requirements libffi,sdl2,pyjnius,kivy,python3,openssl,requests,urllib3,chardet,idna,sqlite3,setuptools,numpy \
     --arch=armeabi-v7a --arch=arm64-v8a --arch=x86_64 --arch=x86 --release
 
+testapps-service_library-aar: virtualenv
+	. $(ACTIVATE) && cd testapps/on_device_unit_tests/ && \
+    python setup.py aar --sdk-dir $(ANDROID_SDK_HOME) --ndk-dir $(ANDROID_NDK_HOME) \
+    --bootstrap service_library \
+    --requirements python3 \
+    --arch=arm64-v8a --release
```
misl6, line 55: There's a reason to only run tests on `arm64-v8a`? (Or is it just for having a faster CI?)

pythonforandroid/toolchain.py
```diff
@@ -1093,6 +1094,8 @@ def _build_package(self, args, package_type):
                     gradle_task = "assembleRelease"
                 elif package_type == "aab":
                     gradle_task = "bundleRelease"
+                else:
+                    gradle_task = "assembleRelease"
```
misl6, line 1098: Looks the same of `package_type == "apk"`, how about avoiding the `else` here ?

misl6 (approved):
Hi @mzakharo,
Thank you, and sorry for the late reply!

Left a couple of minor comments, let me know what do you think about it :)

Makefile
mzakharo: there was no support for multi-arch in `service_library` bootstrap. now fixed, and x86 added to test multi-arch support.

pythonforandroid/toolchain.py
mzakharo: updated the file to match the request

misl6 (approved):
LGTM. Thank you!

misl6:
> Maybe a tutorial like this could be made part of p4a documentation. Comments?

Looks good. Maybe you can write a more generic example, as may also help non-flutter users?

Issue #2609, opened 2022-05-31
Allow multiple python interpreters per process


dbnicholson:
### Checklist

- [x] the issue is indeed a bug and not a support request
- [x] issue doesn't already exist: https://github.com/kivy/python-for-android/issues
- [ ] I have a short, runnable example that reproduces the issue
- [x] I reproduced the problem with the latest development version (`p4a.branch = develop`)
- [x] I used the grave accent (aka backticks) to format code or logs when appropriated

### Versions

- Python: 3.9
- OS: Linux
- Kivy: N/A
- Cython: N/A
- OpenJDK: 11

### Description

Currently the native code [starting the python interpreter](https://github.com/kivy/python-for-android/blob/develop/pythonforandroid/bootstraps/common/build/jni/application/src/start.c) only handles a single interpreter per process. This limits how applications handle services (and soon workers) as the python interpreter always has to be run in a separate process instead of a separate thread in the current process. This is managed in the `AndroidManifest.xml` file by declaring `android:process` in the service declarations.

Python does support this via [sub-interpreters](https://docs.python.org/3/c-api/init.html#sub-interpreter-support), but the setup is more complex. It would be nice to support that, though. Particularly for workers where they currently have to use more complex `RemoteListenableWorker` class.

rtibbles:
@dbnicholson and I have talked about this separately, but just to add in from my experience of attempting to handle this, I think that it wouldn't even be necessary to support sub-interpreters (which is apparently somewhat finicky https://docs.python.org/3/c-api/init.html#bugs-and-caveats), but just properly register the non-Python created threads against the GIL.

However, in pursuing these approaches, I have come across an issue that seemingly is specific to PyJnius, whereby when it attempts to import its interface to Java from the pyi files, it results in an import not found error. The only possible interpretation that I have been able to come up with so far is that somehow the imported modules are garbage collected, because all Python threads have stopped, and this somehow does not get remedied when reinitializing Python in a subsequent invocation. I will report more here as I uncover it.

Issue #2610, opened 2022-05-31
Improve process for stopping python


dbnicholson:
### Checklist

- [x] the issue is indeed a bug and not a support request
- [x] issue doesn't already exist: https://github.com/kivy/python-for-android/issues
- [ ] I have a short, runnable example that reproduces the issue
- [x] I reproduced the problem with the latest development version (`p4a.branch = develop`)
- [x] I used the grave accent (aka backticks) to format code or logs when appropriated

### Versions

- Python: 3.9
- OS: Linux
- Kivy: N/A
- Cython: N/A
- OpenJDK: 11

### Description

Right now the way the `Activity` and `Service` classes stop python is to kill the whole process. That's a very heavy handed way to stop the python thread since it prevents the application taking any action when it has to stop.

I think an one way to do this would be to define a custom python exception and raise it in the python interpreter. The application could catch the exception, perform any actions it needs, and then return gracefully. From the Java side, it would run this via the JNI from a different thread than python is running in. In that case, I believe you could use one of [these](https://docs.python.org/3/c-api/init.html#asynchronous-notifications) [two](https://docs.python.org/3/c-api/init.html#gilstate) mechanisms to [raise the exception](https://docs.python.org/3/c-api/exceptions.html#raising-exceptions) in the python main thread.

PR #2613, opened 2022-06-01
WorkManager support

dbnicholson:
This enables `WorkManager` tasks via the `--worker` p4a option. This is based on the work in #2464, with the following differences:

* Support is added in the common bootstrap rather than the service_library bootstrap so it can be used by p4a apps. I believe it should still work in the service_library context, but I haven't tested.
* I removed the bound service addition. That might still be useful, but my goal was to support `WorkManager` tasks, and that wasn't needed. I believe in #2464 they were using the bound service as the target for the `WorkManager` tasks, but I changed it so that each worker has a generated service class. The generated service class is based on `RemoteWorkerService`, which itself implements a bound service.

An important thing to note is that since p4a only supports one python interpreter per process (see #2609), the work task is always run in a separate process via a `RemoteListenableWorker`. That seems to work well except that I couldn't get [cancellation to work](https://issuetracker.google.com/234211438). The downside of this approach is that the androidx work-multiprocess release including this support requires SDK 30.

The bulk of the work is in the "Support WorkManager tasks" commit. Everything before that is preparation, which could be split out into a separate PR if that's preferred.

There are still a couple things I haven't completed:

- [x] I added a `--worker` test in the on-device tests, but it's only being run in the flask webview app. It would be good to be able to test it in the kivy app, too.
- [ ] I haven't tested foreground handling and the `PythonWorker` class doesn't implement the `getForegroundInfoAsync()` and `getForegroundInfo()` methods.
- [x] Documentation
- [ ] ~~Currently there's ugly `WorkManager.initialize` call in the generated worker service since the androidx startup handling doesn't seem to work on remote processes. It seems to work, but I don't think it's the right way to do it.~~ This isn't ideal, but I think it's fine this way.


dbnicholson:
I fleshed this out some more and got cancellation to work. It turns out that I was just doing it wrong and blocking on the python thread in the worker. That's wrong as it doesn't allow the future to be returned and `WorkManager` doesn't get it until the python process has completed.

Otherwise, I spent some time trying to make sure the worker service always exits since you can get strange errors if you run python more than once in a process. That works most of the time, but sometimes if you cancel work and then run work to completion it won't exit since the service is still bound from the cancelled work. I think you'd need to address #2610 so that the python thread could be stopped without killing the whole process.

Lastly, I have some work in progress on `getForegroundInfoAsync()`, but I ran out of time right now to complete it. I might be able to get to that in the next week.

HackForLive:
Hi @dbnicholson,

I would like to ask you what is the current status of this Task?

dbnicholson:
Sorry, I'm no longer working on any projects that use python-for-android, so I'm not going to finish this. I'm going to close the PR but will leave the branch in place on my fork in case anyone wants to use it for reference.

PR #2615, opened 2022-06-06
use LEGACY_NDK option to build lapack/scipy with a separate NDK

mzakharo:
Google has removed gcc/gfortran support in their latest NDK (r23b), supported by p4a.  Building gfortran out-of-tree is possible, but the sysroot, shipped with NDK r23b does not contain enough libraries to properly link objects, generated by gfortran compiler.

This is a workaround to use older NDK, supporting gcc/gfortran (r19c) to build lapack/scipy, while the rest of the recipes build with r23b by default.

A new environment variable is introduced, `LEGACY_NDK` -> to specify the location of the legacy NDK (r19c), containing gcc/gfortran compilers.

### DIFF
--- a/pythonforandroid/recipes/lapack/__init__.py
+++ b/pythonforandroid/recipes/lapack/__init__.py
@@ -9,6 +9,10 @@ from pythonforandroid.util import current_directory, ensure_dir, BuildInterrupti
 from multiprocessing import cpu_count
 from os.path import join
 import sh
+from os import environ
+from pythonforandroid.util import build_platform
+
+arch_to_sysroot = {'armeabi': 'arm', 'armeabi-v7a': 'arm', 'arm64-v8a': 'arm64'}
 
 
 class LapackRecipe(Recipe):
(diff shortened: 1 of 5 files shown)


AndreMiras (approved):
Nice, thanks for the fix.
I like that you could scope the workaround to the recipes.

So after r19c or so, the NDK would no longer support gfortran at all? What a bummer :confused: 
I took a quick look at scipy repo and bug tracker and it doesn't seem like they have any plans for replacing lapack. I don't know how deep it is integrated in their code base.
Maybe it's worth letting them know they have a community of mobile user/developer, but that gfortran support was dropped some time ago from the NDK.
Maybe they would have suggestions

mzakharo:
> 

Unfortunately, besides lapack, some portions of scipy  are built with gfortran, I dont think fortran is going away any time soon. The only hope is for  LLVM  flang to become a full-fledged compiler/linker of fortran objects, and Google adding support for it officially.

mhsmith:
> Building gfortran out-of-tree is possible, but the sysroot, shipped with NDK r23b does not contain enough libraries to properly link objects, generated by gfortran compiler.

@mzakharo: Do you remember which libraries were missing, and what errors this produced? 

In the [Chaquopy](https://github.com/chaquo/chaquopy/tree/master/server/pypi) project we've had good results using your [prebuilt gfortran](https://github.com/mzakharo/android-gfortran) to build OpenBLAS and SciPy for Android. But we obviously can't stay on an old version of the NDK forever. For example, NDK version 27 is recommended for [16 KB page support](https://developer.android.com/guide/practices/page-sizes).

PR #2619, opened 2022-06-13
lapack/scipy: support NDK r21e, x86/64 archs

mzakharo:


AndreMiras (approved):
Cool thanks, that's a bit better already

Issue #2630, opened 2022-06-30
`tflite-runtime` build every time


Neizvestnyj:
For some reason, the library is compiled every time the application builds
```py
Building tflite-runtime for arm64-v8a
[INFO]:    tflite-runtime apparently isn't already in site-packages
```

**p4a - develop**

PR #2631, opened 2022-06-30
Fix issue #2630

Neizvestnyj:
Now, if the **requirements** has not been changed, the reassembly will not occur

### DIFF
--- a/pythonforandroid/recipes/tflite-runtime/__init__.py
+++ b/pythonforandroid/recipes/tflite-runtime/__init__.py
@@ -1,4 +1,4 @@
-from pythonforandroid.recipe import PythonRecipe, current_directory,\
+from pythonforandroid.recipe import PythonRecipe, current_directory, \
     shprint, info_main, warning
 from pythonforandroid.logger import error
 from os.path import join
@@ -6,7 +6,6 @@ import sh
 
 
 class TFLiteRuntimeRecipe(PythonRecipe):
-
     ###############################################################
     #
     # tflite-runtime README:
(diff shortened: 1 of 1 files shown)


mzakharo:
the fix does not make sense to me. ` install_in_hostpython = True` is a directive to build the recipe twice, once for target arch, and once for host arch. You would only need to build for hostpython if some other recipe  has a build dependency on tflite-runtime, which to my knowledge, is not the cause of issue #2630.

  Many recipes dont have this 'install_in_hostpython' directive, and do not exhibit issue #2630  .

Am I missing something?

Neizvestnyj:
> the fix does not make sense to me. ` install_in_hostpython = True` is a directive to build the recipe twice, once for target arch, and once for host arch. You would only need to build for hostpython if some other recipe has a build dependency on tflite-runtime, which to my knowledge, is not the cause of issue #2630.
> 
> Many recipes dont have this 'install_in_hostpython' directive, and do not exhibit issue #2630 .
> 
> Am I missing something?

It's really strange, but you can check if for some reason the package stops being reassembled

AndreMiras:
Agree, what should be looked into instead is probably the [`PythonRecipe.should_build()`](https://github.com/kivy/python-for-android/blob/v2022.03.13/pythonforandroid/recipe.py#L950-L956).
You can override this function in your recipe, put a breakpoint and step into `super().should_build(arch)`.
My guess would be there's some kind of mismatch between the recipe name and what's to be expected in that folder, well something like that.
But that requires a bit of digging

Neizvestnyj:
> Agree, what should be looked into instead is probably the [`PythonRecipe.should_build()`](https://github.com/kivy/python-for-android/blob/v2022.03.13/pythonforandroid/recipe.py#L950-L956). You can override this function in your recipe, put a breakpoint and step into `super().should_build(arch)`. My guess would be there's some kind of mismatch between the recipe name and what's to be expected in that folder, well something like that. But that requires a bit of digging

Done

misl6 (changes requested):
Hi @Neizvestnyj 

Seems that the PEP8 tests are failing (and so also the other tests), can you take care of it?

tshirtman (approved):

PR #2636, opened 2022-07-18
Fix webview Back button behaviour

### LINKED ISSUE #1907: splash always loading
i build app with bootstrap=webview, 
the application works perfect, but when I press the back button the splah comes back out and stays loaded. What I try is that this does not happen and that when you press the button to go back out of the application, is there any option to access the onResume methods ...? Is there any way to solve this problem?

interlark:
Fix #1907 
Back button currently fall to the start [preloader](https://github.com/kivy/python-for-android/blob/develop/pythonforandroid/bootstraps/webview/build/webview_includes/_load.html) and stuck on it.
We better keep that in mind and not let it happen.

Also fixed bug when app get closed without confirmation when you launch  it and immediately press back.

https://user-images.githubusercontent.com/20641837/179565742-39e1b920-dc1b-421c-8773-b5a214c0343a.mp4

### DIFF
--- a/pythonforandroid/bootstraps/service_only/build/src/main/java/org/kivy/android/PythonActivity.java
+++ b/pythonforandroid/bootstraps/service_only/build/src/main/java/org/kivy/android/PythonActivity.java
@@ -177,19 +177,22 @@ public class PythonActivity extends Activity {
             new File(getApplicationInfo().nativeLibraryDir));
     }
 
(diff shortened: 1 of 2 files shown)


misl6 (approved):
We likely would like to investigate about the need of that feature on `service_only` boostrap, but that's another kind of task 😄 

Tested and the PR LGTM, thank you!

Issue #2623, opened 2022-06-23
Can't apply patches with relative paths for local recipe


dbnicholson:
### Checklist

- [x] the issue is indeed a bug and not a support request
- [x] issue doesn't already exist: https://github.com/kivy/python-for-android/issues
- [x] I have a short, runnable example that reproduces the issue
- [x] I reproduced the problem with the latest development version (`p4a.branch = develop`)
- [x] I used the grave accent (aka backticks) to format code or logs when appropriated

### Versions

- Python: 3.9.2
- OS: Linux
- Kivy: N/A
- Cython: N/A
- OpenJDK: N/A

### Description

A patch with a relative path in a local recipe can't be applied because the patch can't be found:

```
  RAN: /usr/bin/patch -t -d /home/dan/.local/share/python-for-android/build/other_builds/zstandard/x86_64__ndk_target_21/zstandard -p1 -i p4a-recipes/zstandard/preprocessor.patch

  STDOUT:
/usr/bin/patch: **** Can't open patch file p4a-recipes/zstandard/preprocessor.patch : No such file or directory
```

The problem is that `-d` changes the directory before reading the patch passed with `-i`. With a normal recipe, the path to `-i` is absolute because it's derived from `Context.root_dir`, which is absolute. Unless the caller passes an absolute path to `--local-recipes`, the local recipe patch path will also be relative.

### buildozer.spec

N/A

### Logs

```
[INFO]:    Prebuilding zstandard for x86_64
[INFO]:    zstandard has no prebuild_x86_64, skipping
[INFO]:    Applying patches for zstandard[x86_64]
[INFO]:    Applying patch preprocessor.patch
[INFO]:    -> running patch -t -d /home/dan/.local/share/python-for-android/build/other_builds/zstandard/x86_64__ndk_targ...(and 63 more)
           working: /usr/bin/patch: **** Can't open patch file p4a-recipes/zstandard/preprocessor.patch : No such file or directory        Exception in thread background thread for pid 987188:
Traceback (most recent call last):
  File "/usr/lib/python3.9/threading.py", line 954, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.9/threading.py", line 892, in run
    self._target(*self._args, **self._kwargs)
  File "/home/dan/src/kolibri/kolibri-installer-android/venv/lib/python3.9/site-packages/sh.py", line 1683, in wrap
    fn(*rgs, **kwargs)
[...]

PR #2640, opened 2022-07-22
Resolve absolute path to local recipes

dbnicholson:
This is necessary when using patches in a local recipe since
`Recipe.apply_patch` assumes the recipe directory is absolute and uses
`patch -d` to change directories. It could just be fixed there, but this
ensures that recipe directories are always absolute.

Closes: #2623


misl6 (approved):
LGTM. Thank you!

PR #2650, opened 2022-08-01
Create 123

artem123tew:


misl6:
Hi @artem123tew!

That snippet of code doesn't look like to be related in any way to `python-for-android`.

If I'm wrong, feel free to re-open another PR, specifying why this addition would be needed.

PR #2655, opened 2022-08-15
Create tictoe

uslex:


misl6:
Hi @uslex ,

That snippet of code doesn't look like to be related in any way to `python-for-android`.

If I'm wrong, feel free to re-open another PR, specifying why this addition would be needed.

PR #2657, opened 2022-08-16
Remove six and enum34 dependency

misl6:
- Python 2 is not supported anymore, so `six` is not needed.
- The same applies for `enum34` as python versions < 3.4 are not supported anymore.


AndreMiras (approved):
LGTM

PR #2663, opened 2022-09-01
`liblzma`: Use `p4a_install` instead of `install`, as a file named `INSTALL` is already present.

misl6:
Fixes issue #2343 

On macOS with a case-insensitive filesystem (which is the default), `install` and `INSTALL` are the same thing, and `INSTALL` is a file that already exists.

- The folder used to store the build results (`lib` and `headers`) was previously named `install` and now got renamed to `p4a_install` so it doesn't interfere with the file named `INSTALL`

- Removed an `ensure_dir('install')` call, which was creating an unused folder in the wrong place.

- Added a test that may help us to notice if the folder name gets reverted to `install`


AndreMiras (approved):
Thanks for the fix and comprehensive test/comments

PR #2669, opened 2022-09-08
android/activity: Add Application.ActivityLifecycleCallbacks helpers

dbnicholson:
The `Application.ActivityLifecycleCallbacks` interface is used to register a class that can receive callbacks on Activity lifecycle changes. Add a `PythonJavaClass` implementing the interface and helper functions to register python callbacks with it. This can be used to perform actions in the python app when the activity changes lifecycle states.

### DIFF
--- a/doc/source/apis.rst
+++ b/doc/source/apis.rst
@@ -190,6 +190,51 @@ Example::
         # ...
 
 
+Activity lifecycle handling
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. module:: android.activity
+
+The Android ``Application`` class provides the `ActivityLifecycleCallbacks
+<https://developer.android.com/reference/android/app/Application.ActivityLifecycleCallbacks>`_
+interface where callbacks can be registered corresponding to `activity
+lifecycle
+<https://developer.android.com/guide/components/activities/activity-lifecycle>`_
+changes. These callbacks can be used to implement logic in the Python app when
+the activity changes lifecycle states.
+
+Note that some of the callbacks are not useful in the Python app. For example,
+an `onActivityCreated` callback will never be run since the the activity's
+`onCreate` callback will complete before the Python app is running. Similarly,
+saving instance state in an `onActivitySaveInstanceState` callback will not be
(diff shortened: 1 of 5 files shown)


misl6:
Code looks super good and promising, but not tested yet during runtime, meanwhile can you please rebase so we can see a green run from the CI?

dbnicholson:
> Code looks super good and promising, but not tested yet during runtime, meanwhile can you please rebase so we can see a green run from the CI?

Do you want me to exercise it in the test app?

misl6:
> > Code looks super good and promising, but not tested yet during runtime, meanwhile can you please rebase so we can see a green run from the CI?
> 
> Do you want me to exercise it in the test app?

Uh, that would be super nice, but not mandatory.

The rebase was just to include a recent change that will allow the unit tests to run.

dbnicholson:
> > Do you want me to exercise it in the test app?
> 
> Uh, that would be super nice, but not mandatory.

It's fine. I already had something more or less there but I'd punted to work on something else.

dbnicholson:
Well, I'm glad I did because now I found out that it doesn't work on API 31. Trying to figure out why...

dbnicholson:
I rebased and added some callbacks in the test app. It works fine for both sdl2 and webview on API 29 and API 30, but I can't figure out why it doesn't work on API 31. It seems like an Android bug that the callbacks are never sent.

dbnicholson:
> I rebased and added some callbacks in the test app. It works fine for both sdl2 and webview on API 29 and API 30, but I can't figure out why it doesn't work on API 31. It seems like an Android bug that the callbacks are never sent.

False alarm. It does work fine on API 31. It just seems they changed the activity lifecycle a bit. Before the activity would be stopped by going to the overview. Now you have to go into a different activity or back to the home screen to go through the pause/stop cycle. Then it will go back through the start/resume cycle when going back to the app.

So, I think this works fine on all the API levels I tried - 29, 30 and 31.

misl6 (approved):
LGTM.

Thank you for the tests and well-written docs!
Tested both `sdl2` and `webview` APK on API 31, and worked flawlessly!

PR #2672, opened 2022-09-17
Fixes libvpx build

misl6:
Fixes issue #2665 

- `libvpx` build was silently failing.
- `--enable-external-build` is preventing the build to success.
- Disabled `neon-asm` on `armeabi-v7a`
- Removed `ffpyplayer` from `BROKEN_RECIPES`


AndreMiras (approved):
Nice

PR #2673, opened 2022-09-18
Update `SDL2`, `SDL2_ttf`, `SDL2_mixer`, `SDL2_image` to latest releases

misl6:
**SDL2**: `2.0.9` --> `2.24.0`
**SDL2_image**: `2.0.4` --> `2.6.2`
**SDL2_mixer**: `2.0.4` --> `2.6.2`
**SDL2_ttf**: `2.0.15` --> `2.20.1`

~⚠️ Needs some wide runtime testing before merge.~ Tested.

✅ Includes `sdl-gl-swapwindow-nogil.patch` (See: https://github.com/kivy/kivy/pull/8025)
✅ Tested `jpg`, `png` and `webp` support via SDL2_image
✅ Tested `mp3`, `wav` and `ogg` support via SDL2_mixer
✅ Added `packaging` to requirements (`LooseVersion` is deprecated)
✅ Added a logic that allow `kivy` to be patched only if needed. (Is quite common to target `kivy==master`, and the `sdl-gl-swapwindow-nogil.patch` should not be applied in that case, otherwise gets reversed.)


AndreMiras (approved):
Nice stuff thanks a mil

PR #2674, opened 2022-09-21
Resize webview when keyboard is shown

dbnicholson:
When the adjust mode is unspecified, Android will not resize the webview when showing the on screen keyboard to keep the input element on screen. Explicitly instruct it to use `adjustResize` mode. Possibly this is because the webview bootstrap uses an `AbsoluteLayout` as the top level widget. See
https://developer.android.com/guide/topics/manifest/activity-element#wsoft for details.


dbnicholson:
Before:
webview-input-unspecified

After:
webview-input-resize

AndreMiras (approved):
LGTM, thanks for the context and the screenshots

PR #2686, opened 2022-10-12
build: add --android-target-sdk-version arg (defaults to android_api)

SomberNight:
Currently p4a has a variable named `android_api` (set by e.g. buildozer via the `app.android.api` config key). In `AndroidManifest.xml`, both `compileSdkVersion` and `targetSdkVersion` are set to `android_api`.

This PR allows setting the targetSdkVersion and compileSdkVersion independently, by adding a `--android-target-sdk-version` build arg.

`compileSdkVersion` is a compilation setting used by e.g. the NDK at build-time, while `targetSdkVersion` is just a field to set in the manifest, only used at runtime (mainly in the Java world). (see e.g. https://stackoverflow.com/a/26694276 )

-----

note: my current use case is to set `targetSdkVersion` to higher than `compileSdkVersion`, as I am using an older NDK (r22b) that only supports `compileSdkVersion<=30`, but the google play store now requires `targetSdkVersion>=31`. Historically it has sometimes been the case that p4a was using an old NDK, and I had to use this trick once already a few years ago, for the same reason _(but atm p4a is using a very new NDK, I am just having troubles rebasing my fork)_. Regardless, I believe setting the two parameters independently might have other use cases.

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -497,6 +497,12 @@ main.py that loads it.''')
         sdk_dir = fileh.read().strip()
     sdk_dir = sdk_dir[8:]
 
+    if args.android_target_sdk_version == -1:
+        args.android_target_sdk_version = android_api
+    if args.android_target_sdk_version != android_api:
+        print(f'WARNING: android_target_sdk_version ({args.android_target_sdk_version}) '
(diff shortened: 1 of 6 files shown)


SomberNight:
I have a corresponding patch for buildozer to allow setting this in buildozer.spec, but will wait for feedback here first.

accumulator:
bump

This would be great for extra flexibility

PR #2693, opened 2022-10-23
Fixes an issue regarding blacklist and bytecode compile + some cleanup

misl6:
When building an app via `python-for-android` (without using `buildozer` on top of it), the app may fail to build.

`python-for-android` refuses to build due to a python file that is located in a blacklisted folder (via `blacklist.txt`) as that file is failing the byte-compile step.

That should not happen, as a blacklisted folder (E.g. `venv` or `.git`) may contain a python file that can't be byte-compiled.
Also, byte-compiling a blacklisted folder may lead to an increase of the build time.

This PR targets the above-mentioned issue, and the following minor ones:
- **As of Python 3.5, the `.pyo` filename extension is no longer used and has been removed in favour of extension `.pyc`**
All the references to `pyo` has been removed. `--no-compile-pyo` build option has been removed in favor of `--no-byte-compile-python`, which doesn't refer to a specific filename extension.
- `.apk` and `.aab` are not blacklisted by default. The first ever build will work, but the subsequent one will try to also include the `.apk` or `.aab` into the artifact. That will lead to an increase in size and a potential failure during the final gradle step.

### DIFF
--- a/doc/source/buildoptions.rst
+++ b/doc/source/buildoptions.rst
@@ -89,7 +89,7 @@ options (this list may not be exhaustive):
 - ``--service``: A service name and the Python script it should
   run. See :ref:`arbitrary_scripts_services`.
 - ``--add-source``: Add a source directory to the app's Java code.
-- ``--no-compile-pyo``: Do not optimise .py files to .pyo.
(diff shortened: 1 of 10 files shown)


AndreMiras (approved):
Great catch and good clean up thanks!

PR #2701, opened 2022-11-05
Update __init__.py

tct123:
edited line 60.


misl6:
Why is that change needed? Feel free to re-open a new PR describing what are you trying to achieve.

Issue #2704, opened 2022-11-15
Longer app load time, bigger apk size and unnecessary logs


DexerBR:
### Checklist

- [x] the issue is indeed a bug and not a support request
- [x] issue doesn't already exist: https://github.com/kivy/python-for-android/issues
- [x] I have a short, runnable example that reproduces the issue
- [x] I reproduced the problem with the latest development version (`p4a.branch = develop`)
- [x] I used the grave accent (aka backticks) to format code or logs when appropriated

### Versions

- Python: 3.9.7
- OS: Ubuntu 20.04 (WSL2 - Windows)
- Kivy: master
- Cython: 0.29.19
- OpenJDK: openjdk version "17.0.5" 2022-10-18

### Description

When using the kivy + p4a + buildozer development branches to generate an apk, I get this result/behavior:

- The size of the app has increased considerably compared to the stable versions: using one architecture, the app has gone from approximately 12.5MB to 18.5MB (first build) and 26.4MB (later builds)
- The app size grows unexpectedly from 18.5MB to 26.4MB in the second build. From the second to next builds, the sudden size changes seem to stabilize.
- The app's startup time has increased considerably, with the black loading screen (after the splashscreen, which is taking a little longer to load as well) remains for a few seconds before the UI is loaded/displayed.
- A huge amount of log, apparently related to pyjnius is generated (unnecessarily)
- Sometimes logcat quits suddenly with this message: `# org.test.myapp terminated`

### Expected behavior

- Maintain a reasonable app size (in relation to stable versions) and little difference in later builds (with the same app code/architecture).
- Boot time equal to or less than stable versions.
- Only logs needed (as in stable versions)

### main.py

```python
from kivy.app import App
from kivy.uix.button import Button


class TestApp(App):
    def build(self):
        return Button(text="Hello World")


TestApp().run()
```

### buildozer.spec

Command:
```sh
buildozer android debug
```

Spec file:
```
[app]

# (str) Title of your application
title = My Application

# (str) Package name
package.name = myapp

# (str) Package domain (needed for android/ios packaging)
package.domain = org.test

# (str) Source code where the main.py live
source.dir = .

# (list) Source files to include (let empty to include all the files)
source.include_exts = py,png,jpg,kv,atlas

# (list) List of inclusions using pattern matching
#source.include_patterns = assets/*,images/*.png

# (list) Source files to exclude (let empty to not exclude anything)
#source.exclude_exts = spec

[...]

PR #2705, opened 2022-11-19
Gradle: Run the clean task before anything else to make sure nothing is cached.

misl6:
Headache time 🤕⏱️!

As clearly described by @DexerBR in #2704 :
> The app size grows unexpectedly from 18.5MB to 26.4MB in the second build. From the second to next builds, the sudden size changes seem to stabilize.

For a still unknown(likely a Gradle bug) reason, the produced `APK` is bigger on non-first builds.
After some analysis (unpacked the APK, checked and compared the content), I can say that only the size of the APK itself is different, not the content, as when it gets unzipped, the size of the unzipped content is the same between the first build and subsequent ones.

The files md5 hashes are the same, except for the `assets/private.tar` file (and its content), as I'm not building with `SOURCE_DATE_EPOCH` set, which means that [compileall](https://docs.python.org/3/library/compileall.html#cmdoption-compileall-invalidation-mode) uses the `timestamp` as the invalidation mode, making the build non-reproducible.

That's where the 💡 toggled on.

What's the size of my `assets/private.tar`? `17MB`
What's the APK size difference between the first and subsequent builds? `17MB`
Ouch! 😓

For some reason, Gradle creates a bigger APK, which is able to contain all the assets and code + 2x `private.tar` (the old one, and the new one), but then only packages into it the newer one.

I will try to investigate it on Gradle (and Android Gradle Plugin) codebase, but meanwhile, cleaning the build folder (via `./gradlew clean *****`) avoids any caching and ensures a clean build, every time, and doesn't seem to take any considerable amount of time.

EDIT:
Found a discussion on Google Issue Tracker about something similar: https://issuetracker.google.com/issues/172762362
> If an entry has been deleted in the middle of the archive, Zipflinger will not leave a “hole” there. This is done in order to be compatible with top-down parsers such as jarsigner or the JDK zip classes. To this effect, Zipflinger fills empty space with virtual entries (a.k.a a Local File Header with no name, up to 64KiB extra and no Central Directory entry). Alignment is also done via “extra field”.

### DIFF
--- a/pythonforandroid/toolchain.py
+++ b/pythonforandroid/toolchain.py
@@ -1108,7 +1108,10 @@ class ToolchainCL:
             else:
                 raise BuildInterruptingException(
                     "Unknown build mode {} for apk()".format(args.build_mode))
-            output = shprint(gradlew, gradle_task, _tail=20,
+
+            # WARNING: We should make sure to clean the build directory before building.
+            # Looks like our private.tar is causing issues to gradle.
+            output = shprint(gradlew, "clean", gradle_task, _tail=20,
                              _critical=True, _env=env)
         return output, build_args
 
--- a/pythonforandroid/toolchain.py
+++ b/pythonforandroid/toolchain.py
@@ -1110,7 +1110,7 @@ class ToolchainCL:
                     "Unknown build mode {} for apk()".format(args.build_mode))
 
             # WARNING: We should make sure to clean the build directory before building.
-            # Looks like our private.tar is causing issues to gradle.
+            # See PR: kivy/python-for-android#2705
             output = shprint(gradlew, "clean", gradle_task, _tail=20,
                              _critical=True, _env=env)
         return output, build_args


pythonforandroid/toolchain.py
```diff
@@ -1108,7 +1108,10 @@ def _build_package(self, args, package_type):
             else:
                 raise BuildInterruptingException(
                     "Unknown build mode {} for apk()".format(args.build_mode))
-            output = shprint(gradlew, gradle_task, _tail=20,
+
+            # WARNING: We should make sure to clean the build directory before building.
+            # Looks like our private.tar is causing issues to gradle.
+            output = shprint(gradlew, "clean", gradle_task, _tail=20,
```
AndreMiras, line 1114: Nice findings and fix :muscle: 
Do you think we should reference this pull request or initial bug report with a `refs #2705` in the comment since you explained it all well?

AndreMiras (approved):
Thanks for racking your brain on it, that's a very weird bug indeed

pythonforandroid/toolchain.py
misl6: I was waiting to split the original issue,  into multiple chunks (as actually contains multiple issues, and some of them are still being triaged) and then totally forgot to add a reference.

So, I guess that now that we have a PR to reference, I will reference the PR. 👍

PR #2713, opened 2022-12-10
Delete pythonforandroid/recipes/cdecimal directory

RobertFlatt:
Depreciated since Python 3.3, so does not build.

Functionality was included in Python 3.3 https://www.bytereef.org/mpdecimal/
See section "cdecimal - Python module (2.7 - 3.2)"

Nobody could have used this with p4a since p4a Python2 support was depreciated.

Just house cleaning.


AndreMiras (approved):
thanks

PR #2717, opened 2022-12-20
Update cryptography & co (#1)

rnixx:
* Update cryptography recipe to 3.4.8

* Update pyopenssl recipe to 21.0.0

* Downgrade setuptools recipe to 49.6.0

* Add setuptools_rust recipe

* Add semantic_version recipe

* Add typing_extensions recipe

### DIFF
--- a/pythonforandroid/recipes/cryptography/__init__.py
+++ b/pythonforandroid/recipes/cryptography/__init__.py
@@ -3,13 +3,26 @@ from pythonforandroid.recipe import CompiledComponentsPythonRecipe, Recipe
 
 class CryptographyRecipe(CompiledComponentsPythonRecipe):
     name = 'cryptography'
-    version = '2.8'
+    version = '3.4.8'
     url = 'https://github.com/pyca/cryptography/archive/{version}.tar.gz'
-    depends = ['openssl', 'six', 'setuptools', 'cffi']
+
+    # even if rust compilation can be suppressed, setuptools_rust is an
+    # install requirement
+    depends = ['openssl', 'six', 'setuptools', 'setuptools_rust', 'cffi']
     call_hostpython_via_targetpython = False
 
     def get_recipe_env(self, arch):
         env = super().get_recipe_env(arch)
+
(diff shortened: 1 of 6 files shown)


rnixx:
In order to update cryptography to a more recent version, rust cross compiling needs to be figured out, and packages without a setup.py must be installable, which requires the use of pip all over the place instead of ``python setup.py ...``.

I also downgraded setuptools to 49.6.0 due to an import error.

I tried this changes based on current master branch.

misl6:
> I also downgraded setuptools to 49.6.0 due to an import error.

We'll likely need to update `setuptools` to a newer version in future, as other packages may need a newer version.
Is the import error something we can patch?

rnixx:
It was an import error related to build extensions, something changed there in setuptools between 49.6.0 and 50.0.0

I'd also like stick to a newer version. Unfortunately all the pypa stuff is a little bit of a moving target for a while now.

I've seen you created a PR for a recent release, so I'd suggest to fix the pypa story after the recent release is out.

As pointed above, this includes:

- Support packages with no setup.py (setup.py is deprecated anyway in recent python versions)
- Figure out the rust compilation story. It will add another level of complexity but i fear rust extensions will become more common in future.

Once this is in i see no problems updating to very recent versions.

misl6:
@rnixx 

The release is out, but a new one may drop sooner than expected, as we need to fix a couple of "critical" issues.
If you have ideas to fix the pypa story, in order to avoid downgrading `setuptools`, feel free to ping me :)

AndreMiras:
Hi guys :wave: 
Thanks @rnixx for contributing again. Where do we stand with this now?
If we think setuptools can still be an issue, maybe we can break the PR up a little bit and already get some of the non "controversial" recipes merged? Out of the list, which one don't rely on the setuptools downgrade?

T-Dynamos:
@AndreMiras this PR is surpassed by #2962.

misl6:
Closing in favor of #2962.

Feel free to re-open a separate PR for any extra version bumping.

PR #2725, opened 2022-12-29
Support permission properties (`maxSdkVersion` and `usesPermissionFlags`) + remove `WRITE_EXTERNAL_STORAGE` permission, which has been previously declared by default

misl6:
The `--permission` argument historically only accepted something like `--permission android.permission.WRITE_EXTERNAL_STORAGE` and `--permission VIBRATE` which is not enough in certain use cases.

As an example, when dealing with bluetooth permissions and targeting API 31, the legacy `android.permission.BLUETOOTH` and `android.permission.BLUETOOTH_ADMIN` permissions need the property `android:maxSdkVersion` be set to `30`, and 
the `"android.permission.BLUETOOTH_SCAN` needs to strongly assert that the app never derives physical location from Bluetooth scan results via `android:usesPermissionFlags="neverForLocation"`.

Basically, nowadays, android permissions declarations need a lot of fine-tuning, which is not possible at this time via `python-for-android`, unless an extra manifest is used.

This PR introduces a new syntax: `--permission (name=android.permission.WRITE_EXTERNAL_STORAGE;maxSdkVersion=18)` that allows the user to set every single (and currently available/supported) property of the `<uses-permission>` element.

The old syntaxes are migrated before the template gets processed, so users will only notice this change when needed, and when reading the docs.

I also started to add some tests for our `build.py` script. I needed to add a switch to skip some things as ATM are not patchable (at least easily). The whole script needs some refactoring in order to be well-tested.

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -53,10 +53,6 @@ else:
 
 curdir = dirname(__file__)
 
-PYTHON = get_hostpython()
-if PYTHON is not None and not exists(PYTHON):
-    PYTHON = None
-
 BLACKLIST_PATTERNS = [
     # code versionning
     '^*.hg/*',
@@ -75,9 +71,19 @@ BLACKLIST_PATTERNS = [
 ]
 
 WHITELIST_PATTERNS = []
-if get_bootstrap_name() in ('sdl2', 'webview', 'service_only'):
-    WHITELIST_PATTERNS.append('pyconfig.h')
 
+if os.environ.get("P4A_BUILD_IS_RUNNING_UNITTESTS", "0") != "1":
+    PYTHON = get_hostpython()
+    _bootstrap_name = get_bootstrap_name()
+else:
+    PYTHON = "python3"
+    _bootstrap_name = "sdl2"
+
+if PYTHON is not None and not exists(PYTHON):
+    PYTHON = None
+
+if _bootstrap_name in ('sdl2', 'webview', 'service_only'):
+    WHITELIST_PATTERNS.append('pyconfig.h')
 
 environment = jinja2.Environment(loader=jinja2.FileSystemLoader(
     join(curdir, 'templates')))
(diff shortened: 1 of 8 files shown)


pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -646,6 +652,44 @@ def make_package(args):
...
+
+    _permissions = []
+    for permission in args_permissions:
+        if _is_advanced_permission(permission):
+            _permissions.append(_decode_advanced_permission(permission))
+        else:
+            if "." in permission:
+                _permissions.append(dict(name=permission))
+            else:
+                _permissions.append(dict(name=f"android.permission.{permission}"))
```
AndreMiras, line 689: Do we want to add a warning in this case?

tests/test_bootstrap_build.py
```diff
@@ -0,0 +1,86 @@
...
+            os.path.dirname(os.path.abspath(__file__)),
+            "../pythonforandroid/bootstraps/common/build/build.py",
+        )
+
+        buildpy = load_source("buildpy", build_src)
+
+        try:
+            parsed_permissions = buildpy.parse_permissions(args.permissions)  # noqa: F841
+        except ValueError as _e:
+            assert "Property 'propertyThatFails' is not supported." in str(_e)
```
AndreMiras, line 86: If the exception isn't raised, the test won't catch it.
So we either want to make it fail explicitly after the `parse_permissions()`  something like `self.fail("ValueError not raised")` or use something like the pytest.raises context manager e.g. `with pytest.raises(ValueError, match="Property 'propertyThatFails' is not supported."):`

AndreMiras (approved):
Nice thanks!

tests/test_bootstrap_build.py
misl6: Thank you for catching this beginner-like error! 😓
Addressed.

pythonforandroid/bootstraps/common/build/build.py
misl6: Do you mean a deprecation warning? I thought the same, but then I decided to go with the slow path instead in order to avoid a large increase in support requests (users may start panicking 😀).

I did not officially deprecated it by purpose,  instead, I've changed the docs and added a warning regarding "name-only" permissions.

There's a lot of documentation, posts, videos, etc ... which is referring to "name-only" permissions (And we even need to migrate `buildozer` and `python-for-android` examples).

What do you think about it?

AndreMiras: Yeah that makes sense :+1:

AndreMiras (approved):

Issue #2724, opened 2022-12-28
Setting `android:screenOrientation` via `--orientation` has no effect when targeting API 31


misl6:
### Checklist

- [x] the issue is indeed a bug and not a support request
- [x] issue doesn't already exist: https://github.com/kivy/python-for-android/issues
- [x] I have a short, runnable example that reproduces the issue
- [x] I reproduced the problem with the latest development version (`p4a.branch = develop`)
- [x] I used the grave accent (aka backticks) to format code or logs when appropriated

### Versions

- Python: Any
- OS: Any
- Kivy: Any
- Cython: Any
- OpenJDK: Any

### Description

`python-for-android` made possible to set the activity `android:screenOrientation` via `--orientation` argument.
Unfortunately, when targeting API 31 and above, on Android 12 and above, this attribute is ignored, as the activity will run in multi-window mode.

From [Android Docs:](https://developer.android.com/guide/topics/manifest/activity-element#screen)
> android:screenOrientation
The orientation of the activity's display on the device. The system ignores this attribute if the activity is running in [multi-window mode](https://developer.android.com/guide/topics/ui/multi-window).

On Android 12, the multi-window mode is the standard behavior, and additionally:

> On large screens (sw >= 600dp), the platform supports all apps in multi-window mode regardless of app configuration. If [resizeableActivity="false"](https://developer.android.com/guide/topics/manifest/application-element#resizeableActivity), the app is put into compatibility mode when necessary to accommodate display dimensions.

> On small screens (sw < 600dp), the system checks an activity’s [minWidth](https://developer.android.com/reference/android/content/pm/ActivityInfo.WindowLayout#attr_android:minWidth) and [minHeight](https://developer.android.com/reference/android/content/pm/ActivityInfo.WindowLayout#attr_android:minHeight) to determine whether the activity can run in multi-window mode. If [resizeableActivity="false"](https://developer.android.com/guide/topics/manifest/application-element#resizeableActivity), the app is prevented from running in multi‑window mode regardless of minimum width and height.

So, `resizeableActivity="false"`might be great to handle some specific configurations, but is not meant to be a solution for that.
Also, users may want to take advantage of such a feature. (Since the vastly majority of Apps now support multi-window mode)

### But we still want to force a specific orientation, when the app is not in multi-window mode:

Setting `KIVY_ORIENTATION` to a supported SDL configuration, (`Portrait`, `LandscapeRight`, ...) will force the App to keep the selected orientation, even if the device gets rotated.

[...]

PR #2730, opened 2023-01-02
Apply a patch from SDL upstream that fixes orientation settings

misl6:
- inverse landscapeLeft and landscapeRight (now behaves as described in SDL docs)
- fix to match exactly Portrait (as it is contained in PortraitUpsideDown)

See: https://github.com/libsdl-org/SDL/pull/6964 and https://github.com/kivy/python-for-android/issues/2724#issuecomment-1368142168


AndreMiras (approved):
:ship:

PR #2738, opened 2023-01-18
Update __init__.py from `scrypt` recipe

FilipeMarch:
`scrypt` recipe is not working...

Updating the version to `0.8.20` and the url to `https://github.com/holgern/py-scrypt/archive/refs/tags/v{version}.zip` fixes it

You can check the latest release of `scrypt` here: https://github.com/holgern/py-scrypt/releases/tag/v0.8.20

Already tested it on my phone, it works only after applying this fix


AndreMiras (approved):
LGTM thanks

PR #2739, opened 2023-01-21
Implements `--manifest-orientation` and changes how `--orientation` works so we can now pass the setting to the SDL orientation hint

misl6:
As discussed on issue #2724 , when targeting API 31 and above, on Android 12 and above, the attribute `android:screenOrientation` is ignored, as the activity will run in multi-window mode.

This PR takes advantage of changes made on SDL side which have been applied as a patch to `python-for-android` SDL build: https://github.com/kivy/python-for-android/pull/2730.

After these changes, orientation can be controlled in two ways:

`--orientation` will control the allowed orientations (`portrait`, `landscape`, `portrait-reverse`, `landscape-reverse`). The allowed orientations list is then set to `KIVY_ORIENTATION`, which is used during window setup time to populate the `SDL_HINT_ORIENTATIONS` value, as we're doing for other platforms.

`--manifest-orientation` has been added to keep the `android:screenOrientation` personalisation functionality, even if Android will be likely deprecate this attribute in future. Valid values can be find [here](https://developer.android.com/guide/topics/manifest/activity-element.html).

If `--manifest-orientation` is not set, and only (one of multiple) `--orientation` options are passed, the value for `android:screenOrientation` is guessed and synthesised from the `--orientation` option. Since `android:screenOrientation` accepts only 1 value, if multiple `--orientation` are given, `android:screenOrientation` will be set to `unspecified`.

A `buildozer` PR will follow.

P.S. : I'm slowly changing `pythonforandroid/bootstraps/common/build/build.py` in order to make it deeply testable, so some changes may seem to be unrelated, but are needed in order to actually test the code.

⚠️ The changes applied in this PR will likely break existing third party examples and automated builds, if a now unsupported orientation setting is used, so the user will need to migrate, if needed. Specifically, `--orientation sensor` or `--orientation user` are not valid anymore, and an error will be raised, and the user will be forced to migrate.

### DIFF
--- a/doc/source/buildoptions.rst
+++ b/doc/source/buildoptions.rst
@@ -57,12 +57,17 @@ options (this list may not be exhaustive):
 - ``--package``: The Java package name for your project. e.g. ``org.example.yourapp``.
 - ``--name``: The app name.
 - ``--version``: The version number.
-- ``--orientation``: Usually one of ``portait``, ``landscape``,
-  ``sensor`` to automatically rotate according to the device
-  orientation, or ``user`` to do the same but obeying the user's
-  settings. The full list of valid options is given under
-  ``android:screenOrientation`` in the `Android documentation
-  <https://developer.android.com/guide/topics/manifest/activity-element.html>`__.
+- ``--orientation``: The orientations that the app will display in.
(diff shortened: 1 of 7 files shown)


AndreMiras (approved):
Neat, thanks! :clap:

PR #2746, opened 2023-02-10
restrict sh version

HyTurtle:
keep sh version under 2.0 as default behaviour was switched

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ data_files = []
 # https://github.com/kivy/buildozer/issues/722
 install_reqs = [
     'appdirs', 'colorama>=0.3.3', 'jinja2',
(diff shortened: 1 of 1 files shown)


misl6:
Hi @HyTurtle !

Can you please explain why this is needed?

HyTurtle:
https://github.com/amoffat/sh/blob/develop/MIGRATION.md#return-value-now-a-true-string

it now defaults to a string #2745 #2593 related

misl6 (approved):
Thank you for explaining!
Let's wait for the CI pipeline to complete, but LGTM.

I guess we should have an emergency release this weekend.

I will then open an issue to remind us that `sh.Command` usage should be migrated to be future-proof.

HyTurtle:
Sounds like a plan.

Closes #2745

PR #2750, opened 2023-02-13
Add PyAV recipe

DexerBR:
[PyAV](https://github.com/PyAV-Org/PyAV) - is a Pythonic binding for the [FFmpeg](http://ffmpeg.org/) libraries.

I also added a recipe for the codecs in a similar way to ffpyplayer.

### DIFF
--- a/pythonforandroid/recipes/av/__init__.py
+++ b/pythonforandroid/recipes/av/__init__.py
@@ -5,8 +5,8 @@ from pythonforandroid.recipe import CythonRecipe
 class PyAVRecipe(CythonRecipe):
 
     name = "av"
-    version = "v10.0.0"
-    url = "https://github.com/PyAV-Org/PyAV/archive/v10.0.0.zip"
+    version = "10.0.0"
+    url = "https://github.com/PyAV-Org/PyAV/archive/v{version}.zip"
 
     depends = ["python3", "cython", "ffmpeg", "av_codecs"]
     opt_depends = ["openssl"]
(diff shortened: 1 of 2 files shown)


pythonforandroid/recipes/av/__init__.py
```diff
@@ -0,0 +1,25 @@
+from pythonforandroid.toolchain import Recipe
+from pythonforandroid.recipe import CythonRecipe
+
+
+class PyAVRecipe(CythonRecipe):
+
+    name = "av"
+    version = "v10.0.0"
+    url = "https://github.com/PyAV-Org/PyAV/archive/v10.0.0.zip"
```
AndreMiras, line 9: The `{version}` should be used in the URL so it can be overridden dynamically.
```suggestion
    version = "10.0.0"
    url = "https://github.com/PyAV-Org/PyAV/archive/v{version}.zip"
```

pythonforandroid/recipes/av_codecs/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class PyAVCodecsRecipe(Recipe):
+    depends = ["libx264", "libshine", "libvpx"]
+
+    def build_arch(self, arch):
+        pass
+
+
+recipe = PyAVCodecsRecipe()
```
AndreMiras, line 11: Is this recipe doing anything at all then?

AndreMiras (commented):
Thanks for the PR, I left some comments/questions

pythonforandroid/recipes/av_codecs/__init__.py
DexerBR: The main intention is to keep the concept between PyAV and FFPyPlayer codecs distinct. That way, if there are any modifications to the PyAV codecs recipe, they will not interfere with FFPyPlayer.

In short, each recipe (PyAV/FFPyPlayer) in its own context.

AndreMiras (approved):
LGTM thanks

PR #2751, opened 2023-02-16
Add debug loggings for identifying a matching dist

BitcoinWukong:
Sometimes the android dist is being rebuilt every time, and it's hard to understand why that happens. This PR adds some debug logging so that it's easier to investigate when this happens.

In my case, I incorrectly used `Cython` and `Pillow`  in the `requirements` of my `buildozer.spec`, and this caused my dist to be rebuild every time, because `dist.recipes` would have `cython` and `pillow` in it, and p4a thinks it's missing the recipes `Cython` and `Pillow`.

### DIFF
--- a/pythonforandroid/distribution.py
+++ b/pythonforandroid/distribution.py
@@ -140,13 +140,13 @@ class Distribution:
         # If any dist has perfect recipes, arch and NDK API, return it
         for dist in possible_dists:
             if force_build:
-                debug(f"Skipping dist due to forced build")
+                debug("Skipping dist due to forced build")
                 continue
             if ndk_api is not None and dist.ndk_api != ndk_api:
-                debug(f"Skipping dist due to ndk_api mismatch")
+                debug("Skipping dist due to ndk_api mismatch")
                 continue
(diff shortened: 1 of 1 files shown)


pythonforandroid/distribution.py
```diff
@@ -133,17 +140,24 @@ def get_distribution(
         # If any dist has perfect recipes, arch and NDK API, return it
         for dist in possible_dists:
             if force_build:
+                debug(f"Skipping dist due to forced build")
                 continue
             if ndk_api is not None and dist.ndk_api != ndk_api:
+                debug(f"Skipping dist due to ndk_api mismatch")
                 continue
             if not all(arch_name in dist.archs for arch_name in archs):
+                debug(f"Skipping dist due to arch mismatch")
```
AndreMiras, line 149: Minor: linter may complain about that f-string

AndreMiras (dismissed):
Yes why not, thanks for the context.
Let's see if the CI is happy about it

AndreMiras (approved):
LGTM, let's see if the CI is all good before merging

PR #2765, opened 2023-03-06
android_api to integer

kuzeyron:
Smaller fix when `android_api` was given as a string but integer is needed and expected.

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -497,15 +497,9 @@ main.py that loads it.''')
     # Find the SDK directory and target API
     with open('project.properties', 'r') as fileh:
         target = fileh.read().strip()
-    android_api = int(target.split('-')[1])
-    try:
-        int(android_api)
-    except (ValueError, TypeError):
-        raise ValueError(
-            "failed to extract the Android API level from " +
-            "build.properties. expected int, got: '" +
-            str(android_api) + "'"
-        )
+    aapi = target.split('-')[1]
+    android_api = int(aapi) if aapi.isdigit() else -1
+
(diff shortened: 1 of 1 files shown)


pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -497,7 +497,7 @@ def make_package(args):
     # Find the SDK directory and target API
     with open('project.properties', 'r') as fileh:
         target = fileh.read().strip()
-    android_api = target.split('-')[1]
+    android_api = int(target.split('-')[1])
     try:
         int(android_api)
```
misl6, line 502:
```suggestion
    android_api = target.split('-')[1]
    try:
        android_api  = int(android_api)
```

misl6: Just after this call, we have a `try/except` block that checks if the `android_api` string value is parseable as an int.

Parsing the int from string before that call will lead to a crash, without raising the needed error.

misl6 (changes requested):
Hi @kuzeyron !

I've left a suggestion + a comment regarding it.

pythonforandroid/bootstraps/common/build/build.py
```diff
@@ -497,15 +497,9 @@ def make_package(args):
...
         target = fileh.read().strip()
-    android_api = target.split('-')[1]
-    try:
-        int(android_api)
-    except (ValueError, TypeError):
-        raise ValueError(
-            "failed to extract the Android API level from " +
-            "build.properties. expected int, got: '" +
-            str(android_api) + "'"
-        )
```
misl6, line 508: Nice, but maybe we can keep the error message?

misl6 (approved):
LGTM. Thank you!

PR #2769, opened 2023-03-16
fix #2768 smallIconName null can t be compared to String

### LINKED ISSUE #2768: Service don t start anymore, as smallIconName extra is now mandatory
https://github.com/kivy/python-for-android/blob/8cb497dd89e402478011df61f4690b963a0c96da/pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonService.java#L116

```java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.equals(java.lang.Object)' on a null object reference```

We could test if null before.

brvier:


misl6 (approved):
LGTM. Thank you!

PR #2771, opened 2023-03-31
fix tools/liblink: syntax error

SomberNight:
Looks like a regression from 22b7dfee1ddc5f435f8bf016cbf83980a45b4374

To be honest I am not sure how this file is used. Maybe it could be removed altogether seeing as no one complained for three years.


AndreMiras (approved):
Good catch!
Will merge after the CI is done and happy

PR #2784, opened 2023-04-27
Use build rather than pep517 for building

s-t-e-v-e-n-k:
pep517 has been renamed to pyproject-hooks, and as a consequence all of the deprecated functionality has been removed. build now provides the functionality required, and since we are only interested in the metadata, we can leverage a helper function for that. I've also removed all of the subprocess machinery for calling the wrapping function, since it appears to not be as noisy as pep517.


misl6 (approved):
LGTM. Thank you!

PR #2830, opened 2023-06-02
Remove `pytz` recipe, as it's not needed anymore

misl6:
- `pytz` recipe is not needed anymore, as is a dependency that could be installed as-is, from PyPi.
- If we need to install a dependency on `hostpython3`(e.g. to perform builds), that should be handled in another way. (or the installation could be performed from the recipe of the dependants)


AndreMiras (approved):
I'm thinking should the `PythonRecipe` return a warning or something if we don't recommend using recipes for pure Python modules?

misl6:
> I'm thinking should the `PythonRecipe` return a warning or something if we don't recommend using recipes for pure Python modules?

`PythonRecipe` itself is fine e.g. if the package needs some patching, so the warning may be too noisy.
Maybe we should document it somewhere, and have some rules of what can be added as a recipe?

PR #2833, opened 2023-06-04
Optimize CI runs, by avoiding unnecessary rebuilds

misl6:
Well, parallelizing builds, even if most of the work is duplicated, is usually good. (If you're not paying for CI runs)

This is something valid also for `python-for-android` artifacts builds, as parallelizing `APK` and `AAB` builds takes ~14 minutes less than a build that performs APK (debug) + AAB (release) build in the same run.

But ... even if we're not paying for CI runs (thank you GitHub), workers are capped at a maximum concurrent number (and in the case of our Apple Silicon M1 self-hosted runner, only one instance per time is available), which means that other jobs need to wait the completion of previous ones. (making it slower than non-parallelized builds)

Some data, on Ubuntu:
- APK + AAB: `1h 19m 47s` (keeps only 1 instance locked for the entire build)
- APK `1h 5m 21s` + AAB `1h 8m 26s` (keeps 2 instances locked for the entire build)

This PR merges APK and AAB builds, avoiding unnecessary rebuilds.

A lot more could be done to improve our CI wait time, but that's a start.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -53,8 +53,8 @@ jobs:
         parallel: true
         flag-name: run-${{ matrix.os }}-${{ matrix.python-version }}
 
-  ubuntu_build_apk:
-    name: Unit test apk [ ${{ matrix.runs_on }} | ${{ matrix.bootstrap.name }} ]
+  ubuntu_build:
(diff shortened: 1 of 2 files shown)


AndreMiras (approved):
LGTM thanks

Issue #2825, opened 2023-05-31
Enhancement request: cleanup recipe names


RobertFlatt:
### Description

Having worked to get requirements correct, users find the recipe names are different for no apparent reason. WTF?!

And scrolling the recipes directory, looking for a clue, is a challenge because there are also libraries and misnamed items in there.

Its confusing.

Cleanup the names, and in p4a implement a lookup table for backwards compatibility.

```
Wrong name wrt Pypi

dateutil        is python-dateutil
mysqldb         is mysql-python       - this claims to only support Python2
ndghttpsclient  is ndg-httpsclient
opencv          is opencv-python
opencv_extras   is opencv-contrib-python
protobuf_cpp    is protobuf-python
pyleveldb       is leveldb
spine           is spine-cython
```

```
Conflicting Pypi name:

av recipe is not av package, correct name is pyav , av is a different package
```

```
Some oddballs, not clear how a user would know what these are (or how to use them)

audiostream
cymunk
pil
vlc
```

```
Libs that are not named libXXXX (perhaps all the libs could be in a different directory?)

freetype 
harfbuzz
ifaddrs
jepd
openal
openssl
sqlite3
```

misl6:
> av recipe is not av package, correct name is pyav , av is a different package

https://pypi.org/project/av/ isn't the same we support under the `av` recipe?

RobertFlatt:
Sorry, my error. You are correct.

misl6:
Current status:

`dateutil` -> `python-dateutil`
`mysqldb` -> `mysql-python`
`ndghttpsclient` -> `ndg-httpsclient`
`opencv` -> `opencv-python`
`opencv_extras` ->  `opencv-contrib-python`
`protobuf_cpp` ->  `protobuf-python`
`pyleveldb` ->  `leveldb`
`spine` -> `spine-cython`

PR #2839, opened 2023-06-15
change pygame URL to maintained repo, new version

robertpfeiffer:
PyGame development has moved to a new address.


AndreMiras (commented):
Is it really the same software though?
I feel like it deserves its own recipe

misl6:
As suggested via #2825 , recipes should have the same name as the package available on PyPi, and https://github.com/pygame-community/pygame-ce suggest that the correct package name is `pygame-ce`

Maybe @robertpfeiffer can consider trying #2838 ?

PR #2846, opened 2023-06-30
Changed the url to use https as http fails

kuzeyron:
Changed the url to use https as http fails.


misl6 (approved):
`http` works on my side, but why not using `https` when available?

LGTM. Thank you!

kuzeyron:
> `http` works on my side, but why not using `https` when available?
> 
> LGTM. Thank you!

Reason: [Support0 conversation](https://discord.com/channels/423249981340778496/423250272316293120/1124366348655542372)

PR #2852, opened 2023-07-10
Correct sys_platform

Julian-O:
On Window, sys.platform = "win32".

I think "nt" is a reference to os.name.


misl6 (approved):
Nice catch!

LGTM. Thank you!

PR #2857, opened 2023-07-20
Use platform.uname instead of os.uname

Julian-O:
Advantages:

- Works cross platform, not just Unix.
- Is a namedtuple, so can use meaningful  fieldnames.

Also snuck in correction to typo in Readme which doesn't warrant its own review.

[Note: Unit-tests failed, but I don't *think* it had anything to do with this.]

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ to install and begin creating APKs and AABs.
 
 (for the develop branch: `pip install git+https://github.com/kivy/python-for-android.git`)
 
-Test that the install works with:
+Test that theinstall works with:
 
     p4a --version
 
--- a/pythonforandroid/util.py
+++ b/pythonforandroid/util.py
@@ -1,6 +1,7 @@
 import contextlib
 from os.path import exists, join
-from os import getcwd, chdir, makedirs, walk, uname
+from os import getcwd, chdir, makedirs, walk
+from platform import uname
 import shutil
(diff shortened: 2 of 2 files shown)


misl6 (approved):
CI builds are failing (are now fixed via https://github.com/kivy/python-for-android/pull/2862), but everything is clearly handled as expected on non-sdl2 jobs, so there's no need for another run in this case. (like #2858)

LGTM. Thank you!

README.md
```diff
@@ -39,7 +39,7 @@ to install and begin creating APKs and AABs.
 
 (for the develop branch: `pip install git+https://github.com/kivy/python-for-android.git`)
 
-Test that the install works with:
+Test that theinstall works with:
```
AndreMiras, line 42: oops

misl6: Ouch!
For some reason I've read: "Test that install works with:" 🫣

@Julian-O can you take care of it?

Julian-O: I just spent some time trying to figure out how I managed to break this comment.

1. Fix a legitimate typo in a comment of no consequence.
2. Accidentally delete a space in another file.
3. A few days later, at commit time, see that delta for the lost space, but don't notice it is the wrong way. Think "Ah, yeah, the typo I fixed. I will include that in this PR."

Sorry. PR with both inconsequential fixes coming very soon.

PR #2858, opened 2023-07-20
Automatically generate required pre-requisites

Julian-O:
`get_required_prerequisites()` maintains a list of Prerequisites required by each platform.

But that same information is already stored in each Prerequisite class.

Rather than maintaining two lists which might become inconsistent, auto-generate one.

[Note: Unit tests fail, but I don't think it is a result of this change,]


misl6 (approved):
CI builds are failing (but are now fixed via https://github.com/kivy/python-for-android/pull/2862), but the prerequisites are clearly managed as expected, so there's no need for another run in this case.

LGTM. Thank you!

PR #2864, opened 2023-07-22
`build_platform` should be all-lowercase

misl6:
`build_platform` should be all-lowercase. 

FYI: @Julian-O , the issue has been introduced via #2857 (CI wasn't working as expected, and was hard to see)

Now that CI is working as expected, see: https://github.com/kivy/python-for-android/actions/runs/5630635544/job/15256832012?pr=2863


AndreMiras (approved):

Julian-O:
Nice catch.

PR #2866, opened 2023-07-24
Factor out dependency checking. Use modern version handling

Julian-O:
`toolset.py` uses the class `LooseVersion` from `distutils` to implement comparison of versions.

[Distutils](https://docs.python.org/3/library/distutils.html) is deprecated, and won't be in Python 3.12 which is [< 3 months away](https://peps.python.org/pep-0693/).

As a result, deprecations warnings are appearing in p4a's logs.

This PR moves over to use `packaging.Version`. `packaging` is a already an install_req.

Note: `packaging.Version` is a lot more opinionated about what a version number must look like. The current Python dependencies that are checked are [sh](
https://pypi.org/project/sh/#history_) and [colorama](https://pypi.org/project/colorama/#history), both of which adhere to this versioning standard, so there is no issue there.

But, Android SDK versions sometimes contain spaces, so that code has become more complex.

This presented an opportunity to move the  dependency-checking out of `toolchain.py` (another 1300-line file) into its own file.

[This area is expected to need an overhaul to eventually support different python dependencies on different platforms, and Windows Prerequisites.]

Which presented an opportunity to cleanup the imports to follow PEP8 (alphabetical order, standard-then-3rd-party-then-local).

I have added no additional unit-tests. It is tricky to work out a good way to test this.

### DIFF
--- a/pythonforandroid/checkdependencies.py
+++ b/pythonforandroid/checkdependencies.py
@@ -1,0 +1,70 @@
+from importlib import import_module
+from os import environ
+import sys
+
+from packaging.version import Version
+
+from pythonforandroid.prerequisites import (
+    check_and_install_default_prerequisites,
+)
+
+
+def check_python_dependencies():
+    """
+    Check if the Python requirements are installed. This must appears
+    before other imports because otherwise they're imported elsewhere.
+
+    Using the ok check instead of failing immediately so that all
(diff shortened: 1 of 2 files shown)


misl6 (approved):
LGTM. Thank you!

PR #2869, opened 2023-07-27
Python versions: Update documentation & CI testing

Julian-O:
Python 3.11 was released October 2022.
Python 3.7 went end-of-life June 2023

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -32,7 +32,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


AndreMiras (approved):
LGTM, thanks.
I see a few more references of 3.7 within the code base, but I think you're right in not acting on them.
Some are for testing dummy recipes e.g. in `tests/test_recipe.py`, other are to provide support for 3.7 e.g. `pythonforandroid/recipes/python3/__init__.py`. The rest are just comments.
On the second one, it's probably OK to keep indeed as it's just a couple of patches and we don't want to purposely break 3.7 even though we may not invest our time on it if it were to break

Issue #2808, opened 2023-05-30
libmysqlclient recipe build fail


RobertFlatt:
### Logs

```
make[1]: *** [CMakeFiles/Makefile2:999: unittest/libmysql/CMakeFiles/errors.dir/all] Error 2
ld: error: undefined symbol: endpwent
>>> referenced by mf_pack.c
>>>               mf_pack.c.o:(unpack_dirname) in archive ../libmysql/libmysqlclient.a
clang-14: error: linker command failed with exit code 1 (use -v to see invocation)
```

AndreMiras:
I think I've got a different errors in my last CI build https://github.com/kivy/python-for-android/actions/runs/14010036478/job/39228657332
```
...
/home/user/.local/share/python-for-android/build/other_builds/libmysqlclient/x86__ndk_target_24/libmysqlclient/libmysqlclient/extlib/yassl/taocrypt/src/arc4.cpp:163:5: error: invalid operand for instruction
    PROLOG()
    ^
...
/home/user/.local/share/python-for-android/build/other_builds/libmysqlclient/x86__ndk_target_24/libmysqlclient/libmysqlclient/extlib/yassl/taocrypt/src/arc4.cpp:172:5: error: unexpected token in argument list
    AS2(    movzx  edx, BYTE PTR [ecx + 1]  )   // y
    ^
...
```
But yes this recipe deserve some love, it's a big mess.
The recipe isn't even pinned to a given version, no wonder why errors keep changing

Edit: I think it's time to delete this recipe https://github.com/kivy/python-for-android/pull/3130

PR #2871, opened 2023-07-31
Standardise ensure_dir and rmdir

Julian-O:
1) A new version of Flake8 dropped while I was submitting this PR. There are several inconsequential edits to unrelated files, because Flake8 held my tests at gunpoint until I did it.

2) When p4a wanted to create a directory, it used any of several different techniques:
	* it called os.mkdirs()
	* it called shutil.mkdir()
	* it called util.ensure_dir()
	* it called bootstraps.common.build.build.ensure_dir()
	* it called sh.mkdir()
	
The latter has terrible performance and isn't cross-platform.

* I modified util.ensure_dir() to log if and only if it made a change.
* I modified all directory creations to use util.ensure_dir().
* I removed any now unnecessary logs and existence checks from the code, so the clients are now simpler.

3) When p4a wants to remove a directory, it used any of several different techniques:
	* it called shutil.rmtree
	* it called sh.rm("-r")
	* it called subprocess.check_output(["mkdir"..])

The latter two have terrible performance and aren't cross-platform.
	
* I created util.rmdir(). It logs if and only if it makes a change.
* I modified all directory deletions (outside of test code) to use util.rmdir().
* I removed any now unnecessary logs and existence checks from the code, so the clients are now simpler.
* recipe.py had some weird code to delete then create then delete a folder. I simplified it to delete.

4) Whenever I had to touch the imports, I sorted them to make them PEP8 compliant.

5) A couple of the receipes (e.g. sqlite3) weirdly imported shutil indirectly, via toolchain, which caused a problem when toolchain stopped importing it (no longer required). Changed the recipe code to import shutil directly.

Note I only tackled two system calls; I didn't even look at removing single files. There are more to come, but I want to keep each review simple, and this one is already too complex.

### DIFF
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -1,6 +1,5 @@
 from os.path import basename, dirname, exists, isdir, isfile, join, realpath, split
 import glob
-from shutil import rmtree
 
 import hashlib
 from re import match
@@ -10,7 +9,7 @@ import shutil
 import fnmatch
 import urllib.request
 from urllib.request import urlretrieve
-from os import listdir, unlink, environ, mkdir, curdir, walk
+from os import listdir, unlink, environ, curdir, walk
 from sys import stdout
 import time
 try:
@@ -19,7 +18,7 @@ except ImportError:
     from urllib.parse import urlparse
 from pythonforandroid.logger import (logger, info, warning, debug, shprint, info_main)
 from pythonforandroid.util import (current_directory, ensure_dir,
-                                   BuildInterruptingException)
+                                   BuildInterruptingException, rmdir)
 from pythonforandroid.util import load_source as import_recipe
 
 
@@ -218,7 +217,7 @@ class Recipe(metaclass=RecipeMeta):
                     url = url[4:]
                 # if 'version' is specified, do a shallow clone
                 if self.version:
-                    shprint(sh.mkdir, '-p', target)
+                    ensure_dir(target)
                     with current_directory(target):
                         shprint(sh.git, 'init')
                         shprint(sh.git, 'remote', 'add', 'origin', url)
@@ -367,7 +366,7 @@ class Recipe(metaclass=RecipeMeta):
             if expected_digest:
                 expected_digests[alg] = expected_digest
 
-        shprint(sh.mkdir, '-p', join(self.ctx.packages_path, self.name))
+        ensure_dir(join(self.ctx.packages_path, self.name))
 
         with current_directory(join(self.ctx.packages_path, self.name)):
             filename = shprint(sh.basename, url).stdout[:-1].decode('utf-8')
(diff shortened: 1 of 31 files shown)


misl6:
`libmysqlclient` is known to be broken (see: https://github.com/kivy/python-for-android/issues/2808). Can you please add it to `BROKEN_RECIPES_PYTHON3` (`ci/constants.py`) so we can see ✅ on "Updated recipes" checks?

Julian-O:
I don't pretend to understand what the implications of that request, but I went ahead and did it.

misl6:
> I don't pretend to understand what the implications of that request, but I went ahead and did it.

Build tests are performed on pre-defined demo apps (always) and on updated recipes (when a recipe gets updated).

Since you edited `libmysqlclient` recipe file, the CI was trying to build it as part of the "Build Updated Recipes" test.
But, unfortunately, this recipe is broken (since a while), and the whole "Build Updated Recipes" CI pipeline failed.

Adding `libmysqlclient` to `BROKEN_RECIPES_PYTHON3` allows us skip `libmysqlclient` build during the "Build Updated Recipes" test phase.

misl6:
Hi @Julian-O !

Sorry for the late reply, but I was on vacation 😀📷

Seems we still have some issues to perform the build as `boost` recipe is not working anymore. (`boost` is required by `libtorrent`, which has been edited by this PR)

Similarly to `libmysqlclient`, can you please add `boost` and `libtorrent` to `BROKEN_RECIPES_PYTHON3` ?

pythonforandroid/recipe.py
```diff
@@ -423,9 +422,7 @@ def unpack(self, arch):
                 self.name.lower()))
             if exists(self.get_build_dir(arch)):
                 return
-            shprint(sh.rm, '-rf', build_dir)
-            shprint(sh.mkdir, '-p', build_dir)
-            shprint(sh.rmdir, build_dir)
+            rmdir(build_dir)
             ensure_dir(build_dir)
```
misl6, line 426: Uh 😅

misl6 (approved):
Excellent.
LGTM. Thank you!

NomadDemon:
how about to migrate everything to "pathlib" from python ? :) it would be also nice improvement

Julian-O:
@NomadDemon: I agree. Given the amount of path-processing involved, pathlib would be a boon.

I am trying to keep my changes small and incremental to make reviewing easier and to reduce the chance that I break anything. I already have a lot on my plate. So I haven't included it here.

But what I *have* done is make the new Buildops library compatible with path-like parameters, so it is easier to migrate the clients that use it in the future.

Julian-O:
@NomadDemon: Sorry. Ignore that last sentence. That refers to Buildozer, not Python-For-Android.

PR #2884, opened 2023-08-26
Standardise on move

Julian-O:
* Add new util function (and tests) called `move`.
* Change references to sh.mv to use move (as it is faster and cross-platform).
* Change conditional "mv -t" to a for-loop.
* This altered imports which affected rmtree instances too, so cleaned that up.


misl6 (approved):
LGTM. Thank you!

PR #2890, opened 2023-09-14
Add doc tests, make them pass.

Julian-O:
* push.yml
   * Added documentation step which installs [docs], checks links and builds documentation to find any RST file errors.
   * Need to install to get all the dependencies, because p4a code is imported during doc generation.
   * Updated rest of checkout action versions to be consistent with new code.

* apis.rst
   * http -> https
   * Disambiguated two modules with the same name.

* buildoptions.rst
   * Corrected indenting

* conf.py
   * Added language
   * Added two URLS to ignore; link-checker was failing them even though they worked.

* ext/sphinix_rtd_theme/*
   * Old theme did not work with Sphinx 7.0+.
   * Removed it from source control. Install the latest each time.

* launcher.rst
   * This described an old feature that hasn't worked for years. Couldn't fix the links because the whole feature is gone now. Removed it. If the attempts to bring the feature back ever come to fruition, this can be restored and edited.

* index.rst
   * Removed no-longer-existing anchor
   * Removed launcher page

* quickstart.rst
   * Removed launcher

* recipe.rst
   * Fixed syntax error in rst.

* services,rst
   * Fixed broken URL (and linked to project rather than bug-tracker.)

* troubleshooting.rst
   * http->https
   * Link to English rather than Chinese web-sites.

* setup.py
   * Add docs as extra.


doc/source/launcher.rst
```diff
@@ -1,101 +0,0 @@
-.. _launcher:
-
-Launcher
-========
-
-The Kivy Launcher is an Android application that can run any Kivy app
-stored in the `kivy` folder on the SD Card. You can download the latest stable
-version for your android device from the
-`Play Store <https://play.google.com/store/apps/details?id=org.kivy.pygame>`_.
```
misl6, line 9: FYI:
There's a still-in-WIP reboot of the old Kivy launcher, hosted here: https://github.com/kivy/kivy-launcher
 
 But, I agree that launcher docs in `python-for-android` are old and unmaintained.
 
 Also, if the new launcher will do a comeback into the Kivy ecosystem, I feel we should advertise it on kivy.org website, and not into `python-for-android` docs. (as `kivy-launcher` will likely need to have its own docs)

doc/source/apis.rst
```diff
@@ -193,7 +193,7 @@ Example::
 Activity lifecycle handling
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. module:: android.activity
+.. module:: android.activitylifecycle
```
misl6, line 196: `ActivityLifecycleCallbacks` are implemented into `android.activity` module.

See:
https://github.com/kivy/python-for-android/blob/4d50d7d883bb6313f37c22402ecfaefa689bbbed/pythonforandroid/recipes/android/src/android/activity.py#L71

`android.activitylifecycle` module does not exists. Better to just avoid this reference.

misl6 (changes requested):

Julian-O:
Just a note to say haven't forgotten this. On my to-do list to make the changes requested,

misl6:
> Just a note to say haven't forgotten this. On my to-do list to make the changes requested,

Don't worry!

doc/source/launcher.rst
Julian-O: Yes, all agreed. The WIP reboot isn't mature enough to be referenced here, and seems to have stalled.

misl6 (approved):
LGTM. Thank you!

PR #2900, opened 2023-10-17
Update Lottie player version

HugoDaniel:
### Description

Updates the Lottie player version to [the latest 6.1.0](https://github.com/airbnb/lottie-android/releases).

The current version in use (3.4.0) is from 2020, and among the many issues that have already have been fixed since then, it does not allow for color updates in keyframes[0], which is a very common operation when changing Lotties that already exist out there in the public domain.

[0] https://stackoverflow.com/questions/74912770/lottie-animation-error-after-change-color-of-animation


AndreMiras (approved):
LGTM, thanks for the extra context

PR #2915, opened 2023-11-09
enable json1 extenstion in sqlite3

HyTurtle:
Prior to version 3.38.0 the json1 extension is opt-in rather than opt-out. Build flag added to include the extension


NomadDemon:
i can confirm, it works with this flag. e.g peewee can use "jsonfield" on SQLite with this flag enabled

there is no impact on size of .SO still 1.3 mb

AndreMiras (approved):
LGTM

PR #2917, opened 2023-11-10
Introduce  FAQ

Julian-O:
This is largely stolen and copy-edited from python-for-android trouble-shooting section, and Kivy's FAQ.

Future work:

* Remove duplicate info from trouble-shooting section.
* Remove duplicate info from Kivy framework FAQ.
* Link to this from Kivy framework FAQ.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -3,7 +3,7 @@
 ## Introduction
 
 [python-for-android](https://python-for-android.readthedocs.io/) is an open
-source build tool to let you package Python code into standalone android APKs.
+source build tool to let you package Python code into standalone Android artifacts (APKs, AABs, AARs).
 These can be passed around, installed, or uploaded to marketplaces such as the
 Play Store just like any other Android app.
 
@@ -31,13 +31,6 @@ lines to the xml configuration, and you're done.
 
 The following are common problems and resolutions that users have reported.
 
-#### AttributeError: ‘AnsiCodes’ object has no attribute ‘LIGHTBLUE_EX’
-
-This occurs if your version of `colorama` is too low, install version 0.3.3 or
-higher.
-
-If you install python-for-android with pip or via setup.py, this dependency
-should be taken care of automatically.
 
 #### AttributeError: ‘Context’ object has no attribute ‘hostpython’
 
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -0,0 +1,91 @@
...
+
+The following are common problems and resolutions that users have reported.
+
+#### AttributeError: ‘AnsiCodes’ object has no attribute ‘LIGHTBLUE_EX’
+
+This occurs if your version of `colorama` is too low, install version 0.3.3 or
+higher.
+
+If you install python-for-android with pip or via setup.py, this dependency
+should be taken care of automatically.
```
misl6, line 40: I guess we can remove this one, definitely outdated.

```suggestion
```

FAQ.md
```diff
@@ -0,0 +1,91 @@
...
+desirable anyway since you don’t want it in the APK.
+
+#### Requested API target 19 is not available, install it with the SDK android tool
+
+This means that your SDK is missing the required platform tools. You need to
+install the `platforms;android-19` package in your SDK, using the android or
+sdkmanager tools (depending on SDK version).
+
+If using buildozer this should be done automatically, but as a workaround you
+can run these from `~/.buildozer/android/platform/android-sdk-20/tools/android`.
```
misl6, line 64: Better to keep it more generic, like:

```suggestion
#### Requested API target XX is not available, install it with the SDK android tool

This means that your SDK is missing the required platform tools. You need to
install the `platforms;android-XX` package in your SDK, using the android or
sdkmanager tools (depending on SDK version).

If using buildozer this should be done automatically, but as a workaround you
can run these from `~/.buildozer/android/platform/android-sdk-XX/tools/android`
```

FAQ.md
```diff
@@ -0,0 +1,91 @@
...
+
+#### ModuleNotFoundError: No module named ‘_ctypes’
+
+You do not have the libffi headers available to python-for-android, so you need
+to install them. On Ubuntu and derivatives these come from the `libffi-dev`
+package.
+
+After installing the headers, clean the build (p4a clean builds, or with
+buildozer delete the .buildozer directory within your app directory) and run
+python-for-android again.
```
misl6, line 74: I hope that got fixed when `libffi` package was added (and looking at the code it should be) otherwise if we're really using a system-wide dependency for cross-compiling, that's bad.

Better to check, but should not stay in FAQs.

FAQ.md
```diff
@@ -0,0 +1,91 @@
...
+
+On Ubuntu and derivatives:
+
+    apt install libssl-dev
+    p4a clean builds # or with: buildozer `buildozer android clean
+
+On macOS:
+
+    brew install openssl
+    sudo ln -sfn /usr/local/opt/openssl /usr/local/ssl
```
misl6, line 90: `sudo ln -sfn /usr/local/opt/openssl /usr/local/ssl` is not needed anymore, and better to not mess with the developer `/usr/**` folders.
```suggestion
```

misl6 (commented):
I've added some suggestions regarding the wording (and did some cleanup).

Let me know what do you think about it.

FAQ.md
AndreMiras: Yeah agree, we should either remove it completely or say something like:
> There was an old bug/limitation that required to install system-wide libffi-dev package, but this was fixed with the libffi recipe, if you still see this error make a bug report.

Julian-O: If it is old enough, it is no longer a frequently asked question. I am in favour of removing it.

Julian-O:
Thanks for great suggestions. Requested changes made.

misl6 (approved):
LGTM. Thank you!

PR #2923, opened 2023-11-18
Update `python-for-android` prerequisites (`Dockerfile`, `prerequisites.py`, docs)

misl6:
- Updates `Dockerfile` (clean up of old requirements and change of base image)
- Set JDK version `17` as a prerequisite, newer JDK versions such as `21` are incompatible with our version of `gradle`, and `gradle` now requires at least `JDK 17`
- Update docs to reflect the current status (ouch, docs were super-outdated)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@
 # If platform is not specified, by default the target platform of the build request is used.
 # This is not what we want, as Google doesn't provide a linux/arm64 compatible NDK.
 # See: https://docs.docker.com/engine/reference/builder/#from
-FROM --platform=linux/amd64 ubuntu:20.04
+FROM --platform=linux/amd64 ubuntu:22.04
 
 # configure locale
 RUN apt -y update -qq > /dev/null \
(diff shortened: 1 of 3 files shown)


misl6:
Hi @Julian-O ,
Can you please take a look at docs changes?

Dockerfile
```diff
@@ -48,43 +48,37 @@ ENV HOME_DIR="/home/${USER}"
...
+        patch \
+        unzip \
+        zip \
+        pkg-config \
+        python3 \
+        python3-dev \
+        python3-pip \
+        python3-venv \
+        make \
+        sudo \
```
AndreMiras, line 80: This should ideally be ordered alphabetically, making it easier to find dups for instance
https://docs.docker.com/develop/develop-images/guidelines/#sort-multi-line-arguments

doc/source/quickstart.rst
```diff
@@ -56,48 +56,45 @@ You can also test the master branch from Github using::
...
+        unzip \
+        zip \
+        pkg-config \
+        python3 \
+        python3-dev \
+        python3-pip \
+        python3-venv \
+        make \
+        sudo \
+        wget
```
AndreMiras, line 96: Alphabetical order would be a plus here too

Dockerfile
```diff
@@ -48,43 +48,37 @@ ENV HOME_DIR="/home/${USER}"
...
+        ant \
+        autoconf \
+        automake \
+        ccache \
+        cmake \
+        g++ \
+        gcc \
+        git \
+        lbzip2 \
+        libffi-dev \
```
AndreMiras, line 66: I'm wondering if this is still a thing, I thought you updated recipes to pull ffi compile deps

AndreMiras (approved):
LGTM

Dockerfile
misl6: Makes sense. re-ordered.

doc/source/quickstart.rst
misl6: Makes sense. re-ordered.

Dockerfile
misl6: This is for `hostpython3`, which is not cross-compiled. (same applies to `libssl-dev`).

This happens on `kivy-ios`, where we build our own version of `openssl`, even for `hostpython3` (`libffi`, for macOS comes with XCode, so there's no need to build `libffi`).

AndreMiras: Ah yeah makes sense, I forgot that thanks :+1:

AndreMiras (approved):

Julian-O:
Should we be mentioning:

* The need for Cython?
* The need for Cython<=3?
* The packages Cython needs? (It needs `python3-dev` which you have and `build-essential` which is probably covered by the individual packages, but I am not sure.)

Other than that, looks good.

[Reminder: I think we should be *testing* on an instance of this Docker, too.]

PR #2933, opened 2023-12-04
Update __init__.py with new download url. New version (v1.10.37)

excepterror:


excepterror:
Recipe builds successfully.

PR #2935, opened 2023-12-05
Remove redundant append into WHITELIST_PATTERNS

shyamnathp:
- pyconfig.h was appended twice into WHITELIST_PATTERNS
- Regression from commit a636b88


shyamnathp:
https://github.com/kivy/python-for-android/blob/develop/pythonforandroid/bootstraps/common/build/build.py#L86 already does it

feel free to abandon the patch if the removed line was indeed necessary and I missed something,

misl6 (approved):
LGTM. Thank you!

PR #2938, opened 2023-12-08
Remove redundant class variable definition from recipes

shyamnathp:
- Some of the variables like call_hostpython_via_targetpython, install_in_hostpython are redifined in the package recipes with the exact same value as its parent class, making the definitions redundant. This patch removes these redefinitions.


shyamnathp:
> Thanks for the cleanup @shyamnathp , however, the CI is failing.
> 
> The good news is that I'm 100% sure is not your fault, but in order to have a ✅ , can you please add `pygame` [to this list](https://github.com/kivy/python-for-android/blob/0df46c8ccce849da6eb11739e5063653f21352cb/ci/constants.py#L11) ?

As a part of understanding the logs, how did you figure out it was 'pygame' that was causing the crash? From the logs, I see `liblzma` seems to be the culprit and this does not seem to be a dependent recipe of 'pygame'

PR #2939, opened 2023-12-11
Allow text_pil to use latest version of pillow.

akshayaurora:
Update Pillow  To be merged after https://github.com/kivy/kivy/pull/8503


misl6 (changes requested):
Hi @akshayaurora !

The CI run is failing as one of the patches looks incompatible.
Can you take a look?

T-Dynamos:
In latest release version is up-to-date. And text_pil works well on android.

PR #2971, opened 2024-02-11
Pygame-ce on Android

AnGlonchas:
Added a pygame-ce recipe for newer developers, i saw the opportunity to contribute to this community, and i already used this on my project.


pythonforandroid/recipes/pygame-ce/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+        not part of the build. It's usable, but not complete.
+    """
+
+    version = '2.4.0'
+    url = 'https://github.com/pygame-community/pygame-ce/archive/{version}.tar.gz'
+
+    site_packages_name = 'pygame-ce'
+    name = 'pygame-ce'
+
+    depends = ['sdl2', 'sdl2_image', 'sdl2_mixer', 'sdl2_ttf', 'setuptools', 'jpeg', 'png']
```
pmp-p, line 22: sdl2_image could link "webp" dynamically not just jpeg/png. But probably not actually the case.

Julian-O:
I had considered raising an issue on Kivy itself to support Pygame-CE as well as, or instead of, Pygame.

However, pygame support is being deprecated. I am worried what effect that might have on this PR.

T-Dynamos:
@Julian-O It's not about pygame as kivy's backend but we can also run an actually game made in pygame using p4a.

T-Dynamos:
How it is different from #2838?
CC: @Julian-O

PR #2990, opened 2024-03-13
Qt Bootstrap - Adapting to Qt 6.7 release

shyamnathp:
- Make PythonActivity class independent of QtNative.
  QtNative is a private class from Qt 6.7 version. The only use of that class in PythonAcitivity is setting the environment variables.
  This can also be done with the Android Java API and hence the dependency can be removed. 

- add a key array named `bundled_libs` to libs.tmpl.xml
 This is more of a placeholder. As mentioned in the comment, the same purpose can be done with android.add_libs_* in buildozer.spec file. This does not break the previous versions.


misl6 (approved):
LGTM. Thank you!

shyamnathp:
thanks a lot :)

PR #3006, opened 2024-04-14
OpenJDK 11 incompatibility beacuse of Gradle 8.x / cannot build apk on WSL v1

Marcinosoft:
Downgrade gradle-8.0.2 to gradle-7.6.2 and gradle:8.1.1 to gradle:7.4.2 to keep JDK 11 compatibility

### DIFF
--- a/pythonforandroid/bootstraps/common/build/gradle/wrapper/gradle-wrapper.properties
+++ b/pythonforandroid/bootstraps/common/build/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.0.2-all.zip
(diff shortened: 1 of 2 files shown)


cclauss:
Is this still required?
* https://gradle.org/releases

Marcinosoft:
Yes. To be able to build using Java 11, which is required for successfull build on WSL v1.
Details: https://github.com/kivy/python-for-android/issues/3000

kuzeyron:
> Yes. To be able to build using Java 11, which is required for successfull build on WSL v1. Details: #3000

Even with version 17?

Marcinosoft:
> > Yes. To be able to build using Java 11, which is required for successfull build on WSL v1. Details: #3000
> 
> Even with version 17?

Java 11 is the last one able to build using Buildozer on WSL v1.

kuzeyron (commented):
Downgrading the libs is a no go. Instead you should find a way to determine if your system is WSL1 and so on. Only then you can add these changes.

I don't even think WSL1 is getting updates anymore.

cclauss:
Lot's of reasons for continuing to use WSL1...
* https://learn.microsoft.com/en-us/windows/wsl/compare-versions#exceptions-for-using-wsl-1-rather-than-wsl-2
* https://github.com/microsoft/WSL/discussions/4022

PR #3007, opened 2024-04-20
recipe: introduce isolated builds, using `PyProjectRecipe`

### LINKED ISSUE #2776: Ability to use newer version of pandas
This is a feature request, but I don't know how to put on the label...

Is your feature request related to a problem? Please describe.
I can't use the explode function for multiple columns because the current recipe for pandas uses something before version 1.3

Describe the solution you'd like
A pandas recipe that is newer than 1.3

Describe alternatives you've considered
The ability to choose pandas version (remove the numpy patch)

Thank you so much in advance!

### LINKED ISSUE #2904: Feature Request: Pandas>=1.3.0.
### Versions

- Python:3.10
- OS:Windows
- Kivy:2..2.1
- Cython:
- OpenJDK:

### Description

Is that possible that nice people can generate a recipe for Pandas>=1.3.0? Thank you so much in advance. That will help to support tons of other Python modules.

### Logs

  RAN: /usr/bin/patch -t -d /mnt/c/App_Design/plt_example/.buildozer/android/platform/build-arm64-v8a_armeabi-v7a/build/other_builds/pandas/arm64-v8a__ndk_target_21/pandas -p1 -i /mnt/c/App_Design/plt_example/.buildozer/android/platform/python-for-android/pythonforandroid/recipes/pandas/fix_numpy_includes.patch

  STDOUT:
/usr/bin/patch: **** Only garbage was found in the patch input.

T-Dynamos:
Closes #2904, Closes #2776 and maybe more

* `PyProjectRecipe`: support for building packages in isolation which don't have `setup.py` file but `pyproject.toml`.
* `MesonRecipe`: support for cross compiling meson based packages.

Here we need to inform the p4a users that if they are using numpy, they to set minimum ndk api to 24.


misl6:
Please ping me when you're ready

What is gonna be promising is the `PyProjectRecipe`, as we still rely on `setup.py` and non-isolated builds.

Moving to isolated builds will help us quite a lot to mitigate most of the issues that currently require a lot of hacks/setup from the user side (Incompatible Cython versions, etc ...)

T-Dynamos:
~@misl6 Hi, I can't figure out how to bump min `ndk_api` version. Can you please figure out?~

Edit: NVM fixed.

T-Dynamos:
@misl6 Hi!

PR is ready for review.

T-Dynamos:
@misl6 As now we are building wheels for packages, is there any plan to host that wheels? This could save some compile time.

misl6:
> @misl6 As now we are building wheels for packages, is there any plan to host that wheels? This could save some compile time.

This is possible but should be investigated. Feel free to ping me on Discord to discuss it, if you're interested on this topic.

pythonforandroid/recipes/cppy/__init__.py
```diff
@@ -1,14 +1,10 @@
...
+class CppyRecipe(PyProjectRecipe):
     site_packages_name = 'cppy'
-    version = '1.1.0'
-    url = 'https://github.com/nucleic/cppy/archive/{version}.zip'
-    call_hostpython_via_targetpython = False
-    # to be detected by the matplotlib install script
-    install_in_hostpython = True
-    depends = ['setuptools']
+    version = '08366d6fe4f385efde2dc0a38f95d7c3763db577'  # 1.2.1
+    url = 'git+https://github.com/nucleic/cppy'
```
misl6: Why we're not just using `https://github.com/nucleic/cppy/archive/refs/tags/1.2.1.zip` or `https://github.com/nucleic/cppy/releases/download/1.2.1/cppy-1.2.1.tar.gz`?

Also, as it seems it's just needed for the `hostpython`, in case build-isolation is not supported, maybe we can just drop this recipe?

pythonforandroid/recipes/kiwisolver/__init__.py
```diff
@@ -1,10 +1,10 @@
-from pythonforandroid.recipe import CppCompiledComponentsPythonRecipe
+from pythonforandroid.recipe import PyProjectRecipe
 
 
-class KiwiSolverRecipe(CppCompiledComponentsPythonRecipe):
+class KiwiSolverRecipe(PyProjectRecipe):
     site_packages_name = 'kiwisolver'
-    version = '1.3.2'
-    url = 'https://github.com/nucleic/kiwi/archive/{version}.zip'
+    version = '977f94b36b1049125d3d0879548d197df5729a74'  # 1.4.5
```
misl6: Applies what said regarding `cppy` recipe.

misl6 (commented):
I just had a first read, that's a super nice job, but let's try to slowly clean it up / document some changes.

Will apply https://github.com/kivy/kivy/pull/8713 also on `kivy/python-for-android` today, so you'll have less things to worry about. (and CI will be faster, as more runners are available concurrently)

EDIT: Changes applied, you now have access to `macos-13 (Intel)` and `macos-latest (Apple Silicon)`

pythonforandroid/recipes/cppy/__init__.py
T-Dynamos: According to https://github.com/nucleic/cppy/blob/08366d6fe4f385efde2dc0a38f95d7c3763db577/pyproject.toml#L41 it depends on `setuptools_scm`, it requires .git folder for version which is only available after clone.

pythonforandroid/recipes/kiwisolver/__init__.py
T-Dynamos: Same goes here.

pythonforandroid/recipes/cppy/__init__.py
misl6: Got it, but:
- If is just needed for hostpython dependency, we still need a recipe, or we can just use `hostpython_prerequisites` ?
- We can't rely on the tag instead of commit hash for cloning ?

T-Dynamos: Hm, looks like it just only provide build headers used by kiwisolver. I think it can be removed.

misl6: Yeah, I think we actually have a lot of dependencies which are just meant to be installed in `hostpython` as we previously did not allowed installing things into hostpython environment, without a recipe.

With hostpython_prerequisites (and isolated builds, where applicable) these recipes can likely be removed.
(Don't be afraid of cleaning-up things 😅)

T-Dynamos:
@misl6 Done.

pythonforandroid/recipes/scipy/__init__.py
```diff
@@ -87,5 +88,9 @@ def get_recipe_env(self, arch):
         env['LDFLAGS'] += f' -l{self.stl_lib_name}'
         return env
 
+    def build_arch(self, arch):
+        self.install_hostpython_prerequisites()
+        super().build_arch(arch)
```
misl6: Minor (but let's fix it separately): I see we're adding `self.install_hostpython_prerequisites()` in a few places, we can likely avoid it, and just try the installation by default.

pythonforandroid/recipes/numpy/__init__.py
```diff
@@ -1,34 +1,26 @@
...
+from pythonforandroid.recipe import Recipe, MesonRecipe
 from os.path import join
-import glob
-import sh
 import shutil
 
 
-class NumpyRecipe(CompiledComponentsPythonRecipe):
+class NumpyRecipe(MesonRecipe):
+    version = 'v2.0.0rc1'
```
misl6: Why do we need to target a `rc` version?
They are changing how things get built?

Do you know if they have an ETA for the final release?

pythonforandroid/recipe.py
```diff
@@ -478,10 +478,10 @@ def unpack(self, arch):
                 elif isdir(extraction_filename):
                     ensure_dir(directory_name)
                     for entry in listdir(extraction_filename):
-                        if entry not in ('.git',):
-                            shprint(sh.cp, '-Rv',
-                                    join(extraction_filename, entry),
-                                    directory_name)
+                        # if entry not in ('.git',): .git folder is required by `setuptools_scm`
```
misl6: Let's just change this comment with:
```
# Previously we filtered out the .git folder, but during the build process for some recipes
# (e.g. when version is parsed by `setuptools_scm`) that may be needed.
```

So, someone, looking into this in the future, will get a full explaination of what is happening, and we do not keep commented code around.

pythonforandroid/recipe.py
```diff
@@ -978,27 +977,38 @@ def install_hostpython_package(self, arch):
                 '--install-lib=Lib/site-packages',
                 _env=env, *self.setup_extra_args)
 
-    @property
-    def python_version(self):
-        return Recipe.get_recipe("python3", self.ctx).version
+    def get_python_formatted_version(self):
```
misl6: Maybe:
```python
@property
def python_major_minor_version(self):
    parsed_version = packaging.version.parse(self.ctx.python_recipe.version)
    return f"{parsed_version.major}.{parsed_version.minor}"
```

pythonforandroid/recipe.py
```diff
@@ -978,27 +977,38 @@ def install_hostpython_package(self, arch):
...
             "install",
-            *self.hostpython_prerequisites,
+            *packages,
             "--target", self.hostpython_site_dir, "--python-version",
-            self.python_version,
+            self.ctx.python_recipe.version,
             # Don't use sources, instead wheels
             "--only-binary=:all:",
-            "--no-deps"
+            # "--no-deps"
```
misl6: Let's just remove it, instead of commenting it out.

pythonforandroid/recipe.py
```diff
@@ -1157,7 +1167,159 @@ def get_recipe_env(self, arch, with_flags_in_cc=True):
...
-class RustCompiledComponentsRecipe(PythonRecipe):
+class PyProjectRecipe(PythonRecipe):
+    '''Recipe for projects which containes `pyproject.toml`'''
+
+    # Extra args to pass to `python -m build ...`
+    extra_build_args = []
+    call_hostpython_via_targetpython = False
+
+    def __init__(self, *arg, **kwargs):
+        super().__init__(*arg, **kwargs)
```
misl6: Really needed?

pythonforandroid/recipe.py
```diff
@@ -1157,7 +1167,159 @@ def get_recipe_env(self, arch, with_flags_in_cc=True):
...
+        sub_build_dir = join(build_dir, "p4a_android_build")
+        ensure_dir(sub_build_dir)
+        # copy hostpython to built python to ensure correct libs and includes
+        shprint(sh.cp, self.real_hostpython_location, self.ctx.python_recipe.python_exe)
+
+        build_args = [
+            "-m",
+            "build",
+            # "--no-isolation",
+            # "--skip-dependency-check",
```
misl6: Let's just remove these two lines

pythonforandroid/recipe.py
```diff
@@ -1157,7 +1167,159 @@ def get_recipe_env(self, arch, with_flags_in_cc=True):
...
+            shprint(
+                sh.Command(self.ctx.python_recipe.python_exe), *build_args, _env=env
+            )
+            built_wheel = realpath(glob.glob("dist/*.whl")[0])
+
+        info("Unzipping built wheel '{}'".format(basename(built_wheel)))
+
+        with zipfile.ZipFile(built_wheel, "r") as zip_ref:
+            zip_ref.extractall(self.ctx.get_python_install_dir(arch.arch))
+        info("Successfully installed  '{}'".format(basename(built_wheel)))
```
misl6: IIRC there's a way to programmatically install a wheel into a specific `site-packages` folder, without assuming is a zipfile.

But, unless you already have idea on how to, LGTM.

misl6 (commented):
Another (great) step forward!

I've left a few comments around, can you please share your thoughts?

misl6:
Addition:

> Here we need to inform the p4a users that if they are using numpy, they to set minimum ndk api to 24.

Maybe we can just raise an error if they try to do so ?

pythonforandroid/recipes/numpy/__init__.py
T-Dynamos: I did because of https://github.com/pandas-dev/pandas/blob/eadc1294b13bdc661283fcdf3ba54a50f31feeab/pyproject.toml#L9

misl6: As this strict requirement belongs to the build_system, shouldn't we be fine to keep `numpy<2.0.0` ?

misl6: (I'm not worried about rc, that would be super-fine if it was a feature release, but since is a major one, and I have no idea what `numpy` changed between 1.x.x and 2.x.x, I'm worried about breaking things for our users)

T-Dynamos: If I set lower than `2.0.0` then compilation will fail because we are passing numpy includes as:

```python
        env['NUMPY_INCLUDES'] = join(
            self.ctx.get_python_install_dir(arch.arch), "numpy/_core/include",
        )
```

I am afraid if something bad will happen if let pandas to use numpy (build isolation one). But it's worth a try.

misl6: Let's try, otherwise: https://github.com/pandas-dev/pandas/blob/v2.2.1/pyproject.toml

`pandas==2.2.1` does not seem to be affected.

pythonforandroid/recommendations.py
```diff
@@ -162,8 +162,8 @@ def check_target_api(api, arch):
         warning(OLD_API_MESSAGE)
 
 
-MIN_NDK_API = 21
-RECOMMENDED_NDK_API = 21
+MIN_NDK_API = 24
+RECOMMENDED_NDK_API = 24
```
misl6: As per PEP 738, minimum NDK API will be the same of the minimum available API for the latest LTS of the Android NDK. (ATM is `21`)

Can we revert it to `21`? (and keep `RECOMMENDED_NDK_API` to `24`)

setup.py
```diff
@@ -22,7 +22,7 @@
 install_reqs = [
     'appdirs', 'colorama>=0.3.3', 'jinja2',
     'sh>=1.10, <2.0; sys_platform!="win32"',
-    'build', 'toml', 'packaging', 'setuptools'
+    'build', 'toml', 'packaging', 'setuptools', 'wheel'
```
misl6: I'm slowly trying to enforce version pinning for deps anywhere, to avoid potential headaches, can we pin it with `~=` ?

(Renovatebot will send out notifications and open PRs for outdated deps)

pythonforandroid/recipes/numpy/__init__.py
```diff
@@ -37,54 +32,23 @@ def get_recipe_env(self, arch=None, with_flags_in_cc=True):
...
-
+        env["TARGET_PYTHON_EXE"] = join(Recipe.get_recipe(
+                "python3", self.ctx).get_build_dir(arch.arch), "android-build", "python")
         return env
 
     def build_arch(self, arch):
-        self.hostpython_prerequisites = ["setuptools==59.2.0"]
-        self.install_hostpython_prerequisites()
+        # NumPy requires complex maths functions which were added in api 24
+        if self.ctx.ndk_api < 24:
```
misl6, line 41: Minor:

There's a simple way to fail faster? (Otherwise we're gonna implement it later)

wheels/kiwisolver-1.4.5-cp311-cp311-android_arm.whl
misl6, line 1: I guess this is here by mistake 😀

misl6 (commented):
I think we almost reached the final steps, let me know when you're ready for the last round!

pythonforandroid/recipes/numpy/__init__.py
T-Dynamos: You mean before download step?

misl6: That would be perfect

misl6 (approved):
Amazing 🎉

LGTM. Thank you!

PR #3015, opened 2024-05-22
recipes: add new `uvloop` recipe

### LINKED ISSUE #2997: please add uvloop to recipes!
please add uvloop to recipes! uvloop is a fast, drop-in replacement of the built-in 'asyncio' event loop. This has greatly improved the performance of our project. We urgently need it.  please! 🙏🙏🙏
uvloop url: https://github.com/MagicStack/uvloop/tree/master

T-Dynamos:
Closes https://github.com/kivy/python-for-android/issues/2997

### DIFF
--- a/pythonforandroid/recipes/uvloop/__init__.py
+++ b/pythonforandroid/recipes/uvloop/__init__.py
@@ -1,0 +1,16 @@
+from pythonforandroid.recipe import PyProjectRecipe
+
+
+class UvloopRecipe(PyProjectRecipe):
+    version = 'v0.19.0'
+    url = 'git+https://github.com/MagicStack/uvloop'
+    depends = ['librt', 'libpthread']
+
+    def get_recipe_env(self, arch, **kwargs):
+        env = super().get_recipe_env(arch, **kwargs)
+        env["LIBUV_CONFIGURE_HOST"] = arch.command_prefix
+        env["PLATFORM"] = "android"
+        return env
+
+
+recipe = UvloopRecipe()
--- a/pythonforandroid/recipes/libpthread/__init__.py
+++ b/pythonforandroid/recipes/libpthread/__init__.py
@@ -1,0 +1,51 @@
+from os import makedirs, remove
+from os.path import exists, join
+import sh
+
+from pythonforandroid.recipe import Recipe
+from pythonforandroid.logger import shprint
+
+
+class LibPthread(Recipe):
+    '''
+    This is a dumb recipe. We may need this because some recipes inserted some
+    flags `-lpthread` without our control, case of:
+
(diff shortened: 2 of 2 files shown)


pythonforandroid/recipes/uvloop/__init__.py
```diff
@@ -0,0 +1,17 @@
+from pythonforandroid.recipe import PyProjectRecipe
+
+
+class UvloopRecipe(PyProjectRecipe):
+    # 0.19.0
+    version = '6c770dc3fbdd281d15c2ad46588c139696f9269c'
```
AndreMiras: I guess this is to make sure the tag doesn't get replaced in the upstream repo without us noticing?
I think a more common approach is to use the `sha512sum` or `md5sum`.
One advantage I'm seeing is hash verification against a version is enforced, while we don't enforce the "0.19.0" comment above the version vs the commit hash.
This is my personal take and I'm happy to get your opinion on it, but it's not blocking the merge

AndreMiras (dismissed):
LGTM, thanks for splitting and I appreciate the comprehensive LibPthread docstring

pythonforandroid/recipes/uvloop/__init__.py
T-Dynamos: My mistake, forgot to update version :sweat_smile:

pythonforandroid/recipes/uvloop/__init__.py
```diff
@@ -0,0 +1,16 @@
+from pythonforandroid.recipe import PyProjectRecipe
+
+
+class UvloopRecipe(PyProjectRecipe):
+    version = 'v0.19.0'
+    url = 'git+https://github.com/MagicStack/uvloop'
```
AndreMiras, line 6: I can't remember if it will automatically checkout the tag this way, but if you don't want to pull the full repo history you could try that
```suggestion
    url = 'https://github.com/MagicStack/uvloop/archive/refs/tags/{version}.tar.gz'
```

T-Dynamos: Tags were implemented recently: https://github.com/kivy/python-for-android/blob/344fd34d83b3e98c961dcd53b7d39606e1037593/pythonforandroid/recipe.py#L237

And we can't use source distribution because:
https://github.com/kivy/python-for-android/blob/344fd34d83b3e98c961dcd53b7d39606e1037593/pythonforandroid/recipe.py#L482-L486

(they don't contain .git folder)

AndreMiras: Nice thanks for the insight

AndreMiras (approved):

PR #3042, opened 2024-07-26
libsodium-recipe-update to url

gonatienza:
libsodium releases changed their url formatting since version 1.0.18.

Anyone trying to use newer versions of libsodium would receive the error ->

```
[INFO]:    Downloading libsodium from https://github.com/jedisct1/libsodium/releases/download/1.0.18/libsodium-1.0.18.tar.gz
(...)
urllib.error.HTTPError: HTTP Error 404: Not Found
Download failed: HTTP Error 404: Not Found; retrying in 1 second(s)...Download failed: HTTP Error 404: Not Found; retrying in 2 second(s)...Download failed: HTTP Error 404: Not Found; retrying in 4 second(s)...Download failed: HTTP Error 404: Not Found; retrying in 8 second(s)...

```
Suggest updating the url formatting to support newer versions and the default version to start at 1.0.18 instead of 1.0.16.

Background: encounter the above error trying to specify pynacl==1.5.0 as a requirement within my spec file.  The apk compiled and was missing references from the libsodium libs upon execution.  As I specified libsodium==1.0.18,pynacl==1.5.0 I saw the error.

**PS (BACKWARDS COMPATIBILITY CAVEAT)** -> anyone hardcoding as a requirement something lower than 1.0.18 (i.e., 1.0.16 or lower) would receive this error after this change.  Using libsodium without any version specified would result in a successful download of 1.0.18.

### DIFF
--- a/pythonforandroid/recipes/libsodium/__init__.py
+++ b/pythonforandroid/recipes/libsodium/__init__.py
@@ -3,14 +3,23 @@ from pythonforandroid.util import current_directory
 from pythonforandroid.logger import shprint
 from multiprocessing import cpu_count
 import sh
+from packaging import version as packaging_version
 
 
 class LibsodiumRecipe(Recipe):
     version = '1.0.16'
-    url = 'https://github.com/jedisct1/libsodium/releases/download/{version}/libsodium-{version}.tar.gz'
+    url = 'https://github.com/jedisct1/libsodium/releases/download/{}/libsodium-{}.tar.gz'
     depends = []
     patches = ['size_max_fix.patch']
     built_libraries = {'libsodium.so': 'src/libsodium/.libs'}
+
+    @property
+    def versioned_url(self):
+        asked_version = packaging_version.parse(self.version)
+        if asked_version > packaging_version.parse('1.0.16'):
+            return self._url.format(self.version + '-RELEASE', self.version)
+        else:
+            return self._url.format(self.version, self.version)
 
     def build_arch(self, arch):
         env = self.get_recipe_env(arch)


gonatienza:
I can also make it backwards compatible for someone asking for older versions.  Due to the way the url is formatted at the moment of the download it involved re-writing the url properties from the class within the LibsodiumRecipe subclass.  It also involves a urllib request to the github api to build a dictionary with all the releases and their individual urls.  If you would prefer that instead please advise and I will request a new pull or do a new commit over this one.  I find it more future proof but it is a bit more intrusive.

AndreMiras (commented):
Thanks for the PR.
Yeah having it backward compatible would be ideal indeed at least for a little while

gonatienza:
@AndreMiras , figured out a way to make it backward compatible being minimally invasive to the super class.  Tested it building with a specific version or just requiring libsodium, which will maintain the old behavior of using version 1.0.16.  HTH!

gonatienza:
Last two commits were cosmetic updates.  Feel free to merge after your review.

AndreMiras (commented):
Thanks for giving a stab a it :pray: 
I think a more appropriated approach would be to override the `versioned_url()` property.

But now that I'm thinking why don't we simply assume the version can contain the suffix and call it a day?
For instance valid versions could be:
- 1.0.20-RELEASE
- 1.0.17-RELEASE
- 1.0.16

And the `url` attribute would be: 
```
url = 'https://github.com/jedisct1/libsodium/archive/refs/tags/{version}.tar.gz'
```
Note how the `url` is constructed using the `tags` rather than the `releases`

gonatienza:
Hey @AndreMiras , tried messing with versioned_url and url but it felt like a band aid.  No matter what you do there, there's still a hardcoded url format at the subclass attribute level that's poped into _url through the metaclass.  Url and versioned_url indirectly call _url.  Plus playing with the url property caused some recursion issues.  IMHO the right place to make the url right is at the attribute level of the subclass.  If not, any future changes to the recipe.py could have undesired effects.

Regarding the simplification afterwards, I would be happy to push that instead.   I don't think though someone building and specifying the requirements would know that they need to append '-RELEASE' to newer versions.  The libsodium version is 1.0.18 not 1.0.18-RELEASE.  But it is an option.

E.g.,
requirements = libsodium==1.0.18,pynacl==1.5.0,kivy==2.3.0 -> this would fail to download libsodium
requirements = libsodium==1.0.18-RELEASE,pynacl==1.5.0,kivy==2.3.0  -> this would succeed

Another good option could be something like this:

`
from pythonforandroid.util import current_directory
from pythonforandroid.logger import shprint
from multiprocessing import cpu_count
from packaging import version as packaging_version
from os import environ
import sh


class LibsodiumRecipe(Recipe):
    env_version = environ.get('VERSION_libsodium')
    def_version = '1.0.16'
    if not env_version:
        version = def_version
    else:
        version = env_version
    asked_version = packaging_version.parse(version)
    if asked_version > packaging_version.parse(def_version):
        url = 'https://api.github.com/repos/jedisct1/libsodium/tarball/refs/tags/{version}-RELEASE'
    else:
        url = 'https://api.github.com/repos/jedisct1/libsodium/tarball/refs/tags/{version}'
    depends = []
    patches = ['size_max_fix.patch']
    built_libraries = {'libsodium.so': 'src/libsodium/.libs'}

    def build_arch(self, arch):
        env = self.get_recipe_env(arch)
        with current_directory(self.get_build_dir(arch.arch)):
            bash = sh.Command('bash')
            shprint(
                bash,
                'configure',
                '--disable-soname-versions',
                '--host={}'.format(arch.command_prefix),
                '--enable-shared',
                _env=env,
            )
            shprint(sh.make, '-j', str(cpu_count()), _env=env)

    def get_recipe_env(self, arch):
        env = super().get_recipe_env(arch)
        env['CFLAGS'] += ' -Os'
        return env


recipe = LibsodiumRecipe()
`

Let me know what you prefer and I shall push the commit!

:)

gonatienza:
Hey @AndreMiras , tried messing with versioned_url and url but it felt like a band aid.  No matter what you do there, there's still a hardcoded url format at the subclass attribute level that's poped into _url through the metaclass.  Url and versioned_url indirectly call _url.  Plus playing with the url property caused some recursion issues.  IMHO the right place to make the url right is at the attribute level of the subclass.  If not, any future changes to the recipe.py could have undesired effects.

Regarding the simplification afterwards, I would be happy to push that instead.   I don't think though someone building and specifying the requirements would know that they need to append '-RELEASE' to newer versions.  The libsodium version is 1.0.18 not 1.0.18-RELEASE.  But it is an option.

E.g.,
requirements = libsodium==1.0.18,pynacl==1.5.0,kivy==2.3.0 -> this would fail to download libsodium
requirements = libsodium==1.0.18-RELEASE,pynacl==1.5.0,kivy==2.3.0  -> this would succeed

Another good option could be something like this:

```
from pythonforandroid.util import current_directory
from pythonforandroid.logger import shprint
from multiprocessing import cpu_count
from packaging import version as packaging_version
from os import environ
import sh


class LibsodiumRecipe(Recipe):
    env_version = environ.get('VERSION_libsodium')
    def_version = '1.0.16'
    if not env_version:
        version = def_version
    else:
        version = env_version
    if packaging_version.parse(version) > packaging_version.parse(def_version):
        url = 'https://api.github.com/repos/jedisct1/libsodium/tarball/refs/tags/{version}-RELEASE'
    else:
        url = 'https://api.github.com/repos/jedisct1/libsodium/tarball/refs/tags/{version}'
    depends = []
    patches = ['size_max_fix.patch']
    built_libraries = {'libsodium.so': 'src/libsodium/.libs'}

    def build_arch(self, arch):
        env = self.get_recipe_env(arch)
        with current_directory(self.get_build_dir(arch.arch)):
            bash = sh.Command('bash')
            shprint(
                bash,
                'configure',
                '--disable-soname-versions',
                '--host={}'.format(arch.command_prefix),
                '--enable-shared',
                _env=env,
            )
            shprint(sh.make, '-j', str(cpu_count()), _env=env)

    def get_recipe_env(self, arch):
        env = super().get_recipe_env(arch)
        env['CFLAGS'] += ' -Os'
        return env


recipe = LibsodiumRecipe()
```

Let me know what you prefer and I shall push the commit!

:)

gonatienza:
Another good option modifying the versioned_url property only and avoiding any changes in the subclass attributes.  This way we do not touch the url property that causes all the recursion issues nor I check for env exports in the attributes.

```
from pythonforandroid.util import current_directory
from pythonforandroid.logger import shprint
from multiprocessing import cpu_count
import sh
from packaging import version as packaging_version


class LibsodiumRecipe(Recipe):
    version = '1.0.16'
    url = 'https://api.github.com/repos/jedisct1/libsodium/tarball/refs/tags/{version}'
    depends = []
    patches = ['size_max_fix.patch']
    built_libraries = {'libsodium.so': 'src/libsodium/.libs'}

    @property
    def versioned_url(self):
        if self.url is None:
            return None
        if packaging_version.parse(self.version) <= packaging_version.parse(self._version):
            return self._url.format(version=self.version)
        else:
            new_url = self._url + '-RELEASE'
            return new_url.format(version=self.version)

    def build_arch(self, arch):
        env = self.get_recipe_env(arch)
        with current_directory(self.get_build_dir(arch.arch)):
            bash = sh.Command('bash')
            shprint(
                bash,
                'configure',
                '--disable-soname-versions',
                '--host={}'.format(arch.command_prefix),
                '--enable-shared',
                _env=env,
            )
            shprint(sh.make, '-j', str(cpu_count()), _env=env)

    def get_recipe_env(self, arch):
        env = super().get_recipe_env(arch)
        env['CFLAGS'] += ' -Os'
        return env


recipe = LibsodiumRecipe()
```

AndreMiras:
Yes your second solution is what I meant by overriding the `versioned_url()`.
And yes agree with the other solution that it may not be clear upfront for people that they should pin with the suffix.
But that's only if they override the version and if they do they may get it seeing the 404.
Another option is to use this download URL which seems more consistent with the naming:
https://download.libsodium.org/libsodium/releases/

gonatienza:
Hey @AndreMiras , just sent a last commit.

Did not use the 'https://github.com/jedisct1/libsodium/archive/refs/tags/{version}.tar.gz' url as suggested as it results in build errors and the file downloaded is a different one (file name is the same but size is different).  The url that was there in the first place seems like the right one.  Did not use the https://download.libsodium.org/libsodium/releases/ either as all the old unsupported versions (like 1.0.16) are under a different folder too, so url consistency is also a problem.

Let me know...

gonatienza:
Btw, I did  try two builds, one with the default version and one with a newer version and both built.

pythonforandroid/recipes/libsodium/__init__.py
```diff
@@ -3,15 +3,25 @@
...
     built_libraries = {'libsodium.so': 'src/libsodium/.libs'}
 
+    @property
+    def versioned_url(self):
+        asked_version = packaging_version.parse(self.version)
+        recipe_version = packaging_version.parse(self._version)
+        if asked_version > recipe_version:
+            return self._url.format(self.version + '-RELEASE', self.version)
+        else:
+            return self._url.format(self.version, self.version)
```
AndreMiras: Shouldn't we hardcode the value rather than using `recipe_version`?
```
        if asked_version > packaging_version.parse("1.0.16"):
            return self._url.format(self.version + '-RELEASE', self.version)
        else:
            return self._url.format(self.version, self.version)
```
Otherwise if someone bumps the default recipe version it wouldn't work

AndreMiras (dismissed):
Looking good thanks for the follow up and thank you for testing both locally.
I made a minor suggestion, let me know what you think.
Also if you know how to squash and rebase, give it a try, otherwise I can try to do it upon merge.

pythonforandroid/recipes/libsodium/__init__.py
gonatienza: Yes, you are correct.

AndreMiras (approved):
Looking good thank you, will merge after the CI is done

gonatienza:
@AndreMiras , edited with the hardcoded line (good catch btw), added the commit and squashed all the commits.

PR #3051, opened 2024-08-19
Add tiktoken recipe

RobertFlatt:
tiktoken is a fast [BPE](https://en.wikipedia.org/wiki/Byte_pair_encoding) tokeniser for use with OpenAI's models.


AndreMiras (approved):
LGTM thanks!

PR #3066, opened 2024-09-26
reproducible apks: strip file path prefix from .pyc files

SomberNight:
Without this, paths from the build machine will leak into the generated `.pyc` files, which makes the builds more difficult to reproduce.

Note that the [`compileall -s` option](https://docs.python.org/3/library/compileall.html#cmdoption-compileall-s) was only added in python 3.9. Not sure if that is acceptable, as this limits the hostpython3 recipe to 3.9+.

### DIFF
--- a/pythonforandroid/bootstraps/common/build/build.py
+++ b/pythonforandroid/bootstraps/common/build/build.py
@@ -205,7 +205,14 @@ def compile_py_file(python_file, optimize_python=True):
     if PYTHON is None:
         return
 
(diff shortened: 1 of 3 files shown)


AndreMiras:
Thanks for the pull request.
Python 3.8 support ends the 31st of October 2024, refs https://endoflife.date/python and https://devguide.python.org/versions/
So I guess we're probably good to merge, but we need to tag a release before that.
We will try to look this up

PR #3074, opened 2024-10-18
Add ability to use private github repos for recipes

brentpicasso:
Add ability to specify a github_access_token so recipes can download from private github repositories.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -8,6 +8,9 @@ env:
   AAR_ARTIFACT_FILENAME: bdist_unit_tests_app-release-1.1.aar
   PYTHONFORANDROID_PREREQUISITES_INSTALL_INTERACTIVE: 0
 
+concurrency:
+  cancel-in-progress: true
+
 jobs:
 
   flake8:
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -9,6 +9,7 @@ env:
   PYTHONFORANDROID_PREREQUISITES_INSTALL_INTERACTIVE: 0
 
 concurrency:
+  group: build-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -216,7 +216,7 @@ class Recipe(metaclass=RecipeMeta):
                     # Mozilla/5.0 doesnt handle redirection for liblzma
                     url_opener.addheaders = [('User-agent', 'Wget/1.0')]
                     if self.github_access_token:
-                        url_opener.addheaders += [('Authorization', f'token {self.github_access_token}'), ('Accept', 'application/vnd.github.v3.raw')]
+                        url_opener.addheaders += [('Authorization', f'token {self.github_access_token}'), ('Accept', 'application/vnd.github+json')]
                     urlretrieve(url, target, report_hook)
                 except OSError as e:
                     attempts += 1
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -59,9 +59,16 @@ class Recipe(metaclass=RecipeMeta):
               if you want.
     '''
 
-    _github_access_token = None
-    '''Used to access a private git repository. Specify the github-supplied
-    access token in order to download the private repository files.
+    _download_headers = None
+    '''Add additional headers used when downloading the package, typically
+    for authorization purposes.
+
+    Specified as an array of tuples:
+    [("header name", "header value")]
+
+    For example, when downloading from a private
+    github repository, you can specify the following:
+    [('Authorization', 'token <your personal access token>'), ('Accept', 'application/vnd.github+json')]
     '''
 
     _version = None
@@ -176,9 +183,9 @@ class Recipe(metaclass=RecipeMeta):
         return self.url.format(version=self.version)
 
     @property
-    def github_access_token(self):
-        key = "GITHUB_ACCESS_TOKEN_" + self.name
-        return environ.get(key, self._github_access_token)
+    def download_headers(self):
+        key = "DOWNLOAD_HEADERS_" + self.name
+        return environ.get(key, self._download_headers)
 
     def download_file(self, url, target, cwd=None):
         """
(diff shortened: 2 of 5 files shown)


AndreMiras:
Interesting feature thanks.
Alternatively you should already be able to clone using the token like so:
```
git clone https://$TOKEN@github.com/$USERNAME/$REPOSITORY.git
```
Let me know if you think that would be a good enough alternative.
Your approach has a couple of advantages over this one however:
- no full clone needed
- no hardcoded credentials in the recipe (as it can be in the environment variable)
```

**Comment by brentpicasso** [comment]
> Interesting feature thanks. Alternatively you should already be able to clone using the token like so:
> 
> ```
> git clone https://$TOKEN@github.com/$USERNAME/$REPOSITORY.git
> ```
> 
> Let me know if you think that would be a good enough alternative. Your approach has a couple of advantages over this one however:
> 
>     * no full clone needed
> 
>     * no hardcoded credentials in the recipe (as it can be in the environment variable)

Hi, thanks for looking at it. 

Do python recipes already support getting the repository this way via 'git clone'? I only see https download support in the PythonRecipe.

```git clone https://$TOKEN@github.com/$USERNAME/$REPOSITORY.git```

We need this to be automated for our build process, not done by hand as a separate process.

Regarding the hardcoded environment variable- I noticed the PythonRecipe already has a provision for using properties based on enviornment variables: 

```
 @property
    def github_access_token(self):
        key = "GITHUB_ACCESS_TOKEN_" + self.name
        return environ.get(key, self._github_access_token)

```

In my case, my recipe just gets the environment variable and sets it within the recipe, so it is not hard coded.

Overall, if Python recipes support getting the repo via https, then it should also support getting private repos using the same mechanism, by providing the access token.

Edit:
I'd love to have your feedback on the `URL_<recipe>` environment variable approach combined with `https://$TOKEN@github.com/$USERNAME/$REPOSITORY.git`

**Comment by AndreMiras** [comment]
> Do python recipes already support getting the repository this way via 'git clone'? I only see https download support in the PythonRecipe.
> 
> `git clone https://$TOKEN@github.com/$USERNAME/$REPOSITORY.git`
> 
> We need this to be automated for our build process, not done by hand as a separate process.

Yes git cloning is supported in `Recipe`, see https://github.com/kivy/python-for-android/blob/74b891e/pythonforandroid/recipe.py#L221-L242

> Regarding the hardcoded environment variable- I noticed the PythonRecipe already has a provision for using properties based on enviornment variables:
> 
> ```
>  @property
>     def github_access_token(self):
>         key = "GITHUB_ACCESS_TOKEN_" + self.name
>         return environ.get(key, self._github_access_token)
> ```

Good catch for environment variable based properties.
https://github.com/kivy/python-for-android/blob/74b891e/pythonforandroid/recipe.py#L159-L162
```python
    @property
    def url(self):
        key = 'URL_' + self.name
        return environ.get(key, self._url)
```
I'm wondering if you could leverage that already to build to clone URL dynamically and add the token like we mentioned?


> In my case, my recipe just gets the environment variable and sets it within the recipe, so it is not hard coded.

Yes I get that, that's what I meant, it's in the advantage list of your approach.

> Overall, if Python recipes support getting the repo via https, then it should also support getting private repos using the same mechanism, by providing the access token.

Yes agree, one thing however is this is tailor made for GitHub only rather than a generic (header based) authentication.
Also I'm always trying to see if we can meet the requirements without growing the code base with many hidden features.
I'm totally not blocking the feature, just challenging it to see if there're good alternatives

brentpicasso:
Thanks for the feedback. Perhaps we make it generic and let people specify the authorization token and not make it github specific. 

Would that seem better?

AndreMiras:
Yes I was thinking maybe generic header from environment variables could be an option too.
But did you give the git clone + token a try still?

brentpicasso:
Ok, I've updated it to specify headers generically, and enhanced the documentation as well: https://github.com/kivy/python-for-android/pull/3074/commits/e4ade43f942f1fe7c6bd227c5d6e9179eb1800e1

I think this is a pretty clean implementation, thanks for the feedback!

brentpicasso:
@AndreMiras  I also found applied a tiny unrelated fix around a possible missed python 2.7 migration issue: https://github.com/kivy/python-for-android/pull/3074/commits/254440db5c293d9d5cf8ec0da04a1e1e9bc00b2a

pythonforandroid/recipe.py
```diff
@@ -170,6 +182,11 @@ def versioned_url(self):
             return None
         return self.url.format(version=self.version)
 
+    @property
+    def download_headers(self):
+        key = "DOWNLOAD_HEADERS_" + self.name
+        return environ.get(key, self._download_headers)
```
AndreMiras: So in that specific case where we have 2 headers, what would the environment variable be like?
`export DOWNLOAD_HEADERS_recipe=?`
Or maybe it's not supported? Could we have the example in the that property docstring?

AndreMiras (commented):
Beautiful thanks :pray: 
It's not clear to me how the environment variable would make it possible to deal with the GitHub use case of 2 headers.
```
[('Authorization', 'token <your personal access token>'), ('Accept', 'application/vnd.github+json')]
```

pythonforandroid/recipe.py
brentpicasso: Oh, that's a good point. 

I was trying to keep it generic.  For the environmental variables, the alternative I was thinking about would be to have 

DOWNLOAD_AUTHORIZATION_mypackage = "<github_personal_access_token>"
DOWNLOAD_ACCEPT_mypackage  = "application/vnd.github+json"

**OR**

We can do something like this, which could be parsed:

DOWNLOAD_HEADERS_mypackage = "Authorization=foo,Accept=bar"

What do you think?

brentpicasso: Another option is to specify the authorization in the following auth token/accept mime type pairs:

DOWNLOAD_AUTHORIZATION_mypackage="token <my_token>,application/vnd.github+json"

brentpicasso: Ok, I picked a JSON formatted value for the environment variable:

[["header1","foo"],["header2", "bar"]]

https://github.com/kivy/python-for-android/pull/3074/commits/5e14116bf09ca1d48719f7fa7e8c438f1d896ab8

AndreMiras (dismissed):
Sweet thanks I think we're covered :pray: 
Would it be too much to ask to get a test in `tests/test_recipe.py` for the environment variable case? :grimacing: 
I think something like that should do it:
```python
...

    def test_recipe_download_headers(self):
        """Download header can be created on the fly using environment variables."""
        recipe = DummyRecipe()
        with mock.patch.dict(os.environ, <do-the-magic>):
            download_headers = recipe.download_headers
        assert download_headers == [<the-headers>]
```

.github/workflows/push.yml
```diff
@@ -8,6 +8,10 @@ env:
   AAR_ARTIFACT_FILENAME: bdist_unit_tests_app-release-1.1.aar
   PYTHONFORANDROID_PREREQUISITES_INSTALL_INTERACTIVE: 0
 
+concurrency:
+  group: build-${{ github.ref }}
+  cancel-in-progress: true
```
AndreMiras, line 13: Nice addition thanks

AndreMiras (approved):
Thanks for all the follow up :pray: 
This looks good to me.
Let's wait for the CI to complete before we can squash & merge

brentpicasso:
@AndreMiras thanks, happy to help! Looks like all checks have passed.

PR #3076, opened 2024-10-20
GitHub Actions: Add Python 3.12, 3.13, 3.14, and 3.14t  to the testing

cclauss:
* https://docs.python.org/3/using/android.html
* https://www.python.org/downloads/release/python-3130/
* https://pythoninsider.blogspot.com/2024/10/python-3130-final-released.html
* https://devguide.python.org/versions/

> New features
> [PEP 738](https://peps.python.org/pep-0738/): Android is now an [officially supported platform](https://docs.python.org/3/whatsnew/3.13.html#whatsnew313-platform-support), at [tier 3](https://peps.python.org/pep-0011/#tier-3).
> [PEP 738](https://peps.python.org/pep-0738/): Android is now a [PEP 11](https://peps.python.org/pep-0011/) supported platform, with the aarch64-linux-android and x86_64-linux-android targets at tier 3. The 32-bit targets arm-linux-androideabi and i686-linux-android are not tier 3 supported platforms, but will have best-effort support. (PEP written and implementation contributed by Malcolm Smith in [gh-116622](https://github.com/python/cpython/issues/116622).)

* https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down
 * Kivy repos still using macos-13: https://github.com/search?q=org%3Akivy%20macos-13&type=code


AndreMiras (dismissed):
LGTM thanks, I'll merge after the next release as we will probably drop Python 3.8

cclauss:
Why test after making a release?  Usually, tests are added before making a release.

AndreMiras:
In this PR you dropped 3.8 which is a good thing, but that means we officially drop the support for it (which we want to).
However I want to tag a release before we drop the support, so that the tagged release would be the last supporting 3.8

cclauss:
Please let me know if we should bring Py38 back in.

AndreMiras:
> Please let me know if we should bring Py38 back in.

Nope it's perfect keep it this way, I want to make a release in the coming weeks

cclauss:
The reason stated for sitting on this was that it remove the EOL Python 3.8 so let's put it back is so that we are running tests on Python 3.8 plus all currently supported versions of CPython.

cclauss:
Coming up on one year...

AndreMiras (dismissed):
Well build is green, why not.
Will leave it there for a few days and merge if nobody has anything against it.
Yeah build make take longer, but we should remove 3.8 at some point eventually

cclauss:
3.9 will go EOL in a few weeks so I doubt there will be serious protest about dropping support for 3.8.

misl6 (approved):
LGTM. Thank you!

PR #3078, opened 2024-10-20
Keep GitHub Actions up to date with GitHub's Dependabot

cclauss:
Fixes software supply chain safety warnings like at the bottom right of
https://github.com/kivy/python-for-android/actions/runs/11425651890

* [Keeping your actions up to date with Dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot)
* [Configuration options for the dependabot.yml file - package-ecosystem](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem)

@AndreMiras


AndreMiras (approved):
We closed the dup #3223 because it looked like dependabot was configured at org level, but it's not, the only dependabot PR was #3222 which was a security fix, but rest of the actions are out of date, so dependabot is definitely not running.
PR still good to be merged, thanks for it

PR #3085, opened 2024-11-23
Create Калкулятор

Sigmen76:


kuzeyron:
This PR doesn't involve improving python-for-android. Further support can be reached out on Kivy's [Discord](https://chat.kivy.org) channels.

PR #3065, opened 2024-09-24
Bump `FFMpegRecipe` and `PyAVRecipe` versions

DexerBR:
`PyAV` updated to the latest stable release - `13.1.0`
`FFmpeg` updated to `6.1.2` - `ffpyplayer` is still not compatible with FFmpeg `7.x.x.`

PR #3092, opened 2024-12-13
Add hardware acceleration codecs to ffmpeg recipe

brentpicasso:


AndreMiras (commented):
LGTM, just the title is misleading, it says bump recipes version, but all I see are the hardware acceleration codecs enabled right? I imagine the bump was reverted in subsequent commits.
Having the PR rebased & squashed would be a real plus, otherwise no worry I can try to think about it at merge time.
image

brentpicasso:
@AndreMiras I think the original work was to bump the recipe versions, and also to add these codecs. 

It seems that develop received these updated versions, and this PR left behind only the hardware codecs.   @DexerBR do you agree with this assessment?

I updated the title to match the new reality.

DexerBR:
> @AndreMiras I think the original work was to bump the recipe versions, and also to add these codecs.
> 
> It seems that develop received these updated versions, and this PR left behind only the hardware codecs. @DexerBR do you agree with this assessment?
> 
> I updated the title to match the new reality.

I agree, the versions have been updated in this PR: https://github.com/kivy/python-for-android/pull/3065

@brentpicasso How about updating the title to "Add hardware acceleration codecs to `ffmpeg` recipe"?

brentpicasso:
> @brentpicasso How about updating the title to "Add hardware acceleration codecs to `ffmpeg` recipe"?

@DexerBR  done!

AndreMiras (approved):

PR #3126, opened 2025-03-17
Proxy environment variables passthrough to recipes

broccoliboy:
This PR allows recipes to inherit common proxy env vars, which solves build issues on corporate networks that use a proxy.


kuzeyron (approved):

kuzeyron:
Thanks!

PR #3143, opened 2025-04-28
Upgrade SQLAlchemy recipe to 2.0.40

clayote:
This was pretty straightforward.

The syntax for the expression language changed in SQLalchemy 2.0, so this could break some apps that depend on the old behavior.


clayote:
This is already done in the `develop` branch, and better than I had it.

PR #3147, opened 2025-05-05
Fix Cmake compatibility issue on jpeg recipe

Huyston:
The jpeg recipe doesn't compile anymore because CMake dropped compatibility with versions below 3.5.

"Calls to cmake_minimum_required() or cmake_policy() that set the policy version to an older value now issue an error."

Source: https://cmake.org/cmake/help/latest/release/4.0.html


AndreMiras (approved):
LGTM thanks

PR #3151, opened 2025-05-11
`pyjnius`: pin cython version

T-Dynamos:
As pyjnius build fails on latest cython.

```
[1/1] Cythonizing jnius/jnius.pyx
warning: jnius/jnius.pyx:102:0: The 'IF' statement is deprecated and will be removed in a future Cython version. Consider using runtime conditions or C macros instead. See https://github.com/cython/cython/issues/4310

Error compiling Cython file:
------------------------------------------------------------
...
            score += 10
            continue

        if r == 'S' or r == 'I':
            if isinstance(arg, int) or (
                    (isinstance(arg, long) and arg < 2147483648)):
                                     ^
------------------------------------------------------------

jnius/jnius_utils.pxi:323:37: undeclared name not builtin: long
Traceback (most recent call last):
```

PR #3152, opened 2025-05-12
Restrict Cython versions for numpy and pyjnius

### LINKED ISSUE #3150: numpy build fails for lack of `numpy.math`
`buildozer debug` seemed to work fine yesterday, but then I added `armeabi-v7a` to its archs and now this is happening:

```
cython -M --fast-fail -3 /home/sanotehu/src/lisien/.buildozer/android/platform/build-arm64-v8a_armeabi-v7a/build/other_builds/numpy/armeabi-v7a__ndk_target_24/numpy/numpy/random/_common.pyx -o numpy/random/_common.cpython-311-x86_64-linux-gnu.so.p/numpy/random/_common.pyx.c
Ignoring "sys._home = value" override

Error compiling Cython file:
------------------------------------------------------------
...
from collections import namedtuple
from cpython cimport PyFloat_AsDouble
import sys
import numpy as np
cimport numpy as np
cimport numpy.math as npmath
        ^
------------------------------------------------------------

/home/sanotehu/src/lisien/.buildozer/android/platform/build-arm64-v8a_armeabi-v7a/build/other_builds/numpy/armeabi-v7a__ndk_target_24/numpy/numpy/random/_common.pyx:8:8: 'numpy/math.pxd' not found
```

[Full log, as well as my buildozer.spec, are at this gist.](https://gist.github.com/clayote/0f1a375b89074567e0da6960a71c0671)

clayote:
These recipes both broke recently because of changes to Cython. I've restricted what Cython versions their hostpython can use.

Fixes #3150. Thanks to @T-Dynamos for identifying the problem.

### DIFF
--- a/pythonforandroid/recipes/numpy/__init__.py
+++ b/pythonforandroid/recipes/numpy/__init__.py
@@ -9,7 +9,7 @@ NUMPY_NDK_MESSAGE = "In order to build numpy, you must set minimum ndk api (mina
 class NumpyRecipe(MesonRecipe):
     version = 'v1.26.5'
     url = 'git+https://github.com/numpy/numpy'
-    hostpython_prerequisites = ["Cython>=3.0.6"]  # meson does not detects venv's cython
+    hostpython_prerequisites = ["Cython>=3.0.6,<3.1.0"]  # meson does not detects venv's cython
     extra_build_args = ['-Csetup-args=-Dblas=none', '-Csetup-args=-Dlapack=none']
     need_stl_shared = True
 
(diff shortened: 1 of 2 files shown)


clayote:
Sorry, this is redundant with #3151

AndreMiras:
Isn't that one a bit better actually since it's a bit more relaxed on the version vs hard pinning?
What do you think @T-Dynamos

clayote:
Maybe, but I didn't feel like testing a bunch of versions in the range.

PR #3136, opened 2025-03-27
scipy: update to v1.16.2

T-Dynamos:
Closes #3116 

Cons:

1. Can't build on macos
2. Needs android 7 and 7+
3. Only 64 bit arches supported

Extra Notes:

Removes need to use `LEGACY_NDK` as it uses fortran from https://github.com/termux/ndk-toolchain-clang-with-flang.

PR #3164, opened 2025-06-17
Update: `numpy`, `pandas`, `sdl2` to newer versions which support `ndk28c`

### LINKED ISSUE #3163: Numpy recipe
Please increase numpy version from 1.26.5 to 2.0.2, because 1.26.5 version is not available

T-Dynamos:
Closes https://github.com/kivy/python-for-android/issues/3163

Split #3136

### DIFF
--- a/ci/makefiles/android.mk
+++ b/ci/makefiles/android.mk
@@ -1,12 +1,12 @@
 # Downloads and installs the Android SDK depending on supplied platform: darwin or linux
 
 # Those android NDK/SDK variables can be override when running the file
-ANDROID_NDK_VERSION ?= 25b
+ANDROID_NDK_VERSION ?= 28c
 ANDROID_NDK_VERSION_LEGACY ?= 21e
 ANDROID_SDK_TOOLS_VERSION ?= 6514223
 ANDROID_SDK_BUILD_TOOLS_VERSION ?= 29.0.3
 ANDROID_HOME ?= $(HOME)/.android
-ANDROID_API_LEVEL ?= 35
+ANDROID_API_LEVEL ?= 36
 
 # per OS dictionary-like
 UNAME_S := $(shell uname -s)
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -155,6 +155,11 @@ class Recipe(metaclass=RecipeMeta):
         starting from NDK r18 the `gnustl_shared` lib has been deprecated.
     '''
 
+    min_ndk_api_support = 20
+    '''
+    Minimum ndk api recipe will support.
+    '''
+
     def get_stl_library(self, arch):
         return join(
             arch.ndk_lib_dir,
@@ -375,6 +380,9 @@ class Recipe(metaclass=RecipeMeta):
     # Public Recipe API to be subclassed if needed
 
     def download_if_necessary(self):
+        if self.ctx.ndk_api < self.min_ndk_api_support:
+            error(f"In order to build '{self.name}', you must set minimum ndk api (minapi) to `{self.min_ndk_api_support}`.\n")
+            exit(1)
         info_main('Downloading {}'.format(self.name))
         user_dir = environ.get('P4A_{}_DIR'.format(self.name.lower()))
         if user_dir is not None:
--- a/pythonforandroid/recipes/numpy/__init__.py
+++ b/pythonforandroid/recipes/numpy/__init__.py
@@ -1,5 +1,4 @@
 from pythonforandroid.recipe import Recipe, MesonRecipe
-from pythonforandroid.logger import error
 from os.path import join
 import shutil
 
@@ -7,11 +6,12 @@ NUMPY_NDK_MESSAGE = "In order to build numpy, you must set minimum ndk api (mina
 
 
 class NumpyRecipe(MesonRecipe):
-    version = 'v1.26.5'
+    version = 'v2.3.0'
     url = 'git+https://github.com/numpy/numpy'
-    hostpython_prerequisites = ["Cython>=3.0.6"]  # meson does not detects venv's cython
+    hostpython_prerequisites = ["Cython>=3.0.6", "numpy"]  # meson does not detects venv's cython
     extra_build_args = ['-Csetup-args=-Dblas=none', '-Csetup-args=-Dlapack=none']
     need_stl_shared = True
+    min_ndk_api_support = 24
 
     def get_recipe_meson_options(self, arch):
         options = super().get_recipe_meson_options(arch)
@@ -35,13 +35,6 @@ class NumpyRecipe(MesonRecipe):
         env["TARGET_PYTHON_EXE"] = join(Recipe.get_recipe(
                 "python3", self.ctx).get_build_dir(arch.arch), "android-build", "python")
         return env
-
-    def download_if_necessary(self):
-        # NumPy requires complex math functions which were added in api 24
-        if self.ctx.ndk_api < 24:
-            error(NUMPY_NDK_MESSAGE)
-            exit(1)
-        super().download_if_necessary()
 
     def build_arch(self, arch):
         super().build_arch(arch)
--- a/pythonforandroid/recipes/pandas/__init__.py
+++ b/pythonforandroid/recipes/pandas/__init__.py
@@ -3,10 +3,10 @@ from pythonforandroid.recipe import MesonRecipe
 
 
 class PandasRecipe(MesonRecipe):
-    version = 'v2.2.1'
-    url = 'git+https://github.com/pandas-dev/pandas'  # noqa
+    version = 'v2.3.0'
+    url = 'git+https://github.com/pandas-dev/pandas'
     depends = ['numpy', 'libbz2', 'liblzma']
-    hostpython_prerequisites = ["Cython~=3.0.5"]  # meson does not detects venv's cython
+    hostpython_prerequisites = ["Cython<4.0.0a0", "versioneer", "numpy"]  # meson does not detects venv's cython
     patches = ['fix_numpy_includes.patch']
     python_depends = ['python-dateutil', 'pytz']
     need_stl_shared = True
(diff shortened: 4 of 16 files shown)


T-Dynamos:
Ready for review. Hope it can be merged soon!


(One action failed, probably due to an internet connection error)

pythonforandroid/recipe.py
```diff
@@ -375,6 +380,9 @@ def get_recipe_dir(self):
     # Public Recipe API to be subclassed if needed
 
     def download_if_necessary(self):
+        if self.ctx.ndk_api < self.min_ndk_api_support:
+            error(f"In order to build '{self.name}', you must set minimum ndk api (minapi) to `{self.min_ndk_api_support}`.\n")
+            exit(1)
```
AndreMiras, line 385: Wait, shouldn't we leverage `pythonforandroid/recommendations.py`?
Maybe @misl6 has better insights

pythonforandroid/recipes/numpy/__init__.py
```diff
@@ -1,17 +1,17 @@
...
 
 NUMPY_NDK_MESSAGE = "In order to build numpy, you must set minimum ndk api (minapi) to `24`.\n"
 
 
 class NumpyRecipe(MesonRecipe):
-    version = 'v1.26.5'
+    version = 'v2.3.0'
     url = 'git+https://github.com/numpy/numpy'
-    hostpython_prerequisites = ["Cython>=3.0.6"]  # meson does not detects venv's cython
+    hostpython_prerequisites = ["Cython==3.0.6"]  # meson does not detects venv's cython
```
AndreMiras: Do we have to be that restrictive? Maybe it doesn't work with too recents Cython versions, but in this case maybe we should use a range at least

pythonforandroid/recipes/numpy/__init__.py
```diff
@@ -36,13 +36,6 @@ def get_recipe_env(self, arch, **kwargs):
                 "python3", self.ctx).get_build_dir(arch.arch), "android-build", "python")
         return env
 
-    def download_if_necessary(self):
-        # NumPy requires complex math functions which were added in api 24
-        if self.ctx.ndk_api < 24:
-            error(NUMPY_NDK_MESSAGE)
-            exit(1)
-        super().download_if_necessary()
```
AndreMiras, line 44: Oh OK I see what you did there by moving it to `recipe.py`
@misl6 would have better insights, but I'd be tempted to keep it in this recipe only for this PR unless we have a clearly cleaner approach, what do you guys think?

T-Dynamos: I did this because it was also required in `scipy` recipe.

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -141,7 +141,6 @@ class Python3Recipe(TargetPythonRecipe):
 
     site_packages_dir_blacklist = {
         '__pycache__',
-        'tests'
```
AndreMiras: Why do we keep tests in site packages? Or is it already handled upstream and no longer part of it?

pythonforandroid/recipes/numpy/__init__.py
AndreMiras: Thanks for the context. But I feel the current implementation in the core recipe is a bit hacky isn't it? It was hacky in the numpy recipe, but at least a bit more contained. I don't mind having dups in scipy until we find a cleaner approach.
What do you guys think?

pythonforandroid/recipes/python3/__init__.py
T-Dynamos: numpy has some tests dir, without that `import numpy as np` fails.

AndreMiras: Thanks for the context. How about mutating/patching the recipe from numpy instead?
Because in the current approach we will end up impacting bundle size for everyone. Plus someone may add tests back to the blacklist not realizing numpy needs it. Having it handled within the numpy recipe would solve both issues.

T-Dynamos:
@AndreMiras any updates on this?

pythonforandroid/util.py
```diff
@@ -48,7 +48,7 @@ def temp_directory():
                               temp_dir, Err_Fore.RESET)))
 
 
-def walk_valid_filens(base_dir, invalid_dir_names, invalid_file_patterns):
+def walk_valid_filens(base_dir, invalid_dir_names, invalid_file_patterns, excluded_dir_exceptions=[]):
```
AndreMiras: I know this looks convenient, but could you update to `excluded_dir_exceptions=None` instead. And then in the function body you can `excluded_dir_exceptions = [] if excluded_dir_exceptions is None else excluded_dir_exceptions`.
There's a gotcha with mutable default arguments in Python that could lead to bugs painful to track.
https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments
I know we don't currently mutate it in your case, but I think it's a good rule of thumb to generally avoid using a mutable in default arguments at all so nothing weird happens if someone thinks about mutating this list later.

But more here than for the `min_ndk_api_support` change I would feel more comfortable to have this handled outside of the core code if possible. It's a very specific case from numpy, don't you think we could find a way to handle it from numpy directly in an acceptable way?

AndreMiras (commented):
Yeah thanks for addressing the site package exclude thingy in a different way, at least now it doesn't impact all the other recipes.
However, generally speaking I'm cautious with changes in the core code that introduce a bit more complexity specially if it's for some very rare cases, I hope you understand.

While the changes in `download_if_necessary()` are easy to grasp and could easily help for more than a couple of recipes, I still want to believe we could try a different approach for the site package exclude dir exception case.
Maybe addressing it from within the Numpy recipe wouldn't be the sexiest, but at least it would be more contained.
Would you mind giving this specific part an extra thought otherwise I can give it a try when I get a chance.
This might as well not be easy at all or would in fact introduce much more complexity and completely off balance the benefits. I haven't explored, but it might be worth thinking about it and giving it a try.
Edit: my naive thought on that one is to `Recipe.get_recipe('python3', self.ctx)` from numpy `prebuild_arch()` to then mutate `site_packages_dir_blacklist` there. It's definitely hacky, but much more contained.

Also if @misl6 thinks otherwise, I totally trust him to make the right call, he's much more active and knowledgeable that anyone else regarding this code base

pythonforandroid/util.py
T-Dynamos: > It's a very specific case from numpy

Also for scipy.

T-Dynamos:
@AndreMiras 

> Recipe.get_recipe('python3', self.ctx) from numpy prebuild_arch()

calls:
```python
def load_source(module, filename):
    # Python 3.5+
    import importlib.util
    if hasattr(importlib.util, 'module_from_spec'):
        spec = importlib.util.spec_from_file_location(module, filename)
        mod = importlib.util.module_from_spec(spec)
        spec.loader.exec_module(mod)
        return mod
    else:
        # Python 3.3 and 3.4:
        from importlib.machinery import SourceFileLoader
        return SourceFileLoader(module, filename).load_module()
```
However, this dynamic loading of the python3 recipe does not affect `self.ctx.python_recipe`, because the `python_recipe` module was already loaded earlier in the build process.


as `python_recipe` used in `pythonforandroid/bootstraps/_sdl_common/__init__.py`:
```python
                site_packages_dir = self.ctx.python_recipe.create_python_bundle(
                    join(self.dist_dir, python_bundle_dir), arch)

```

Edit: My opinion, it's fine as it is :)

ci/makefiles/android.mk
```diff
@@ -1,12 +1,12 @@
 # Downloads and installs the Android SDK depending on supplied platform: darwin or linux
 
 # Those android NDK/SDK variables can be override when running the file
-ANDROID_NDK_VERSION ?= 25b
+ANDROID_NDK_VERSION ?= 27c
```
misl6: Can't we target the latest available here?

ci/makefiles/android.mk
```diff
@@ -1,12 +1,12 @@
...
 ANDROID_NDK_VERSION_LEGACY ?= 21e
 ANDROID_SDK_TOOLS_VERSION ?= 6514223
 ANDROID_SDK_BUILD_TOOLS_VERSION ?= 29.0.3
 ANDROID_HOME ?= $(HOME)/.android
-ANDROID_API_LEVEL ?= 27
+ANDROID_API_LEVEL ?= 33
```
misl6: Let's target API 36 :)

pythonforandroid/recipes/pandas/__init__.py
```diff
@@ -3,10 +3,10 @@
 
 
 class PandasRecipe(MesonRecipe):
-    version = 'v2.2.1'
-    url = 'git+https://github.com/pandas-dev/pandas'  # noqa
+    version = 'v2.3.0'
+    url = 'git+https://github.com/pandas-dev/pandas'
     depends = ['numpy', 'libbz2', 'liblzma']
-    hostpython_prerequisites = ["Cython~=3.0.5"]  # meson does not detects venv's cython
+    hostpython_prerequisites = ["Cython<4.0.0a0", "versioneer", "numpy"]  # meson does not detects venv's cython
```
misl6, line 9: We're really that optimistic that works with Cython<4.0.0 ?

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -143,9 +143,15 @@ class Python3Recipe(TargetPythonRecipe):
         '__pycache__',
         'tests'
     }
-    '''The directories from site packages dir that we don't want to be included
+    '''the directories from site packages dir that we don't want to be included
     in our python bundle.'''
 
+    site_packages_excluded_dir_exceptions = [
+        'numpy',
+    ]
```
misl6: I see a discussion about that, can we add a comment that explains why is that needed?

misl6 (commented):
Thank you!

Left some comments, can you take a look at it?

pythonforandroid/recipes/pandas/__init__.py
T-Dynamos: According to [pandas' pyproject.toml](https://github.com/pandas-dev/pandas/blob/628c7fb/pyproject.toml#L8), they also specify Cython <4.0.0a0.

misl6: Ok, if they're optimistic on their package, we're too on them behalf 😅

Agree on keeping 4.0.0a0

misl6 (approved):
LGTM. Thank you!

PR #3167, opened 2025-06-30
Update broadcast.py

Sahil-pixel:
To avoid crash on_pause() on_resume() if we do call start() and stop() there 

jnius.jnius.JavaException: JVM exception occurred: java.lang.IllegalThreadStateException


Sahil-pixel (commented):
I have Tested This is working fine . No crash of thread on _pause on_resume of kivy app

Issue #3188, opened 2025-08-19
pybind11 ModuleNotFoundError: No module named 'cmake'


Sahil-pixel:
p4a==master 
NDK r25b

 Building pybind11 for arm64-v8a
[INFO]:    pybind11 apparently isn't already in site-packages
[INFO]:    Installing pybind11 into site-packages
[INFO]:    -> directory context /home/My_Kivy/.buildozer/android/platform/build-arm64-v8a/build/other_builds/pybind11/arm64-v8a__ndk_target_21/pybind11
[DEBUG]:   -> running python3 setup.py install -O2 --root=/home/My_Kivy/.buildozer/android/platform/build-arm64-v8a/build/python-installs/dlb/arm64-v8a --install-lib=.
[DEBUG]:   	Traceback (most recent call last):
[DEBUG]:   	  File "/home/.local/bin/cmake", line 5, in <module>
[DEBUG]:   	    from cmake import cmake
[DEBUG]:   	ModuleNotFoundError: No module named 'cmake'
[DEBUG]:   	Traceback (most recent call last):
[DEBUG]:   	  File "/home/My_Kivy/.buildozer/android/platform/build-arm64-v8a/build/other_builds/pybind11/arm64-v8a__ndk_target_21/pybind11/setup.py", line 117, in remove_output
[DEBUG]:   	    yield
[DEBUG]:   	  File "/home/My_Kivy/.buildozer/android/platform/build-arm64-v8a/build/other_builds/pybind11/arm64-v8a__ndk_target_21/pybind11/setup.py", line 139, in <module>
[DEBUG]:   	    subprocess.run(cmd, c

Novfensec:
`depends = ['setuptools', 'cmake']`

`cmake` missing in depends.

Sahil-pixel:
```python 
from pythonforandroid.recipe import PythonRecipe
from os.path import join


class Pybind11Recipe(PythonRecipe):
    """
    Recipe to build pybind11 for Android using cmake.
    """

    version = "2.11.1"
    url = "https://github.com/pybind/pybind11/archive/refs/tags/v{version}.zip"

    # Make sure we depend on setuptools and cmake
    depends = ["setuptools", "cmake"]
    # Don’t let p4a try to run setup.py through targetpython
    call_hostpython_via_targetpython = False
    # Install into hostpython, because pybind11 is a header-only lib
    install_in_hostpython = True

    def get_recipe_env(self, arch):
[...]

PR #3192, opened 2025-08-20
Update __init__.py

Sahil-pixel:
#3188

### DIFF
--- a/pythonforandroid/recipes/pybind11/__init__.py
+++ b/pythonforandroid/recipes/pybind11/__init__.py
@@ -3,15 +3,31 @@ from os.path import join
 
 
 class Pybind11Recipe(PythonRecipe):
+    """
+    Recipe to build pybind11 for Android using cmake.
+    """
 
-    version = '2.11.1'
-    url = 'https://github.com/pybind/pybind11/archive/refs/tags/v{version}.zip'
-    depends = ['setuptools']
+    version = "2.11.1"
+    url = "https://github.com/pybind/pybind11/archive/refs/tags/v{version}.zip"
+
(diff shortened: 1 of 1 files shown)


pythonforandroid/recipes/pybind11/__init__.py
```diff
@@ -3,15 +3,31 @@
...
+    depends = ["setuptools", "cmake"]
+    # Don’t let p4a try to run setup.py through targetpython
     call_hostpython_via_targetpython = False
+    # Install into hostpython, because pybind11 is a header-only lib
     install_in_hostpython = True
 
+    def get_recipe_env(self, arch):
+        env = super().get_recipe_env(arch)
+        # Ensure cmake is found inside hostpython
+        env["PATH"] = "/usr/bin:" + env.get("PATH", "")
```
AndreMiras, line 23: This is weird, why would you need to add `/usr/bin` to the PATH, that should come as the default

AndreMiras (changes requested):
Thanks for the pull request.
Could update it to simply add `cmake` to the `depends` list and not change all the linting related stuff. Because this would keep changing back and forth until we integrate black or ruff

T-Dynamos:
Actually this recipe is not even required, it can be replaced with one line:
```python
hostpython_prerequisites = ['pybind11']
```

Novfensec:
> Actually this recipe is not even required, it can be replaced with one line:
> ```python
> hostpython_prerequisites = ['pybind11']
> ```

Exactly 😅

AndreMiras:
Sweet, thanks for the heads up, someone wants to make a pull request to remove the recipe then?

PR #3208, opened 2025-09-14
`android`: shift to PPR

T-Dynamos:
Also enables sound module.

### DIFF
--- a/pythonforandroid/recipes/android/__init__.py
+++ b/pythonforandroid/recipes/android/__init__.py
@@ -1,11 +1,11 @@
-from pythonforandroid.recipe import CythonRecipe, IncludedFilesBehaviour
+from pythonforandroid.recipe import PyProjectRecipe, IncludedFilesBehaviour
 from pythonforandroid.util import current_directory
 from pythonforandroid import logger
 
 from os.path import join
 
 
-class AndroidRecipe(IncludedFilesBehaviour, CythonRecipe):
+class AndroidRecipe(IncludedFilesBehaviour, PyProjectRecipe):
     # name = 'android'
     version = None
     url = None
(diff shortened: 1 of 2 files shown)


pythonforandroid/recipes/android/__init__.py
```diff
@@ -1,23 +1,24 @@
...
-class AndroidRecipe(IncludedFilesBehaviour, CythonRecipe):
+class AndroidRecipe(IncludedFilesBehaviour, PyProjectRecipe):
     # name = 'android'
     version = None
     url = None
 
     src_filename = 'src'
 
     depends = [('sdl3', 'sdl2', 'genericndkbuild'), 'pyjnius']
+    hostpython_prerequisites = ["cython~=3.0.12"]
```
AndreMiras: Do you think this could be improved?

T-Dynamos: Can you suggest a better version range?

AndreMiras: I haven't tested the recipe, but would something like <3.1 be OK?
Edit:
```
hostpython_prerequisites = ["Cython>=3.0,<3.2"]
```

pythonforandroid/recipes/android/__init__.py
```diff
@@ -1,23 +1,24 @@
...
-class AndroidRecipe(IncludedFilesBehaviour, CythonRecipe):
+class AndroidRecipe(IncludedFilesBehaviour, PyProjectRecipe):
     # name = 'android'
     version = None
     url = None
 
     src_filename = 'src'
 
     depends = [('sdl3', 'sdl2', 'genericndkbuild'), 'pyjnius']
+    hostpython_prerequisites = ["Cython>=3.1,<3.2"]
```
AndreMiras: Agr my bad, I didn't realise the Cython 3.1 build was failing on some arch e.g. x86
I played with it a bit and I got it working with Cython 0.29 and 3.0 with x86, so I guess we should change it to:
```
hostpython_prerequisites = ["Cython>=0.29,<3.1"]
```
Sorry for misleading you with my initial comment thinking 3.1 would do it

AndreMiras: Thanks for the quick fixes and the neat rebasing as always :muscle:

AndreMiras (approved):
Looks good will merge as soon as the CI goes green

PR #3221, opened 2025-10-05
:construction_worker: Runtime test the app from the CI

AndreMiras:
Leverage `reactivecircus/android-emulator-runner` to run the testapps artifact on an emulator and check the ADB logs.
We're currently checking for the following pattern from the logcat output:
- "I python: Initialized python"
- "I python: Ran 14 tests in"
- "I python: OK"

Note that we're keeping things simple for the first iteration and we only run the `ubuntu-latest-sdl2-artifacts` artifact.
In the future we may want to test the different bootstraps, build host and even updated recipes dynamically.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -80,7 +80,7 @@ jobs:
             target: testapps-qt
     steps:
     - name: Checkout python-for-android
-      uses: actions/checkout@v4
+      uses: actions/checkout@v5
     - name: Build python-for-android docker image
       run: |
         docker build --tag=kivy/python-for-android .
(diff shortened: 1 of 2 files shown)


.github/workflows/push.yml
```diff
@@ -162,6 +162,26 @@ jobs:
           name: ${{ matrix.runs_on }}-${{ matrix.bootstrap.name }}-artifacts
           path: dist
 
+  test_on_emulator:
+    name: Run App on Emulator
+    needs: ubuntu_build
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
```
cclauss:
```suggestion
      - uses: actions/checkout@v5
```
Or better yet, automate these changes via:
* #3078

AndreMiras: Yeah I was using whatever version was there already.
But I bumped all this file to v5 then.
Regarding dependabot, isn't it already running on open source repositories by default already?

cclauss: Dependabot is not on by default.  It can be turned on at an org level or with a `dependabot.yml` file.

cclauss (approved):

cclauss:
@T-Dynamos  When you think a pull request is useful and is ready to be merged, ___please consider giving it a positive review___.

Every check mark ✔️ at the top right of this page gives project maintainers confidence that the proposed changes have been read through and deemed both useful and safe to merge into the codebase.  Lots of 👍 and "_what is the ETA?_" comments are easier for maintainers to ignore than ✔️✔️✔️✔️✔️ from several different reviewers.

Anyone can review a pull request on GitHub.  To do so here:
1. Scroll to the top of this page.
2. Click the `Files changed` tab and read through each file carefully looking for potential issues.
3. Click the `Review changes` button.
4. Click `Approve` (or one of the other options) and add comments only if they have not already been stated in the PR.
5. Click `Submit review` so that your ✔️ will be added to the list.

kuzeyron (approved):

PR #3236, opened 2025-10-17
Use `pip install` instead of `setup.py install` in recipes

clayote:
On Python 3.14, recipes such as msgpack wouldn't build anymore, because the way `setup.py install -O2` did its byte-compiling called into `distutils`, a package that no longer exists in Python 3.14. This PR replaces `setup.py install`, but not any other calls to `setup.py`, with `pip install`.


clayote:
Wrong target, belongs in T-Dynamos' branch

PR #3241, opened 2025-10-20
Patch out the `ast.Str` reference in Kivy, for Python 3.14 compatibility

clayote:
The basic `kivy` recipe didn't work on Python 3.14, because Kivy 2.3.1, still the version in the recipe, has a line in `kivy.lang.parser` that refers to `ast.Str`, which no longer exists in Python 3.14. This PR adds a patch to remove that before building Kivy.

I suppose this breaks Python 3.6 compatibility, but I don't think p4a supports that anymore.


clayote:
I thought the changes to the py3.14 recipe itself in [6559b2a](https://github.com/kivy/python-for-android/pull/3241/commits/6559b2af1dad52ae8438de21c291a1494969285c) were out of scope,  but would not be opposed to putting them back in.

AndreMiras (approved):

AndreMiras:
LGTM thanks, I'll let it there for a couple of days and merge it if nobody has anything against it

T-Dynamos (approved):

PR #3242, opened 2025-10-20
Update to Python 3.14, remove distutils

clayote:
Python 3.14 finally removes the `distutils` package, which breaks the way p4a byte-compiles all its recipes using `setup.py install`. This PR replaces all instances of `setup.py install` with `pip install`, but leaves all other uses of `setup.py`.

Two recipes, `android` and `freetype`, used distutils in other ways. I've replaced those uses with `setuptools`, too.


clayote:
I bumped the freetype versions just because I happened to be looking at them. If this isn't good practice, I'll put them back.

T-Dynamos:
@clayote Can we port `freetype-py` to `PyProjectRecipe`?

clayote:
hmmm... sure, why not

clayote:
The way I had to change the source url to get the newly PyProject based FreetypePy recipe to build suggests that I'll need to make similar changes to every other recipe. Well, it had to happen eventually.

clayote:
[The version of cffi that adds Python 3.14 support also drops Python 3.8 support](https://github.com/python-cffi/cffi/releases/tag/v2.0.0).

I think that means p4a will need to pick one or the other, and I'd certainly favor Python 3.14.

T-Dynamos:
@clayote Let's wait for https://github.com/kivy/python-for-android/pull/3233

T-Dynamos (approved):
LGTM

clayote:
Any recipe not using a pythonhosted.org url will likely need to be revised, because eg. GitHub hosted bundles don't have version metadata that setuptools can use.

T-Dynamos:
I think those recipes aren’t a top priority since they’re not widely used and most of the popular ones are already up to date. Let’s handle the rest in another PR since this one already includes 15 files.

kuzeyron (approved):

kuzeyron:
Thanks!

PR #3244, opened 2025-10-22
Remove some stray uses of `distutils`

clayote:
We're all `setuptools` all the time now.

Apparently, [the `'distutils.commands'` entry points in `setup.py` are unchanged](https://setuptools.pypa.io/en/latest/userguide/extension.html#customizing-commands), despite `distutils` itself not existing in Python 3.14. Weird!


clayote:
I was expecting to find `distutils` in more recipes than just `android` and `freetype-py`, but that was all.

clayote:
Looks like this won't work without #3241, so I may as well include it in that PR.

PR #2465, opened 2021-06-16
Adds libm to Pillow recipe

Sahil-pixel:
https://github.com/kivy/buildozer/issues/1020
I have put requirements=pillow directly without any modification
.buildozer/android/platform/python-for-android/pythonforandroid/recipes/Pillow/init.py
Results
In Android 8 running well without any problem

In Android 5 crashing
ImportError: dlopen failed: cannot locate symbol "log" referenced by "_imaging.so"...

and after the change ,In both device running well.
env['CFLAGS'] += cflags + " -lm"

PR #3256, opened 2025-11-10
Fix `libm` linkage for Pillow recipe builds

DexerBR:
This PR has the same intention of fixing the issue addressed by this PR previously: https://github.com/kivy/python-for-android/pull/2465


AndreMiras (approved):
Thank you 🙏

PR #3279, opened 2026-02-13
Fix Qt/PySide6 based android service

AshkanV:
Fix android service containing PySide6 codes crashes the service due to Qt preparations not done and Qt libs not loaded


AndreMiras:
Thanks for the changes :pray: 
It looks like the linter is failing, make run `make docker/java-lint-fix` to fix and let the CI run fully.
Also try to rebase the branch since we made a few CI fixes recently that might be helpful

PR #3280, opened 2026-02-14
Add support for prebuilt wheels

T-Dynamos:
### `--extra-index-url`

Additional package indexes to search for prebuilt wheels.
Can be used multiple times.

**Example:**

```bash
--extra-index-url https://chaquo.com/pypi-13.1/ \
--extra-index-url https://anshdadwal.is-a.dev/p4a-wheels/p4a/
```

### `--skip-prebuilt`

Do not use prebuilt wheels; always build from source.

### `--use-prebuilt-version-for`

For these packages, ignore recipe-pinned versions and use the latest prebuilt from the extra index if available.
Only applies to packages with a recipe.
Can be used multiple times.

**Example:**

```bash
--use-prebuilt-version-for numpy
```

### `--save-wheel-dir`

Directory where wheels built by `PyProjectRecipe` are saved.

If set, any wheel produced during a recipe build is copied to this directory.

## New behavior

When a recipe (any `PyProjectRecipe`) is built, it checks whether a prebuilt version is available. If so, it uses it; otherwise, it falls back to building from source.

test requirements: `requirements = python3, kivy==master, numpy, openai`

with `--use-prebuilt-version-for kivy`

```
[INFO]:    Building numpy for arm64-v8a
[INFO]:    numpy apparently isn't already in site-packages
[INFO]:    Writing cross file at: /tmp/android.meson.cross
[DEBUG]:   -> running pip3 install numpy==2.3.0 --ignore-installed --disable-pip-version-check --python-version 3.14.2 --only-binary=:all: --no-deps --platform=android_24_arm64_v8a --platform=android_24_aarch64 --extra-index-url https://anshdadwal.is-a.dev/p4a-wheels/p4a/ --extra-index-url https://chaquo.com/pypi-13.1/ --dry-run
[DEBUG]:   	Defaulting to user installation because normal site-packages is not writeable
[DEBUG]:   	Looking in indexes: https://pypi.org/simple, https://anshdadwal.is-a.dev/p4a-wheels/p4a/, https://chaquo.com/pypi-13.1/
[DEBUG]:   	Collecting numpy==2.3.0
[DEBUG]:   	 Downloading https://github.com/T-Dynamos/p4a-wheels/releases/download/1.2/numpy-2.3.0-cp314-cp314-android_24_aarch64.whl (6.9 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 6.9/6.9 MB 683.1 kB/s  0:00:09
[...]

PR #3272, opened 2025-12-13
`ppr`: fix issue with `extra_build_args` initialization

T-Dynamos:
Closes #3285

@mriscoc Please let me know if this fixes the issue.

PR #3283, opened 2026-02-21
Fixed Python version support.

GvozdevLeonid:
Fixed a bug when selecting libpythonХХХ.so version is decreasing (3.14 -> 3.5), but the check for the absence of the Python library ends at version 3.14, I changed it to 3.5. I also slightly changed the Python recipe to return support for version selection via buildozer

### DIFF
--- a/pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonUtil.java
+++ b/pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonUtil.java
@@ -59,8 +59,8 @@ public class PythonUtil {
             addLibraryIfExists(libsList, name, libsDir);
         }
 
-        for (int v = 14; v >= 5; v--) {
-            libsList.add("python3." + v + (v <= 7 ? "m" : ""));
+        for (int v = 14; v >= 8; v--) {
+            libsList.add("python3." + v);
         }
 
         libsList.add("main");
@@ -85,7 +85,7 @@ public class PythonUtil {
                 // load, and it has failed, give a more
                 // general error
                 Log.v(TAG, "Library loading error: " + e.getMessage());
-                if (lib.startsWith("python3.5") && !foundPython) {
+                if (lib.startsWith("python3.8") && !foundPython) {
                     throw new RuntimeException("Could not load any libpythonXXX.so");
                 } else if (lib.startsWith("python")) {
                     continue;
--- a/pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonUtil.java
+++ b/pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonUtil.java
@@ -59,7 +59,7 @@ public class PythonUtil {
             addLibraryIfExists(libsList, name, libsDir);
         }
 
-        for (int v = 14; v >= 5; v--) {
+        for (int v = 5; v <= 3.14; v++) {
             libsList.add("python3." + v + (v <= 7 ? "m" : ""));
         }
 
--- a/pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonUtil.java
+++ b/pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonUtil.java
@@ -59,7 +59,7 @@ public class PythonUtil {
             addLibraryIfExists(libsList, name, libsDir);
         }
 
-        for (int v = 5; v <= 3.14; v++) {
+        for (int v = 14; v >= 5; v--) {
             libsList.add("python3." + v + (v <= 7 ? "m" : ""));
         }
 
@@ -85,7 +85,7 @@ public class PythonUtil {
                 // load, and it has failed, give a more
                 // general error
                 Log.v(TAG, "Library loading error: " + e.getMessage());
-                if (lib.startsWith("python3.14") && !foundPython) {
+                if (lib.startsWith("python3.5") && !foundPython) {
                     throw new RuntimeException("Could not load any libpythonXXX.so");
                 } else if (lib.startsWith("python")) {
                     continue;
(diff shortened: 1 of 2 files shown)


T-Dynamos (approved):
LGTM! I was trying to do something same in https://github.com/kivy/python-for-android/pull/3280.

GvozdevLeonid:
I haven't figured out yet if it's possible to somehow remove the dependency on sqlite (previously, you could blacklist sqlite in buildozer and p4a, but now my build crashes because of it)

Sorry for such a large number of commits. I edited through GitHub Web and made a ton of mistakes...

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -64,28 +63,6 @@ class Python3Recipe(TargetPythonRecipe):
...
-
-    if _p_version.minor >= 14:
-        patches.append('patches/3.14_armv7l_fix.patch')
-        patches.append('patches/3.14_fix_remote_debug.patch')
-
-    if shutil.which('lld') is not None:
-        if _p_version.minor == 7:
-            patches.append("patches/py3.7.1_fix_cortex_a8.patch")
-        elif _p_version.minor >= 8:
-            patches.append("patches/py3.8.1_fix_cortex_a8.patch")
```
AndreMiras, line 87: Wow why was this at the root of the class in the first place? 🤷
Thanks for flagging.
This clean up could deserve its own pull request.

GvozdevLeonid: I just wanted to fix my build 😅

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -210,6 +182,32 @@ def link_version(self):
...
         )
 
+    def apply_patches(self, arch, build_dir=None):
+
+        _p_version = Version(self.version)
+        if _p_version.major == 3 and _p_version.minor == 7:
+            self.patches += [
+                'patches/py3.7.1_fix-ctypes-util-find-library.patch',
+                'patches/py3.7.1_fix-zlib-version.patch',
+            ]
```
AndreMiras, line 192: We may have a bug here, `apply_patches()` is called once per architecture and each call appends to `self.patches`. So if you build for both arm64-v8a and armeabi-v7a, the patches list accumulates duplicates on the second call.
The same issue may happen with `self.configure_args.append('--with-build-python=...')` in `set_libs_flags()`

pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonUtil.java
```diff
@@ -85,7 +85,7 @@ public static void loadLibraries(File filesDir, File libsDir) {
                 // load, and it has failed, give a more
                 // general error
                 Log.v(TAG, "Library loading error: " + e.getMessage());
-                if (lib.startsWith("python3.14") && !foundPython) {
+                if (lib.startsWith("python3.5") && !foundPython) {
```
AndreMiras: I guess that bug got introduced in #3242 (as we changed the for loop from ascending to descending) and may deserve it's own PR or is the rest of the change also related?

pythonforandroid/recipes/python3/__init__.py
GvozdevLeonid: You're probably right. It might make sense to add a local variable named patches to the function so it starts from scratch each time.

pythonforandroid/bootstraps/common/build/src/main/java/org/kivy/android/PythonUtil.java
GvozdevLeonid: I only changed the version from 3.14 to 3.5. I initially reverted to an incremental version, but then I thought you probably made the right change, since it makes sense to test with the highest version.

GvozdevLeonid: Does version 3.5 make sense? I think the latest updatable version is 3.10.

Besides, almost all the recipes use libraries with versions that don't support such an old version of Python. If someone still needs it, they can fork it.

AndreMiras: Yeah I also think the descending makes sense for the reason you gave even though it's error prone.
As for the version let's maybe go with 3.8 even though it's not supported, this is what we have until now in the CI. Then when we (eventually) make a new p4a release we can ditch the older versions.

GvozdevLeonid: Okay. I'll fix the issues.

kuzeyron (approved):

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -210,6 +182,34 @@ def link_version(self):
...
+        if _p_version.minor >= 11:
+            self.patches.append('patches/cpython-311-ctypes-find-library.patch')
+
+        if _p_version.minor >= 14:
+            self.patches.append('patches/3.14_armv7l_fix.patch')
+            self.patches.append('patches/3.14_fix_remote_debug.patch')
+
+        if shutil.which('lld') is not None:
+            if _p_version.minor == 7:
+                self.patches.append("patches/py3.7.1_fix_cortex_a8.patch")
```
AndreMiras, line 206: Nit: probably not needed anymore since we dropped that version

pythonforandroid/recipes/python3/__init__.py
```diff
@@ -210,6 +182,34 @@ def link_version(self):
...
+            elif _p_version.minor >= 8:
+                self.patches.append("patches/py3.8.1_fix_cortex_a8.patch")
+
+        self.patches = list(set(self.patches))
```
AndreMiras, line 210: Patching order is no longer preserved with the set, just something to have in mind in case patch depend on one another (which I'm not sure would be a good idea anyway)

AndreMiras (approved):
Thanks for the bug fix and the follow ups.
I left two comments for the records, but nothing that absolutely needs to be addressed in this pull request.

pythonforandroid/recipes/python3/__init__.py
GvozdevLeonid: It seems that this will not be relevant after closing https://github.com/kivy/python-for-android/pull/3272

I suggested that all lists responsible for the assembly be divided into internal (for a specific assembly and reset before each build) and external for writing the recipe itself
https://github.com/kivy/python-for-android/issues/3285

PR #3308, opened 2026-04-28
Fix pyjnius linking with Qt bootstrap libmain

lazycat11:
When using the Qt bootstrap, ndk-build generates the main library as libmain_<abi>.so, for example libmain_arm64-v8a.so. pyjnius links with -lmain, so the linker cannot find the ABI-suffixed library.

This change creates a libmain.so alias before building pyjnius and adds the bootstrap libs/<abi> directory to LDFLAGS.

Tested with Buildozer / python-for-android on arm64-v8a.

### DIFF
--- a/pythonforandroid/recipes/pyjnius/__init__.py
+++ b/pythonforandroid/recipes/pyjnius/__init__.py
@@ -2,7 +2,7 @@ from pythonforandroid.recipe import PyProjectRecipe
 from pythonforandroid.toolchain import shprint, current_directory, info
 from pythonforandroid.patching import will_build
 import sh
-from os.path import join
+from os.path import exists, join
 
 
 class PyjniusRecipe(PyProjectRecipe):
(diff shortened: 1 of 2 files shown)


T-Dynamos:
Wouldn't this increase app size? I think it would be better to patch `pyjnius` to use `-lmain_` instead.

lazycat11:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我最近正在繁忙中，无法亲自回复您的邮件。我将在繁忙结束后，尽快给您回复。

lazycat11:
Thanks, agreed.

I updated the PR so it no longer creates a libmain.so alias. Instead, pyjnius now uses PYJNIUS_ANDROID_LIBMAIN with a fallback to "main". The recipe only sets this to "main_<arch>" when libmain_<arch>.so exists in the bootstrap output, so existing libmain.so cases should continue to link via -lmain unchanged.

T-Dynamos:
@lazycat11 `flake8` tests are failing can you have a look?

lazycat11:
Thanks for the heads-up. The failing check was caused by a Flake8 continuation indentation issue in `pythonforandroid/recipes/pyjnius/__init__.py`.

AndreMiras (approved):

PR #3309, opened 2026-04-29
Update SDL3_mixer recipe to version 3.2.0, remove libgme patch (not needed as SDL3_mixer disables it by default) and fix compilation crash

kengoon:
This pull request updates the `sdl3_mixer` recipe to use the official 3.2.0 release of SDL_mixer instead of a specific commit, and removes the custom patch that disabled libgme support. This simplifies the recipe and ensures it uses an official, stable release.

Dependency update:

* Updated the `version` to `"3.2.0"` and changed the `url` in `pythonforandroid/recipes/sdl3_mixer/__init__.py` to point to the official SDL_mixer 3.2.0 release tarball.


AndreMiras (approved):
Looking good, thanks for the cleanup, did you get a chance to test it at runtime?

pythonforandroid/recipes/sdl3_mixer/disable-libgme.patch
```diff
@@ -1,12 +0,0 @@
-diff -Naur SDL3_mixer.orig/Android.mk SDL3_mixer/Android.mk
---- SDL3_mixer.orig/Android.mk	2025-03-16 21:05:19
-+++ SDL3_mixer/Android.mk	2025-03-16 21:06:33
-@@ -31,7 +31,7 @@
- WAVPACK_LIBRARY_PATH := external/wavpack
- 
- # Enable this if you want to support loading music via libgme
--SUPPORT_GME ?= true
-+SUPPORT_GME ?= false
```
AndreMiras, line 9: Are we sure we want to skip that one?
Edit: my bad it's disabled by default indeed, ignore the comment

kengoon:
> Looking good, thanks for the cleanup, did you get a chance to test it at runtime?

Not yet, I'm about to do that.

kengoon:
@AndreMiras Just tested it and it works as expected

PR #3310, opened 2026-04-30
Fix tflite-runtime numpy include path

lazycat11:
tflite-runtime still looks for NumPy headers under numpy/core/include, but current NumPy installs them under numpy/_core/include.

This matches the include path already used by the numpy, pandas, scipy, and opencv recipes. The change first tries numpy/_core/include and falls back to numpy/core/include for older layouts.

Validation:
- python3 -m py_compile pythonforandroid/recipes/tflite-runtime/__init__.py
- git diff --check
- verified local NumPy build output contains numpy/_core/include


AndreMiras (approved):
LGTM thanks

PR #3316, opened 2026-05-08
Added msgspec recipe

GvozdevLeonid:
Added a recipe for building the msgspec library. Tested on Python 3.11 + arm64-v8a.


AndreMiras (approved):
Thanks, looking good.
Python 3.8 build fails, but support should have been dropped quite a while ago, so it's accepted to let it fail

Issue #3326, opened 2026-05-10
[docs]: Documentation missing for new recipe types


T-Dynamos:
Documentation about new recipe classes - `PyProjectRecipe`, `MesonRecipe` and `RustCompiledComponentsRecipe` is missing. 

https://python-for-android.readthedocs.io/en/latest/recipes.html

Also users must be informed that they should use `PyProjectRecipe` by default and why it is important for saving and building wheels.

AndreMiras:
I've started with `PyProjectRecipe`, let me know what you think https://github.com/kivy/python-for-android/pull/3334

PR #3334, opened 2026-05-21
:memo: Document PyProjectRecipe usage

AndreMiras:
Add PyProjectRecipe guidance for modern pyproject.toml packages, including the wheel-oriented build flow and notes on dependency, host build, patch, and environment customization, refs #3326

### DIFF
--- a/doc/source/recipes.rst
+++ b/doc/source/recipes.rst
@@ -123,10 +123,11 @@ This is the core of what's necessary to write a recipe, but has not
 covered any of the details of how one actually writes code to compile
 for android. This is covered in the next sections, including the
 `standard mechanisms <standard_mechanisms_>`_ used as part of the
-build, and the details of specific recipe classes for Python, Cython,
-and some generic compiled recipes. If your module is one of the
-latter, you should use these later classes rather than reimplementing
-the functionality from scratch.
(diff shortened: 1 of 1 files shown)


doc/source/recipes.rst
```diff
@@ -380,6 +385,40 @@ patching to remove this import or make it fail quietly.
...
+the older ``setup.py install`` flow.
+
+For instance::
+
+  from pythonforandroid.recipe import PyProjectRecipe
+
+  class ExampleRecipe(PyProjectRecipe):
+      version = '1.2.3'
+      url = 'https://example.com/example-{version}.tar.gz'
+      depends = ['python3']
```
T-Dynamos: `PyProjectRecipe` is always dependent on `python3` - this may confuse users into thinking it's required by default. So let's change to something else maybe like `pybind11`, `requests`

AndreMiras: Thanks for the heads up, I've updated the depends with `['pycparser', 'libffi']` and added the following:
```
You do not need to include ``python3`` in ``depends``; even when a recipe
overrides ``depends``, ``PythonRecipe`` adds the ``python3`` dependency.
```
Le me know what you think, maybe that extra is overkill

T-Dynamos (approved):

PR #3340, opened 2026-05-25
:bug: Resolve Meson tools through recipe env

AndreMiras:
MesonRecipe installs meson and ninja as hostpython prerequisites and PyProjectRecipe exposes those executables by prepending the hostpython bin directories to the recipe environment PATH.
Recipes that call sh.meson or sh.ninja directly can fail before _env is applied when those tools are not also visible on the current Python process PATH.

Add shared MesonRecipe helpers that resolve commands against the recipe env PATH and return explicit sh.Command instances.
Use those helpers in the libthorvg and libcairo manual Meson/Ninja build steps.

Add a regression test that reproduces the environment-sensitive lookup failure and verifies the helper resolves meson from env PATH.

### DIFF
--- a/pythonforandroid/recipe.py
+++ b/pythonforandroid/recipe.py
@@ -1567,6 +1567,18 @@ class MesonRecipe(PyProjectRecipe):
             if arg not in self.extra_build_args:
                 self.extra_build_args.append(arg)
 
+    def get_recipe_env_command(self, command, env):
+        command_path = shutil.which(command, path=env["PATH"])
+        if command_path is None:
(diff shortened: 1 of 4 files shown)


AndreMiras:
Should hopefully fix recent buildozer integration tests issues when trying to build kivy -> libthorvg -> `sh.CommandNotFound: meson`, refs: https://github.com/kivy/buildozer/actions/runs/26367835861/job/77629193613

T-Dynamos (approved):

kuzeyron (approved):

PR #3343, opened 2026-06-04
Fix Qt bootstrap handling of non-back key events

lazycat11:
## Summary
- handle only Android Back key events in the Qt bootstrap double-tap-to-exit logic
- pass all other key events through to Qt via `super.onKeyDown`
- checked the other bootstraps: webview/service_only already guard KEYCODE_BACK, and the remaining bootstraps do not use this double-back exit logic

## Testing
- `git diff --check`
- fork GitHub Actions workflow `Unit tests & build apps` run 26929233140 completed successfully, including Flake8, Java Spotless, pytest matrix, recipe tests, emulator run, and Qt app build

### DIFF
--- a/pythonforandroid/bootstraps/qt/build/src/main/java/org/kivy/android/PythonActivity.java
+++ b/pythonforandroid/bootstraps/qt/build/src/main/java/org/kivy/android/PythonActivity.java
@@ -123,8 +123,10 @@ public class PythonActivity extends QtActivity {
 
     @Override
     public boolean onKeyDown(int keyCode, KeyEvent event) {
-        // If it wasn't the Back key or there's no web page history, bubble up to the default
-        // system behavior (probably exit the activity)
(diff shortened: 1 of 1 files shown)


pythonforandroid/bootstraps/qt/build/src/main/java/org/kivy/android/PythonActivity.java
```diff
@@ -123,8 +123,10 @@ public void onDestroy() {
 
     @Override
     public boolean onKeyDown(int keyCode, KeyEvent event) {
-        // If it wasn't the Back key or there's no web page history, bubble up to the default
-        // system behavior (probably exit the activity)
+        if (keyCode != KeyEvent.KEYCODE_BACK) {
```
kuzeyron, line 126: This seems not to be the best approach as it's colliding with the rest of its code and its purpose. Instead of doing this you should try another approach from the app itself on how it handles the back key.

PR #3345, opened 2026-06-08
Update recipe version

RobertFlatt:
Update version to 3.5.1


AndreMiras (approved):
Thanks LGTM, if it's expected that macOS build fails, then maybe have it more explicit  with a raise exception or documenting with docstring like we have for gevent.
https://github.com/kivy/python-for-android/blob/v2026.05.09/pythonforandroid/recipes/gevent/__init__.py#L1-L10
I can probably look this up in a follow up.

PR #3350, opened 2026-06-25
Fix `setup.py` parser config.

T-Dynamos:
Move `--use-setup-py` and `--ignore-setup-py` to `generic_parser`.                                      
                                                                                                 
*Problem* : `--ignore-setup-py` (and `--use-setup-py`) were defined only on `parser_packaging`, which is a parent of the apk/aar/aab subcommands but not of create. 

When running p4a create `--ignore-setup-py` ..., the flag was silently dropped into `unknown_args` by `parse_known_args`, and args.   
        
`ignore_setup_py` was never set (always defaulted to False). This made the flag a no-op for the    
create command.

PS: AI assisted PR.

### DIFF
--- a/pythonforandroid/toolchain.py
+++ b/pythonforandroid/toolchain.py
@@ -333,6 +333,20 @@ class ToolchainCL:
             default='')
 
         generic_parser.add_argument(
+            '--use-setup-py', dest="use_setup_py",
+            action='store_true', default=False,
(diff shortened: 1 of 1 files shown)


kuzeyron (approved):

kuzeyron:
LGTM! Thanks!

Issue #3050, opened 2024-08-17
Tiktoken is for EM_X86_64 | ImportError: dlopen failed ()


sensoryfox:
The issue is indeed a bug and not a support request.
I have a short, runnable example that reproduces the issue.
I used the grave accent (aka backticks) to format code or logs when appropriated.

**Versions**

- Python: 3.11.5
- OS: Ubuntu
- Kivy: 2.3.0
- Cython: 3.0.11

**Description**

I'm trying to use the `tiktoken` library in my Android app built with Python-for-Android. After building the APK and running it on my device, I get the following error:
**ImportError: dlopen failed: "/data/data/org.test.sensory_app/files/app/_python_bundle/site-packages/tiktoken/_tiktoken.so" is for EM_X86_64 (62) instead of EM_AARCH64 (183)**

This error means that the `_tiktoken.so` library, which is a native library compiled for ARM64 architecture, is not compatible with my device's architecture (ARM64). It seems like the library was compiled for x86_64 instead of ARM64.

I believe that a recipe for `tiktoken` is needed to properly build and install it for Android.

**buildozer.spec**
Command:
buildozer android debug

Spec file:
[app]
title = SensoryContext
package.name = sensory_app
package.domain = org.test
android.private_storage = True
source.dir = .
source.includes = Plugin/*
version = 0.1
source.include_exts = json
entrypoint = main.py
requirements = python3,kivy,kivymd,setuptools-rust>=1.5.2,setuptools>=62.4,python-telegram-bot>=21.4,openai,httpx,pillow,pydub,tenacity,sniffio,httpcore,h11,trio,socksio,regex>=2022.1.18,requests>=2.26.0, tiktoken
android.permissions = INTERNET,WRITE_EXTERNAL_STORAGE
android.ndk = 25b
android.ndk_api = 21

android.archs = arm64-v8a
android.cflags = -Wno-cast-function-type-strict

[buildozer]
log_level = 2
warn_on_root = 1

I would be very grateful if you could create a recipe for `tiktoken` to solve this issue.

RobertFlatt:
> I believe that a recipe for tiktoken is needed to properly build and install it for Android.

yes, because ticktoken contains this Rust code https://github.com/openai/tiktoken/blob/main/src/lib.rs

As far as I know there are no examples of p4a recipes using Rust.

sensoryfox:
[...]

PR #3366, opened 2026-08-15
Remove venv creation for python package install stage

T-Dynamos:
* Virtual environment creation is not required now as we are using our own custom pip now.
* This PR strictly prevents to install wrong arch modules like it was allowed previously causing issues like: https://github.com/kivy/python-for-android/issues/3050
* It might break for some users which are dependent on this behavior.
* It also changes how setup.py runs and uses a more simplified approach.
* Now full compatible with native android packages on pypi like `charset-normalizer` are installable without recipes (it directly picks up android wheels from pypi)

### DIFF
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -479,9 +479,6 @@ class Context:
                 exists(join(site_packages_dir, name + '.so')) or
                 glob.glob(join(site_packages_dir, name + '-*.egg')))
 
-    def not_has_package(self, name, arch=None):
-        return not self.has_package(name, arch)
-
 
 def build_recipes(build_order, python_modules, ctx, project_dir,
                   ignore_project_setup_py=False
@@ -729,9 +726,6 @@ def run_pymodules_install(ctx, arch, modules, project_dir=None,
     # Reuse the state constructed
     pip, platforms, indices, env = state
 
-    # It always runs with --upgrade so this is not required as it skips if module already exists
-    # modules = [m for m in modules if ctx.not_has_package(m, arch)]
-
     # We change current working directory later, so this has to be an absolute
     # path or `None` in case that we didn't supply the `project_dir` via kwargs
     project_dir = abspath(project_dir) if project_dir else None
@@ -787,7 +781,7 @@ def run_pymodules_install(ctx, arch, modules, project_dir=None,
         with current_directory(project_dir):
             # TODO: It will only work for basic python projects with no compiled components
             shprint(
-                pip, 'install', ".",
+                pip, 'install', ".", "--no-deps", "--only-binary=:all:",
                 '--target', ctx.get_site_packages_dir(arch),
                 '--disable-pip-version-check', '--upgrade',
                 *platforms, *indices, _env=env
@@ -796,13 +790,14 @@ def run_pymodules_install(ctx, arch, modules, project_dir=None,
         info("No setup.py found in project directory: " + str(project_dir))
 
     # Strip object files after potential Cython or native code builds:
-    if not ctx.with_debug_symbols and env.get("STRIP", None) is not None:
+    arch_env = arch.get_env()
+    if not ctx.with_debug_symbols and arch_env.get("STRIP", None) is not None:
         info('Stripping object files')
         shprint(
             sh.find, '.', '-iname', '*.so',
-            '-exec', env['STRIP'].split(' ')[0],
+            '-exec', arch_env['STRIP'].split(' ')[0],
             '--strip-unneeded', '{}', ';',
-            _env=env
+            _env=arch_env,
         )
 
 
--- a/pythonforandroid/build.py
+++ b/pythonforandroid/build.py
@@ -637,6 +637,9 @@ def process_python_modules(ctx, modules, arch):
     # add extra index urls
     for index in ctx.extra_index_urls:
         indices.extend(["--extra-index-url", index])
+
+    state = [pip, platforms, indices, env]
+
     try:
         shprint(
             pip, 'install', *modules,
@@ -646,7 +649,7 @@ def process_python_modules(ctx, modules, arch):
         )
     except Exception as e:
         warning(f"Auto module resolution failed: {e}")
-        return processed_modules
+        return processed_modules, state
 
     with open(path, "r") as f:
         try:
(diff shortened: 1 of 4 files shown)


RobertFlatt:
Strangely using this pr and specifying one of `chardet` or `charset-normalizer` in requirements, appears to result in both in the build (yes I appclean'd in between).

RobertFlatt:
Pinning charset-normalizer is ignored, this loads 3.5.1

`requirements = python3,kivy,requests,urllib3,charset-normalizer==2.1.1,idna`

T-Dynamos:
@RobertFlatt Fixed now.

Auto resolution was introduced in: https://github.com/kivy/python-for-android/pull/3271

With:

```python
requirements = python3, kivy==master, requests, charset-normalizer==2.1.1

```

image


With:

```python
requirements = python3, kivy==master, requests
```
In case of 32bit arm:
image

In case of 64 bit arm:
image


Notice the difference of `charset_normalizer-3.5.1-cp314-cp314-android_24_arm64_v8a.whl` and `charset_normalizer-3.5.1-py3-none-any.whl` (as no wheel for 32 bit arm)

T-Dynamos:
Even works with kivymd but need to specify my custom index: https://anshdadwal.is-a.dev/p4a-wheels/p4a/

```python
requirements = python3, kivy==master, kivymd
```

image

AccelQuasarDragon:
btw, I am just commenting here +1 to this PR to be merged because several users and I are hitting this bug since uv package manager pulls latest charset-normalizer which then hits you with the "charset_normalizer-3.5.1-cp314-cp314-android_24_arm64_v8a.whl is not a supported wheel on this platform."

If you search "charset_normalizer" you get 3+ hits on the kivy discord android support channel alone. 

1. me
2. Fector101 https://discord.com/channels/423249981340778496/712344698559397895/1540144039918051358
3. bob(robertflatt in this thread) https://discord.com/channels/423249981340778496/712344698559397895/1538320450298581052
4. davekri https://discord.com/channels/423249981340778496/712344698559397895/1538101151927504948
5. TLBritts https://discord.com/channels/423249981340778496/712344698559397895/1532125837560778885

this is a legit silent error that is probably causing p4a to freak out for many users and many people will not report it, but there have been 5 people in the support channel alone that have reported this issue and it is caused by charset updating with android wheels. please merge so the noobs don't suffer! thanks.

T-Dynamos:
`markupsafe` needs a recipe now.

kuzeyron (dismissed):

pythonforandroid/build.py
```diff
@@ -859,84 +754,56 @@ def run_pymodules_install(ctx, arch, modules, project_dir=None,
...
+
+    # Afterwards, run setup.py if present:
+    if project_has_setup_py(project_dir) and not ignore_setup_py:
+        with current_directory(project_dir):
+            # TODO: It will only work for basic python projects with no compiled components
+            shprint(
+                pip, 'install', ".",
+                '--target', ctx.get_site_packages_dir(arch),
+                '--disable-pip-version-check', '--upgrade',
+                *platforms, *indices, _env=env
```
AndreMiras: Wouldn't the platform flag also require `--no-deps` and/or `--only-binary=:all:` like we use to have?
Also refs https://github.com/pypa/pip/blob/26.2.1/src/pip/_internal/cli/cmdoptions.py#L123-L129

pythonforandroid/build.py
```diff
@@ -833,9 +720,17 @@ def run_pymodules_install(ctx, arch, modules, project_dir=None,
...
+        if (m_version := os.environ.get(f'VERSION_{module}', None)) is not None:
+            modules[index] = f'{module}=={m_version}'
 
-    modules = [m for m in modules if ctx.not_has_package(m, arch)]
+    modules, state = process_python_modules(ctx, modules, arch)
+    # Reuse the state constructed
+    pip, platforms, indices, env = state
+
+    # It always runs with --upgrade so this is not required as it skips if module already exists
+    # modules = [m for m in modules if ctx.not_has_package(m, arch)]
```
AndreMiras: Does this mean we redownload everytime and break offline build?
Maybe that's fine.
Also `not_has_package()` is now dead code and can be removed

pythonforandroid/build.py
```diff
@@ -859,84 +754,56 @@ def run_pymodules_install(ctx, arch, modules, project_dir=None,
...
+                pip, 'install', ".",
+                '--target', ctx.get_site_packages_dir(arch),
+                '--disable-pip-version-check', '--upgrade',
+                *platforms, *indices, _env=env
             )
+    elif not ignore_setup_py:
+        info("No setup.py found in project directory: " + str(project_dir))
+
+    # Strip object files after potential Cython or native code builds:
+    if not ctx.with_debug_symbols and env.get("STRIP", None) is not None:
```
AndreMiras: env here comes from `env = environ.copy()` in process_python_modules() which is os.environ.copy(), not the recipe env, so strip is likely never set hence this is now dead code

AndreMiras (dismissed):
Nice effort thank you! Looks good overall.
I didn't get a chance to test it, but I've raised a couple of concerns that might be worth a check and maybe a coverage if that wasn't caught by tests.
Let me know

pythonforandroid/build.py
T-Dynamos: > Does this mean we redownload everytime and break offline build?

That only happens if requirements are changed, else this stage doesn't run.

ydirson:
I was hit by the charset-normalizer issue during build, and can confirm this PR allowed me to build Kivy

AndreMiras (approved):
Thanks for all the good work and follow up fixes, thanks also for everyone who provided feedback and tested the PR locally.
One non-blocking follow-up: the stripping command currently runs `find .`, but run_pymodules_install() current working dir isn't ctx.build_dir.
Previously, stripping explicitly targeted ctx.build_dir.
We should follow up by giving find an explicit root, probably ctx.get_site_packages_dir(arch) or ctx.build_dir and also ideally restore the test coverage for with_debug_symbols.
I'm happy for this to merge and have the stripping handled in a follow-up.

AndreMiras:
FWIW, I've created the stripping and test coverage follow up here https://github.com/kivy/python-for-android/pull/3368

AccelQuasarDragon:
just wanna say thanks for the quick fix+merge it is greatly appreciated!

wefzg:
the error  "charset_normalizer. ..." is gone and the apk file is assembled again. much thanks for your help :-)

kuzeyron (approved):

kuzeyron:
Darn it. GitHub for Android is so messy. It doesn't sort comments/commits..

PR #3372, opened 2026-09-05
fix for freetypehost host

SNEHASISHROY-125:
updated freetype source url
to https://download-mirror.savannah.gnu.org/releases/freetype/freetype-2.14.1.tar.gz

### DIFF
--- a/pythonforandroid/recipes/freetype/__init__.py
+++ b/pythonforandroid/recipes/freetype/__init__.py
@@ -25,7 +25,8 @@ class FreetypeRecipe(Recipe):
     """
 
     version = '2.14.1'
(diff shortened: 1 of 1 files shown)


Novfensec:
Maybe its better if we start using https://gitlab.freedesktop.org/freetype/freetype/-/archive/VER-2-14-1/freetype-VER-2-14-1.tar.gz?ref_type=tags

kuzeyron:
@T-Dynamos Are you up for a task? Having a list with urls so when the first link fails then next one tries to be downloaded.

T-Dynamos:
@kuzeyron Okay!

On it.

kuzeyron (approved):

kuzeyron:
LGTM! Thank you!

Issue #3173, opened 2025-07-14
Introduce `--extra-manifest-application-xml` option


Novfensec:
This is a feature request.

- Add an option to directly write inside the `<manifest><application>` element of the AndroidManifest.xml.

Sample `buidlozer.spec` equivalent to the option in p4a:
```spec
# (str) Extra xml to write directly inside the <manifest><application> element of AndroidManifest.xml
# use that parameter to provide a filename from where to load your custom XML code:
# android.extra_manifest_application_xml = ./src/android/extra_manifest_application.xml
```

Sample `src/android/extra_manifest_application.xml`:

```spec
<activity
    android:name="org.myapp.android.MyActivity"
    android:process=":my_process"
    android:exported="true"
    android:configChanges="orientation|keyboardHidden"
    android:launchMode="singleTask" />

<receiver
    android:name="org.example.MyBroadcastReceiver"
    android:enabled="true"
    android:exported="false">
    <intent-filter>
        <action android:name="android.intent.action.BOOT_COMPLETED" />
    </intent-filter>
</receiver>

<service
    android:name="org.example.MyBackgroundService"
    android:enabled="true"
    android:exported="true" />
```

Novfensec:
Closed by https://github.com/kivy/python-for-android/pull/3377

PR #3377, opened 2026-09-10
Add --extra-manifest-application-xml for injecting elements inside <application>

copiousfish:
There's currently no way to add an arbitrary child element (like a <provider>, needed for e.g. a FileProvider) inside <application> in the generated AndroidManifest.xml:

- --extra-manifest-xml writes at the <manifest> root, as a sibling of <application> -- a <provider> there is invalid, providers must be nested inside <application>.
- --extra-manifest-application-arguments only splices into the <application ...> opening tag's attribute list, so it can't hold a full child element either.

This is a real gap: a FileProvider (needed to hand another app, e.g. a camera app, a content:// Uri instead of a file:// one and avoid FileUriExposedException on API 24+) requires exactly this kind of <provider> entry, and there's currently no supported way to add one without patching the rendered manifest after the fact via a p4a.hook.

Adds --extra-manifest-application-xml, rendered immediately before </application> in every bootstrap's manifest template, alongside the two existing options.

### DIFF
--- a/pythonforandroid/bootstraps/_sdl_common/build/templates/AndroidManifest.tmpl.xml
+++ b/pythonforandroid/bootstraps/_sdl_common/build/templates/AndroidManifest.tmpl.xml
@@ -142,6 +142,7 @@
     {% for a in args.add_activity  %}
     <activity android:name="{{ a }}"></activity>
     {% endfor %}
+    {{ args.extra_manifest_application_xml }}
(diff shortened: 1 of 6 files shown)


Novfensec:
Thanks https://github.com/kivy/python-for-android/issues/3173

T-Dynamos (approved):

PR #3380, opened 2026-09-13
android: Upgrade Hardware.java APIs for API 26+/23+/24+ compatibility

Novfensec:
Updates `Hardware.java` to use modern Android APIs, replacing several
deprecated calls that produce lint warnings on API 26+ targets.

### Vibration (`vibrate`)
- Use `VibrationEffect.createOneShot()` on API 26+
- Fall back to the legacy `Vibrator.vibrate(long)` on older devices

### Network detection (`checkNetwork`)
- Use `ConnectivityManager.getActiveNetwork()` + `NetworkCapabilities`
  on API 23+ instead of the deprecated `getActiveNetworkInfo()`
- Detects Wi-Fi, cellular, and Ethernet transports explicitly

### Network listener (`registerNetworkCheck`)
- Register a `ConnectivityManager.NetworkCallback` on API 24+
  instead of the deprecated `CONNECTIVITY_ACTION` broadcast
- Falls back to the broadcast receiver on older devices

### Context / View access
- Expose `context` and `view` as `public` (were package-private)
- Add `getContext()` helper that falls back to `PythonActivity.mActivity`
- Add `getView()` helper that falls back to the activity's decor view

### WiFi scanning (`enableWifiScanner`)
- Obtain `WifiManager` via `getApplicationContext()` (avoids memory leak
  with non-application contexts)
- Wrap `getScanResults()` in a `SecurityException` catch block
  (required permission `ACCESS_FINE_LOCATION` on API 29+)

### Misc
- Replace `String +=` concatenation loops with `StringBuilder`
- Remove redundant `(float[])` casts on `readSensor()` return values
- Add `sSensor == null` guard in `changeStatus()` for devices that lack
  a particular sensor type

## Testing
- Verified no behavioural regression on a device running API 36.
- Legacy code (API < 23/24/26) remain unchanged via SDK version checks.
- Same Hardware.java file is being used for ksproject toolchain. (All of these are tested and improved maintaining backward compatibility.)

### DIFF
--- a/pythonforandroid/bootstraps/common/build/src/main/java/org/renpy/android/Hardware.java
+++ b/pythonforandroid/bootstraps/common/build/src/main/java/org/renpy/android/Hardware.java
@@ -9,9 +9,13 @@ import android.hardware.SensorEvent;
 import android.hardware.SensorEventListener;
 import android.hardware.SensorManager;
 import android.net.ConnectivityManager;
+import android.net.Network;
+import android.net.NetworkCapabilities;
 import android.net.NetworkInfo;
 import android.net.wifi.ScanResult;
 import android.net.wifi.WifiManager;
+import android.os.Build;
+import android.os.VibrationEffect;
 import android.os.Vibrator;
 import android.util.DisplayMetrics;
 import android.view.View;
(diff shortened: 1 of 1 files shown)


Novfensec:
I did spotless with different version but I don't know which version this expects. And So there should be a pre-commit hook that applies spotless as expected itself.

